PR #1642, opened 2014-10-24
Package GaussianMixtures

davidavdav:
Hello, 

This is a package that I have been working on for a very long time now, GMMs or Gaussian Mixture Models.  It focuses on fairly big models (~2000 Gaussians of ~60 dimensions are not uncommon, where training data might encompass ~ 10^9 or more data points).  It has support for training the models on a computing cluster (using ClusterManagers) where data is passed to the training algorithm by streaming from disc. 

This week I accomplished some more inner-loop efficiency improvement and I further incorporated the package BigData that seemed not to be viable to live on its own, unfortunately.  I even managed to put together a travis automatic test.  (I tend to first write a package before I package it so that I can't call Pkg.generate()).  I finally changed the license to MIT. 

Cheers, 

---david


spencerlyon2:
I'm working on a package that does "the other kind of GMM" (econometrics a la Hansen 1982). (https://github.com/gragusa/GMM.jl)

Should we also try to avoid the name `GMM.jl` in favor of something a bit more verbose?

johnmyleswhite:
Yeah, let's not share that name. I think the rule should be: "don't claim disputed territory".

spencerlyon2:
Sounds good, any suggestions? `GeneralizedMethodOfMoments.jl` is too long!

johnmyleswhite:
I think that's not such a bad name, so you probably need feedback from somebody else.

PR #1933, opened 2015-01-01
Pull request/9f37988e

dostodabsi:


dostodabsi:
I might have mixed up some things. I'd like to merge this package in: https://github.com/dostodabsi/V.jl

ViralBShah:
How about calling it VStatistic.jl?

dostodabsi:
yeah, I thought you'd say that, but I quite like the subtle V for Vendetta reference. It's pretty hard to find this way, though. Weird tradeoff - will rename it!

PR #1936, opened 2015-01-02
Remove PValueAdjust

dirkschumacher:
I would like to remove my package until I made some adjustments.


kmsquire:
Hi @dirkschumacher, removing packages is pretty discouraged, as it has the tendency to screw up people's local package repositories.  

Why do you need to remove your package until you make the adjustments?  Can you describe the type of adjustments you plan to make?

kmsquire:
(Closing so this isn't actually merged.  If this really needs to happen for some reason, it can be reopened.)

dirkschumacher:
I changed the license and made some breaking api changes. While doing this I accidently removed the tags from the public github repo. I thought until I find time to fix this it is better to not list the package for new users.

kmsquire:
Hmmm.. okay, I see that that is a problem.  (tags == "tagged versions")

There's a reasonably good chance that @IainNZ has a recent copy that he could upload to github, and that we could temporarily point the URL at.

If not, a couple of thoughts:

1. `~/.julia/.cache` contains a cached version of most/all installed repositories.  It's a raw repository, so you can't see the files, but it *might* have the tagged versions.  Some googling and/or a post to julia-users might help (or someone copied below might post here).

2. If that doesn't work, looking at https://github.com/dirkschumacher/PValueAdjust.jl/stargazers, I'm wondering if any of them have a recent clone of the repo that they could extract and upload.  

Cc: @simonster (because he starred your repo), @StefanKarpinski

dirkschumacher:
I think I have a backup somewhere on a network drive but can only work on it later this weekend.

dirkschumacher:
I could restore the tagged version. Thanks.

simonster:
In case it's useful, I pushed my local repo to https://github.com/simonster/PValueAdjust.jl.

PR #1938, opened 2015-01-02
Pull request/1cbd954c

jlapeyre:
https://github.com/jlapeyre/ZChop.jl


milktrader:
Coming in at 13 lines of code in the `src/` directory, this must rank as the lightest weight package we have. Impressive parsimony. 

I think this is good to merge but quickly, curious in what domains `zchop` is useful.

milktrader:
Also needs a tagged version so it can be used with `Pkg.add("ZChop")`

jlapeyre:
I originally wrote this to remove visual noise in big lists of numbers, some of which were zero plus ~ eps(Float64).

I edited the README to include semi-realistic examples. These are similar to the examples in the docs for the [Mathematica function `Chop`](http://reference.wolfram.com/language/ref/Chop.html).

[Here]( https://answers.yahoo.com/question/index?qid=20080506193134AAOtwlt) is someone asking for this kind of function to zero small values in a sparse matrix. I also added an example of this to the readme for `ZChop`.

Here are some usage [examples in Maple](http://www.maplesoft.com/support/help/Maple/view.aspx?path=fnormal) of the similar function `fnormal`.

There are many examples in forums of people asking for something like `Chop` in matlab.

sympy provides [similar functionality](http://docs.sympy.org/0.7.2/modules/matrice/matrices.html#sympy.matrices.matrices.MatrixBase.eigenvects), but there [does not seem to be such a function](http://stackoverflow.com/questions/14419290/set-very-low-values-to-zero-in-numpy) in numpy.

Some earlier versions of matlab [had a function with the name `chop`](http://es.mathworks.com/matlabcentral/newsreader/view_thread/42587), but it did something different. The Julia function 'chop' does the same thing as the perl function `chop`. I think overloading `chop` is wrong because a) the functions are semantically completely distinct , b)  `zchop` should act as the identity on strings.

Having `zchop` act on "leaves" in an expression is Mathematica-like, which is perhaps not always a good choice in a language like Julia. But, I think it makes sense here. For instance, `lufact` returns a tuple of different objects, some of which can sensibly be "chopped".

jlapeyre:
Thanks, I will add a tagged version now.

The current [Julia manual](http://julia.readthedocs.org/en/latest/manual/packages/#publishing-your-package) and even more so the manual for v0.3 seem to suggest doing `Pkg.publish()`  before tagging and then doing it again  sometime later after tagging.

The v0.3 manual seems to suggest Pkg.publish() to put package in the METADATA. Then after some testing period, create a tag and do Pkg.publish() again.

milktrader:
Sounds good to me. How about a version, say `0.0.1`

jlapeyre:
Right, I just used the recommended `Pkg.tag("ZChop")` and got v0.0.1.  I did `Pkg.publish()` and am now waiting  with  "INFO: Validating METADATA".  This is running for several minutes and using 100% cpu. I aborted and tried again. Same thing. I'll keep waiting.

jlapeyre:
I'm reading about possible bugs in Pkg.publish() right now. I will try to find a work around and will comment when I find something.

jlapeyre:
I issued another PR for the tagged version. I hope everything is in order now!

IainNZ:
Superseded by https://github.com/JuliaLang/METADATA.jl/pull/1943 I believe

jlapeyre:
@IainNZ   Thanks for closing.

PR #1939, opened 2015-01-02
Add Watcher.jl

rened:


milktrader:
This is cool. Look forward to using it. Is it fast?

rened:
The implementation is quite simplistic (but using the watch_file function, i.e. using the OS' notification system). So I guess there should be no noticeable delay as long as there are only a few hundred files. But apparently "watched" files count towards the file-handle count: while testing with 1000 files I got a "Too many open files" error. (on OSX)

PR #1943, opened 2015-01-03
Register ZChop: v0.0.1

jlapeyre:
Here is a tagged version.


milktrader:
Ok, great. Maybe in the future this might belong in base? Not sure but you could do a PR down the road if there is a wider need for it.

jlapeyre:
Thanks! I somehow didn't see your last message till now. Yes, I think it may belong in Base. Maybe it will get some review since it is a module. --John

PR #1946, opened 2015-01-04
Register Latex + DictFiles

rened:


kmsquire:
Perhaps LaTeX.jl?

DictFiles/url
```diff
@@ -0,0 +1 @@
+ssh://git@github.com/rened/DictFiles.jl.git
```
mlubin: Why is this URL using the ssh protocol?

ViralBShah: Yeah - should use the git protocol.

rened: Thanks, was not aware of this.

rened:
I thought Latex.jl was simpler, but I can change that of course! @ViralBShah @mlubin what do you think?

ViralBShah:
I think `LaTeX` is the standard spelling, and looks more familiar to my eyes than `Latex`. I don't have a strong opinion otherwise.

rened:
Ok, thanks, I renamed it to LaTeX.

PR #1947, opened 2015-01-04
Register JuMPChance

mlubin:
A JuMP extension for optimization with probabilistic constraints.
@IainNZ @joehuchette


IainNZ:
I guess I should tag JuMPeR before class this semester!

ViralBShah:
Would you want it on JuliaBox? Feel free to open an issue there in that case. Usually, lead time is a week between image updates.

mlubin:
This is probably too specialized, not really for general consumption. I don't mind asking users to install it with Pkg.add.

PR #1951, opened 2015-01-06
Pull request/2a54505d

sawcordwell:


IainNZ:
Woo, NZ-made packages, with an operations research flavor, my favourite :D

IainNZ:
I think the name is good, even though we try to avoid acronyms. I can't imagine a strictly better alternative, and Google searching confirms my hunch there isn't anything else as popular for the acronym `MDP`. I'll let this sit for a bit in case there is a dissenting opinion.

sawcordwell:
Yeah, I did consider MarkovDecisionProcesses but since MCMC and others use acronyms I thought I would try this first.

PR #1962, opened 2015-01-09
Register PolarFact.jl

weijianzhang:
A Julia package for the polar decomposition. (https://github.com/weijianzhang/PolarFact.jl)

Currently, it provides the following algorithms:

* (Scaled) Newton's method
* Halley's method
* the SVD method 

More algorithms will come soon. The design of the package is inspired by NMF.jl.

### DIFF
--- a/PolarFact/url
+++ b/PolarFact/url
@@ -1,0 +1,1 @@
+git://github.com/weijianzhang/PolarFact.jl.git


milktrader:
This looks cool. Can you tag a version for it?

````julia
Pkg.tag("PolarFact")
````

weijianzhang:
Sure, I will do it now.

PR #1963, opened 2015-01-09
Tag PolarFact v0.0.1

weijianzhang:
A Julia package for the polar decomposition.


milktrader:
Looks good to me. Maybe we can have some local experts take a peak at it before we merge. @jiahao

weijianzhang:
Sure.

PR #1964, opened 2015-01-10
Register GraphCentrality: v0.0.1, v0.1.0

sbromberger:


ViralBShah:
Can this instead be part of graphs.jl ?

sbromberger:
Unfortunately not (yet). My modifications to DijkstraStates is causing concern. See https://github.com/JuliaLang/Graphs.jl/pull/129 and https://github.com/JuliaLang/Graphs.jl/pull/155

My bigger concern, frankly, is that anything that imposes ANY sort of performance cost within Graphs.jl will be summarily denied; this means that development of new functionality will not be approved until all optimizations are finished, and I don't necessarily want to wait that long in order to start using new code to do "real work".

ViralBShah:
Fair enough to keep it separate until a later point.

johnmyleswhite:
@sbromberger, I think you might find it useful to maintain your own fork of Graphs.jl if you feel that it's hard to win people over to your viewpoint. In general, I think we haven't found a good way to balance (a) encouraging people to work together with(b) allowing people to do good work without having to reach consensus first. I've come to think we should do more to encourage people to work a bit on their own as there are projects whose value is only obvious when they're fully built.

ViralBShah:
The downside with a fork is that it is not easy to distribute the functionality.

ViralBShah:
I also think that while a fast graph centrality measure eventually belongs in Graphs.jl, there will be lots of graph algorithms that extend the functionality in Graphs.jl. This package is building on Graphs.jl - so I am merging it for now.

ViralBShah:
@sbromberger BTW, I think that the early effort towards having graph generators for benchmarking that you have in this package could perhaps live in a separate package.

sbromberger:
@ViralBShah re: the generators / readers - yes. I know someone was working on graph persistence, and I didn't want to collide with those efforts, so I hacked together something that would allow me to do quick tests. The code is really there for convenience and I'm happy to pull it out or incorporate it directly into the test suite. There's no reason for it to be directly accessible, and it's the reason I haven't exported the functions.

Frankly, they're not good enough to be their own package, and I don't plan on refining them - it would be wasted effort since they do what they need to do now and since there should be some better code forthcoming via some other effort.

PR #1970, opened 2015-01-11
(Maybe don't merge?) Package for measuring test coverage of base

timholy:
It's not at all obvious to me that this needs to be a package: when the results are sent to Coveralls, this "package" only needs to be run by one person in the entire world. So maybe it shouldn't even be registered. But in case others feel differently, I'm submitting it for inclusion.

Either way, the code will be available from my GitHub account, and I'm happy to run it nightly on my server.


jiahao:
perhaps the coverage testing can be run by @staticfloat's buildbots?

staticfloat:
@timholy if you'd like to distribute the load, I'd be happy to set something up on the buildbots to run nightly coverage analysis.  Heck, we could even run coverage analysis on a per-commit basis, like we do for packaging.  It doesn't matter to me; this stuff is easy to setup.

timholy:
I'm not sure what kind of per-commit analysis you already do, but if you are interested in that, it sounds cool. I kept this off Travis in part because Travis was taking approximately 50+-n minutes to run this, and 50 is of course their timeout, so there were many failures right at the end of testing. (You have to run everything in a single process, and turn off inlining, so it's slower than our usual tests.) I think it was a fair bit faster on my own server, but it's still enough of a time commitment that daily seemed sufficient to me.

Basically, I'm fine with whatever. I set this up to get it going, but if you want to pick up the ball and run further, be my guest. Why not check out https://github.com/timholy/CoverageBase.jl and tell me what you want to do; until I hear otherwise, I'll just keep the job running on my server.

staticfloat:
Sounds great.  I'll add it to my TODO list.  :)

PR #1973, opened 2015-01-12
Register WORLD: v0.0.1

r9y9:
WORLD.jl is a wrapper for WORLD; a high-quality speech analysis, modification and synthesis system.


ViralBShah:
Name nitpick - Could we name it WORLDSpeech or something? I would me much more interested with Speech in the package name as a package browser. But given that WORLD is the upstream name, I will just make this a suggestion.

r9y9:
From a discussion of naming issue in https://github.com/JuliaLang/METADATA.jl/pull/1729, I would prefer to use a creative name for this, since the original WORLD (written in C++)  already have a creative name. If we need to change the name to more descriptive one, I think WORLDVocoder is better than WORLDSpeech.

StefanKarpinski:
I'd maintain this objection to the name `using WORLD` just really looks like you're importing all packages or something. If you see `using WorldVocoder` then it's fairly clear that it's a speech thing.

johnmyleswhite:
Given that we're trying to curate the names of Julia packages, but can't curate the names of non-Julia code, I think we need to adopt a special strategy for library-wrapper packages. Otherwise, we're going to have this debate over and over again since the outside world doesn't care about our naming conventions.

joehuchette:
What about just using a uniform suffix for packages that are thin wrappers around external libraries, e.g. Foo-wrap.jl? (Is that a valid package name with the ``-``?)

ViralBShah:
How about a `Lib` prefix?

ViralBShah:
Of course, that would only apply to projects that do not already have lib in the name.

johnmyleswhite:
You can't put `-` in an identifier, but I otherwise like both Wrap as a suffix and Lib as a prefix.

And, yes, we'd have to do this retroactively.

r9y9:
I think wrapper packages doesn't always need to have a descriptive name; packages have its description, README and docs. Adapting a special strategy for library-wrapper packages seems good idea. I'm not sure how the naming issue to be resolved, but I'm fine with changing the name to `WorldVocoder` and can make a PR. I understand a package name should be self-descriptive in genereal.

IainNZ:
Reasserting my support for WORLD.jl and GR.jl and CPLEX.jl and all such things. I don't understand when and where the confusion would arise.

* Looking through pkg.julialang.org? Descriptions are provided
* Reading someone elses code? Presumably if you are reading the code, it makes sense in the context. Even if not, its no more confusing than Gadfly - you just go look it up.
* Reading through METADATA? Only a very small number of people will do that, and its the same as my second point.

ViralBShah:
While I personally like a prefix, I can live without it for the reasons mentioned. However, avoiding 2 and 3 letter package names otherwise, is a rule worth considering as it leads to fewer cryptic names.

PR #1974, opened 2015-01-12
Tag Watcher and register ShapeModels

rened:


milktrader:
I'm interested in how you approached plotting for ShapeModels. You don't REQUIRE a plotting package but you do test if PyPlot is installed. 

I'm not sure why I like this design but maybe you can elaborate?

Also, can you plot with other packages and how would you do so?

rened:
I didn't want to impose a specific plotting package on anyone. I, for example, use Gadfly or PyPlot depending on what is easier to use in the context.
In several packages I have the following:
```jl
isinstalled(a) = isa(Pkg.installed(a), VersionNumber)
if isinstalled("PyPlot")
    ...
end
```
Currently I see the provided plot functions just a simple helpers - once you know that the models you estimate are doing the right thing you just include it in your actual algorithm / pipeline. At which point I believe everyone has their own preferences how to visualize things.
But, actually, it would of course be neat to have kind of meta-drawing primitives for simple stuff (just plotting points and lines in 1 plot for example) and then have them draw with whatever package is already installed / loaded. 
Haven't thought much about it that though, yet. What do you think?

milktrader:
I like this approach and will need to adopt it.

PR #1975, opened 2015-01-12
Pull request/847ddc98

joshuamiller:


StefanKarpinski:
That URL is only valid on your local machine. The wikipedia entry for [BEncode](http://en.wikipedia.org/wiki/Bencode) is inconsistent about capitalization, but given that Julia packages tend to be capitalized and the `bencode` protocol is pronounced "B Encode", I think that `using BEncode` would be clearer.

joshuamiller:
@StefanKarpinski Yeah, I just caught that I'd set url wrong. Will make that change and the naming change, then resubmit.

PR #1978, opened 2015-01-13
new package DustExtinction

kbarbary:
This is an astronomy-specific package for dealing with interstellar dust extinction. Hopefully the package name is specific enough, but I'm open to alternatives.


ViralBShah:
The name suggested astronomy to me right away, but maybe not for everyone. Let's go with it anyways for now.

PR #1980, opened 2015-01-13
Register TypeInheritance

DANA-Laboratory:
julia dose not support inheritance like standard object oriented langs e.g.(c++ or java).
Inheritance really is a good feature when creating mathematical models of real world objects.
I solve this problem for myself (when modeling process equipments) by writing a simple module that simulates Type Inheritance.
This trick creates a direct way to address all variables of member variables with special prefix before their names (at this time prefix is *`__base__`*) by using their Symbols. 
Although it's not really inheritance, now i can define a _Special Pump_ type that have all variables of a _Basic_ one. 
and evaluate expressions like:
`:(mypump.anSpecialPump_variable*mypump.aBasicPump_variable)` 
in place of  
`:(mypump.anSpecialPump_mv*mypump.basicPumpModel.aBasicPump_mv)` 
this really simplifies creating large models of various equipment each with many equations .

### DIFF
--- a/TypeInheritance/url
+++ b/TypeInheritance/url
@@ -1,0 +1,1 @@
+git://github.com/DANA-Laboratory/TypeInheritance.jl.git


DANA-Laboratory:
i would like to mention :
https://groups.google.com/forum/#!topic/julia-dev/eA4VkFAD-yQ
https://groups.google.com/forum/#!topic/julia-dev/p9bV6rMSmvY
#4935

milktrader:
Is this part of the inheriting behavior versus inheriting structure debate?

DANA-Laboratory:
@milktrader
i dont think so, mostly because this way is not really a solution for inheritance. but a simple way for evaluating expressions when u think having inheritance relationships between involved types could simplifies expressions and help modeler do better.

ViralBShah:
Is there any objection to the name of the package? I feel it is a bit general, but ok for now. If there are no further objections, let's merge, say tomorrow.

IainNZ:
I find the package kind of odd, as it seems to depend on `eval`ing things, which doesn't seem like it'd work. Is there any other code using this? Will a version be tagged?

DANA-Laboratory:
@ViralBShah you are right, i am thinking of it. may be a better name exist.
@IainNZ thanks for your advice, thanks @tshort julia already have a good simulation package for equation based modeling. TypeInheritance is a part of another project for equation based modeling that i am working on it. https://github.com/DANA-Laboratory/dana-laboratory.jl.

DANA-Laboratory:
rename  to *DotPlusInheritance* and *not depend on eval*

PR #1981, opened 2015-01-13
Add GR module

jheinen:
I'd like to make the GR module available to Julia users. I will provide further information on http://gr-framework.org as soon as the module is published.


IainNZ:
The Julia convention would be to capitalize it to GR.jl

jheinen:
I'll create a new GitHub repo 'GR' and make a new pull request with capitalized names

IainNZ:
You can rename a github repo, no need to create a new one.

jheinen:
I changed the module/repository names to 'GR'.

IainNZ:
It seems like the folder in METADATA (in this PR) is still in lower case

jheinen:
Folder name should be ok now.

IainNZ:
Looks good, thanks for your patience!

jheinen:
I think the problem was at my side. Thanks for YOUR patience :-)

jheinen:
How can I change the sha1 of GR/v0.1.0 to 18c05885c5cd31e698d6a2b33508c0e0b17c0451 ?
After renaming all this stuff, I forgot to update it.
Thanks in advance.

ViralBShah:
Sorry coming late here - but I feel that 2 letter names are very difficult to figure out the meaning of. I wish we can have a slightly more descriptive name. Perhaps just GRFramework here?

jheinen:
The idea was to keep the names as close as possible to the Python package (gr) or other GR bindings (C, Swift, Fortran, Pascal. If it's not acceptable, I could change the module name to GRFW.

ViralBShah:
Let's wait for others to chime in.

IainNZ:
I think its fine for wrappers to just be the same as their libraries - why would anyone *need* to figure out what GR.jl is without the ability to look at it in pkg.julialang.org or in Github or just googling it?

ViralBShah:
Yes, for upstream libraries, that would be ok.

jheinen:
Thanks a lot. This makes migrating Python scripts to Julia much easier :-)

StefanKarpinski:
I have to say I'm not thrilled about such a short, opaque package name.

scidom:
GR is also an international abbreviation for Greece, so expect at least 10 million of Greek natives to misinterpret the name of the package at a first glance :)

milktrader:
GR is also "hail" in pilot weather information IMETARs). I think it's one of those French abbreviations that snuck in because there is no G or S in hail. 

In any case, I likewise would prefer something more descriptive.

milktrader:
This discomfort is similar to [WORLD](https://github.com/JuliaLang/METADATA.jl/pull/1973) where upstream libraries become Julia packages by appending `.jl` to them. I'm really not sure how this issue gets resolved. I suppose if you don't know what GR is in the Python world, you'd not be troubled by using it so maybe the point is moot after all.

rfourquet:
My problem with a non-descriptive name like this is that if the description on the pkg page is not descriptive either, I have to click on the link to know what it is about (I want to click b/c sometimes a package I never heard about turns out to be useful for me). Here the description "The GR module for Julia" sounds like a tautology (or pleonasm?). While there are no categories/searchability in the pkg page, useful descriptions are even more needed.

jheinen:
The Julia package list ist not up-to-date. I changed the description to "Plotting for Julia based on GR, a framework for visualisation applications" and we should wait for the next package update. If this is not acceptable, please let me know.

https://github.com/jheinen/GR.jl/issues/1#issuecomment-70297329

rfourquet:
Oh sorry for not having checked, the new description is nice.

PR #1982, opened 2015-01-13
Register JFVM

simulkade:


IainNZ:
Seems to be replaced by #1983

PR #1983, opened 2015-01-13
Pull request/eb45eaac

simulkade:


IainNZ:
So with the name, "Julia Finite Volume Model"? Seems like this would fall under the multiple-approaches-to-the-problem category, so idiosyncratic names are fine.

simulkade:
Indeed. I thought FVM (finite volume method) or FEM (finite element method) are quite familiar to those who work with partial differential equations. I could have chosen juliaFVM or JuliaFiniteVolume as well.

IainNZ:
No I think you made a good choice, I don't like package names having Julia spelled-out in them because it makes them seem "official".

PR #1989, opened 2015-01-14
Update sha1

jheinen:


jheinen:
Next time, I will use `Pkg.tag()` and `Pkg.publish()`. Sorry for the inconvenience caused.

PR #1998, opened 2015-01-16
Tag GR v0.3.0

jheinen:
I'd also like to use "Plotting for Julia based on GR, a framework for visualisation applications" for the package description, but I still don't know how to change the metadata information in the Julia packages repo.


aviks:
@jheinen where would you like to see that description? If you are talking about pkg.julialang.org ,  you need to set the description for the Github repository.  It is currently set to "The GR module for Julia " in your repo.

PR #2000, opened 2015-01-17
Register DotPlusInheritance

DANA-Laboratory:
#1980 
changes:
- rename general looking *TypeInharitance* to *DotPlusInheritance*
- re define `(.+)`  so it evals the expression itself
- add another method to `dotplus()`
- review tests


DANA-Laboratory:
not ready for publish yet.

PR #2005, opened 2015-01-19
Register ThermodynamicsTable

DANA-Laboratory:
Thermodynamics Table, indexed by CasNo
only physical facts, no more calculations.
i think it's legal please check-> https://groups.google.com/forum/#!topic/julia-dev/IvHAGWM75wI


DANA-Laboratory:
update repo and tag v0.0.1

ViralBShah:
I think this is ok based on the discussion and that this is limited to facts.

PR #2008, opened 2015-01-20
Tag Convex v0.0.3

karanveerm:


IainNZ:
You sure about it requiring SCS?

karanveerm:
We want to provide at least one solver by default when Convex is installed. I changed ECOS to SCS, mostly from the perspective of the class we're TAing, EE364A where we want students to just stick with one solver, and ECOS can't solve SDPs. 
What do you think? 
cc @madeleineudell

IainNZ:
My only thinking is that it might confuse people (not EE364A students) into thinking SCS is required for Convex to work, and anyone who wants to use Convex would have to download SCS even if they don't want to. It isn't a big deal, but just didn't quite feel right to me. Totally your call though!

madeleineudell:
Well in some sense SCS *is* required, because there's no other solver that
can do exponential cones and sdp cones at the same time, which is necessary
to solve problems with eg logdet. (Unless I underestimate the powers of
MOSEK.)

Iain, would you prefer if we installed no solver by default, and just
raised very articulate errors? eg if you call solve!() without explicitly
specifying a solver, it runs through all your installed solvers to see if
any can solve this problem type. If not,

"No solver found for conic problems with cones <...>. To solve this
problem, please install either Mosek or SCS using Pkg.add("...")."

(That is, we'd encode in code a mapping from sets of cones to appropriate
solvers.)

(Such a mapping would be nice to have regardless. eg a
`canbesolvedby(p::problem)` method.)

Madeleine

On Mon, Jan 19, 2015 at 9:10 PM, Iain Dunning <notifications@github.com>
wrote:

> My only thinking is that it might confuse people (not EE364A students)
> into thinking SCS is required for Convex to work, and anyone who wants to
> use Convex would have to download SCS even if they don't want to. It isn't
> a big deal, but just didn't quite feel right to me. Totally your call
> though!
>
> —
> <https://github.com/JuliaLang/METADATA.jl/pull/2008#issuecomment-70593728>
> .
>



-- 
Madeleine Udell
PhD Candidate in Computational and Mathematical Engineering
Stanford University
www.stanford.edu/~udell

IainNZ:
Yeah that was my reasoning, related back to this actually: https://github.com/JuliaOpt/JuMP.jl/issues/83 (see bottom for supportedcones discussion). You are absolutely write about needing SCS if you want to solve those problem classes, but I imagine most people are probably only interested in, say, SOCPs, so might choose ECOS, Gurobi, whatever - they don't need SCS.

madeleineudell:
KV writes on that thread that he was planning to write a function for
Convex that takes in the array of cones and returns some solver that
1. the user has installed
2. can support those cones

(or throws a supremely clear error otherwise.)

KV, any progress on this? It definitely should be written, and would be
super useful for Convex; though whether it should like in Convex or MPB is
another question.

On Mon, Jan 19, 2015 at 9:52 PM, Iain Dunning <notifications@github.com>
wrote:

> Yeah that was my reasoning, related back to this actually:
> JuliaOpt/JuMP.jl#83 <https://github.com/JuliaOpt/JuMP.jl/issues/83> (see
> bottom for supportedcones discussion). You are absolutely write about
> needing SCS if you want to solve those problem classes, but I imagine most
> people are probably only interested in, say, SOCPs, so might choose ECOS,
> Gurobi, whatever - they don't need SCS.
>
> —
> <https://github.com/JuliaLang/METADATA.jl/pull/2008#issuecomment-70596843>
> .
>



-- 
Madeleine Udell
PhD Candidate in Computational and Mathematical Engineering
Stanford University
www.stanford.edu/~udell

karanveerm:
Sorry, I haven't had time for that- things have been very busy. For now, I'm closing this PR and will make a new one without a solver installed by default.

PR #2012, opened 2015-01-20
Register OCCA

ReidAtcheson:
OCCA is a cross platform shared memory threading library that is retargetable to multiple backends such as pthreads, openmp, opencl, and cuda. OCCA.jl is a Julia interface into OCCA.


kmsquire:
Looks interesting!  Would you care to also tag a version, so people could install it with `Pkg.add("OCCA")`?

ReidAtcheson:
Thanks for response. I added the sha1 for the git commit, is that what you meant for tagging the version?

kmsquire:
Yes. You could also use `Pkg.tag("OCCA")` in Julia to do this, which would start numbering at 0.0.1.  We're encouraging packages to start with that version (more stable packages sometimes choose 0.1.0). Would you also be able to make that change?

kmsquire:
Also, it would be great if you squashed the commits together. Cheers!

ReidAtcheson:
Ok I rebased and changed version to 0.0.1.

kmsquire:
Thanks!  LGTM.

PR #2013, opened 2015-01-21
Register and tag Voting package

tchajed:
Voting provides implementations of voting schemes, or more technically, social choice functions. A set of votes is expressed as permutations of a ballot, a generic array, and several functions give the top candidate (or candidates, in the case of ties) based on different voting schemes.


StefanKarpinski:
I wonder if `VotingSchemes` wouldn't be a better package name? One could imagine a package that actually implements voting via some kind of API being named `Voting`.

milktrader:
They could call that package `Vote` maybe?

IainNZ:
Or `Votes.jl`, `Elections.jl`, or whatever. Lets cross that bridge when we get there.

PR #2018, opened 2015-01-22
Pull request to add ShowSet module

scheinerman:
This is a pull request to add my ShowSet module. The ShowSet module outputs Julia `Set` and `IntSet` objects as lists of their elements enclosed in curly braces. Here's a before-and-after example:
```julia
julia> A = Set(["alpha", "beta", "gamma", "delta"])
Set{ASCIIString}({"alpha","gamma","delta","beta"})

julia> B = IntSet(5,2,6,11)
IntSet([2, 5, 6, 11])

julia> using ShowSet
Warning: Method definition show(IO,Set{T}) in module Base at set.jl:10 overwritten in module ShowSet at /Users/ers/.julia/v0.3/ShowSet/src/ShowSet.jl:25.
Warning: Method definition show(IO,IntSet) in module Base at intset.jl:176 overwritten in module ShowSet at /Users/ers/.julia/v0.3/ShowSet/src/ShowSet.jl:26.

julia> A
{alpha,beta,delta,gamma}

julia> B
{2,5,6,11}
```


milktrader:
Looks great. If you want to suppress the `Method definition show ...` warning you need to import the `Base.show`. Do this in front of *your* `show` method like this:

    import Base.show

Easy. If you have a list of Base methods you can also do this

    import Base: show, length

milktrader:
Oh, I see you have already imported the Base.show and Base.string methods ... Not sure why we're getting this warning ...

IainNZ:
Because he is really overriding the Base methods and replacing them with his own.

milktrader:
Yeah, I see that now. There are actually base methods for that type signature.

milktrader:
I'll submit a PR to fix this, but my solution involves a change to the API since the way I figured to get around this was to create a new type.

kmsquire:
I would think that something like an `@override` or `@nowarn` macro would be better.  Either of these would act as a flag to the compiler (like `@simd` or `@fastmath`).

milktrader:
https://github.com/scheinerman/ShowSet.jl/pull/1

scheinerman:
I rather like @kmsquire's proposal for an `@nowarn` macro.

milktrader:
@kmsquire could you file an issue for creating such a macro? The more I think about this the more I'm __not__ in favor of such an idea. The reason being is that a warning is put in place to sound an alarm of abhorrent behavior. To remove it, one should either stop the code from doing what it's doing, or change the parameters of the warning. It's like putting a fire alarm in your house and removing the batteries.

I do agree that warnings can get excessive and annoying, but maybe that's motivation for getting the code back in bounds on the playing field.

milktrader:
Actually, I found some discussion of this already ...

https://github.com/JuliaLang/julia/issues/4895

PR #2020, opened 2015-01-23
Pull request/b23d3890

jgoldfar:
Adding a preliminary (works for me; perhaps 1.0 is too much...) version of Silo.jl, a wrapper for the [LLNL Silo data library](https://wci.llnl.gov/simulation/computer-codes/silo). Package tests are running (works on release and nightly Julia builds.) Welcoming comments and issue requests!


ViralBShah:
I do think that 1.0 is too much. Can we start at 0.1? Can rapidly go towards 1.0 though. Of course, if you feel that the wrappers are 1.0 quality, we will be happy to merge.

ViralBShah:
I didn't know about Silo before. Looks like it has some pretty cool mesh generation capabilities.

jgoldfar:
I have no problem starting at 0.1.0, completely agree with that. Is there an easy way to remove the current version tags and replace them? I usually use mercurial but would like to use git more; hg-git causes a lot of history duplication somehow, even though it works.

StefanKarpinski:
Git delete the tag: `git tag -d v1.0.0` or something like that. Also delete the METADATA version directory.

jgoldfar:
Made those changes; thanks for the help! Not sure if I did it right (should have just pushed changed METADATA.jl from git, rather than in julia?), but the updated PR is at #2027

PR #2040, opened 2015-01-27
New version of DASSL

pwl:


IainNZ:
I don't know how this passed Travis, but there is no `sha1` file - instead there is a `version` file!

pwl:
Sorry, my bad.  Thanks for the fix.

PR #2041, opened 2015-01-27
Register Jumos, a package for molecular simulations !

Luthaf:
Well, I think it is time to make this package available to the world !

[Jumos](https://github.com/Luthaf/Jumos.jl) is a package to do molecular simulation in Julia. The main goal here is to be as extensible as possible — each algorithm is a type and can be subtyped and replaced — and to provide a toolbox for molecular simulations. 

Only the basic of molecular dynamic simulations are implemented, together with some simple analysis.


Luthaf:
Well, that was quick !

PR #2042, opened 2015-01-27
Register GeoJSON: v0.0.1

yeesian:


garborg:
@yeesian I believe the SHA should be 344844e806b161d0ae9be532b15c6be4be1a2b96. I don't see the included SHA in the commit history.

garborg:
Shoot, a moment too late. Since it's a SHA that (I think) couldn't possibly be checked out, it's kosher to update the SHA without tagging a new version.

yeesian:
oh boy.. I could have messed things up by accident; how do I resolve this?

aviks:
Yikes, yes, I clicked the button a little too eagerly. I'll try and fix this.

yeesian:
Thanks! and sorry about it ><

garborg:
It's fine to just submit a PR that updates the file in place. (It's generally not allowed to change an existing tag, unless it has only been merged for a few moments, or it was not valid. No harm done here.)

PR #2044, opened 2015-01-27
Update GeoJSON SHA

aviks:
Initial version SHA was incorrect.


aviks:
cf: #2042 
cc: @garborg @yeesian

garborg:
Thanks, @aviks!

PR #2045, opened 2015-01-28
Update tikzpictures to 0.2.0

mykelk:


IainNZ:
Please don't merge PRs like this with 20 merge commit and two actual commits.

mykelk:
Not sure where all the merges came from, but the Pkg.tag("TikzPictures") and Pkg.publish() failed a number of times before it went through.

IainNZ:
You tagged v0.0.2 in August by the looks of it, and then ran Pkg.update 20 times between now and when you just tagged v0.2.0

johnmyleswhite:
We should maybe revert these changes.

mykelk:
@johnmyleswhite That's fine with me. But we still want TikzPictures to go to v0.2.0. What is the cleanest way to do this?

johnmyleswhite:
If it's not too late, we need to do a proper revert. One thing we need is to stop having new commits going in from other packages. So let's hold off on anything else you've been working on. After that, it's probably easiest to do the revert for you.

mykelk:
It's not too late. I made the PGFPlots.jl tag after this one, but Pkg.publish() went through without all those merges. You can go ahead with the revert. Sorry for messing this one up.

johnmyleswhite:
Looks like Iain (or someone else) may have fixed this already.

Don't worry about the problems with this one. I'd just try to keep in mind that you need to read through exactly what changes are being made before hitting the merge button.

PR #2050, opened 2015-01-28
Register Plotly

bpostlethwaite:


milktrader:
Julia packages are named with a `.jl` extension, so this would best be renamed as Plotly.jl

milktrader:
Once that's done, a version has to be tagged so users can add the package with the `Package.add()` method. To do this:

````julia
julia> Pkg.update()
# get some update info here
julia> Pkg.tag("Plotly", :patch)  # this will start versioning at 0.0.1
````

bpostlethwaite:
okay, repo name has changed to Plotly.jl and I have added a tag.

milktrader:
Great, now we just need to update the PR. It still shows the old url and no tags.

ViralBShah:
It's exciting to have a plotly interface.

bpostlethwaite:
Okay I think I made a mess of this PR. I attempted to put control of the METADATA fork under the Plotly organization and got confused what Pkg.publish() was doing under the hood. I think I will make the PRs manually from now on.

The new PR is at: JuliaLang/METADATA.jl/pull/2054

PR #2051, opened 2015-01-29
Register MatpowerCases.jl

kersulis:
Hello,

I am a PhD student studying Power and Energy. I used to use MATLAB, but @mlubin converted me to Julia last summer.

One of the few things I miss about MATLAB is easy access to transmission network data via the GPL-licensed [MATPOWER][2]. The package I would like to register, [MatpowerCases][1], makes it easy to load MATPOWER's transmission network data (along with data for other popular networks) into a Julia environment.

Grateful for your consideration,
Jonas Kersulis

[1]: https://github.com/kersulis/MatpowerCases.jl
[2]: http://www.pserc.cornell.edu//matpower/


mlubin:
:+1:

PR #2052, opened 2015-01-30
Register CoinOptServices: v0.0.1, v0.0.2, v0.0.3 

tkelman:
The first open-source MINLP solver[s] connected to Julia, I believe. Should be functional, though I need to put it through some more comprehensive tests.

cc @mlubin @joehuchette @IainNZ


joehuchette:
This is great, Tony. We can plug in ``OsilSolver`` in the JuMP tests to stress test it a bit.

tkelman:
I was thinking the same thing. (Also your in-progress baron interface was a useful reference, so thanks for that.) And once I figure out the right way to implement osil reading, that will provide a few more test problems. I think Gams has some test sets on their website in osil format, and there's https://github.com/coin-or/OS/tree/master/OS/data/osilFiles.

Building all of Bonmin, Couenne, and OS might be a bit taxing for JuMP's Travis setup to do on every commit though.

mlubin:
Nice!
Maybe just run JuMP tests as part of the OS tests? We can add ``OsilSolver`` to the list of recognized solvers to test but exclude it from JuMP's ``test/REQUIRE`` so it doesn't run on travis.

joehuchette:
We can try out adding it as a dependency and see how much it slows down. Even if we don't add it to Travis, we can still add it to the tests so that anyone who has it built locally will test it whenever the tests are run. 

Am I correct in my understanding that this supports:
* LP
* IP
* QP
* MIQP
* MIQCP
* SOCP
* NLP
* MINLP

but not
* SDP
* SOS constraints
* MIP callbacks

(that mess of acronyms is pretty revolting)

mlubin:
Does OS support SOCP? That would be surprising since OSI doesn't.

joehuchette:
I meant the SOC _qua_ quadratic constraints we test from JuMP, not conic form.

tkelman:
I don't currently have any code for specific handling of quadratic objectives or constraints (osil does have a format for them in the schema), but they should work through the general nonlinear expression tree interface. There's some stuff in the schema for cones and SDP's are being actively worked on by Horand Gassman, but I don't have those connected yet. SOS constraints could probably be made to work but I haven't done so yet.

mlubin:
OSI doesn't support plain quadratic constraints either.

joehuchette:
It should through the NLP interface, at least.

mlubin:
@tkelman, out of curiosity, have you run into any bugs in OS so far?

tkelman:
Yes. And reminding myself about bugs that I reported some time ago but don't look like they're fixed yet: https://projects.coin-or.org/OS/ticket/52

I think that one is a problem in the hessians for maximization problems. There could easily be others. And there's a major performance difference in function evaluation time between my cross-compiled WinRPM binaries and the Win32 MSVC binaries from http://www.coin-or.org/download/binary/OS/OS-2.4.1-win32-msvc9.zip. I should try a CoinAll binary and see how that looks.

mlubin:
> And there's a major performance difference in function evaluation time between my cross-compiled WinRPM binaries and the Win32 MSVC binaries

Which are faster?

tkelman:
The MSVC ones, annoyingly. By a substantial enough margin that I suspect there's a bug somewhere.

IainNZ:
We can merge this right?

tkelman:
Unless anyone wants to bikeshed the name. I've got more feature and bugfix updates coming but that doesn't mean the already-tagged versions aren't useful.

mlubin:
I'd vote actually for ``CoinOptimizationServices`` or something similar. The name as it is seems overly broad.

tkelman:
That's getting pretty long to type, there are only 3 current examples of packages with names at least that long:
```
julia> readdir(Pkg.dir("METADATA"))[map(length,readdir(Pkg.dir("METADATA"))) .>= length("CoinOptimizationServices")]
3-element Array{Union(UTF8String,ASCIIString),1}:
 "DirichletProcessMixtures"
 "ExtremelyRandomizedTrees"
 "InformedDifferentialEvolution"
```

I had originally thought `CoinOS` might work, but OS is not the greatest acronym.

IainNZ:
Yeah `CoinOS` is almost-good... `CoinOptServices`? I'm ok with `OptimizationServices`, I think its weird enough no one would make assumptions about what it is.

mlubin:
``CoinOptServices`` is okay with me. ``OptimizationServices`` doesn't really describe what the package does, most people will use it as a MathProgBase solver, some will use it to read in osil models, and even fewer if any will use the remote services aspect.

tkelman:
`CoinOptServices` works, I can start the renaming train going if no one objects to that. grep and sed, rename the folder and url here, anything else I need to do?

quinnj:
Rename the actual github repo? (probably obvious, but you didn't mention
that in your list)

tkelman:
Thanks, yeah that much is clear. And `git mv` the main `src/OptimizationServices.jl` file too.

tkelman:
Err I guess the tagged 0.0.1 and 0.0.2 most likely won't function, so there's no point in registering them. I'll overwrite this with just 0.0.3.

tkelman:
Rename done, look okay?

PR #2057, opened 2015-01-30
Force package repo names to end in .jl

jiahao:
We have enforced this policy manually up to now. Perhaps adding a test for this naming policy to the METADATA tests will help cut down on the need to check URLs of newly registered packages.

I have tested the logic locally and it seems to work. In fact, it picked up several packages whose repository names do not have a `.jl` extension, thus causing Travis to fail:

- [ ] NamedArrays git://github.com/davidavdav/NamedArrays.git
- [ ] PEGParser git://github.com/abeschneider/PEGParser.git
- [ ] JFVM git://github.com/simulkade/JFVM.git
- [ ] CompressedSensing git://github.com/dahlend/CompressedSensing.git

Perhaps we can ping the authors to get the repos renamed.

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -14,9 +14,12 @@ for (pkg, versions) in Pkg.Read.available()
     @assert ismatch(r"git", scheme) "Invalid url scheme $scheme for package $(pkg). Should be 'git' "
     if ismatch(r"github\.com", host)
         m2 = match(gh_path_reg_git, path)
-        @assert m2 != nothing "Invalid github url pattern $url for package $(pkg). Should satisfy $gh_path_reg_git"
+        @assert m2 != nothing "Invalid GitHub url pattern $url for package $(pkg). Should satisfy $gh_path_reg_git"
         user=m2.captures[1]
         repo=m2.captures[2]
+        
+        @assert endswith(repo, ".jl") "Repository name $repo does not end in .jl"
+        
         sha1_file = "METADATA/$pkg/versions/$(maxv)/sha1"
         @assert isfile(sha1_file)
     end


IainNZ:
Neat. While you are mucking around with these (and understand how they work), can you fix the tests so things like what happened here: https://github.com/JuliaLang/METADATA.jl/pull/2040 don't make it through?

jiahao:
The problem that caused #2040 is that because `DASSL/versions/0.0.4/sha1` did not exist, `Pkg` did not recognize `v"0.0.4"` as an available version of `DASSL`, and hence the test that `0.0.4/sha1` is a file never gets run. So problems with misnamed files have to be detected by exhaustively walking through all the files.

tkelman:
+1 for the uniformity. I'm a little surprised there are only 4 examples of registered packages where this isn't the case, I thought it was more.

jiahao:
cc @davidavdav @abeschneider @simulkade @dahlend

milktrader:
Yay!

tkelman:
Did you intend to merge this already, while there are still failure cases? Should we put in temporary exception cases for them, to avoid failing on other PR's that get submitted before those 4 packages can be fixed?

jiahao:
Oh, git.

I think the simplest thing is to turn the errors into warnings for now.

davidavdav:
OK, I renamed NamedArrays to NamedArrays.jl.  Does this have a consequence for METADATA.jl?

jiahao:
I think GitHub's URL redirection will handle older METADATA information.

ViralBShah:
Perhaps we can add a test to check that packages have a .jl extension and the right URL format.

jiahao:
@ViralBShah that is exactly what this PR does!

ViralBShah:
Duh - of course.

simulkade:
I renamed JFVM to JFVM.jl

kmsquire:
@simulkade, can you also update the URL in this repository and submit a pull request?

jiahao:
Ref #2068 #2070

SylvainCorlay:
Is there a technical reason why the repo URL should end with .jl ?

It is a CI check that appears not to be required by `Pkg` or the other tools and is also not documented.

I have a github organization for a library that has bindings for multiple languages (-python, -julia, etc). It seems odd to be requires to add .jl in the url.

PR #2058, opened 2015-01-30
Better error if package has no tagged versions

jiahao:
Currently if there are no tagged versions one would just get a `BoundsError`.


milktrader:
+1

PR #2059, opened 2015-01-30
Check version subdirectories

jiahao:
Traverses each `versions` subdirectory for each package

Ensures that:
- each subdirectory of `versions` is an available version
- the only allowed subdirectories must be semvers, i.e. `VersionName(dirname)` must succeed
- the only allowed files within those subdirectories are 'sha1' and 'requires'

Also found a misnamed file hanging around.


IainNZ:
Excellent!

milktrader:
Another +1

PR #2061, opened 2015-01-31
RFC: Checks most recent version's requirements

### LINKED ISSUE #1012: Test to ensure that METADATA REQUIRE is equal to the package REQUIRE.
It is easy to get the REQUIRE file from
`https://raw.githubusercontent.com/{{pkg.url}}/{{SHA1-hash}}/REQUIRE` (if the package is hosted on github). It would be nice if we could make inconcistencies trigger a travis error.

See: https://groups.google.com/forum/#!topic/julia-dev/9T6SF3L9YUc

jiahao:
For tagged version ver with hash sha1:

- Checks that version/$ver/requires has a tagged Julia version
- Checks that package repo's REQUIRE at hash sha1 has a tagged Julia version
- Checks that requires and REQUIRE agree

Note: the test now uses Requests to download the necessary REQUIRE files and
will install the Requests package automatically to do so.

Note 2: The test script does NOT error out when requirements are inconsistent;
there are way too many inconsistent packages!

Closes #1012

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -1,6 +1,12 @@
 const url_reg = r"^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?"
 const gh_path_reg_git=r"^/(.*)?/(.*)?.git$"
 
+#We don't have a mechanism for installing packages required for testing
+#purposes only when the repo being tested is METADATA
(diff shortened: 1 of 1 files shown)


aviks:
I'm wondering if making that many HTTP requests will scale on Travis?

jiahao:
Good question; I don't know. There will be many tiny requests.

jiahao:
Mostly superseded by #3582

PR #2062, opened 2015-01-31
linop: A Julia Linear Operator Package

dpo:
### Synopsis

Operators behave like matrices but are defined by their effect when applied to a vector. They can be transposed, conjugated, or combined with other operators cheaply. The costly operation is deferred until multiplied with a vector.

See https://github.com/dpo/linop.jl


dpo:
The build error seems to be related to something else:
````
ERROR: LoadError: assertion failed: Repository name NamedArrays does not end in .jl
````

jiahao:
Sorry, error was triggered by some recent changes to the test script. Should be rectified now.

In any case, we already have https://github.com/Jutho/LinearMaps.jl . Is there a way we can merge the functionality?

tkelman:
Temporary hiccup, should be fixed if you restart the build (I tried manually but Travis didn't like me). Could you rebase out the merge commits while you're at it?

dpo:
Seems like `Pkg.publish()` made a real mess. My commits are interspersed with Jiahao's. Not sure how to rebase in those conditions.

tkelman:
It's just the 2 so it's not so bad, but we've had a few examples recently of pull requests with dozens of merge commits since `Pkg.update()` automatically adds merge commits. This should be a Julia issue, but I'm wondering whether `Pkg.publish()`, `Pkg.register()`, `Pkg.tag()` etc should do their business in new branches to try and avoid this.

dpo:
I just tried creating a new pull request and it has even more merge commits...

tkelman:
My usual recourse there is doing `git checkout -b savemywork && git checkout metadata-v2 && git rebase -i origin/metadata-v2`, removing my local commits to at least get the local metadata-v2 branch back in line with the upstream one.

jiahao:
cc @Jutho

Jutho:
I agree that the functionality seems to be largely overlapping and would be in favor of extending LinearMaps.jl with any new functionality @dpo would like to add. I am also sure that some of the existingcode in LinearMaps.jl can be improved.

Note that anyhow, the package naming guidelines would suggest that this package is called LinearOperators.jl and that linear operators are endomorphisms and should therefore actually be restricted to square matrices:
http://en.m.wikipedia.org/wiki/Linear_map

dpo:
I agree that LinearOperators.jl would be a better name. But I don't rely on Wikipedia's "mathematics". In analysis linear operators between Banach spaces aren't restricted to endomorphisms. Also the LinearMaps code looks quite impenetrable to me, so I'd rather keep this a separate package, at least for now.

PR #2064, opened 2015-01-31
New packages must have at least one tagged version

jiahao:
A package with no tagged versions cannot be installed by `Pkg.add()`, so including them on the official package list seems to have limited value. I believe the question of registering a package with no tagged version has come up several times.

This PR ~~adds warnings to Travis for packages with no tagged versions.~~ enforces a new policy that new PRs to METADATA that register a new package must also tag at least one version of the package.

21 currently registered packages have no tagged versions and are made exempt from this new policy. They are:

- [ ] AuditoryFilters
- [ ] CorpusTools
- [ ] Elemental
- [ ] ErrorFreeTransforms
- [ ] Evapotranspiration
- [ ] GtkSourceWidget
- [ ] HiRedis
- [ ] KrylovMethods
- [ ] KyotoCabinet
- [ ] LatexPrint
- [ ] MachO
- [ ] MathLink
- [ ] Mimi
- [ ] ObjectiveC
- [ ] OffsetArrays
- [ ] Processing
- [ ] RaggedArrays
- [ ] RationalExtensions
- [ ] SignedDistanceFields
- [ ] SolveBio
- [ ] SortPerf

[@jiahao: updated 2015/09/30]

PR #2067, opened 2015-01-31
Tag DictFiles/ExtemelyRandomizedTrees/Shapemodels/InformedDE/LaTeX

rened:


aviks:
Could you possibly squash the two Latex commits? Not sure if they might cause issues with case sensitive file systems.

rened:
Sure! You mean like this?

aviks:
Yes, that is what I meant, thanks!

But looking at the commit I now understand what you are doing, and I am wondering, if someone on a Mac already has `Latex.jl` installed, and now runs a `Pkg.update()` will that cause problems? I'm wary because we've had issues in the past with case-renames on the mac. 

On the other hand, this is clearly the right thing to do, so maybe we should just do this and handle any support issues as and when they come in?

@IainNZ @jiahao  any thoughts?

jiahao:
I don't remember what had to be done in the past...

IainNZ:
Yeah something might break... but I think its probably best to just do it anyway.

kmsquire:
I think any existing users on case insensitive filesystems will have to remove the package entirely, and then reinstall. 

If true, I wonder if this will happen often enough that we should try to detect this situation and do that for the user...?

jiahao:
Looks like we should merge this and deal with the fallout.

jiahao:
Ref #2064

PR #2068, opened 2015-01-31
Added .jl to NamedArrays URL

davidavdav:
As requested and discussed in https://github.com/JuliaLang/METADATA.jl/pull/2057#issuecomment-72294827


jiahao:
Thanks!

PR #2071, opened 2015-02-01
Tag LightXML v0.1.10

tkelman:
adds `free` and `unlink` for `XMLElement` and `XMLNode`

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -36,7 +36,7 @@ for (pkg, versions) in Pkg.Read.available()
         #Ref: #2040
         verinfodir = "METADATA/$pkg/versions"
         for verdir in readdir(verinfodir)
-            if VersionNumber(verdir) in keys(versions)
+            if convert(VersionNumber, verdir) in keys(versions)
                for filename in readdir(verinfodir * "/" * verdir)
                    if !(filename=="sha1" || filename=="requires")
                        error("Unknown file $verinfodir/$verdir/$filename encountered. Valid filenames are 'sha1' and 'requires'.")
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,8 @@
-language: cpp
-compiler: 
-    - clang
+language: julia
+julia:
+    - nightly
 notifications:
     email: false
-before_install:
-    - sudo add-apt-repository ppa:staticfloat/julia-deps -y
-    - sudo add-apt-repository ppa:staticfloat/julianightlies -y
-    - sudo apt-get update -qq -y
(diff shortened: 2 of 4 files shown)


tkelman:
@jiahao what's up with your tests failing? On my local machine with Julia 0.4 the test script just hangs...

jiahao:
It's not "my tests" that are failing. Look at the history of how long it takes to run the METADATA tests; you can verify for yourself that most of the time is spent in `Pkg.Entry.check_metadata()`.

jiahao:
The only thing I can think of is that the new tests do traverse a lot of files (it has to walk practically the entire directory tree of METADATA). Maybe it's leaving too many open file handles? I don't see where it would happen.

tkelman:
You are right that I was just impatient with running the tests locally, they do pass fine there. But
```
/home/travis/build.sh: line 41:  6270 Killed                  julia METADATA/.test/METADATA.jl

The command "julia METADATA/.test/METADATA.jl" exited with 137.
```

I'm verifying now whether it still fails using an older version of `.test/METADATA.jl` from before your recent changes, which will tell us whether or not it's your additions causing the failure.

tkelman:
Not your fault, nevermind, sorry. Does happen with old version of the tests too - https://travis-ci.org/tkelman/METADATA.jl/builds/49050734

I suspect we might be hitting the Travis OOM killer. `Pkg.Entry.check_metadata()` takes a slightly obscene amount of memory, peaks at over 90% of my local machine's memory.

jiahao:
Yeah... the recently added tests may simply have pushed memory consumption over the edge to the point where we're seeing it more frequently.

tkelman:
It's all in `Resolve.sanity_check`. New tests taking some memory to begin with, plus running the tests with nightly which might mean new gc, it's apparently going way overboard. I'm trying with 0.3 to see if things are any better there.

tkelman:
Dunno why, but at least on my fork it seems slightly less likely to fail when using `language: julia` instead of the ppa.

tkelman:
Okay that was just Travis being mean and coincidentally passing on my fork, but failing here.

When I run with 0.3 though (requires one little syntax tweak), then the tests never take more than about 5% of my local machine's memory. I suspect this is a bug with the new gc or something else on 0.4.

jiahao:
Good sleuthing. I would be fine with running the METADATA tests with release-0.3.

tkelman:
Just did https://github.com/JuliaLang/julia/issues/9998

tkelman:
Okay to merge this then?

jiahao:
lgtm

PR #2072, opened 2015-02-01
Pull request/58d19923

eraviart:


tkelman:
Neat package, but the name's pretty vague. Traditional Julia package-name-bikeshedding would suggest either something more specific, or something more creative.

eraviart:
I just renamer the package to "Biryani", like its Python version. So forget this pull request. It is replaced by #2089

PR #2086, opened 2015-02-05
Register ALPS.jl

yomichi:


ViralBShah:
I think the package is pretty bare bones, with little docs and no tests. Perhaps we can add this to METADATA once it has seen some more development?

yomichi:
OK, I develop more and will send new PR later.
Thanks for comment!

ViralBShah:
Thank you. Looking forward to adding this with more readiness.

PR #2095, opened 2015-02-07
Register DeepNets.jl by @yarlett

yarlett:
DeepNets aims to provide a simple interface to allow people to specify "deep" stacks of neural network layers, and then train them on input-output data using gradient descent over various loss / error functions. The library is at an early stage but I am hoping that it is close to being useful to people working in the machine learning community.


IainNZ:
Hey @yarlett, thanks for submitting DeepNets.jl. Without tagging a version, no one is going to be able to `Pkg.add()` this so it might be best just to hold off until you feel you can tag a v0.1.

How would you say this compares/contrasts with @pluskid's Mocha.jl?

yarlett:
Hi @IainNZ, thanks for the response. Sounds reasonable; I'll get back to you when I've tagged a v0.1 and then we'll see where we're at.

Thanks for bringing Mocha.jl to my attention. I was not aware of that library, and it looks good. I'd say that in terms of (intended) functionality there's actually a lot of overlap between the 2 libraries, although Mocha is obviously a lot more mature and has a number of additional logging and DB niceties built into it (not to mention a cool coffee-based theme).

I don't know what Julia's policy is with regards to overlapping libraries, but obviously if you guys would rather not have overlaps in the system libraries then it probably wouldn't make sense to include DeepNets. I imagine in another week or so I'll be ready to tag a v0.1 and so maybe you can take a look then and let me know what you think? Thanks again!

IainNZ:
People in the community have different views I think on overlapping functionality. Usually its most acceptable when the two packages represent quite different takes on the same issue, which is more common on newer topics like this. Things like plotting, for example, are very opinionated so they too can naturally have duplicates. Something like linear regression though, probably doesn't need more than one.

My personal preference is to unify and join forces, especially given that its such early days for Julia still, but I know thats not always practical. Anyway, now you know about Mocha, maybe it'll influence your work over the next week or two - might push you in a different direction, or maybe it really is best to just add anything that is missing from Mocha.

ViralBShah:
It would be great to merge efforts, if the functionality are similar. Overlapping functionality is usually ok, if the approaches are completely different - but otherwise we prefer to have collaboration.

johnmyleswhite:
FWIW, I wouldn't worry much about overlap here. Deep learning is clearly a place where there's a lot of design decisions for which there aren't established best practices.

ViralBShah:
How about merge this now, and if there is collaboration, the packages can be merged later on.

IainNZ:
@yarlett suggested waiting a week until its 0.1 ready - lets just wait for that.

yarlett:
Hi again, all. Thanks for the comments. I have now tagged a v0.0.1 of DeepNets. It took a little longer than I had anticipated, so thanks for bearing with me, but I have the outline of what I hope turns out to be a somewhat useful and flexible API, even if performance isn't super stellar right now.

I'm a bit uncertain about whether I need to update the original pull request, or whether you have everything you need at this point, but do you think you would be able to pull DeepNets into METADATA.jl at this point?

johnmyleswhite:
Can we give it a name that doesn't make it seem like the definitive DeepNets package? Assuming that were done, I'd say this should get merged.

yarlett:
That should be fine. You'll have to give me a couple of days to think of a
new name though. I always find naming things the hardest part of
programming.

milktrader:
`DeepNetAlgorithms`  :bike: :house:

yarlett:
OK, I have renamed DeepNets.jl -> StackedNets.jl and made corresponding
changes in the codebase.

@johnmyleswhite, can you please advise wrt what I need to do now, if
anything, to get this merged? Thanks.

johnmyleswhite:
You'll need to update this PR so that you're submitting metadata for the renamed package. You need a new URL, a new directory in METADATA and a version subdirectory so that people can install StackedNets.jl

yarlett:
OK @johnmyleswhite, I've updated the PR request to reflect the repository/package name change. I inserted the version tag and SHA1 manually so let me know if you see a problem, but otherwise I'm hoping that we're in good shape for a merge! Cheers.

milktrader:
The tests *should* fail because there is no version tag

milktrader:
Build 5002 passed with no tagged versions. This is the test that I would've expected to have failed

````julia
 @assert length(versions) > 0 "Package $pkg has no tagged versions."
````
@jiahao

milktrader:
Ok, now I'm confused. Build 5003 failed __after__ a version was added ...

milktrader:
Ah, that explains it

yarlett:
Yep, sorry for the confusion, this is the first time I'm doing this.

milktrader:
Yeah, I think the idea behind the `Pkg*` family of methods is to avoid this hit-and-miss with manual updates. I think we're good now though.

milktrader:
Future updates to the code follows this progression in REPL

````julia
Pkg.update() # get the latest commit
Pkg.tag("StackedNets", :patch) # increments version by 0.0.1 (:minor increments by 0.1.0)
Pkg.publish()  # generates a PR
````
A browser window then pops up with a PR loaded and ready to submit.

yarlett:
That's very helpful, thanks. I think I tried something along those lines but it didn't work for me, possibly because I used Pkg.clone("StackedNets") to locally clone the package from Github, or maybe because I changed the repo name.

In any case I'll follow that flow next time round and see if it works...

milktrader:
Ok, I guess the discussion of merging has satisfied the main participants so I'll just merge this. Still unsettled that travis passed with no tagged versions.

milktrader:
You're not alone in running into trouble when first-time registering. There is a strong motivation to update a PR manually and just *fix* things. 

@StefanKarpinski I'm wondering if `Pkg.publish()` should enforce a tagged version.

jiahao:
@milktrader see #2064

jiahao:
In dbb7787 the package manager did not recognize any tagged versions and therefore none of the tests ran. I can add a test to see if `versions` were tagged but not recognized.

milktrader:
Okay, got it. It's a work in progress.

milktrader:
I'll not complain about more tests. In this edge-case it appears a `version` directory was created but not with an actual tagged version directory beneath it.

jiahao:
Please _do_ complain. Here we found an edge case not detected by the test.

PR #2108, opened 2015-02-08
Register RandomTestMatrices by @mschauer

mschauer:
https://github.com/mschauer/RandomTestMatrices.jl

A small module providing random test matrices. These are for testing and debugging purposes unlike the random matrix ensembles with known limiting distributions in RandomMatrices.jl

### DIFF
--- a/RandomTestMatrices/url
+++ b/RandomTestMatrices/url
@@ -1,0 +1,1 @@
+git://github.com/mschauer/RandomTestMatrices.jl.git
--- a/RandomTestMatrices/versions/0.0.1/sha1
+++ b/RandomTestMatrices/versions/0.0.1/sha1
@@ -1,0 +1,1 @@
+ec57ed1f0c7a4fdc1fd208208c9c1efa66ca9c41


jiahao:
I think all the functionality in this package is either already provided by RandomMatrices or can be incorporated without difficulty. Is there any reason not to use RandomMatrices instead other than a different preference for spelling and faster loading times? 

I admit that the spelling can be difficult: `randorth(d)` is equivalent to `RandomMatrices.rand(Haar(1), d)` but probably not many people would know that. In `RandomMatrices` I have tried to follow the `Distributions` nomenclature instead of writing a host of `rand*` functions (which I did in earlier versions).

What can we do to make these features more discoverable and easier to use?

cc @andreasnoack

mschauer:
I came up with half of the constructions ad hoc, so I did not expect them to have named distributions and did not consider it material for RandomMatrices. But before writing the documentation, I figured out that at least ``randorth`` and ``randposdef`` have famous distributions and ``randsym`` and ``randantisym`` are marginally standard normal or zero, so have rather clear distributions. Still, regarding the random stable and the random normal matrix I cannot say anything about their distribution.  But now, if you have no quarrels, then we can add ``Distribution``s for them add include this in the RandomMatrix package, we should at least be able to figure out their mean and marginal variance.

jiahao:
Always happy to have more random matrix distributions!

mschauer:
Great, currently I am also busy with my version of "issue 8839" but I then this should be straight forward to tackle.

PR #2110, opened 2015-02-08
Register FDAsrsf

jdtuck:


IainNZ:
Hi @jdtuck,

- Could you maybe check out the naming guidelines for packages: http://docs.julialang.org/en/latest/manual/packages/#guidelines-for-naming-a-package, I think it'd be worth making the package name longer to make it less of a letter stew.
- You'll likely want to tag a version

jdtuck:
I have changed the name to ElasticFDA, I don't want to make it too much longer as FDA is the standard in the statistics community for functional data analysis and I feel to expand makes it too long. I also tagged a version, thanks for noticing that. I will resubmit it if you feel the name works

IainNZ:
Seems pretty good to me, go for a resubmit.

jdtuck:
Thanks!

PR #2149, opened 2015-02-14
Tag Convex v0.0.4

karanveerm:
Main changes:

- added `matrix_frac` atom
- added `diagm` atom
- added `conv` (convolution) atom
- `norm(x, p)` supports all p >= 1
- made `sum_squares` more efficient
- updated documentation

cc @davidlizeng @madeleineudell


ViralBShah:
@karanveerm @madeleineudell You both should have commit access, so that you can merge METADATA PRs. Thanks!

PR #2152, opened 2015-02-14
Tag KDTrees v0.5.3

KristofferC:
Fixed the julia 0.2- now


IainNZ:
Please read http://docs.julialang.org/en/latest/manual/packages/#man-package-requirements - I don't think you want to support Julia 0.2 prerelease

KristofferC:
You are right, I will fix it.

PR #2175, opened 2015-02-19
Register JuMPeR: v0.1.0

IainNZ:
Happily to *finally* release JuMPeR, an extension of JuMP to robust optimization. First MIT class using it is tomorrow, so that was motivation to get it done :D

@mlubin @vgupta1 @joehuchette


mlubin:
:100:

StefanKarpinski:
Does robust mean "resistant to outliers"? How does that work in the context of optimization?

IainNZ:
Indeed. So say you have a linear constraint:
```
a^T x <= b
```
and you aren't too sure on what `a` is precisely. One school of thought would be, put a probability distribution on `a` and turn it into something like
```
P(a^T x <= b) >= 0.95
```
Thats supported by `JuMPChance.jl`.

The robust optimization view says, let `a` live in some set `U`, and then say that
```
a^T x <= b  for all  a in U
```
JuMPeR lets you model these sorts of problems, and handles the transformation from this infinite-sized problem to a finite size problem.

PR #2180, opened 2015-02-20
Pull request/274652ed

joshday:
Add package `AverageShiftedHistograms` for David Scott's ASH estimator.


joshday:
The Travis CI builds [attached](https://travis-ci.org/joshday/AverageShiftedHistograms.jl) to my repo are building fine.  I don't understand the details above as to what is failing.

milktrader:
I restarted the build, let's give it a sec. It appears there wasn't a `.jl` associated with a repo named that didn't match (even close) your repo.

milktrader:
@joshday there are yor repo's tests and then the travis build for METADATA.jl given your pull request. The second one is the one that failed. We just included some tests to make sure certain things were in place for a PR, such as the `.jl` extension, an existing url, etc.

milktrader:
Hmm, well there it is again.

````bash
WARNING: Repository name PEGParser does not end in .jl
````

milktrader:
iirc, there were a few packages that had been in METADATA prior to the new test for `.jl` and I'm thinking `PEGParser` is one of them. The solution short term (again, iirc), was to provide exceptions to those packages while the authors were contacted. Not sure why it caught your PR the first time but then the second time when I re-ran it the error was still there but the build passed. 

In any case, your package is not causing any trouble. 

Thanks!

joshday:
Good deal!  Thanks for the help.

PR #2190, opened 2015-02-23
Tag Bokeh v0.1.0

samuelcolvin:
I believe this is all I need, let me know if there's anything else I need.


milktrader:
Issues are turned off at Bokeh.jl and I cannot get any example to render in Safari.

samuelcolvin:
Sorry to hear your having trouble with Safari.

I've just checked and issues are not turned off in Bokeh.jl, please submit an issue there: https://github.com/bokeh/Bokeh.jl/issues/new and I'll look into it as soon as I can.

Would be great if you could provide details on the version of safai and webkit (I know for example that windows safari uses every old webkit). Also do you get the same ues with the standard python Bokeh package?

milktrader:
Thanks, I guess I was the first issue so that's why it *appeared* turned off. This package looks exciting so I can't wait to get it running. 

Filed issue at Bokeh.jl

PR #2205, opened 2015-02-24
Register FunctionalDataUtils and tag v0.0.1 +  v0.0.2

rened:


FunctionalDataUtils/versions/0.0.2/requires
```diff
@@ -0,0 +1,6 @@
+SHA
+FactCheck
```
IainNZ: For future reference, you can probably just put this in your `test/REQUIRES` file instead of your main one

rened: You are right, absolutely - just in this case I need it in the main REQUIRE because the package provides some convenience wrappers around FactCheck functionality.

PR #2206, opened 2015-02-25
Pull request/a889282b

timothyrenner:
Registering ColorBrewer package.


IainNZ:
Hey @timothyrenner, I just saw this issue https://github.com/JuliaLang/Color.jl/issues/77 and I was curious why this wasn't added to `Color.jl`?

IainNZ:
In fact some of it already seems to be in there, and theres even a file that seems to depend on this package.

timothyrenner:
I didn't add it to `Color.jl` because in my original search for colorbrewer palettes in Julia I noticed they've been discussed in various places (dcjones/Gadfly.jl#21, JuliaLang/julia#2278, nolta/Winston.jl#105). In these cases the authors preferred algorithmic solutions. At that point I assumed it would be easier to just write my own.

The file in `Color.jl` that requires this package was written by @ikirill to generate the colormaps for the READMEs. It's present in `ColorBrewer.jl` and `Color.jl`.

IainNZ:
OK, I see. I'm just bugging you about it because we have a preference wherever possible to reduce duplication. It definitely seems like a difference of perspective then, I didn't realize before you actually have the maps stored as data. CC @timholy @dcjones @jiahao, the people I think of when I think of colors.

I won't merge this just yet, because I think you probably should tag a version (`v"0.1"`) so people can `Pkg.add` it.

dcjones:
This looks good. Ultimately it would be nice to unify this with the palette code in Color, possibly splitting it into a separate ColorPalettes package.

timothyrenner:
@dcjones I agree - something like `brewer("Greens", 5)` for a colorbrewer palette would be a nice way to integrate it with algorithmic palettes.

@IainNZ I'm going to tag shortly.

PR #2207, opened 2015-02-25
Run version check on all subdirectories of METADATA

jiahao:
This PR addresses an edge case observed in #2095 in which a package was
registered and the user attempted to register versions manually but
placed sha1 under `"METADATA/$pkg/versions"` instead of
`"METADATA/$pkg/versions/v0.01"`.

The version check test introduced by #2040 is now placed in a new loop
that runs for each subdirectory of METADATA, regardless of whether each
subdirectory corresponds to a package entry recognized by `Pkg.Read`.
Doing so allows the tests to detect package entries that are
misconfigured and hence not recognized as such by the package manager.

Additionally, `joinpath` is used consistently instead of manually splicing
strings together.

Furthermore, as PEGParser is the only package left without a .jl
extension in #2057, the .jl extension check is now upgraded to an
assertion error except for PEGParser, which is grandfathered in.

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -21,7 +21,7 @@ for (pkg, versions) in Pkg.Read.available()
 
         for (ver, avail) in versions
             #Check that all sha1 files have the correct version hashes
-            sha1_file = "METADATA/$pkg/versions/$ver/sha1"
+            sha1_file = joinpath("METADATA", pkg, "versions", string(ver), "sha1")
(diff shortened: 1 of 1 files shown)


jiahao:
I ran this new test against dbb7787 and verified that it does correctly fail the test with error

```
ERROR: LoadError: Invalid version number requires found in METADATA/StackedNets/versions
```

milktrader:
Nice work!  Thanks @jiahao

PR #2212, opened 2015-02-25
Add LRUCache to metadata

jcrist:


IainNZ:
@jcrist if you don't tag a version, no one will be able to `Pkg.add` it. Also I notice in your package that there is no `REQUIRE` file containing `julia 0.4-`, might want to add it before tagging.

jcrist:
Woops, I wasn't sure if I had to do that or not. Tagged, and added a REQUIRE file.

PR #2236, opened 2015-03-02
Register IndexedArrays: v0.0.1

garrison:
As announced at https://groups.google.com/d/msg/julia-users/bMftxcvtVbA/Vr_Hjf3WLOUJ


IainNZ:
Can we bikeshed the name a bit?! Aren't all `Arrays` `Indexed`?

garrison:
Another name I considered was `ReverseIndexedArrays`.  Do you have any other ideas?  The word "unique" also comes to mind, as all elements are required to be unique.  It's also a bidirectional mapping (or bijection) between 1..N and some set of elements, but I'm having a hard time condensing that concept to a name.

garrison:
Or perhaps the word `Set` should be in there somewhere.  `IndexedSet` could work, but the type does not implement the usual `Set` operations, such as union, intersection, complement; it actually implements more of the array methods.

garrison:
Per discussion here and at https://github.com/garrison/IndexedArrays.jl/issues/4, I plan to rename this package `UniqueVectors`.

PR #2254, opened 2015-03-06
Tag TimeSeries v0.5.6

milktrader:
Adds support for non-dot mathematical operations between two TimeArrays and between floats, integers and TimeArray, except for `/`, which has confusing semantics and has been excluded in Base.


milktrader:
So where is the travis button?

milktrader:
Well, I don't see a travis test being done so I'm just going to merge this now since there is a significant change to the API and the docs already assume this change.

PR #2255, opened 2015-03-06
Tag MatrixDepot v0.2.0

weijianzhang:
Main change:

Add a Julia interface to the UF sparse matrix collection


jiahao:
:+1:

jiahao:
Failure is caused by Travis timeout.

weijianzhang:
@jiahao probably due to downloading data from the web. I will change that soon.

PR #2271, opened 2015-03-11
Register OptimPack

emmt:
OptimPack.jl is a Julia package to solve large scale optimization problems (with millions or even billions of variables) possibly with constraints.

### DIFF
--- a/OptimPack/url
+++ b/OptimPack/url
@@ -1,0 +1,1 @@
+git://github.com/emmt/OptimPack.jl.git


IainNZ:
Hi @emmt 
Looks like interesting work, would be interested to see benchmarks vs NLOpt and Optim.jl one day.
Do you want to tag a version so people can easily install it?

emmt:
Le 11/03/2015 17:24, Iain Dunning a écrit :
> Hi @emmt <https://github.com/emmt>
> Looks like interesting work, would be interested to see benchmarks vs
> NLOpt and Optim.jl one day.
> Do you want to tag a version so people can easily install it?
>

Yes the idea is of course to compare to other similar packages. Some 
colleagues reported that OptimPack compared favorably but more tests are 
needed. The OptimPack.jl is based on the C library OptimPack which has 
been successfully used to solve very large image reconstruction problems 
and has been found to be at least as fast as other alternatives.

I have tagged a version (sorry, 3 in fact because I've done a mistake 
with git and fixed a minor bug in the mean time).

Thanks!

PR #2302, opened 2015-03-14
Register KShiftsClustering and QuickShiftClustering

rened:
Depending on https://github.com/JuliaStats/Clustering.jl/issues/40 it can still be included there later on.


ViralBShah:
Registering for now, but hopeful that this can become part of the Clustering package.

PR #2308, opened 2015-03-17
CRC32: remove package

fhs:
There are faster implementations in Zlib.jl and CRC.jl.
No other package depends on this package.

Fixes https://github.com/fhs/CRC32.jl/issues/3


IainNZ:
Packages can't be removed (yet), so I'm closing this to prevent any accidental merges.

For one thing
`ERROR: package ZipFile v0.1.0 requires a non-registered package: CRC32`

Here is how to deprecate a package:
- In the package REQUIREs, set a maximum version of 0.4
- Tag a new release and final release.
- Retroactively change the METADATA entries for past releases to also have a maximum version of 0.4
- Then you're done!

IainNZ:
I see there is nothing in the REQUIRE right now, so I'd put the following in there: `julia 0.2 0.4`

PR #2321, opened 2015-03-19
Tag Homebrew v0.1.14

jakebolewski:


tkelman:
Can you please push the tags to JuliaLang/Homebrew.jl too? Or start using `Pkg.publish` again, it's not slow any more on 0.4.

PR #2329, opened 2015-03-20
Package updates and URL transfer

bearnado:
In these packages I've corrected depreciated logic.

In NMEA - swapped out my try_parse function for the newly implemented tryparse function in Julia 0.4.  I also made NMEA type fields Nullable as I kept running into NMEA message streams that didn't report certain fields on various GPS receivers.

I also transferred ownership from my organization (Templar Labs) to myself because I'm a selfish selfish being.

Thanks!


jiahao:
Tests timed out but updates look fine.

PR #2330, opened 2015-03-20
Register CUBLAS

denizyuret:


IainNZ:
@denizyuret maybe tag a version?

PR #2338, opened 2015-03-23
Tag ExtremelyRandomizedTrees v0.0.2

rened:


ViralBShah:
Just curious how this is different from DecisionTrees. Love the `Extemely` part!

rened:
The difference that there is no optimization of the split threshold for each feature. The thresholds are complete random as well. Works just as well, with much faster training. Tradeoff: you need a larger number of trees for the same performance, i.e. you have slightly larger prediction times.
I borked 0.0.2, will submit 0.0.3 shortly...

ViralBShah:
Thanks. Also, you have commit access too - so should be easy to merge the new version quickly.

PR #2341, opened 2015-03-23
add TextMining.jl to metadata

mtabor150:


IainNZ:
Hey @mtabor150 , this looks like it'll be a great package.
I'm a little worried about the name - its very sweeping. Not the worst thing though. Any thoughts?
Sorta related, looking at the README, this package seems to have very wide scope, including clustering and naive bayes. There are already packages for clustering and naive bayes, and it'd seem to me to be a better idea to utilize those.

IainNZ:
Also CCing @Kevin-Damazyn who seems to be a big contributor to this, and maybe @ninjin and @johnmyleswhite as they are maybe also familiar with the area.

ninjin:
Giving it a quick read through it seems like a fairly generic library to work with text and some machine learning methods.  If it is intended to be some sort of mega library along with the lines of [NLTK](http://www.nltk.org/), the name would be fine.  But given the current scope it would most likely both have to cover text processing **and** machine learning, which would be quite the effort.  Personally, I think there are far too many ways to make a library like this and would vote for a name that is not already used to describe a sub-field in academia, like for example [Mocha](https://github.com/pluskid/Mocha.jl).

mtabor150:
The package is being developed to be used specifically by Digital Humanists and English academics at Saint Louis University.  We opted not to utilize the existing packages to provided greater control to those involved at the University, and to provide experience to the students developing the tools in machine learning and Julia.  The long term intent is for the SLU Computer Science department to maintain and expand the package based on research needs.  Anything to add @kscanne ?

IainNZ:
@mtabor150 given that, it might not make sense to register in METADATA, or at least if it is it should be given a more idiosyncratic name. Packages in METADATA are, in spirit if it not always in reality, intended to be generic and useful to the wider community at large. Perhaps the users could do a one-time `Pkg.clone("$theurl")` instead of `Pkg.add`?

ViralBShah:
You never know who will use the code once released, even though that wasn't the original intention. How about calling it SLUTextMining for now?

mtabor150:
We definitely want the package to be generic and useful to communities outside of SLU, and we want to encourage collaboration from outside parties.  That being said, I don't think anyone will have an issue with giving the package a more idiosyncratic name like SLUTextMining.  I'll discuss this with the team and resubmit a pull request.

StefanKarpinski:
I would just point out that the first four letters of the proposed package name are slightly unfortunate.

IainNZ:
Thanks @mtabor150 - I always hate to discourage someone registering a package in any way, thank you for your patience!

mtabor150:
@IainNZ it's no problem.  We chose TextMining as the package name simply to try to adhere to the [JuliaLang docs](http://docs.julialang.org/en/release-0.3/manual/packages/#creating-a-new-package).  Also, we are intending on producing a lot of supporting functionality around the Feature Space Model we are developing for Julia. Would it be preferable for you if we split these up into separate packages (SLUFeatureSpace, SLUClustering, SLUClassification, etc.), or to keep it as a mega library?

IainNZ:
My true preference, and I feel silly even talking about it because its basically telling you how to spend your own precious time, is that you contribute to the existing clustering and classification packages where needed, and only make packages for new functionality. Theres some strong reasoning behind this:
* limited resources are focussed
* more eyes -> less bugs, better maintenance
* users know where to go to get what they need
* improves in one package propagate to all dependent packages.

That aside, and more to your question, is that the way METADATA has been evolving would not really seem to suit having multiple packages in a "namespace" like that, but I can see the need, and in general it does make a lot more sense to split it up (to me at least).

ViralBShah:
Yes, I think splitting up does make more sense - as it may likely encourage collaboration/merging with other packages.

StefanKarpinski:
I kind of disagree since I think that overly granular packages often end up like the web stack where there are five or so packages, each having about one source file and so interdependent that you can't use one without using the others anyway.

quinnj:
See the discussion [here](https://github.com/JuliaStats/Roadmap.jl/issues/11) for a somewhat lengthy discussion on the machine learning landscape.

I'd say since @mtabor150 and co. will be managing/maintaining this functionality, so do what would work best for them. Whether that's packing everything in one package, or splitting into more specific packages with a "meta" package that brings them all in.

On the other hand, if there are more general methods (clustering, etc.) that are not text-mining focused, certainly look into contributing/enhancing other packages.

ViralBShah:
+1 to doing what's most convenient for maintaining the codebase and for the users. I just felt that something like clustering and classification is worth splitting out because those may merge with existing packages later and have an independent life.

PR #2351, opened 2015-03-26
Tag BoundingBoxes v0.1.0

sjkelly:
Fixes issues in my prior commit.


IainNZ:
Looks like Travis timed out due to no output - maybe it should print something?

sjkelly:
The remote changed for this package. How should I update it?

garborg:
I believe I've always just changed it [here](https://github.com/JuliaLang/METADATA.jl/blob/7ba318dd1131afa01ed750f995ad737d616b39e0/BoundingBoxes/url) the next tag after the remote changed.

sjkelly:
I tried it here: https://github.com/JuliaLang/METADATA.jl/pull/2352

garborg:
Missed that -- then I'd guess the timeout there isn't the fault of your PR.

PR #2352, opened 2015-03-26
BoundingBoxes moved to voxel8

sjkelly:


sjkelly:
@IainNZ this should fix the issue.

PR #2362, opened 2015-03-28
Test: diagnostics output during each stage of testing

### LINKED ISSUE #2355: Lots of recent timeouts
Is there something we can do to increase the 10 minute timeout and/or make Pkg run faster?

jiahao:
Hopefully will avoid Travis 10 min timeout and fix #2355

Also removes special-casing of legacy packages in #2057

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -29,11 +29,9 @@ for (pkg, versions) in Pkg.Read.available()
             @assert sha1fromfile == avail.sha1
         end
 
-        #TODO Replace warnings with assertions below once packages in Issue #2057 have been addressed.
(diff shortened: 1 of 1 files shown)


jiahao:
It appears that `Pkg.Entry.check_metadata` by itself takes dangerously close to 10 min already. So this is pretty much the best we can do without surgery on `Pkg`.

https://travis-ci.org/JuliaLang/METADATA.jl/builds/56258815#L114

tkelman:
Oh, right, that function was majorly sped up on 0.4, so this is partially my bad for suggesting we run these tests with Julia release. Running with nightly should be faster, but runs much more risk of stability issues from base Julia.

PR #2369, opened 2015-03-30
Register TrainingImages

juliohm:


kmsquire:
Since these are for geostastical simulation, perhaps it would be good to choose a more specific name.  Maybe GeoTrainingImages.jl or GeoStatImages.jl?

juliohm:
@kmsquire , I was trying to come up with a more specific name, but none sounded right to me at the time. `GeoStatsImages.jl` seems like a good one. Is there any way to change the name of the repository on Github without major impacts? I see this option under Settings in the project where I can rename the repo, have you ever tried it before?

kmsquire:
Yep, the rename option under Settings should work fine--in fact, the old repo will redirect to the new one. I would rename it there, re-clone, and rename `src/TrainingImages.jl` to `src/GeoStatsImages.jl`.  I think that'll be all you need.

juliohm:
Will do, thanks.

PR #2381, opened 2015-04-01
Pull request/abda235f

tawheeler:
New package Discretizers.jl
Supports encoding / decoding data to and from bins.
Support for discretization algorithms.


ViralBShah:
Perhaps tag a release?

PR #2383, opened 2015-04-01
Tag UUID v0.0.4

WestleyArgentum:
I think there's one minor thing keeping UUID from being 0.3 and up -- no time atm though


WestleyArgentum:
I believe the failure here is related to https://github.com/JuliaLang/METADATA.jl/issues/2355

Maybe that needs to be reopened? This build got stuck on the `Verifying METADATA...` step, I guess maybe we could add some verbose flag there that prints out the name of each package as it check it or something?

PR #2391, opened 2015-04-02
Run METADATA test on Julia nightly

jiahao:
The Travis status is now completely unusable on Julia release, since the internal consistency check reliably times out Travis, exceeding 10 min with no output.

This PR runs the tests on Julia nightly, which has a speedier Pkg.Entry.check_metadata.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: julia
 julia:
-    - release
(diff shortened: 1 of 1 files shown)


jiahao:
9.8 seconds vs 600+ seconds...

PR #2395, opened 2015-04-02
Tag GnuTLS v0.0.4

timholy:
I have nothing to do with this package, but I noticed that submission of Coverage results is broken on 0.4. It looks like this will be fixed just by tagging a new version.


quinnj:
Yeah, we need to tag most of the JuliaWeb repos; I went over and did a bunch of 0.4 compatibility fixes, but I'm just on master for all of them.

timholy:
I noticed your commit, and I appreciate the compatibility fixes. (I've spent my entire morning doing the same for other packages.) Is there any reason not to merge this---i.e., do you know of any bugs, etc?

Issue #347, opened 2013-08-07
Clarification/Documentation of policy for deprecating packages


kmsquire:
There are two two issues I would like to see clarified:

~~1) sha1 bumping should (almost) always accompany a new version (with exception of version 0.0.0).  (Does current policy allow bumping of any x.x.0 version?  Should this only be in effect for released versions of Julia?)~~

~~Rationale: no matter how careful one is, some new version will break on someone's system, and it's important to allow the user to revert to a previous version.  Without bumping the version, this is hard for the naive user.~~

2) Old versions of packages should (almost) never removed from METADATA.jl

Rationale: Removing a package version from METADATA.jl causes problems when 1) a package depends on a removed version of a package, or 2) a user wants to replicate some results produced with a particular version of a package, and that version is removed.

---

My impression is that these are policy, or at least should be, but I don't think they're documented, and both have been violated

Any suggestions/objections/refinements?  Should I file an additional issue for Pkg2 to support updating packages?

jiahao:
I think what happened in #2346 is an extreme example where the right thing to do really is to scrub the entirre record from METADATA.

hayd:
Is METADATA supporting 0.2 (or -) users indefinitely??

Perhaps similar to @IainNZ's great work on the pkg tests, there could be a package health heuristic for automatically putting "unhealthy" packages up for deprecation consideration. E.g.

- recent release
- tests pass (supports most recent julia release)
- has documentation
- recent git activity
- latest git issue(s) responded to

If these were all/mostly bad then first ping the maintainers to fix things, and if the health does not improve then flag for the community to consider removal.

*Some packages may fail some of the above but needn't be removed, e.g. they are important or failing X is no big deal, or it may be they can be moved (e.g. to a julia org).*

wildart:
As issue [PkgDev/#21](https://github.com/JuliaLang/PkgDev.jl/issues/21) suggested, there should be some functionality to deregister/unmaintain/unlist registered packages.

I propose to add `unlist` empty file to the root of a package folder which would indicate that the particular package should not be installed/used. If developer will decide to resume development this file will be deleted.

All necessary functionality will be provided by `PkgDev`.

IainNZ:
Whats wrong with the current approach?

wildart:
Zeroing version disables only a particular version of package on which this operation was performed. If package has multiple versions, then other versions can be accessed, unless you set requirements to zero in every version.

IainNZ:
[...]

PR #2416, opened 2015-04-10
Register IQSIM

juliohm:


jiahao:
See http://julia.readthedocs.org/en/latest/manual/packages/#guidelines-for-naming-a-package

How about calling it ImageQuilting?

juliohm:
Mainly due to historical reasons. Previous algorithms developed in the
field follow this naming convention: ccsim, sgsim, ...

Also I want to emphasize that this package is for geostatistical
simulation, not only image quilting as in my previous package
ImageQuilting.jl.

juliohm:
Could you please merge the pull request?

jiahao:
I don't think the name 'IQSIM' is sufficiently meaningful to qualify as a package name.

Again, see http://julia.readthedocs.org/en/latest/manual/packages/#guidelines-for-naming-a-package

> Package names should be sensible to most Julia users, even to those who are not domain experts.

juliohm:
Such naming convention rules shouldn't be applied to all packages,
specially those that are meant for domain experts. Probably less than 10
people in the world are interested in this package.

Naming it differently to a more verbose sentence will just make researchers
in the field confused and will also take the new name out of the list of
available names to choose from in the future.

The other package ImageQuilting.jl that I registered months ago is more
useful to the general scientific community and as so deserves a clear name.
This new package unfortunately doesn't have a clear short mnemonic name.

Could you please reconsider it?

tkelman:
> Probably less than 10 people in the world are interested in this package.

Then does it really need to be registered? It can still be easily installed via `Pkg.clone` with the url.

juliohm:
Registration means having versions in this case.

juliohm:
This package won't be merged because of its name? I'm about to present the
code in a conference, please tell me if it won't be merged.

Keno:
Nobody said anything about not merging this package. Since METADATA is a global namespace we have to make sure that names are appropriate and recognizable. That is a normal part of any pull request. I suspect we may end up merging it with the current name, even if it's not descriptive. This has happened for a few other packages, but all of those had discussion.

dpsanders:
> Registration means having versions in this case.

You can still tag versions in your personal repository even if the package is not registered. Is that what you meant by "versions"?

I agree that the name IQSIM is too non-descriptive. Googling "IQSIM" and "IQSIM image quilting" don't seem to throw up anything.

StefanKarpinski:
The description of the package is "3D image quilting simulation"... so it seems pretty specific to image quilting. What's the difference between this package and the other registered ImageQuilting package?

juliohm:
This package is about Geostatistics. It's about generating samples that
reproduce the patterns in an image and honor Q-Q plots, trends, etc. It
fits in the statistics organization.

The previous package is simply a 2D draft on how to paste images together
without loosing texture. It can be considered an image processing algorithm.

dpsanders:
So how about the package name `GeostatisticsIQSIM`?

juliohm:
I don't think the change makes the name clearer.

I have plans to deprecate this package in the future and create a whole set
of geostatistical algorithms in Julia (GeoStats.jl). Don't have the time to
do it now because of other priorities.

Think of this package as an intermediate step to a larger contribution from
the geostats community. I'm trying to bring them to start using Julia.

jiahao:
I don't think it is a good idea to register a package, knowing full well that it will be deprecated in the future. I'm sorry you feel that our rules are annoying, but one of the reasons why we are so strict about package names now is because we don't have a good mechanism for handling deprecations and renaming (#347).

juliohm:
So it won't be merged?

jiahao:
Since you plan to have a more comprehensive package planned, how about registering this package as `GeoStats.jl` and provide IQSIM as a function within that package?

juliohm:
I was trying to avoid this responsibility of maintaining a more
comprehensive package at this moment. I will rename it to GeoStats.jl by
tomorrow if that is the only way to get the code merged.

tkelman:
I wouldn't consider it a whole lot of responsibility to claim a recognizable name for the initial package, as long as you don't disappear completely. If someone else comes along in the future who's interested in geostatistics and wants to see more functionality in the package than what you're able to write by then, then you can add them as a collaborator and figure out how to jointly make the most effective progress. That's what has happened informally with various organizations like JuliaOpt, JuliaStats, JuliaQuantum, JuliaWeb, etc which is the closest thing we've got at the moment to namespacing.

StefanKarpinski:
I don't really get it. This package is only of interest to ten people in the world and is already planned to be deprecated. Why does it make sense to have it in the metadata registry? How is not having it registered preventing you from doing anything?

juliohm:
Hi Stefan, I would like the package to be part of the Julia package
ecosystem and also take advantage of versions.

I'll just rename it to GeoStats.jl

ViralBShah:
Fantastic. Closing this one then.

juliohm:
Reconsidering all the discussion, I'll just release a new version of the previous ImageQuilting.jl package instead of creating a new repository.

PR #2422, opened 2015-04-11
Register SFML

zyedidia:


IainNZ:
Oh neat! I used to use SFML quite a bit. Why did you wrap the C++ interface and not CSFML?

zyedidia:
I did wrap CSFML but it is functionally equivalent to SFML and SFML is better known so I named it that.

IainNZ:
Oh! I was wondering why there was CSFML code and a dynamic library included. Maybe we can talk about it on issues page.
Anyway, is there any point to registering it without a tagged version?

zyedidia:
I wasn't sure it was "official" yet but I am happy to tag it now. Should I do that and then send another pull request? Or should I do it after the current pull request gets accepted?

dpsanders:
Could the package be renamed to something that is more intelligible for the
non-expert?

zyedidia:
I named it SFML.jl because in the Julia manual it says that packages "that wrap external libraries or programs should be named after those libraries or programs," for the purposes of easy Google searching. And they give the example of CPLEX.jl being a wrapper for the CPLEX library.

IainNZ:
Yeah the name absoultely has to be SFML. I'll merge this and then you can tag whenever you are ready.

PR #2440, opened 2015-04-17
Pull request/a840ec2e

njwilson23:
This adds a package that wraps the TEOS-10 routines for the Gibbs Sea Water equation of state. So far, now the package "works" and passes the GSW test suite.


milktrader:
Cool, how about tagging a version so it can be added with `Pkg.add`.

milktrader:
I missed that you had the versions already.

njwilson23:
I tried this in #2442 - I'm not sure that's quite right.

milktrader:
Yep, works

PR #2441, opened 2015-04-17
Tag BinDeps v0.3.10

ihnorton:
ref https://github.com/JuliaLang/BinDeps.jl/pull/140


ihnorton:
Seems to work properly on 0.3, and on 0.4 with https://github.com/JuliaLang/julia/pull/10832 applied.

PR #2442, opened 2015-04-17
Pull request/a2df63d9

njwilson23:
I think this tags a version (0.0.1).


milktrader:
@njwilson23 when you have new versions to tag, it would be better to run `Pkg.update()` first and then run `Pkg.tag("GibbsSeaWater", :patch) to avoid the `Merge branch 'metadata-v2'` cruft.

njwilson23:
Got it, thanks.

PR #2455, opened 2015-04-18
Register Recsys: v0.0.1, v0.0.2

filipebraida:
Recommender System Framework.


IainNZ:
Hey @filipebraida, interesting package. Naming-wise, I guess there would be multiple approaches to this so I suppose this falls under the "idiosyncratic names" category. Will wait for a possible dissenting opinion before merging.

filipebraida:
Thanks @IainNZ! Today the framework only works with collaborative filtering, but I would like to work other approaches.

milktrader:
I was thinking `Recommend` or `Recommender` since `Recsys` is ambiguous and I think first of Recreational.

milktrader:
So @filipebraida have you considered possibly changing the name from `RecSys`?

filipebraida:
Yes! The name could be CF.jl or ColaborativeFiltering.jl. What do you think? @milktrader

milktrader:
I like CollaborativeFiltering much, much better. CF is too terse and could mean just about anything. I'm not a big fan of acronyms but there is a place for them when the package is a wrapper for a library that shares the acronym.

milktrader:
While we're at it, it would be good to add `julia 0.3` to your REQUIRE file. This is newly documented. Also, this assumes that you support Julia `v0.3`. If you only support Julia `v0.4-dev` then the correct line in REQUIRE would be `julia 0.4-`.

filipebraida:
@milktrader I'll change the REQUIRE file and I'll add julia 0.3. Thanks!!!

What do you think about CollaborativeFiltering.jl? @IainNZ

IainNZ:
Sounds good to me. Is this a project you intend to be working on and maintaining @filipebraida ? If this is research code for experimenting with ideas, you should not register it - people can still install it from your Github!

filipebraida:
@IainNZ  in fact it is both. I think It's better to separate. Most algorithms are baselines. I would like to continue the maintenance and documentation, but I'm enrolled in a PhD. So, I don't have time to do this now. What do you think?

IainNZ:
@filipebraida I'm also in a PhD, so I know the time pressure. Maybe advertise for collaborators on julia-users? I'm sure many are interested in the area. I'll close for now, feel free to reopen with a new name.

PR #2473, opened 2015-04-20
Register SystemImageBuilder: v0.0.1

rened:


tkelman:
Please don't fork code from base and maintain separate modified copies of it out in a package. I don't see a problem with making wrappers around the existing scripts in contrib, but if there are problems with them and you need to make changes, please submit issues and PR's against the copies in base.

rened:
yes, this was more of a test until I got it working - I'll reintegrate it into master + ask for backport and adapt this package



Am 20.04.2015 um 15:55 schrieb Tony Kelman <notifications@github.com>:

PR #2475, opened 2015-04-20
Register SystemImageBuilder

rened:
This time without duplication of code from julia/contrib. Thanks @tkelman for your help!


tkelman:
Cool, that makes more sense, now it's just convenience wrappers.

PR #2477, opened 2015-04-20
Register AbstractDomains

zenna:


milktrader:
How about doing

````julia
Pkg.update()
Pkg.tag("AbstractDomains", :patch)
````
Otherwise looks interesting. Is this mostly theoretical or are there some applications for it?

zenna:
I've tagged it now.  I'm new to this, does I need to create a new pull request using `Pkg.publish()`?

Abstract domains are used in abstract interpretation which can be used for static analysis of programs.  For instance, we can tell if a some behaviour of a program is possible in all possible executions of the program.  This in contrast to testing, where you try a few inputs to your program and see if it breaks.

Also Abstract Interpretation allows you to do probabilistic inference.  All of the code in AbstractDomains is from my probabilistic programming [Sigma.jl](https://github.com/zenna/Sigma.jl).  I'm moving it out of Sigma because it's useful in its own right.  But check out there if you want more practical examples of what abstract domains can enable.

milktrader:
Yes.

Probably need to do another Pkg.update()

````julia
Pkg.update()
Pkg.publish()
````
I'm actually unsure if this updates this PR or creates a new one so I'd like to find out :smile:

milktrader:
Closing this since it's merged now in a separate PR.

PR #2490, opened 2015-04-23
Tag Jacobi v0.1.0

pjabardo:
New functionality, improved documentation and many more tests.


dpsanders:
Given the large number of things named after Jacobi, I had no idea what this package was about from the name `Jacobi.jl`. Perhaps it could be changed to e.g. `JacobiPolynomials.jl`?

tkelman:
It's probably fine for now, but it would be great if in the future this could be consolidated in a larger package that provides more families of orthogonal polynomials. There is https://github.com/daviddelaat/Orthopolys.jl but it hasn't been touched in some time.

dpsanders:
Yes, good idea, with the name `OrthogonalPolynomials.jl`...

pjabardo:
I can change the name but it would be better to split it up in two packages, one for orthogonal polynomials and the other for Quadrature/derivative related stuff

PR #2492, opened 2015-04-23
tag Compat.jl 0.4.2

tkelman:
I'm not 100% sure what the consequences of removing a tag in metadata are...


IainNZ:
Is the tag actively harming anything?

tkelman:
It's a bit misleading to have an older tag refer to newer code, I think. But I could remove the second commit if that would be safer.

I'm still wondering how Pkg allowed that, unless Stefan didn't use `Pkg.tag` at all for this? Or if `Pkg.tag` has a bug that allows tags to end up mis-ordered?

It would be fine if say we had made a release-0.3 branch in Compat.jl before merging the changes that were worth bumping the version number to 0.4.0 (which is a somewhat arbitrary choice, other than it being nice to have the Tuple changes start at an easy-to-remember version number for the purposes of other packages' REQUIRE), then tagged 0.3.8 based on some backport to that branch. But that's not quite what happened here.

kmsquire:
Maybe he tagged before running Pkg.update(), and it didn't complain because
there was no 3.8 tag in the repo.

IainNZ:
So this is good to go right?

tkelman:
yeah I took out the reversion, this leaves the out-of-order 0.3.8 tag alone

PR #2493, opened 2015-04-23
Pull request/09c5750c

reworkhow:


johnmyleswhite:
There's a *very* popular text analysis Python library with the name GenSim.

reworkhow:
That is what we are considering. We are trying to come up with a new name.

IainNZ:
For cross-reference, I found:
* GenSim: http://reworkhow.github.io/GenSim/

I couldn't find the reason why JWAS is named the way it is, but I don't know the area at all

dpsanders:
https://github.com/reworkhow/JWAS.jl

"Julia for Whole-genome Analyses Software"

This name should also be reconsidered. "Julia" isn't necessary, since that's specified by the `.jl` ending.
How about `WholeGenomeAnalysis.jl`?

reworkhow:
We call it "JWAS" because "GWAS", which is short for genome-wide associate study, is well-known in genetics area.

kmsquire:
JWAS is actually a pretty clever name for a genetics package.

dpsanders:
The question is should package names be "clever" and meaningful only to people in a certain field, or should they be generally understandable? As I interpret the current naming scheme, the latter is preferred. (Except that in certain popular fields, idiosyncratic names are preferred, such as Mocha. Which seems to contradict this rule...) On the other hand, as more packages are published, it will perhaps be more difficult to avoid  similar clever names and abbreviations.

IainNZ:
Its more like, if there are multiple ways to handle something and they're likely to be opinionated, custom names a fine. JWAS seems like kinda a clever name now.

dpsanders:
> Its more like, if there are multiple ways to handle something and they're likely to be opinionated, custom names a fine. JWAS seems like kinda a clever name now.

OK, in that case I have a suggestion for @IainNZ and <pkg.julialang.org>: Make the taglines (or whatever the longer descriptions are called) searchable as well as just the package names, since the package names are going to be increasingly opaque; and make it compulsory to have a tagline that actually explains what the package does...

kmsquire:
Yeah, I think it's hard and somewhat arbitrary defining the separation
between what's meaningful and what isn't.

 `ODE.jl` probably isn't understandable to most of the (non-Julia using)
population, but no one is going to suggest that package be renamed.

Pretty much everyone in genetics knows what GWAS means, though most of them
aren't using Julia at this point (because a lot of tools are
already written in other languages).

Does this mean that we can only choose names (real or clever) that are
meaningful to engineers and scientists comfortable with linear algebra?
Sufficiently clever names excepted, of course. (But not too clever, or no
one will understand! :-)

Cheers!

ihnorton:
@reworkhow unrelated to the naming question: just in case you've not seen it, do have a look at [BioJulia](https://github.com/BioJulia). There is a great core of folks there (@dcjones @blahah, several others) and they are starting to have some really nice tooling in place. I'm sure they would welcome input/contribution/collaboration in this area.

kmsquire:
Bump.  

From the discussion above, it seems that JWAS.jl, at least could be registered, but the other parts of this PR should be removed, or it should be submitted separately.

It also seems that GenSym.jl was renamed or removed?

mlubin:
The commits could use some squashing

kmsquire:
Bump.

reworkhow:
GenSim was renamed to XSim. A paper is published [here](http://www.g3journal.org/content/early/2015/05/07/g3.115.016683.abstract).

IainNZ:
I think something went horribly wrong here, the entire package seems like it was added to the PR?

mlubin:
Yes, a tidy PR would help.

PR #2498, opened 2015-04-23
Tag CasaCore v0.0.2

mweastwood:
See #1935 for a brief description of CasaCore.


milktrader:
Thanks for the reference for the original package merge.

PR #2499, opened 2015-04-23
Register QuDirac

jrevels:
This PR publishes [QuDirac.jl](https://github.com/JuliaQuantum/QuDirac.jl).


milktrader:
Interesting package *and* organization.

PR #2518, opened 2015-04-28
Pull request/12034a56

oxinabox:
A package that implements my suggestions for JuliaLang/julia#5571
via macro
--
v0.02
Fixed a  bug in Pipe.jl
from not escaping macro returns.


IainNZ:
I'm hitting merge, it doesn't seem like there is any opposition

PR #2519, opened 2015-04-28
Tag AbstractDomains v0.0.1

zenna:
Wasn't tagged before.

Also, I added HyperBoxes.


IainNZ:
For your reference, the `-` in the versions doesn't mean "greater than or equal to", it actually means "allow pre-release versions". I realize that is isn't super obvious though!

PR #2522, opened 2015-04-29
Register Interfaces.jl

Rory-Finnegan:
A simple implementation of interfaces for julia. I'll note that it doesn't currently work on the latest nightlies due to [this](https://github.com/JuliaLang/julia/issues/11052) issue with `methodswith`.


johnmyleswhite:
Why not use Traits.jl instead?

Rory-Finnegan:
Traits.jl seemed overly complicated to use with the `@traitdef` and `@traitfn` and it seemed like it was fighting against the language with the tick to make dispatching work.

StefanKarpinski:
For what it's worth I think further experiments with how to do interfaces in the language are worthwhile, with the understanding that we'll eventually pick one and go with it.

Rory-Finnegan:
Just realized Traits.jl isn't in METADATA. Alternatively, I could withdraw my pull request and just mention it in julia-users.

johnmyleswhite:
I don't think that's necessary. Stefan's point about experimenting with different approaches is good.

IainNZ:
Lets merge?
(PS @Rory-Finnegan, seems like your Github email and your local git config'd email don't match?)

Rory-Finnegan:
Oops, yeah. I'll fix that right away.

PR #2534, opened 2015-05-01
Remove Multirate

JayKickliter:
The code base for Multirate as been absorbed by DSP.


tkelman:
Does anything depend on it? I think the right way to do this is set an upper Julia version bound so it won't be installable by future Julia versions.

JayKickliter:
As far as I can tell, no other packages depend on it. I don't even think anyone has used it besides me.

tkelman:
See https://github.com/JuliaLang/METADATA.jl/issues/347 for some discussion, I'll let the usual suspects of "stewards of metadata" chime in

JayKickliter:
Ok, thanks. I’ll just limit it to a maximum Julia version of 0.3.

PR #2537, opened 2015-05-02
Pull request/19845a6e

fairbrot:


IainNZ:
Hi @fairbrot,
Looks like a neat package. Two things:
* The name. There is a preference to [err on the side of clarity](http://docs.julialang.org/en/latest/manual/packages/#guidelines-for-naming-a-package), so `GaussianProcesses.jl` would be a really great choice for a name.
* Tagging. You actually need to tag a version to be able to `Pkg.add` a package. This can be done after registration or at the same time.

kmsquire:
Is this meant to be the canonical Gaussian Process package?  If not necessarily, then it might not be good to take that name.

kmsquire:
It might also be fine.  There are many possible ways to implement decision trees, and one of them happens to be implemented in the DecisionTrees.jl package.

ViralBShah:
I think the generic name is ok here.

ViralBShah:
I mean `GaussianProcesses` is ok.

johnmyleswhite:
I think names like `GaussianProcesses` are fine if we develop some community standards that reflect who owns what. If we allow some kind of eminent domain rulings on names like `DataFrames` or `Stats`, then the names are safe to use.

fairbrot:
Hi everyone, thanks for the feedback. We're happy to change the name of the package to GaussianProcesses. We'll update this over the next few days,

fairbrot:
Pull request has been recreated at #2561

PR #2542, opened 2015-05-03
Register OSXNotifier

jonasrauber:


IainNZ:
Wooah, blast from the past, thought I remembered this: https://github.com/JuliaLang/METADATA.jl/pull/711

ViralBShah:
Pretty cool! Is this good to merge @IainNZ ?

jiahao:
I think we'd want to see a tagged version first.

jonasrauber:
@IainNZ Yes, finally added some more options and the `terminal-notifier` dependency is now handled by `Homebrew.jl`
@jiahao trying `Pkg.tag("OSXNotifier")` followed by `Pkg.publish()` right now, but `publish` just prints `INFO: Validating METADATA` and then takes forever. Am I doing something wrong?

jiahao:
Hm, I've only seen that happen when one of the git commands invoked by `Pkg.publish` times out on a network connection.

tkelman:
No, `Pkg.publish()` is really slow on 0.3, that's expected and unfortunate but will get better once 0.4 is stable enough for normal users to work with.

jiahao:
Oh, right, I forgot about 0.3 being slow. :/

dpsanders:
I had success doing everything on 0.3 *except Pkg.publish(), which I did from 0.4, having a soft link so the v0.4 directory in .julia is just a link to the v0.3 one.

jonasrauber:
Using v0.3 it doesn't give any feedback, just uses a lot of CPU resources and doesn't seem to stop. However, pushing the version tag manually worked.

PR #2569, opened 2015-05-09
Register Slug: v0.1.0

eraviart:


jiahao:
Slug is not a very descriptive name. Are there other names you could consider?

eraviart:
Do you prefer a name like Slugify or Slugifier?
http://stackoverflow.com/questions/5574042/string-slugification-in-python

cbenz:
Since "slugify" this is a common operation in particular when dealing with URLs in web development, I appreciate the existence of a dedicated package. 

In the Python world I see many packages named "xxx-slugify", so I would vote for Slugify.jl

StefanKarpinski:
+1 for Slugify.jl

eraviart:
Slug has been renamed to Slugify. I close this pull request because it is now replaced with #2593.

PR #2590, opened 2015-05-15
Pull request/e4bdb386

andrewcooke:


aviks:
Hi @andrewcooke, Please wait for Travis tests to complete before merging PR's. If you look above now, you will see a red cross next to the git commit hash. That should be a green arrow before merging any PR.

andrewcooke:
ok, thanks for explaining.  there was no "merge with caution" when i submitted the above.

andrewcooke:
do i have to do something to fix this, or is the error because i didn't wait?

also is there any documentation that explains this?  i've been using the julia package docs, but they don't mention travis.

IainNZ:
The error is because you didn't wait, don't worry about it now.

We've started using Travis more since you were last active to validate changes to METADATA (to avoid any accidently introduced corrupted entries.) Travis is a continuous integration service that is used extensively by Julia and most Julia packages to validate changes against the package's tests as they are made.

PR #2592, opened 2015-05-15
RudeOil, a crude attempt a playing with Julia and Docker

mdavezac:


jiahao:
How much overlap is there with https://github.com/Keno/Docker.jl?

@Keno

mdavezac:
Frankly, I'm not sure. There is no documentation, no test, no usage example in Keno/Docker.jl yet. So I'm not sure which way end to grab it from. That's why RudeOil exists.

Keno:
Docker.jl is currently unmaintained, as I do not currently have a use case that would require Docker any more. I'm happy to give up the name if you actually have a use case and want to maintain the package.

tkelman:
FWIW I think RudeOil is a clever, funny, awesome package name. Maybe not the most discoverable name in the world, but at least Docker's in the sub-description.

ViralBShah:
It's a funny name - but if this is going to be a maintained package, it does make sense to call it Docker.jl.

mdavezac:
For what it's worth, I would rather keep the name. Julia seems to have a tradition of names that are not quite on topic, e.g. Julia itself, then Winston, Gadfly...

IainNZ:
@mdavezac we do have a naming guide in the docs, that prefers literal names if possible. Something like plotting as no one right way to do it, hence the galaxy of names... would have been good to re-purpose the Docker.jl name for something useful, but oh well.

milktrader:
+1 for clever names

PR #2593, opened 2015-05-15
Register Slugify: v0.1.0

eraviart:


milktrader:
Curious why this is licensed under Apache License Version 2.0

eraviart:
No special reason: For small projects like this one, I use Apache
License 2.0 by default, but it could be any non-copyleft license. Is
this license a problem for you?

StefanKarpinski:
Apache 2.0 may not be compatible with GPLv2: http://www.apache.org/licenses/GPL-compatibility.

milktrader:
The standard license for Julia packages is the MIT license.

kmsquire:
(Which doesn't mean new projects are required to be MIT licensed, just encouraged.)

eraviart:
License changed to MIT in Slugify 0.1.1 https://github.com/eraviart/Slugify.jl/commit/a8191225799beaace0355bc97eca0c0cd1e9abef

PR #2596, opened 2015-05-16
Making travis happy :)

dfdx:
New versions for several packages with fixed travis settings. This should help to fix package status on `pkg.julialang.org`.


mlubin:
Passing tests also make users happy :)

PR #2604, opened 2015-05-18
Tag Winston v0.11.10

timholy:
@nolta, I'll give you a chance to decide if this is OK. Winston needs a version bump if Tk is going to start passing its tests, see https://github.com/JuliaLang/Tk.jl/pull/100.


nolta:
Ok with me.

timholy:
Thanks @nolta & @IainNZ.

PR #2607, opened 2015-05-18
Tag MarketData v0.2.3

milktrader:
Needed to change sdata to seconds_data since Base defined two new methods by the same name.


milktrader:
Whoa, that was fast!

milktrader:
Now the chicken-egg problem with getting tests to pass :smile:

PR #2617, opened 2015-05-20
Register GreatCircle: v1.0.0

acrosby:
Just trying to register a basic package that will accurately compute distances on a spheroid earth.


milktrader:
This is a cool idea that I thought about doing some day. Are you implementing the Haversine algorithm?

acrosby:
@milktrader This implements the Vincenty algorithm which is supposedly more accurate than Haversine. It would be cool to include the option to use either, if only to allow for comparisons.

milktrader:
Curious about this dispatch example:

````julia
(f::Number, a::Number, phi1::Array, lembda1::Array, alpha12::Array, s::Array)
````
Typically you can provide more information to the compiler by using `Float64` and `Array{Float64}` unless there is something I'm missing.

milktrader:
Here is an exported method:
````julia
function great_circle(distance, azimuth, latitude, longitude, rmajor=6378137.0, rminor=6356752.3142)
    """
        Named arguments:
        distance = distance to travel, or numpy array of distances
.
.
.
````
Is a numpy array important? And how do you get one into Julia in the first place? Would it be okay to dispatch something more idiomatic along these lines?

````
function great_circle{T,N}(distance::Array{T,N}, azimuth::Array{T,N}, latitude::Array{T,N}, longitude::Array{T,N}; rmajor=6378137.0, rminor=6356752.3142)
````
Notice the `;` separates kwargs with their default values.

Also, you can define `2.0 * pi` in Julia as simply `2pi` or even 2π`

acrosby:
* Certainly *could* provide more information to the compiler, but I'm not sure it is necessary at least for my uses. I'm not super familiar with the JIT optimizations, but is there anything substantial to be gained by specifying the type, other than more code to debug down the line?

* Numpy reference is from the doc strings of the project that the port started from, i can scrub them out

* Would the Array{} dispatch allow for scalar args?

* The `;` is only to denote the start of kwargs right? Is there something inherently problematic without the use of kwargs? Either way is fine with me.

milktrader:
>but is there anything substantial to be gained by specifying the type, other than more code to debug down the line?
Yes, that's a big idea behind Julia, to provide the compiler with type information.

It would be nice to remove numpy references

There really isn't scalars _per se_ but single element arrays.

By not including the `;`, anytime a user chooses to substitute a default value, _both_ values would need to be input in the original order. You couldn't do something like `(..., rmajor=6378137.11)` for example.

Overall the project looks great with what looks like a good test suite. Just trying to encourage a little more idiomatic Julia I hope you don't mind.

johnmyleswhite:
FWIW, I wouldn't burden a new developer with some of these concerns.

milktrader:
Ok, fair enough. 

These are things you can do later. Probably better to raise these sorts of things in the project's Issues.

milktrader:
I'd merge this but not sure if we have gotten particular about having the very first tagged version start at `1.0.0` vs `0.0.1`. 

@IainNZ

acrosby:
I appreciate the feed back, this project was just to test the waters with
Julia with something that I actually use reasonably often. I actually had
*no idea* that a scalar was just an array of length 1. I may have to
rethink the way I structured some things to handle that more elegantly with
different dispatches.

How does the METADATA.jl pull request work if I end up making changes in
the actual code base?

> I'd merge this but not sure if we have gotten particular about having the
very first tagged version start at 1.0.0 vs 0.0.1.

Oh, I thought versions could follow a variety of forms, but looking at at
the current list of packages I see that this is not the case. I can make
some changes and resubmit the METADATA.jl  PR at a v0.0.1 tag, if that is
the best approach.

johnmyleswhite:
> I appreciate the feed back, this project was just to test the waters with Julia with something that I actually use reasonably often. I actually had *no idea* that a scalar was just an array of length 1. 

That's not true. I think @milktrader may have mildly miscommunicated how Julia works.

johnmyleswhite:
The argument against 1.0.0 versions is that Julia itself is going to change substantially before 1.0, so your package is likely to change a lot before 1.0 as well.

milktrader:
Yes, sorry about the confusion regarding scalars.

IainNZ:
Sorry just got to this. @acrosby yes please do submit another PR when you are ready, I'd say `0.1` is a very safe place to start when you are testing the waters. Looking forward to hitting the merge button!

PR #2635, opened 2015-05-24
Register Pardiso: v0.0.1

KristofferC:


tkelman:
I think this is mature enough that it could be moved to JuliaSparse whenever you want.

KristofferC:
Ok, thanks. After I have created the repo in JuliaSparse what do I do? Is it enough to tag a new release to update the repo location in METADATA?.

tkelman:
It would be best to do a repository ownership transfer, that way github will redirect https://github.com/KristofferC/Pardiso.jl to https://github.com/JuliaSparse/Pardiso.jl automatically. Then we could adjust the url at some future point in METADATA but it's not urgent to do so.

KristofferC:
Done.

PR #2642, opened 2015-05-26
Update Expect, register Bond

wavexx:
- Update Expect to 0.1.2
- Register new package (Bond.jl) and tag it as 0.0.1


jiahao:
Also, I'm hoping to find a more descriptive name than "Bond".

wavexx:
On 05/26/2015 10:08 PM, Jiahao Chen wrote:
> Also, I'm hoping to find a more descriptive name than "Bond".

I've used the same module name I'm using in Python.
Indeed, it's generic, although I cannot think of anything else now
(probably spoiled).

Suggestions?

I guess renaming the package later is a bad idea.

jiahao:
Let's wait until we get consensus on this package name.

Having cross-language support for PHP and Perl is really neat!

wavexx:
On 05/26/2015 10:12 PM, Jiahao Chen wrote:
> Let's wait until we get consensus on this package name.
> 
> Having cross-language support for PHP and Perl is really neat!

I have CL (sbcl/clisp) and Julia itself in the works too ;)

IainNZ:
Huh that is pretty neat. I'm not too opposed to Bond, I feel like the other uses of the word are in the physical sciences and they can always write e.g. MolecularBond. Out of curiosity, why register now? The README makes it sounds still very experimental?

wavexx:
On 05/26/2015 10:16 PM, Iain Dunning wrote:
> Huh that is pretty neat. I'm not too opposed to Bond, I feel like the
> other uses of the word are in the physical sciences and they can
> always write e.g. MolecularBond. Out of curiosity, why register now?
> The README makes it sounds still very experimental?

Indeed, I based the named off molecular bonds.
The python's doc reads "ambivalent bonds between languages"

The package is actually fully working and tested (except one missing
function: interact()). I wanted to register it to encourage usage and
have more feedback about the api. Likewise for Expect.

I feel it could be more Julia-ish, but maybe it's my impression.

StefanKarpinski:
+1 for MolecularBonds.jl – I was planning on releasing a package called JamesBond.jl

milktrader:
I first thought molecular bond, and then financial bond. How about Polyglot? (just a thought).

Also, at the risk of sounding pedantic, I'm wondering if the main function `make_bond` might be changed to `bond!`. Just a little easier to type. 

Agreed it looks to be an interesting project.

wavexx:
On 05/26/2015 11:47 PM, milktrader wrote:
> I first thought molecular bond, and then financial bond. How about
> Polyglot? (just a thought).

Would you still keep the bond! a b* prefix for the functions if renamed?

make_bond -> pspawn maybe?
beval -> peval ...

> Also, at the risk of sounding pedantic, I'm wondering if the main
> function `make_bond` might be changed to `bond!`. Just a little
> easier to type.

I like the suggestion.

On a related side, in the project, the src/drivers directory is actually
a git submodule (since the drivers are shared).

In deps/build.jl I have put "@unix_only run(`git submodule update
--init`)". Is this the correct/best way to do it? I couldn't find other
examples except HttpParser that does the same.

I don't actually need to build anything, I just need to extract the data
inplace (and yes, I would like to avoid a git subtree here).

milktrader:
If you like the name Polyglot I don't see why not keep the current method naming in place, with the suggestion of changing `make_bond` -> `bond!`.

There are some methods that you can simplify naming as well, such as calling `bcall` simply `call`. You won't be the first to share this particular function name.

````julia
julia> call
call (generic function with 938 methods)
````
Other things such as `eval`, `import`, export` probably need you to append or prepend a letter. 

This is entirely up to you of course, just throwing out some ideas.

wavexx:
On 05/27/2015 11:10 PM, milktrader wrote:
> If you like the name Polyglot I don't see why not keep the current
> method naming in place, with the suggestion of changing `make_bond`
> -> `bond!`.

Any more takes on the name? If there's no strong opposition, I would
personally keep "Bond" as it would simplify doc writing to me, as I
could swap a few more stuff and naming scheme with the python
implementation.

But Polyglot would be fine otherwise.

> There are some methods that you can simplify naming as well, such as
> calling `bcall` simply `call`. You won't be the first to share this
> particular function name.

Yes, I was hoping to use some generic function name, but you cannot do
that with eval/import, so I just stuck a prefix onto it to make it at
least uniform.

StefanKarpinski:
I like `Polyglot` as a package name for this, personally. It would be interesting to see if we can factor out some common infrastructure for the various packages that call other languages – PyCall, JavaCall, RCall.

StefanKarpinski:
My ideal vision would be that each of these packages depend on `Polyglot` and lots of shared functionality goes in there and each package to call an external language just provides the language-specific bits.

wavexx:
On 05/28/2015 02:54 PM, Stefan Karpinski wrote:
> My ideal vision would be that each of these packages depend on
> `Polyglot` and lots of shared functionality goes in there and each
> package to call an external language just provides the
> language-specific bits.

I'll rename the package to Polyglot then.

I don't know how much can be shared between pycall and bond, though.
pycall actually loads an actual interpreter as a dso (and I assume is
the same for rcall/javacall). As such, julia/<pycall> share the address
space and you can only have one instance, but it's very efficient.

With bond I actually setup a serial protocol between two processes. I
*do* need a handle to the instance to do anything with it.

What we can do here is define some generic functions / a common protocol
so that we can say:

# r for "remote"
reval(instance::type, ...)
rcall(instance::type, ...)
importfn(instance::type, ...)
exportfn(instance::type, ...)

and all packages just overload those basic types and/or use a dumb type
as the first argument, so that would avoid the *call proliferation.

I'm currently using importfn/exportfn for the last two, since I do not
allow anything except functions to be shared. Other packages maybe allow
actual types to be shared as well?

StefanKarpinski:
In that case I'm more ambivalent about the name. But I'm still hopeful that we can factor out more common functionality between the various inter-language interfaces.

milktrader:
I like the big idea approach, too bad there isn't a clear path to how to make this work. It would be also be cool to rename all the __other language__ packages simply by their host language. 

````julia
julia> using R, Python
````

milktrader:
One Language to rule them all, One Language to find them,
One Language to bring them all and in the interpreter bind them

StefanKarpinski:
> @milktrader: `using R, Python`

Yes, I like that too :-)

wavexx:
I realized ``base/multi.jl`` has most of the same interface.

It has remotecall(handle, fn...) for example, although it returns a remote reference by default. rcall in my package returns the equivalent of remotecall_fetch.

RemoteRef is my 'BondRef' analogue.

fetch *could* be logically the same as reval(ref).

The main difference is that remotecall takes actual AST as the second argument, I just instead take a string.

wavexx:
Closing this PR.

PR #2651, opened 2015-05-28
Minor version bump for Interpolations

timholy:


timholy:
I'm pretty surprised that it would take Travis more than 7 hours to get to this. Any ideas about what's up?

IainNZ:
I've just cancelled and restarted all the outstanding Travis builds, we'll see what happens I suppose

IainNZ:
I'm not entirely sure I'm reading https://www.traviscistatus.com/ right, but it seems like maybe they are at capacity?

timholy:
Thanks for pointing that page out, @IainNZ.

tkelman:
We could try adding `sudo: false` to `.travis.yml` to make the builds use docker workers, which should have a quicker queue time.

PR #2652, opened 2015-05-28
Tag OIFITS v0.2.0

emmt:
This release is to work with FITIO.jl version >= 0.6


OIFITS/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3-
```
tkelman: note that the `-` actually means "any prerelease of 0.3," which is probably not what you want here

emmt: Le 29/05/2015 12:30, Tony Kelman a écrit :
> In OIFITS/versions/0.2.0/requires
> <https://github.com/JuliaLang/METADATA.jl/pull/2652#discussion_r31313850>:
>
>> @@ -0,0 +1,2 @@
>> +julia 0.3-
>
> note that the |-| actually means "any prerelease of 0.3," which is
> probably not what you want here
>
> —
> <https://github.com/JuliaLang/METADATA.jl/pull/2652/files#r31313850>.
>

you are completely right I've fixed this (for the two entries in fact)

thank you for the advice

tkelman: No problem.

PR #2654, opened 2015-05-29
Change julia requires of Winston 0.11.10 to 0.4-

tkelman:
ref https://github.com/nolta/Winston.jl/pull/222, this tag does not work correctly on 0.3, so prevent it from being installed there

cc @nolta


tkelman:
no longer strictly necessary since there's a newer tag that works, but still might be good to do anyway

IainNZ:
Seems harmless, lets just merge

PR #2656, opened 2015-05-29
Register PAINTER: v0.1.0, v0.1.1

andferrari:
``Painter.jl`` is a julia implementation of PAINTER: Polychromatic opticAl INTErferometric Reconstruction algorithm.


IainNZ:
The package name/code all use Painter, but the MeTADATA is PAINTER. It should be one or the other consistently.

andferrari:
Thank your for the help.
Do I have to register again the package with a new name: Pkg.register("Painter") ?

StefanKarpinski:
I would just edit the METADATA files manually, do a git amend and then push to this branch again.

IainNZ:
I actually preferred the PAINTER version given that its an acronym/name of an algorithm, but Painter is fine too I suppose

dpsanders:
On 29 May 2015 17:53, "Iain Dunning" <notifications@github.com> wrote:
>
> I actually preferred the PAINTER version given that its an acronym/name
of an algorithm,

I agree.

> but Painter is fine too I suppose

I don't think so, since this will lead people to think that the package is
about graphics or even (possibly?) art criticism. At least the capital
letters are a giveaway that the name is not to be taken literally.

andferrari:
Thank you for your comments. I have renames the package PAINTER.

IainNZ:
Closed by #2660

PR #2659, opened 2015-05-29
Tag MixedModels v0.3.22

dmbates:


MixedModels/versions/0.3.22/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.3-
```
tkelman: Not sure if you're aware, but note that the trailing `-` means "all pre-releases of 0.3" (http://docs.julialang.org/en/latest/manual/strings/?#version-number-literals, same for the packages below) which is probably not what you're intending.

dmbates: Thanks for catching that, Tony.

IainNZ:
I'm sure @dmbates can fix that on the next version, so I'll merge this

PR #2670, opened 2015-05-31
Register FilterMD: v0.1.0

timholy:
This is a new package for filtering and linear algebra operations along particular dimensions of multidimensional arrays. It has a goal of high performance. It's julia 0.4 only, as it requires many of the new iteration capabilities of 0.4.

I will not merge this immediately, to allow the traditional bikeshedding of the name.


tkelman:
since you asked... MD sounds like markdown or metadata to me

dpsanders:
I thought of Molecular Dynamics.
How about `FilterMultiDim` or `MultidimensionalFiltering` as the package name? I prefer the latter. (Though is "filtering" the right word here? That sounds like some kind of time series or image manipulation.)

timholy:
@dpsanders, the motivation for this package is to serve Interpolations and Images, so indeed your guess is accurate. While I haven't written it yet, this package will eventually contain a `filter_md!` algorithm to perform IIR filtering along an arbitrary dimension. But the term doesn't really capture the "linear algebra" part of the functionality, unless you think of LU backsubstitution as essentially filtering (which, for a tridiagonal system, is not a bad analogy, but it's less good for matrix multiplication, which this also does).

I agree the MD should probably be expanded. "MultidimensionalAlgorithms" sounds way too broad. "MultidimensionalLinearOperations" is quite a mouthful. Perhaps punning off of `mapslices` would be better, since in effect this is a cache-friendly way of doing things you'd otherwise use `mapslices` for. Maybe just "MapSlices"? "ApplyToDim"? ["SeparableAlgorithms"](https://en.wikipedia.org/wiki/Separability)?

timholy:
Hmm, what about "AxisAlgorithms"? Currently, every algorithm in the package gets applied along a single axis (dimension) of a multidimensional array.

mbauman:
:+1: I need to find time to get back to AxisArrays. Seems fitting.

dpsanders:
> Hmm, what about "AxisAlgorithms"? Currently, every algorithm in the package gets applied along a single axis (dimension) of a multidimensional array.

Seems like a good option.

timholy:
OK, two votes is enough for me. I'll go with AxisAlgorithms.

timholy:
Oops, accidentally did this by the back door (pushed to the wrong branch). Let me know if anyone experiences problems.

milktrader:
+1 for AxisAlgorithms

timholy:
To clarify, it's already registered and tagged as [AxisAlgorithms.jl](https://github.com/timholy/AxisAlgorithms.jl).

milktrader:
A post-registration thumbs up for an alliterative and descriptive package name.

PR #2678, opened 2015-06-02
Register Polyglot: v0.0.1

wavexx:
See #2642 for old request.

- Renamed package from "Bond" to "Polyglot".
- "experimental" nature toned down (package is fully functional)
- Update prefix to r* (for generic "remote"), resulting in reval, rcall, etc.
- Renamed make_bond to bond!
- Better docs.


IainNZ:
Neat!

milktrader:
:+1:

wavexx:
What do you think if I did an announce into julia-users?
I'm looking for feedback.

milktrader:
Either that or julia-dev (but not both). The reason for julia-dev is maybe toss around the idea of the package being central to other-language packages.

PR #2684, opened 2015-06-03
Pull request/35c7e7f3

andrewcooke:


IainNZ:
No ParserCombinator**s**.jl?

andrewcooke:
don't people normally say "it's a parser combinator library" rather than "it's a parser combinators library"?  i don't know why - it just seems to be normal use.  i think?

IainNZ:
More just a general trend in Julia pacakges to use plurals for the package name, is all. Not an actual problem :+1:

andrewcooke:
well, it's on it's third name already.  maybe this is good enough.

PR #2685, opened 2015-06-03
Tag WORLD v0.1.0

r9y9:
WORLD.jl is now based on the latest version of WORLD v0.2.0_1, which was released at May 31, 2015.


r9y9:
@jiahao I forgot to fix it, sorry. Fixed.

PR #2686, opened 2015-06-03
Bumping Images, ImageView, Grid, Compat, Interpolations, CUDArt

timholy:

PR #2687, opened 2015-06-03
Bump Images, ImageView, Grid, Compat, Interpolations, CUDArt

timholy:
Better version of #2686


jiahao:
I'm sure His Holiness would approve of Prof. Holy's work.

getty_012914_popefrancis

timholy:
Way better than an green checkmark from Travis. With that, how can I _not_ hit the merge button? I'll try to hold myself back...

PR #2695, opened 2015-06-04
Tag StochasticSearch v0.0.2

phrb:
StochasticSearch.jl is a Julia package that provides tools and optimization algorithms for implementing different Stochastic Local Search methods, such as FocusedILS and Racing Algorithms. It is an ongoing project, and currently offers a limited interface to Optim.jl. In the future, StochasticSearch.jl will implement its own optimization and local search algorithms. The API will provide tools for implementing parallel and distributed program autotuners.

https://github.com/phrb/StochasticSearch.jl


IainNZ:
Hey @phrb this looks good, sorry it went unnoticed. I can merge this if you'd like? Let me know

phrb:
@IainNZ, thanks! I thought this was for known contributors only. If you could merge it would be awesome!

It is an ongoing project, so it is changing a lot every week. Should I submit another pull request only when things are more stable?

IainNZ:
Probably a good idea to tag 0.1 when it is more stable, yeah.

PR #2698, opened 2015-06-04
Pull request/233182ff

fairbrot:


GaussianProcesses/versions/0.1.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.3-
```
tkelman: Note that this means "Any prerelease of Julia 0.3" which is probably not what you want here. See http://docs.julialang.org/en/latest/manual/strings/?#version-number-literals.

PR #2699, opened 2015-06-04
Bumped version to 0.1.7

goedman:
HyperDualNumbers Julia 0.4 compatibility update (thank you Tim!)


timholy:
Thanks!

HyperDualNumbers/versions/0.1.7/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3-
```
tkelman: Note that this means "Any prerelease of Julia 0.3" which is probably not what you want here. See http://docs.julialang.org/en/latest/manual/strings/?#version-number-literals.

goedman: Thanks Tony, I'll fix that!

PR #2700, opened 2015-06-05
Tag LightGraphs v0.1.14

sbromberger:
Improved Dijkstra and betweenness centrality; edge distances can be different types.


sbromberger:
Thank you very much!

PR #2714, opened 2015-06-08
Register OEIS

MurrayT:
Add a package to allow for interface with [OEIS](http://oeis.org) from within Julia

### DIFF
--- a/OEIS/url
+++ b/OEIS/url
@@ -1,0 +1,1 @@
+git://github.com/MurrayT/OEIS.jl.git


milktrader:
Cool. How about tagging a version (i.e., `0.0.1`) so the package can be used with `Pkg.add`.

Also, would you mind maybe a quick demo in the README of using the `oeis` method and what the output would look like?

MurrayT:
Do I need to close this pull request before opening the new one?

PR #2717, opened 2015-06-08
Register AutoHashEquals

andrewcooke:


tkelman:
Not much use in registering a package until it has at least one tag, otherwise it can't be added via `Pkg.add`.

andrewcooke:
well i was just working through the package documentation, which has tag afterwards.  it's worked before this way, although i guess it's an extra build request.  why isn't it running?

tkelman:
Not sure, maybe a backlog on Travis.

PR #2722, opened 2015-06-09
Pull request/4d0e5e7c

mauro3:
To add https://github.com/mauro3/Parameters.jl


tkelman:
Should this add a `julia 0.3` requirement?

mauro3:
Done.  Should I be more specific about the versions of other requirements as well?

tkelman:
Cool, will also need to add to https://github.com/mauro3/Parameters.jl/blob/master/REQUIRE before tagging the next version to avoid having to do it manually next time.

If you happen to know which versions of your deps added the features you need then it's nice to add, but not strictly necessary.

mauro3:
I added `julia 0.3` to REQUIRE, thanks. And no, I don't know which package versions.  So I'll leave it like this for now. Thanks.

PR #2723, opened 2015-06-09
Register new package ImplicitEquations

jverzani:
This package provides a means to graph two dimensional implicit equations and inequalities following a strategy of [Tupper](http://www.dgp.toronto.edu/people/mooncake/papers/SIGGRAPH2001_Tupper.pdf)


tkelman:
Should this add a `julia 0.3` requirement?

jverzani:
Thanks. Added to master. (Is there an easy way to update this pull request that doesn't involve just tagging a new version? Does editing through Github work?)

tkelman:
push to the same branch. editing through github would likely work, if you're careful that what you push matches what `Pkg.tag` would do

jverzani:
I'm wary of my git skills messing up the METADATA. :(  If all I have to use are the `Pkg.*` commands, I think I'll just tag a new version once this gets added.

PR #2724, opened 2015-06-09
Tag Meshes v0.1.0

sjkelly:
It has been a while since we last made a release. We are quickly maching towards the 0.2.x series that require 0.4+ and it was requested we make a tag before we break everything :)


Meshes/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.3-
```
tkelman: Note that this means "Any prerelease of Julia 0.3" which is probably not what you want here. See http://docs.julialang.org/en/latest/manual/strings/?#version-number-literals.

sjkelly: Yes, thanks! that was a copying mistake from a 0.4 package

sjkelly:
Will reopen soon.

PR #2728, opened 2015-06-10
Tag Escher v0.0.3

shashi:


Escher/versions/0.0.3/requires
```diff
@@ -0,0 +1,11 @@
+julia 0.3-
```
tkelman: Note that this means "Any prerelease of Julia 0.3" which is probably not what you want here. See http://docs.julialang.org/en/latest/manual/strings/?#version-number-literals.

shashi: You're right. Will change this in the next release.

PR #2737, opened 2015-06-13
Pull request/4eb7b7f6

kibaekkim:
This module provides an interface for an optimization solver DSP.


IainNZ:
Hi @kibaekkim!

This package might not be appropriate for going into METADATA. It definitely can't go in right now as it depends on StochJuMP, which is also not in METADATA. @joehuchette would need to register that - so lets see what he thinks.

Thanks,
Iain

IainNZ:
Also cc @mlubin

PR #2743, opened 2015-06-17
Tag Compat v0.4.6

kmsquire:
Added Timer compatibility code.


tkelman:
Looks good, though we should probably change the `0.3-` to `0.3` at some point. Otherwise if someone had an 0.3-pre build sitting around from just after 0.2.0 was tagged, we'd be claiming to support it here which we almost certainly aren't. Mostly a cleanliness concern since `julia 0.3-` looks different than what it means for new users who are going by examples set by other packages. ref https://github.com/JuliaLang/julia/pull/11716

milktrader:
This says it all
> Otherwise if someone had an 0.3-pre build sitting around from just after 0.2.0 was tagged, we'd be
 > claiming to support it here which we almost certainly aren't.

PR #2744, opened 2015-06-17
Tag MsgPack v0.0.3

kmsquire:
Store tuples as arrays


tkelman:
Should probably put a julia version in your require at some point, if you aren't testing with 0.2 any more.

milktrader:
What is the consequence of not putting in a julia version in a REQURE at all? Does it trigger testing by travis for all the versions?

tkelman:
Nothing too major, but if anyone has Julia 0.1 or 0.2 around and tries to add the package, they'll get the latest code without any warning, which is probably an untested combination for most package developers by now. Travis currently has no support for testing anything older than roughly 0.3.1, we didn't have the infrastructure in place to build generic Linux binaries before that point. Maybe prior builds from the juliareleases ppa could be installable by requesting a version number from apt-get but I don't know if anyone's tried that.

PR #2747, opened 2015-06-18
Tag GR v0.9.5

jheinen:


tkelman:
Out of curiosity, are you using `Pkg.tag` for these? I would expect it to be picking up the contents of your `REQUIRE` file automatically and including them here.

PR #2748, opened 2015-06-19
Register RandomFerns

rened:


rened:
thanks, good point - done

milktrader:
Ok, thanks! Working on expanding the package docs to explain this better.

tkelman:
Oh, and if you use `Pkg.publish`, then your git tags will get pushed to the package repository too, so users can easily see them under github's releases tab.

rened:
Thanks, will to that in the future. Old habit of doing things manually.


Am 22.06.2015 um 01:32 schrieb Tony Kelman <notifications@github.com>:

tkelman:
`Pkg.publish` used to be annoyingly slow, it isn't any more

PR #2749, opened 2015-06-20
Tag MsgPack v0.0.4, VideoIO v0.1.4

kmsquire:
Add minimum Julia version (ref: https://github.com/JuliaLang/METADATA.jl/pull/2744#issuecomment-113356425)

In truth, the code is extremely simple, and probably still works on v0.2.  So I'm leaving v0.0.3 untagged.  That said, I truly doubt there are many (or any) users on Julia 0.2 who use this package.


milktrader:
I think in this case, it's fine to have `julia 0.2` in the REQUIRE. The only catch is that you'd probably need to be able to test against the last 0.2.x to make sure it still works. It would be really nice if we had an environment management system similar to Ruby's rvm that helps one run different versions of the same programming language.

kmsquire:
I wasn't planning to push these in the same commit, but travis is taking a while, and I pushed before I remembered this was here.  Anyway...

PR #2752, opened 2015-06-21
Tag Mamba v0.4.10

brian-j-smith:


milktrader:
This looks like a really nice solution. I'm considering doing the same with my packages. I'm a little fuzzy about how this is done though. I understand that `master` (currently Julia 0.3) holds code for the stable Julia and it seems easy enough to make that into a separate brach once Julia 0.4 is released with a `git checkout -b julia-0.3`, but how do you move the head of `master` to `julia-0.4`?

This is where my git-fu is lacking. I can do a quick search and figure it out though. 

Overall, it seems like a well-organized way to manage packages across Julia versions.

brian-j-smith:
@milktrader - I can't claim to be a git zen master, but here's how I'm thinking about doing the transition.

Create a ``julia-0.3`` branch from ``master``:
```sh
$ git checkout master
$ git branch julia-0.3
```

Merge the ``julia-0.4`` branch into ``master`` and automatically resolve any conflicts in favor of those in the 0.4 branch:
```sh
$ git checkout master
$ git merge -s theirs julia-0.4
```

Disclaimer: I have not tried this strategy before and, when the time comes, plan to do a bit testing of it on a local copy of my repository to check that resulting files and git history look ok before pushing things up to GitHub.  In other words, just be careful if trying this.

milktrader:
Yeah, I thought first about the `git merge` idea but then envisioned a tangled mess of conflicts. I think the easiest part of the puzzle is moving the master to a branch (I think you need the `-b` tag, no?). 

Isn't `master` an arbitrary name that defines the `HEAD` anyway? So maybe there is a way to move the `HEAD` from `master` to `julia-0.4` in this case and going forward not define `master` at all?

Just brainstorming here. Haven't tried this idea myself. But I think you're onto something.

milktrader:
I think I'm conflating `HEAD` and the _main_ branch here. `HEAD` is the point where you are currently working. The task is to define the _main_ branch as the current Julia version.

brian-j-smith:
I think you're right about that being another possible approach - just rename the branches like:

```sh
$ git branch -m master julia-0.3
$ git branch -m julia-0.4 master
```

The simplicity is appealing, but I wonder if the rename actions show up in the git history.  If not, it might be difficult to tell what was done with the branches.  This approach might also require more communications with other developers on the project to let them know the rename is happening and that (special) steps need to be taken to propagate the changes to their local repositories.

milktrader:
I started a sandbox here: https://github.com/milktrader/BranchExperiment where maybe we can pin down this process. I'm adding @brian-j-smith @tkelman as collaborators and you're invited to experiment at will.

tkelman:
This is one valid approach. The other approach is conditional includes so only part of the source differs between Julia versions, I believe JuMP is already doing that.

milktrader:
I'll have to ask @IainNZ about that tomorrow then.

PR #2753, opened 2015-06-21
Register SerialPorts: v0.0.1

sjkelly:
This is a serial IO package similar to Arduino.jl, except it wraps PySerial for the time being.


SerialPorts/versions/0.0.1/requires
```diff
@@ -0,0 +1 @@
+PyCall
```
tkelman: If you aren't testing with julia versions prior to 0.3, it would be best practices to add a julia 0.3 line here, ref JuliaLang/julia#11716

PR #2754, opened 2015-06-21
Tag Neovim v0.0.2

bfredl:


Neovim/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+Compat
```
tkelman: If you aren't testing with julia versions prior to 0.3, it would be best practices to add a `julia 0.3` line here, ref https://github.com/JuliaLang/julia/pull/11716

PR #2756, opened 2015-06-23
Tag SFML 0.0.7; Tag Chipmunk 0.0.2

zyedidia:


zyedidia:
The travis failure here appears to also be caused by JuliaLang/julia#11801.

jiahao:
We can remove the `@time` macro usage now. @tkelman could you add this to #2757?

tkelman:
yeah will do

PR #2757, opened 2015-06-23
back to julia release for running travis tests?

tkelman:
more stable and hopefully fast enough now that pkg speedups were backported


IainNZ:
Good thinking

PR #2759, opened 2015-06-23
Pull request/574a175e

kersulis:
I added a function to make it easier for users to see which MATPOWER cases are supported.


IainNZ:
Cool

PR #2760, opened 2015-06-23
Register GeoInterface: v0.0.1

yeesian:
A Julia protocol for geospatial data, similar to the geo_interface for
python, with aspirations towards the [OGC Simple Features
standard](https://en.wikipedia.org/wiki/Simple_Features).

See also: https://github.com/JuliaGeometry/GeometryTypes.jl/issues/3

CC: @sjkelly , @shashi, @rohitvarkey, @KristofferC, @garborg @wkearn


IainNZ:
:+1:

PR #2774, opened 2015-06-25
Register FixedEffectModels: v0.0.1

matthieugomez:
correct git url and version


tkelman:
This does look like a really nicely done package! Though you probably need to keep an eye on the case of the main file name https://github.com/matthieugomez/FixedEffectModels.jl/blob/master/src/fixedeffectmodels.jl - looking at your Travis, you probably need RDatasets in your `test/REQUIRE`, and you might not be setting the exist status from FactCheck quite right yet.

matthieugomez:
Thanks! Everything is better now. I'm not sure how to push a better version without opening a new pull request though.

tkelman:
You should be able to manually push a `versions/0.0.2/requires` and `versions/0.0.2/sha` file to this same branch and I think that would work fine. You'd also want to manually push the git tag for the corresponding commit to the package repository so people can see tags under the releases tab on github.

tkelman:
sorry, typo, meant `versions/0.0.2/sha1` for the second file

matthieugomez:
I changed the info in v0.0.1 instead. Thanks!

PR #2776, opened 2015-06-25
Tag SCS v0.0.4

joehuchette:


mlubin:
Won't this break Convex.jl if not bumped simultaneously?
@madeleineudell

joehuchette:
Likely.

joehuchette:
Should Convex get bumped?

PR #2780, opened 2015-06-26
Register Electron

jkroso:


IainNZ:
@one-more-minute is there any relation to your work here?

milktrader:
A couple of housekeeping issues to resolve besides the ping to @one-more-minute about code duplication.

* Add to your empty REQUIRE file a minimum Julia version that the code supports (guessing this will be `julia 0.3` but it *might* be `julia 0.4-`)
* Tag the package with `Pkg.tag("Electron") to instantiate it and make it available to those using the `Pkg.add("Electron")` call. (The package is not available with this call if it doesn't  have a tag).

one-more-minute:
This looks like it's basically repeating the build step of Blink.jl, that is, downloading Electron and making it available to run.

@jkroso obviously you're welcome to say if you have a grand plan for Electron.jl which diverges from what Blink does, but if not, perhaps it's best to just have Blink for this stuff? I'm going to update the Electron version soon if that's a concern.

jkroso:
@one-more-minute Blink.jl will have new releases for all kinds of reasons. My intention with Electron.jl is to mirror the version number of the atom/electron releases

tkelman:
If mirroring an upstream version number is really important, that would usually call for a non-semver extra digit I think. And you might need to fix bugs in the package setup occasionally, but be delayed in being able to release those bugfixes until upstream has a new version ready? `Blink.jl` could more clearly document (and expose an API to query) the version of Electron that it's wrapping vs its own version numbers. Or if there's a strong need for a just-unpack-Electron Julia package, then Blink could build on top of it too. I don't think exactly mirroring upstream version numbers is necessarily highly valuable though.

jkroso:
Its not important but it might be important one day. It might never matter who cares Electron.jl is just a better name. I didn't even know to look for Bink.jl

tkelman:
> Its not important but it might be important one day

What might be? The version number matching? Could you explain what you think that helps with?

> who cares Electron.jl is just a better name. I didn't even know to look for Blink.jl

This is true. I think when Blink started, Electron was still called atom-shell (or was using a different tool? I don't know the details). And package discoverability is something we do need to get better at.

jkroso:
Version matching enables you to choose the version of electron you want to run. Sure there are other ways to do that but this is clearly the best way

jkroso:
@milktrader I couldn't figure out how to choose my own tags which is a shame but doesn't really matter. I've tagged it at 0.0.1 and taken care of the REQUIRE file issue

milktrader:
The `Pkg.tag()` method defaults to `:patch` that increments the third digit. Other increments are `:minor` and `:major`. So `Pkg.tag("Electron", :major) will tag it as `1.0.0` for example.

jiahao:
bump

so what's the verdict here on overlap in functionality?

jkroso:
I think if overlap is an issue then Blink.jl should depend on Electron.jl

EricForgy:
Hi guys,

I'm a little surprised (in a good way) to find this. I actually have an Electron.jl package I was working on so I obviously agree this is a good thing to have and would love to contribute.

I have worked a lot with Blink.jl and it is not Electron.jl. I actually agree that Blink.jl should be built on top of something like Electron.jl

Blink is great. I have gone over every line of code several times and have a few PRs over there (some merged, some waiting... and waiting). However, Blink.jl has other stuff that take it away from a pure simplistic Electron implementation and I don't see myself building any serious application on top of Blink.jl, but I would build something serious on top of a well-crafted minimalist Electron.jl.

I hope this (or some Electron.jl package) gets added to METADATA. 

I'll also have a look at this implementation and compare notes. Cool stuff @jkroso !

tkelman:
We've frozen metadata for packages that support julia 0.3. If you'd like to register a version for 0.4, feel free to open a new PR.

PR #2787, opened 2015-06-26
Register JuliaFEM

ovainola:
Registeration of JuliaFEM, FEM project that develops open-source software for reliable, scalable, distributed Finite Element Method.

### DIFF
--- a/JuliaFEM/url
+++ b/JuliaFEM/url
@@ -1,0 +1,1 @@
+git://github.com/JuliaFEM/JuliaFEM.jl.git


tkelman:
Cool. It's usually recommended to tag a package if you think it's ready to be registered, since otherwise it can't be installed with `Pkg.add` yet. The package documentation for 0.4-dev has been updated with more information here (thanks to @milktrader) and is hopefully more clear.

ViralBShah:
Should we check for the package tagging in the tests?

tkelman:
Ref #2064 - so far this is an informal suggestion, not really a strict requirement.

Also any comments about the package name? That's the other major decision to make at register time.

jiahao:
Usually we don't have package names with `Julia` in the name; it is implied already.

tkelman:
Generally organizations do have Julia in the name, but packages have at most a J or Ju. Maybe keep the organization JuliaFEM, but the package could be named something like FiniteElements,jl ?

ahojukka5:
@jiahao 

"Usually we don't have package names with Julia in the name; it is implied already."

This is new information for us. Is this requirement documented somewhere? At this point it would be a very hard decision to rename package, we have already registered domain juliafem.org. Therefore I'm asking an exception with this naming convention.

StefanKarpinski:
The package does not have to have same name as the domain – you can call the package FEM.jl and host the website at `juliafem.org` – seems like a good fit.

ahojukka5:
But what's the actual reason driving you to demand changing the package name?

I've been working with open source softwares for a while and first time I see something like this. After talking with other contributors none of us have seen any documentation telling that there's some package naming conventions we have to follow. Otherwise we would have chosen something else for package name.

Now you tell us that we need to change name of the package in order to get it registered. I'm having a very frustrated feeling about this package naming issue because this is something unnecessary jargon which is not having anything to do about making open source. Instead of this stupid posting I would do something useful like writing new code to make world better.

I always though that making open source is making world better for free. We do not get paid for giving our code and expertise for open source community. I used to think that when working with open source no one can force me to do anything but everything is based on my free willing to do so. And now we face a situation where "people above us" is commanding that we need to do X and Y to get package registered. Is there any further requirements maybe not mentioned yet?

I don't feel very "open sourced" at the moment. But don't get me wrong; I'm very good in doing things other people tell me to do, but at least I get paid for that. I just can't understand that our code is not valid because of some package naming convention which is not clearly described and is pointless in my point of view.

IainNZ:
Well, there is some documentation about package naming convention, but yeah, it is unfortunate. I think its clear why they were asking if the name could be changed: its not a very conventional name for us. They aren't saying you HAVE to rename it to merge it.

IainNZ:
The namespace of METADATA belongs to the community, so when a package is registered its natural that people would like to discuss a name. It isn't a personal attack on you.

milktrader:
Here is the naming packages reference http://docs.julialang.org/en/release-0.3/manual/packages/#guidelines-for-naming-a-package

StefanKarpinski:
It's not a demand. If you really want to use the name JuliaFEM.jl, that's fine.

milktrader:
The process of _negotiating_ package names is bike shedding by the community to a curated repository. It could be that after some back and forth, nobody comes up with a better name that the original.

ahojukka5:
We need to discuss about package name tomorrow (it's 2 AM local time here). But if we decide to change package name, would for example FEM.jl be ok?

milktrader:
Ok, this is my _opinion_ on FEM. Yes, it is an acronym for Finite Element Method, but also could be confused with Field Emission Microscopy. The acronym would make sense in your domain, but it would mean something different to those working in _other_ scientific domains. 

Also, I think dropping Julia is a good idea from the package name. Imagine someone calling your package in REPL

````julia
julia> using JuliaFEM
````
It seems awkward, no? 

I would like something more descriptive such as `FiniteElement` or `FiniteElementMethod`. Now others might not agree but that is completely fine. It's the community part of open source :smile:

joehuchette:
+1 on ``FiniteElement``

ahojukka5:
@milktrader I see your point and it's very good. However I think that when talking about solving partial differential equations FEM is a very common and understandable acronym. At least my google points to finite element method wiki, not for field emission microscopy. Not even in second page. But point is good. Is there any realistic possibility that user confuses these two (or more) concepts?

kshyatt:
FWIW I immediately thought of "finite element methods" when I read FEM but I think `FiniteElement` or `FiniteElementMethod` is more effectively Google-able than `FEM`.

milktrader:
May I introduce you to duckduckgo? :smile: 

screen shot 2015-06-27 at 7 24 42 pm

I didn't mention the Norwegian television station because there _isn't_ a realistic possibility that your package might refer to it in a repository focused on technical and scientific computing, but to quote the wiki on field emission microscopy

>Field emission microscopy is an analytical technique used in materials science to investigate molecular surface structures and their electronic properties. Invented by Erwin Wilhelm Müller in 1936...

ahojukka5:
I just drop this here:
http://www.googlefight.com/finite+element+method-vs-field+emission+microscopy.php

milktrader:
Hmm, 100 to 55? So two out of three scientists would pick finite element method vs field emission microscopy. I find this chart interesting though, the amount of space occupied by the winner is disproportionate to its actual advantage.

screen shot 2015-06-27 at 7 38 21 pm

There is a name for this sort of representation, I don't recall it though.

ahojukka5:
Comment of @kshyatt reveals that FEM is intuitively connected to method of discretizing and solving of partial differential equations. I think talking about field emission microscopy is more like academic example how acronyms can confuse people. But aren't we actually doing a favour to field emission microscopy people when not owning the namespace "FEM" but instead use "JuliaFEM"? Field emission microscopy guys can then develop their own package and have this important namespace for they use ;)

StefanKarpinski:
http://ualuealuealeuale.ytmnd.com/

StefanKarpinski:
JuliaFEM is fine.

ViralBShah:
FEM is clearly finite element method. No confusion there IMO.

tkelman:
No, but to riff off of "Explicit is better than implicit," I think for the purposes of package names, unambiguous moderately-long-yet-clear terms are better than acronyms. And this package still needs to be tagged before anyone can add it without `Pkg.clone`.

milktrader:
> unambiguous moderately-long-yet-clear terms are better than acronyms.

Agreed, and perhaps the docs can be more clear on this? @ahojukka5 mentioned in this PR that this is the first time he and his collaborators have read about naming conventions ...

> After talking with other contributors none of us have seen any documentation telling that there's some package naming conventions we have to follow. Otherwise we would have chosen something else for package name.

So if it's important enough to discuss, perhaps we make it more front-and-center in the Package documentation? What I see as a valid style guideline to improve clarity sometimes devolves into pedantic quibbling that usually takes the package developer by surprise.

btw, the package has now been tagged.

milktrader:
Moving discussion to here: https://github.com/JuliaLang/julia/pull/11910

mauro3:
(This concerns the actual name, not the discussion about naming conventions, thus I posted it here.)

Sorry, a bit late to the game, too late maybe.  There are a few other Julians (@billmclean, @PetrKryslUCSD,  @KristofferC, me, and probably others) interested in finite elements, so bike-shedding about package names does indeed makes sense.  In particular, as @IainNZ pointed out, METADATA.jl is a shared name-space.

I think, JuliaFEM would have been a good name (maybe the best) for a Julia group around finite elements.  So this package name (and the associated group) does clash with that.  Well, maybe the group could still be used as a group for all finite element interested Julians?  But that might clash with the plans of @ovainola, having already http://www.juliafem.org/ registered?  Well, I'm sure we'll all be happy either way :-)

PR #2789, opened 2015-06-27
Tag SIUnits v0.0.4

timholy:


SIUnits/versions/0.0.4/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.2-
```
tkelman: I strongly suspect this is no longer accurate.

timholy: True enough. Hopefully someone will remember to fix on the next tag.

PR #2792, opened 2015-06-27
Bump Reactive, Patchwork and Escher

shashi:
Bug fixes.


tkelman:
Please wait until your pull requests pass travis before merging them.

Reactive/versions/0.2.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.2-
```
tkelman: unless you're testing on the first commit after 0.1.0 was released, this is not accurate

Patchwork/versions/0.1.5/requires
```diff
@@ -0,0 +1,4 @@
+JSON
```
tkelman: could use a `julia 0.3` line here, assuming you aren't testing with anything earlier

PR #2798, opened 2015-06-29
Register LCS: v0.0.1

WestleyArgentum:
Simple package for finding longest common and longest contiguous subsequences of strings and arrays


milktrader:
Would be great to avoid a non-descriptive acronym and instead use a descriptive longer name. In this case, `LongestContiguousSubsequence` might be a bit long, but maybe `LongestSequence`?

Also, it's good to add a REQUIRE file with a minimum Julia version that the package supports. This would typically be `julia 0.3` in the current development cycle of Julia.

WestleyArgentum:
Ya, I was hoping `lcs` was common enough terminology that it would be recognizable. Does it catch other people off guard too?

Also good catch on the `REQUIRE` file, totally forgot it because I auto-generated this one

quinnj:
Even being familiar with the functionality, LCS wasn't immediately obvious to me, and is easy to imagine something else using the same acronym. A few ideas:
* Subsequences.jl
* Sequences.jl (maybe too general)
* StringSequences.jl
* StringSearch.jl

milktrader:
Ah, so auto-generate doesn't provide a REQUIRE file. I'm unfamiliar with this method but that would be a good issue to file at Julia. Is that the `Package.generate()` method?

milktrader:
https://github.com/JuliaLang/julia/issues/11931

WestleyArgentum:
Ok, sounds like a rename is in order. I like the `Subsequences.jl` option from above, unless there are other objections I'll rename tonight or tomorrow.

dpsanders:
Perhaps `CommonSubsequences.jl`?  Just `Subsequences` seems a bit too general to me.

quinnj:
I like `Subsequences.jl` the most. While a hair general, I don't think there's necessarily a great potential of conflict where that name would be the "perfect" fit for another package/functionality more-so than this. `CommonSubsequences.jl` is also a bit limiting since the functionality here also deals with longest *contiguous* subsequence as opposed to just the longest *common* sequence.

dpsanders:
As far as I understand, the contiguous subsequences are contiguous subsequences that are *common* between two given sequences, which is why I think `CommonSubsequences.jl` is a good name. (Contiguous subsequences of a single sequence do not require a new package!)

PR #2808, opened 2015-07-01
Tagged Clustering.jl v0.4

slundberg:
Bumps the version of clustering to include hclust.


johnmyleswhite:
LGTM

Clustering/versions/0.4.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.3-
```
tkelman: unless you're testing with a prerelease, this should drop the trailing `-`

slundberg: I just copied how things are in the current 0.3.3 version. If you want me
to change it I can.

milktrader: Yeah, the dash cruft is everywhere! We're just trying to mop it up :smile:

PR #2809, opened 2015-07-01
Register Subsequences.jl

WestleyArgentum:
Renamed per discussion here: https://github.com/JuliaLang/METADATA.jl/pull/2798

I chose to go with `Subsequences` on the grounds that, if in an alternate universe I came across metadata with both `Subsequences` and `CommonSubsequences`, I'd be dubious. Surely those packages would end up sharing a lot of basic methods, and they seem like they'd be similar enough that the functionality could be covered by just one.

Hope that's agreeable!


quinnj:
Thanks for going through the hassle of renaming. +1

IainNZ:
Merge when green :+1:

milktrader:
+1

PR #2811, opened 2015-07-01
Register Levenshtein

rawrgrr:
Small package for measuring the Levenshtein distance between two Strings.

### DIFF
--- a/Levenshtein/url
+++ b/Levenshtein/url
@@ -1,0 +1,1 @@
+git://github.com/rawrgrr/Levenshtein.jl.git


IainNZ:
Neat, just submitted a PR to make it 20x faster - the wonders of type stability :+1:

milktrader:
The docs have recently been updated to include some information about including a `REQUIRE` file with a minimum Julia version and also the relevance of tagging a version. You can add a require with just the line `julia 0.3` since it doesn't need other packages. To tag, run the following:

````julia
Pkg.update()
Pkg.tag("Levenshtein")
Pkg.publish()
````

**Comment by WestleyArgentum** [comment]
Awesome, I've been copy pasting a slow implementation of `levenshtein` into projects, really glad you made this!

**Comment by rawrgrr** [comment]
I'm somewhat confused by what the commands are supposed to do, so I'm not sure where to debug from here. This is what I see after trying to tag once I pulled in the change to add a `REQUIRE` file:

```
julia> Pkg.update()
INFO: Updating METADATA...
INFO: Computing changes...

julia> Pkg.tag("Levenshtein")
INFO: Tagging Levenshtein v0.0.2
INFO: Committing METADATA for Levenshtein

julia> Pkg.publish()
ERROR: Levenshtein v0.0.1 is incorrectly tagged – 0b2c59605ed804986669a091c5402ef56e33fb73 expected
 in error at error.jl:21
 in publish at pkg/entry.jl:328
 in anonymous at pkg/dir.jl:28
 in cd at /opt/homebrew-cask/Caskroom/julia/0.3.10/Julia-0.3.10.app/Contents/Resources/julia/lib/julia/sys.dylib
 in __cd#229__ at /opt/homebrew-cask/Caskroom/julia/0.3.10/Julia-0.3.10.app/Contents/Resources/julia/lib/julia/sys.dylib
 in publish at pkg.jl:57
```

What do I do from here?

**Comment by milktrader** [comment]
Odd. Can you show what you get with `$ git show v0.0.1`? (i.e., from the command line)

**Comment by rawrgrr** [comment]
I got
```sh
$ git show v0.0.1
fatal: ambiguous argument 'v0.0.1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Should I be able to see the tags with `git tag`? I get no output from that from within the `Levenshtein.jl` project.

rawrgrr:
I'm not sure what went wrong the first time, but I was able to reset the project, untag v0.0.1, redo the tag via `Pkg.tag("Levenshtein")` in julia and opened a pull request for adding the tag and `REQUIRE` file at #2813 

Thanks for all the help!

milktrader:
Cool!

StefanKarpinski:
Not sure if it's helpful at all, but at some point I wrote this somewhat toyish implementation:

```jl
function levenshtein(
    a :: String,
    b :: String,
    d :: Matrix = Array(Int, length(a)+1, length(b)+1),
)
    m = length(a)
    n = length(b)

    d[1:m+1, 1] = 0:m
    d[1, 1:n+1] = 0:n

    for i = 1:m, j = 1:n
        d[i+1,j+1] = min(d[i  , j+1] + 1,
                         d[i+1, j  ] + 1,
                         d[i  , j  ] + (a[i] != b[j]))
    end

    return d[m+1, n+1]
end
```

Has the disadvantage of using a full matrix for computation but the advantage of allowing you to pre-allocate it and pass it into the algorithm, which allows avoiding allocation when doing lots of comparisons.

rawrgrr:
I have some concerns about having the full `n x m` matrix since only the "last row" is needed to compute the "current row" (in the case where someone wants to compare two really long strings), but I do like the idea of avoiding allocation when doing multiple comparisons. I'll see if I can figure out how to add that in without increasing the space complexity.

PR #2812, opened 2015-07-01
Tag Compose and Gadfly 0.3.13

dcjones:


tkelman:
If you're not testing on 0.3-pre, the `-` should be deleted from the julia line in REQUIRE

PR #2820, opened 2015-07-02
Pull request/939cf3f3

tinybike:


IainNZ:
Hi @tinybike. All Julia packages are in the same namespace, so `Simulator` is way too general a name to add into METADATA. `AugurSimulator` would be a good idea.

dpsanders:
Or perhaps `AugurSimulation.jl` or even maybe just `Augur.jl`?

tinybike:
Oh, I see your point!  Yeah, `Augur.jl` would probably be easiest -- `AugurSimulator` makes sense but is a bit long.

dpsanders:
Actually in my opinion, longer is better -- for a non-expert, `Augur` does not mean anything. In any case in the REPL or IJulia etc. you can use tab completion.

PR #2823, opened 2015-07-02
Pull request/8b5208bf

tinybike:


tkelman:
Could you squash the commits? Also you shouldn't have `julia 0.3-` in your REQUIRE file at this point, unless you're testing with prerelease versions of julia

PR #2825, opened 2015-07-02
Pull request/a7b44ea7

reworkhow:


reworkhow:
A tidy PR is created for two packages JWAS and XSim. The related old PR can be found [here](https://github.com/JuliaLang/METADATA.jl/pull/2493). A related paper is published [here](http://www.g3journal.org/content/early/2015/05/07/g3.115.016683.abstract).

tkelman:
Does the updated documentation still need to be clarified further? `julia 0.3-` means you're claiming to support prerelease versions of Julia master from before 0.3.0 was released, which I doubt you're testing with.

reworkhow:
I agree. I'll create a new pull request.

tkelman:
For future reference, note that if you force push to the same branch you can update or replace an existing pull request.

PR #2826, opened 2015-07-02
Register and tag Augur v0.0.1

tinybike:
I think I did the rebase correctly, but please let me know if this broke anything.

### DIFF
--- a/Augur/url
+++ b/Augur/url
@@ -1,0 +1,1 @@
+git://github.com/AugurProject/Augur.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
Looks better now, though sorry I was unclear earlier - you should have `julia 0.3`, not `julia 0.3-`, to state that your package depends on version 0.3.0 or newer of Julia. The trailing `-` is very counterintuitive, it looks like it means "and up" but that's not what it means at all. If you check the package documentation for latest Julia this has been updated recently.

tinybike:
Oh, ok!  Added `julia 0.3` to the REQUIRE file.

PR #2827, opened 2015-07-02
Pull request/80a8f2e9

reworkhow:
A tidy PR is created for two packages JWAS and XSim. The related old PR can be found [here](https://github.com/JuliaLang/METADATA.jl/pull/2493). A related paper is published [here](http://www.g3journal.org/content/early/2015/05/07/g3.115.016683.abstract).


IainNZ:
The names are obviously not super informative, but its a specialized field and they don't clash with anything else I can google, so I'm merging this. Glad to see Julia being used in your field!

johnmyleswhite:
Maybe better not to merge 11 commits?

tkelman:
Only the 3 merge commits are superfluous, this is registering plus 3 tags each for 2 packages. Too late anyway.

PR #2828, opened 2015-07-03
Tag Coverage v0.1.0

IainNZ:
Supports @stevepeak's http://codecov.io, offering us an alternative to http://coveralls.io.

Thanks to @kshyatt for pushing the last stage of this through!


kshyatt:
[Check it out!](https://codecov.io/github/kshyatt/julia?ref=master)

tkelman:
Partial? Branches? Methods? How do we make that happen?

kshyatt:
It does partial line coverage by default. If you're running on your local machine, it will just read in the git commit info to determine the branch. Just `git checkout` and then run the coverage script.

tkelman:
Does it mean git branches, or control-flow branches?

IainNZ:
@tkelman in their JSON api, partial is expressed as "1/3". Just need to figure out what we want on the Julia side now I guess.

IainNZ:
No idea about the other two, but its not git branches, its a language thing.

kshyatt:
Oh! It means control flow branches. I guess `Coverage.jl` needs to spoof up our `Dicts` a little.

IainNZ:
One step at a time :D At least this adds an epsilon more motivation for getting more clever.

kshyatt:
We can also put a GRAPH of coverage increase/decrease directly in README. Truly a glorious day.

IainNZ:
Thats the kind of feature that I use to trick people into thinking I'm a valued member of the community ;)

tkelman:
graphs => pillar

StefanKarpinski:
This is gonna be pretty sweet :-)

stevepeak:
Thank you all for the feedback and love! Anything we need please do not hesitate to ask! 

We have a strong roadmap of feature and enhancements which you can [find here](https://github.com/codecov/support/issues).

Cheers!

carnaval:
@stevepeak One feature that would help a lot would be support to display coverage for different specializations of the same code under different argument types. We could emulate this by outputting partial coverage results but it would then be unclear which specialization is responsible. Can codecov do that in some form ?

stevepeak:
@carnaval thank you for the feedback! Can you checkout [this example](https://codecov.io/github/pyca/cryptography) which uses our environment variables and build matrix. Does this solve the issue?

Another improvement coming out soon is enhanced messages: https://github.com/codecov/support/issues/47

Let me know if this is what you are looking for. Thank you!

carnaval:
I guess we could use the message feature to display the coverage breakdown per specialization. Let me explain our use case a bit more.

We might have a single function, but depending on the argument types you call it with, it may have very different behavior (much like c++ templates). This is pervasive in idiomatic julia code. We can emit coverage hit information separatly for each "version" of the same function that was called, but they will map back to the same lines in the file. Using partial coverage for this is not very satisfying because going through the UI it would not be clear which version of the function is not covered.

I guess the best thing would be to have a way of specifying coverage for separate versions  of a single line and be able to visualize the different coverage results of a function's "versions" in the UI. It may represent a significant amount of work on the UI side thought :-)

stevepeak:
Can you show me an example output of this version testing concept? Plus we can move this conversation to https://github.com/codecov/support/issues/54. Thanks!

PR #2830, opened 2015-07-03
Register Bezier: v0.0.1

rawrgrr:
Adding a package that does interpolations for Bézier curves of arbitrary order.

I'd like to also get some feedback for how I might structure this better. Julia is still very new to me and I'm sure there's something I'm getting wrong. To give some context, I was thinking about this from a graphics perspective of drawing Bezier curves, but perhaps there's other usages that may be useful to people that I should try to support. Thanks in advance for taking the time to look at this!


IainNZ:
You should post it on julia-users!

PR #2831, opened 2015-07-03
Pull request/e0f7cc52

tbreloff:


IainNZ:
Is there any reason this couldn't have gone into DataStructures.jl? Its seems like that'd be better long term.

tbreloff:
No reason... more that I wasn't ready to take the time to merge, and that
it currently only works on 0.4 master.  I don't mind if you reject the PR
for now and sometime in the future I'll consider a PR to DataStructures.jl.

IainNZ:
I mean, I want to get the code out there, but it seems like a very general name, and registering packages is (right now) a permanent thing, so getting it into DataStructures would seem like the best long term option.

tbreloff:
Understood completely... that's the reason why I haven't registered any of
my packages yet.  This was the first one (because it's pretty simple with
almost no dependencies), and I mainly just wanted to see what the process
was like.  I'll work on a DataStructures PR when I have time.

milktrader:
How is this different from the Linked List in DataStructures.jl? And is there a specific time series application you use this for?

tbreloff:
The LinkedList type holds a reference to the first data item and the rest of the list, so it's a little clunky to do certain operations (inserting/deleting in the middle for example) without looping through part of the list.  Also iteration returns the data, as opposed to the LinkedList object, so you can't easily change the list while iterating.  I was going to rewrite most of the interface anyways, so I just started from scratch.  It's not to say that I couldn't wrap LinkedList with additional functionality... This is just what I did. If no one wants to use this, it's fine by me. Most people will be fine sticking to Arrays and LinkedList. I use the list types primarily to represent order books. The double version allows iteration/updates both forwards and backwards. 

As for CircularBuffer and RollingWindow, I use them primarily for holding lagged streaming data that are inputs to online statistical/learning time series models. 

I think these all could be optimized more than they are. Maybe a few more eyes on the code will be helpful...

IainNZ:
If you haven't already, I'd definitely advertise this on julia-users, its good work and I'd hate for it to be missed. I'd also be happy to merge this package as-is except for a more targeted name, but while we are still vetting names for METADATA.jl I think we'll have to close this for now.

milktrader:
Yeah, I was thinking about this and this data structure is *similar* but different to the one in DataStructures.jl. I could see this package as a stand alone but with a different name. Maybe something like `OnlineDataStructures`? What do you think @tbreloff ?

tbreloff:
I can't decide whether these belong standalone or not.  I think
CircularBuffer belongs in DataStructures, but I'm not sure about the
others.  I don't want to confuse people... they have very specific uses.
OnlineDataStructures is a little long... I was thinking Juju.jl instead (jk
;)

milktrader:
Ha, Juju is taken! 

Would this data structure work with mapping input streams to output streams, as would be used in a liquid state machine? Your LinkedList seems more flexible than the one in DataStructures.jl but I'm not sure it either would be computationally superior to a plain matrix, or an implementation like AxisArrays.jl

milktrader:
`StreamStructs` (`StreamingStructs`)
`StreamLists` (`StreamingLists`)

Just spitballing ...

tbreloff:
To clarify... the linked lists aren't only meant for streaming data.  I use
them for as a sorted container that can insert/delete anywhere in the list
with consistent complexity.   I would consider OnlineStructs.jl, since I
seem to like working on packages with "Online" in the name.

PR #2835, opened 2015-07-04
Register SnJulia

GDPlumb:
I'm trying to publish a package called "SnJulia".  However, Pkg.publish() didn't work for me so I followed the directions to pull manually.


milktrader:
Presumably, a package in METADATA is written in Julia, so including "Julia" in the package name is redundant. This leaves us with `Sn`, which is cryptic. I would first think an element in the periodic table. (I peeked and it's Tin). 

But really this is code about
>  harmonic analyis on the symmetric group of order n, 

I'm not sure what a good alternative name would be. Maybe `SymmetricGroup`?

Also, it would be good to include a `REQUIRE` file with the minimum Julia version that package supports. If it works with Julia v0.3, then put `julia 0.3` in this file.

GDPlumb:
I was considering that issue as well, but we were far enough through the project that changing the name seemed a little late.  I'll ask about "SymmetricGroup", it is a descriptive name.  Is there a way to change the package name or should I resubmit with a new name?

I'll add the require file.  It runs in 0.3 but has some syntax that is deprecated in 0.4 (but it still works).

GDPlumb:
The group would prefer not to change the name.  Is the change mandatory?

IainNZ:
No, but you don't *need* to register your package - people can still just download it directly from your Github page too.

IainNZ:
I'm ok with registering it.

johnmyleswhite:
It's not mandatory, but we'd like you to reconsider your preferences. We generally think names like SnJulia are bad for the community's long-term welfare.

GDPlumb:
Would "SnFFT" be an acceptable name?  It avoids the redundancy of "SnJulia" but is a little nicer sounding then "SymmetricGroup" and is more descriptive.  The focus of the library is much more on the Fourier transform part then the actual group operations.  

Is the easiest (I don't think the version history of the package is significant) way to change the name to simply start a new package and copy everything else into it?

johnmyleswhite:
That seems much better to me.

johnmyleswhite:
The way you suggested to do a renaming works. It's also easy to just change the name of the package inside the existing files and rename the GitHub repo. It mostly depends on what feels more natural to you.

GDPlumb:
Sounds good, I'll get the names changed and resubmit as a new package (it will get the version numbers of everything synced).  

Thanks for your help

johnmyleswhite:
Thank you for being willing to change the name. It's a huge favor to us.

PR #2846, opened 2015-07-06
Adds new SIE.jl package

dlfivefifty:
This adds SIE.jl v0.0.1, which is a new Julia package for solving Singular Integral Equations.  Examples include acoustic scattering and fracture mechanics.


tkelman:
-1 for the acronym name, naming guidelines and most of the community would prefer something longer and spelled out

dlfivefifty:
SIE is a standard acronym for Singular Integral Equations.  So its comparable to ODE.jl.

tkelman:
Standard acronym within a very specialized field doesn't make the package name any easier to understand at a glance by non-experts. ODE is also a bad package name that doesn't conform to the naming guidelines, but it's also been in metadata for several years and would have been renamed by now if renaming an already-registered package were less disruptive. The cost of a longer name is a few extra keystrokes (and tab-completion can mitigate this). The benefit is users having a better idea of whether or not your package would be useful to them.

dlfivefifty:
would SingularIntegralEquations.jl be preferable then?

milktrader:
+1 for longer spelled-out package names.

ViralBShah:
+1 for the longer name in this case. I didn't know about these equations before the package name.

dlfivefifty:
OK we'll change the name

dlfivefifty:
New pull request is https://github.com/JuliaLang/METADATA.jl/pull/2853

PR #2848, opened 2015-07-06
Tag LightGraphs v0.1.15

sbromberger:
Lots of (major) changes here.


IainNZ:
Are there any breaking changes? Maybe should have been 0.2

sbromberger:
Hm. No breaking changes (well, one API change in `induced_subgraphs`, but that's minor - we just got rid of the forward mapping in the return value). Lots of big performance improvements and new functionality.

I will be merging @zaccranko's work shortly (hopefully in the next week) and we should probably go to 2.0 then since we're redefining what "simple graphs" are - we're going to start allowing self-loops!

sbromberger:
PS - thank you for the quick merge. :)

PR #2850, opened 2015-07-06
Tag Interfaces v0.0.4

Rory-Finnegan:
Lots of major changes. NOTE: breaks julia v0.3 compatibility due to dependencies on the 0.4 version of the julia C libraries.


milktrader:
`julia 0.4-` in the REQUIRE should not update this package for those using Julia 0.3 so everything looks in order.

PR #2854, opened 2015-07-07
Add MLKernels.jl to Metadata

trthatcher:


milktrader:
Have you looked at the MachineLearning.jl package?

IainNZ:
Looks cool, thanks.

milktrader:
Thanks @IainNZ this package looks very well done. How does it relate to SVM.jl in JuliaStats?

IainNZ:
I think thats actually about fitting SVMs, this is more a foundational package like Distributions.jl but for machine learning kernels.

trthatcher:
@milktrader Thanks! That should be `julia 0.3`. I'm new to Github... do I just make the change manually (for METADATA) and send a new pull request with that change? Do I need to reference this one somehow?

IainNZ  is correct - it would be like Distributions.jl for kernels used in machine learning (or the covariance functions in Gaussian processes). We wanted to provide a simple `Kernel` type for efficiently computing kernel matrices. Right now we've only implemented a few of the common kernels, but we plan to extend it.

IainNZ:
Just fix that in the package, then next time you tag a version it'll propagate through to METADATA

milktrader:
Thanks for the package @trthatcher!  The `-` can be removed on your next tagged version, no problem. Our documentation has been a little misleading and we're working to clean it up.

PR #2873, opened 2015-07-09
Register Decimals

tinybike:


Decimals/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3-
```
tkelman: no `-` for 0.3

tinybike: Sorry about that.  I corrected it for 0.0.2 but looks like I forgot for 0.0.1.  Will fix.

tkelman:
Thanks for fixing the require. Looks okay to me now, any other feedback from the peanut gallery?

You've had this around but unregistered for a while, right?

tinybike:
Yep!  I thought that BigFloat provided the same functionality as Decimal, so this package didn't seem useful/interesting enough to register.  A user pointed out to me that [this is not the case](https://github.com/tinybike/Decimals.jl/commit/3e9b48f2a0e0a7d8b4b0fcbb7a4f8b0b42b0423e), so I decided to go ahead and register it.

Worth noting: so far it only has add, subtract, multiply, divide.  When time permits, I'm going to put in exponents, logs, mod, round-half-X rules, etc.  (All the stuff that Python's decimal package has, basically.)  If you guys would prefer I wait until it has a full arithmetic suite to register it, that's totally fine too -- just let me know.

Issue #2800, opened 2015-06-29
Rename METADATA.jl to CuratedPackages.jl


milktrader:
This will make the curated repository of repositories more descriptive, thereby complying with its own naming guidelines :smile:

IainNZ:
Well, assuming we want it to be a curated, which is still ?!. I think a curated repository actually should be a new one, because otherwise we'd have to throw things out of this one.

StefanKarpinski:
Totally possible, but we need to come up with a coherent strategy and not just rename the repo.

milktrader:
Yep, just changing the name would a bit of a mess. I like the idea of simply creating a new repo of repos though and beginning to migrate packages to it. Also, the new package could be explicit in what it is, and the guidelines for inclusion. This repo could begin enforcing minimum requirements (License, REQUIRE with a Julia version, passing tests (?), reasonable package name, etc) right from the start.

So maybe decide on these requirements, migrate packages from METADATA that meet them, and loosen up inclusion rules for METADATA.

tkelman:
Then a question is do we make CuratedPackages the default, and free-for-all METADATA opt-in? The free-for-all could be called something like OpenRegistry.jl if we also feel like renaming it too.

milktrader:
I like OpenRegistry as a name. Yes, I think CuratedPackages should be the default. OpenRegistry could include a dozen implementations of gradient descent, many of which are students just doing their homework assignments in the form of a package. CuratedPackages would have the best-of-breed package.

milktrader:
This requires a small change to the `Pkg.add()` method to include a kwarg for `repo="CuratedPackages"`.

milktrader:
https://github.com/JuliaLang/julia/issues/11914

quinnj:
-1. This sounds like potentially a *lot* of work + resulting breakage/issues for not a lot of gain. I feel like I'm really against having an "OpenRegistry" repo. That's something that someone could start curating on their own if they really wanted without the need to provide something "official" (i.e. Julia.jl). I think it'd be much better to, for example, make `Pkg.clone()` resolve declared dependencies; then `Pkg.clone()` becomes my true free-for-all that acts almost exactly like `Pkg.add` while `Pkg.add` can continue to pull from METADATA where things are more carefully curated (particularly with #2777)

ssfrr:
[...]

Issue #2777, opened 2015-06-25
Require a package's own tests to pass before allowing a version bump


quinnj:
As discussed at JuliaCon 2015 Thursday evening:
- To bump a version of a package, a PR would be submitted, a bot/machine would download the proposed version of the package along with the proposed compatible version of julia and run the tests of the package
- On passing tests, the PR would be automatically merged and version bumped
- On failing tests, PR is not merged (locked?) and PR submitter is notified of failing tests or version mismatch

On Julia version bumps (minor or major), all packages would be re-run if the package has no upper limit on julia version compatibility to ensure that its tests still pass:
- If tests still pass, a new version is automatically bumped pointing to the same previous version
- On failing, the previous version is amended with an upper limit on the previous working version of julia; trying to `Pkg.add()` that package on the newer julia version would result in a `no version compatible with your julia version exists` or something like that

Obviously this is a lot of buildbot/testing infrastructure, but it seems like this would allow the state of METADATA packages to be 100% passing their own tests, with full transparency if changes to julia occur that break the package. The ideal goal here is that a user would never be able to `Pkg.add()` and get a package that "doesn't work" (at least according ot their own tests).

A related issue to discuss would be:
- minimum package testing standards that would at least be "eye-balled"

timholy:
@jiahao pointed me to this. There's a lot to like about this proposal. The only problem I see with it is that I occasionally find myself reluctant to comment out tests that are failing for reasons "beyond my control," by which I mean highly-nontrivial upstream bugs that do not affect any of the platforms I depend on ([like this one](https://github.com/timholy/Images.jl/issues/284)). I nevertheless sometimes would prefer to leave the test there to deliberately annoy myself. But maybe I just need to dig into more flexible testing frameworks?

milktrader:
If you use the FactCheck package, you can use `@pending` instead of commenting out tests. This doesn't fail the tests but it does remind you that there is a test that needs to be dealt with.

StefanKarpinski:
We should add that feature to the base testing support. It should not error but print a loud warning.

tkelman:
[...]

PR #2883, opened 2015-07-13
Tag Lazy v0.9.0

one-more-minute:


tkelman:
Should put `julia 0.3` in require if you aren't testing against anything earlier

one-more-minute:
Sure, PRs are welcome.

milktrader:
@tkelman I wonder if we could get a list of registered packages that either don't have `julia` in the REQUIRE or have dash cruft. @IainNZ probably knows how to do this.

tkelman:
Ref #2800 and #2777 - the consensus is towards enforcing a higher standard automatically as we go on. Merging without addressing valid review is somewhat poor form.

PR #2888, opened 2015-07-14
Register MacroTools

one-more-minute:


tkelman:
needs `julia` dependency

PR #2901, opened 2015-07-15
Split JLD from HDF5; update Images and Options

timholy:


jiahao:
That's a lot of JLD versions :)

jiahao:
Also, shouldn't JLD 0.5.0 require HDF5 0.5.0?

mbauman:
:tada:!  Do you mean to tag 0.5 as the `teh/split_jld` branch?  Or have you just not pushed your local master (with that branch merged) yet?

timholy:
Oh, crap. Yes, I messed this up big time.

It occurs to me that since I'm going to be traveling for the next ~18hrs, maybe I should just wait to do this until I can deal with the fallout.

PR #2905, opened 2015-07-16
Merge pull request #1 from JuliaLang/metadata-v2

ewiger:
getting latest version


tkelman:
Updating forks like this is much better done via a local clone, using git from the command line or a GUI like sourcetree, rather than using the github web interface to do pull requests for this.

PR #2907, opened 2015-07-16
Releasing v0.1.0

ewiger:


MatlabCompat/versions/0.1.0/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.3-
```
aviks: this should be `julia 0.3` `0.3-` means *any prerelease version of 0.3*, which is probably not what you meant.

ewiger: Do I need to specify 0.4- in addition?

milktrader:
Looks good now.

PR #2910, opened 2015-07-17
Tag GR v0.9.8

jheinen:


tkelman:
Should probably add `julia 0.3` to a REQUIRE file if you aren't testing with any earlier julia versions

milktrader:
The package has a REQUIRE but the tag does not. Odd.

tkelman:
Oh, that is very strange. @jheinen are you using `Pkg.publish`, or making these metadata commits manually?

jheinen:
I didn't know about ``Pkg.publish``. - I make metadata commits manually. If I'm doing something wrong, pls let me know ...

milktrader:
I'm not sure off-hand how `Pkg.publish` handles it, but somehow it generates a `requires` file that is stored in the version directory, along with the `sha1`. Presumably it reads the REQUIRE file in your package for this. This `requires` file is used to define the floor (and ceiling) of Julia versions and the packages in that file, which is useful information for users when they invoke a `Pkg.update()`

wildart:
To get a correct METADATA commit (which includes `requires` file), you need to run `Pkg.tag` before publishing manually or through `Pkg.publish`.

milktrader:
Yeah, probably the best way going forward is to run this ...

````julia
Pkg.update()                #get latest METADATA
Pkg.tag("GR", :patch)  # :patch is the default, :minor and :major are other options
Pkg.publish()                # generates the PR for METADATA and other magical things
````

PR #2911, opened 2015-07-17
Register IDRs:v0.1

mschauer:


milktrader:
Any tests planned? Also wondering if we could avoid an acronym name here.

tkelman:
`julia 0.4` won't work until it gets released. Would also strongly prefer a longer name here.

mschauer:
InducedDimensionReduction would be possible. Is it still such a pain to rename a package?

milktrader:
I think renaming is more of a cumbersome task once the package is in METADATA, but renaming the actual repo (and package) is fairly straight-forward with github.

mschauer:
See #2919.

PR #2919, opened 2015-07-18
Register IDRsSolver:v0.1.1

mschauer:
Redo of #2911 .


mschauer:
At the end, I found IDRsSolver.jl  better than InducedDimensionReduction.jl because the former gives information about what the package does and keeps the naming closer to ``idrs.mat`` and ``idrs.py`` from the website.

tkelman:
Still doesn't help anyone know what the package does at a glance.

mschauer:
Come on...

tkelman:
We can register it with this name anyway if you really don't want to spell out the acronym, but the whole point of disliking acronym package names isn't that they are too short, it's that they are opaque and incomprehensible.

mschauer:
Thanks.

tkelman:
Did you get the original authors' permission to distribute a derived work under the MIT license? The original Matlab code doesn't looks like it has a clear license, just a copyright.

mschauer:
Yes, via @astudillor. PS: do you have a preferred name?

tkelman:
Would be good to write that down explicitly somewhere in the documentation.

If this is just a Krylov method, would it be possible to incorporate into https://github.com/JuliaLang/IterativeSolvers.jl instead of having to be its own tiny package?

mschauer:
I think that would be possible, but I like that the interface allows for abstract linear operators and abstract vectors (as long as ``vec``, ``dot`` and ``axpy!`` are defined).

mschauer:
To clarify, the tiny package will do, until this is figured out.

PR #2926, opened 2015-07-20
Tag MathProgBase v0.3.14

IainNZ:
Enables compilation of JuMP and probably Convex.jl as well

@mlubin @joehuchette @madeleineudell


tkelman:
yay for fast-load jump

PR #2939, opened 2015-07-23
Pull request/33442cef

ChrisTimperley:
Added pre-release Wallace and Push packages.


milktrader:
These look like really cool projects, thanks!

As a matter of tiny detail, remember to remove the `-` from your `julia 0.3-` requirement in a future tag. You no doubt saw this documented at one time and we're trying to clean that up. The `-` is designed to capture pre-release versions but once the version has been officially released, there is no need to support pre-release versions.

ChrisTimperley:
Thanks :-) No worries, I'll keep that in mind for the future.

PR #2944, opened 2015-07-24
ProbabilisticNetwork package

tpoisot:
(first PR on METADATA, so I feel like pre-emptive apologies are needed)


johnmyleswhite:
Since there's going to be some name-bikeshedding, maybe throw in a reference to ecology? I would have assumed this was a general purpose DAG library.

tpoisot:
@johnmyleswhite Good call. I worry that `ProbabilisticEcologicalNetworks` is a bit too long. Let me think of something and I'll come back with a better name.

IainNZ:
I'd say its not too long, but thats your call. Could shorten to `Eco`, or `Prob`, or even `Nets`? Cool package, regardless.

joehuchette:
``ProbabilisticEcoNets``?

PR #2953, opened 2015-07-25
Register StatsFuns

lindahua:
This is package collects all numerical special functions from `StatsBase` and `Distributions`. 

Purposes:
- This package serves the similar role as `Rmath`. `Distributions` will depend on this package instead of on `Rmath` directly. The efforts of replacing `Rmath` computational routines will happen here.
- By moving the numerical functions to this package, the development on downstream packages can focus on higher-level functionalities, while implementing/testing/optimizing the computational routines would happen in this package. 
- There are packages that may need some special stats-functions, but don't need the distributions.

### DIFF
--- a/StatsFuns/url
+++ b/StatsFuns/url
@@ -1,0 +1,1 @@
+git://github.com/JuliaStats/StatsFuns.jl.git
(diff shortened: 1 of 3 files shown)


ViralBShah:
This is a great idea.

PR #2956, opened 2015-07-25
EcologicalNetwork (1.0.0) AND Brim (1.0.0)

tpoisot:
(Two packages for the price of one, sorry if it's not best practice)

## `EcologicalNetwork`

Previous PR was for `ProbabilisticNetwork` -- since the package actually works for all sorts of EcologicalNetworks, I gave it a more general name.

## `Brim`

Bipartite recursively induced modules, a general purpose way of finding the more modular partition for bipartite networks. Currently comes with BRIM, and previous preliminary steps to create an initial partition (including the very efficient LP)


ViralBShah:
2 for the price of 1 is always good. However, you probably want to start at version numbers like 0.0.1 or 0.1.0.

tpoisot:
@ViralBShah I would usually do this, but not in this case -- since these two packages have software notes that are going to be published, and I like the first release to start at 1.0.0, to make it clear that users can expect something stable (this is not 0.x.x material, in other words).

IainNZ:
Sounds good!

tkelman:
The main reason to be hesitant about marking packages as 1.0 is that Julia itself isn't at 1.0. Breaking changes in the language could easily cause breaking changes in your package. Not the end of the world, but keep it in mind.

Your `REQUIRE` files could also use a little bit of cleanup. It's not good practice to depend on `julia 0.3-` since the `-` means a prerelease of 0.3, which has not been necessary for nearly a year. EcologicalNetwork could also use a `julia 0.3` line in its `REQUIRE` if you are not testing it against anything earlier.

tpoisot:
@tkelman Thanks for the advice. I had neglected the fact that Julia is not 1.0 itself, something to keep in mind for future releases.

PR #2960, opened 2015-07-26
Tag Coverage v0.2.0

IainNZ:
This is a breaking change for anyone who was using Coverage "manually", but should not affect packages using it in Travis (and should fix some things as a bonus).


tkelman:
Hm, yep there's some breakage on the buildbots http://buildbot.e.ip.saba.us:8010/builders/coverage_ubuntu14.04-x64/builds/712 but since they're checking out master it doesn't make too much difference whether this is tagged yet

IainNZ:
I've never looked at that page before, but this:
http://buildbot.e.ip.saba.us:8010/builders/coverage_ubuntu14.04-x64/builds/712/steps/Gather%20inlined%20test%20results/logs/stdio
seems to be a JuliaParser related error, which should be unrelated to this.

tkelman:
Yeah may not be your bug, but `in amend_coverage_from_src! at /home/ubuntu/.julia/v0.4/Coverage/src/Coverage.jl:123` makes it looks newly-triggered by something you did change

IainNZ:
Funny enough, that function is probably the only one I didn't touch, *except* for changing it to not return anything (probably shouldn't have done that? unrelated though surely).

PR #2962, opened 2015-07-26
Tag StatsBase v0.7.0

lindahua:


IainNZ:
@lindahua where is the issue discussing this reorganization of StatsBase/StatsFun?

lindahua:
There was a brief discussion in https://github.com/JuliaLang/METADATA.jl/pull/2953.

This is mainly an internal code re-organization, and it does not affect the interface. Hence, did not raise a RFC in advance.

kmsquire:
It still might affect other people working directly on related packages.  Probably good to do an RFC in the future, to get broader input.

PR #2968, opened 2015-07-27
Tag Nettle v0.1.9

staticfloat:
Silences some warnings for Julia v0.4, compatibility with Nettle versions >= 3, enable precompilation.


tkelman:
Tag seems mismatched between here and the package repo's git tags?

tkelman:
Did you delete the 0.1.10 tag at the Nettle repo? I think that was fine.

tkelman:
@staticfloat I think we went with opposite fixes here - you deleted v0.1.10 from Nettle and made v0.1.9 point to latest master, but I swear yesterday I saw v0.1.9 pointing to https://github.com/staticfloat/Nettle.jl/commit/bfec54cfb314818c87c58d19cf6e85b49b981661 so I modified the METADATA commit to be consistent with that before merging this. I'm less worried about modifying the package's git tags than I am about modifying the content of METADATA post-merging this, so I went and made the package consistent with METADATA again. Hopefully you can update your local copy to reflect that?

staticfloat:
Yeah, we did go with opposite fixes.  :P  That's fine for now, I just in general don't like to have holes in METADATA versions, but I hadn't realized you modified the commit.  :)

PR #2974, opened 2015-07-28
Tag SparseVectors v0.3.2

lindahua:


tkelman:
Please don't merge PR's until Travis has passed on them. Is it really necessary to have so many tags in rapid succession, a handful of commits apart?

lindahua:
It is actually on average 10+ commits apart. For each new patch, a bunch of new functions are added to the package. There are other packages that are actually depending on this one.

tkelman:
No registered packages yet.

I guess it doesn't hurt anything to have so many tags, but it does take a fair amount of queue time on Travis to verify. Perhaps publish multiple tags together at the same time?

PR #2975, opened 2015-07-28
Pull request/2a7ebac1

Lanzafame:
Created the begins of a sdk for the [Connect](getconnect.io) platform.


tkelman:
That is a really generic name - do they go by anything slightly more specific, a second word in there to disambiguate from the million other things Connect could refer to?

Also you should add a `REQUIRE` file stating the minimum version of Julia that your package is tested against.

Lanzafame:
Connect is the name of the [product](getconnect.io).

Will add a `REQUIRE` now.

tkelman:
That link isn't working. If you don't have a leading http, github assumes it's a link to a file in this repo.

That's a really terribly named product then. We've had precedent for having Julia packages with longer names than the upstream project they're wrappers for - this is a shared namespace, and getconnect.io might not exist in 6 months whereas deprecating packages in metadata takes quite some effort.

tkelman:
`GetConnect` ?
`ConnectClient` ?
`ConnectSDK` ?
`ConnectAPI` ?

What does the product/package even do?

Lanzafame:
That link is fixed.

ConnectSDK is probably the best fit.

Lanzafame:
I will close this then, change the name and open it with the new repo.

tkelman:
Thanks for bearing with the name bikeshedding

StefanKarpinski:
Closest thing I could find to an about page:

http://docs.getconnect.io/v1.1/discuss/558595e76a83f52f009b309f

The choice of name "Connect" for the product is kind of generic for a product as well, I guess `ConnectSDK` might be a good name for this.

aviks:
The JavaScript and .NET client objects are called `Connect` while the iOS, Ruby and Java objects are some variation of `ConnectClient`. So `ConnectClient` is also a good name for this package. With `Connect

aviks:
I just noticed that your object is called ConnectClient. It might be better to then invert the names `ConnectClient` as the package name, and `Connect` as the object name, since the former is in a global namespace, while the latter is within its own namespace.... just one option.

StefanKarpinski:
I think that `ConnectSDK` is a less confusing name – `ConnectClient` still sounds very generic, whereas `ConnectSDK` at least indicates that "Connect" might be a product name.

aviks:
Fair enough.

PR #2976, opened 2015-07-28
Tag GaussianMixtures v0.0.8

davidavdav:
As requested [here](https://github.com/davidavdav/GaussianMixtures.jl/pull/13).

A missing JLD dependence, my feeling is that JLD must have broken loose form HDF5 or something like that.


tkelman:
Yeah, you must've missed the announcement post on the mailing list. Would also be good to add a `julia 0.3` line to your `REQUIRE` if that's the minimum version you're testing against.

kmsquire:
julia-users link: https://groups.google.com/d/topic/julia-users/e6Sp4v9-pCw/discussion

Cc: @timholy, this is where I thought having a deprecation period might help.

PR #2979, opened 2015-07-28
Tag FactCheck v0.2.9

jakebolewski:


IainNZ:
Does this include the new syntax? Seems like it should be v0.3, also the NEWS.md seems to think is in v0.2.4

jakebolewski:
Makes sense, I'll make those changes.

PR #2980, opened 2015-07-28
Tag FactCheck v0.3.0

jakebolewski:


dpsanders:
The examples should be updated to use the new syntax.

tkelman:
I don't think this should've been tagged as 0.2.9, the syntax change is worth a minor bump.

PR #2981, opened 2015-07-28
Register VLFeat and tag 0.0.2

stemann:


IainNZ:
Hi @stemann, just making sure you do mean the `julia` version dependency you have there? It means "at least 0.3, but not the release version of 0.4 (but prerelease 0.4 ok)". Is that correct?

IainNZ:
(it isn't harmful, as long as you release another version before 0.4 comes out)

stemann:
Yes, I know.

stemann:
Hold on for a bit - need to transfer repo from @kmsquire.

stemann:
OK to proceed.

PR #2992, opened 2015-07-30
Register SoArrays

simonster:


tkelman:
So... what are SoArrays? Not an immediately obvious name to me.

StefanKarpinski:
Agreed, not immediately obvious. I read it as "So..., Arrays."

tkelman:
`HowBoutThemArrays.jl`

johnmyleswhite:
Maybe better to go with the wordier `StructOfArrays`?

StefanKarpinski:
`StructsOfArrays` to keep with the pluralization thing?

johnmyleswhite:
Oh yeah, that's more correct.

simonster:
Then should I rename the type itself to `StructOfArrays`? Or leave it as `SoArray`?

StefanKarpinski:
That could go either way, but `StructOfArrays` is nice and obvious :-)

simonster:
Registered as StructsOfArrays.

PR #2996, opened 2015-07-30
Publish initial GreatCircle.jl at v0.0.1

acrosby:
Attempting to publish GreatCircle.jl as an update to this PR: https://github.com/JuliaLang/METADATA.jl/pull/2617


tkelman:
You should add a `REQUIRE` file to the package that states `julia 0.3` is the minimum version of Julia you're testing the package against.

acrosby:
I updated the repo with a `REQUIRE` file at the repo's root. What is the best way to re-attempt to publish/tag within Julia's `METADATA.jl` pull requirest system?

tkelman:
We may as well merge this now, and the require will make its way to metadata at the next tag when you do `Pkg.tag("GreatCircle"); Pkg.publish()`.

What it means is if there's anyone out there who ever tries to install your package using Julia 0.2 or earlier, they'll get this first tag and it will likely not work. We could modify this first tag to add the require here so such a user would get a message to the effect of "Package GreatCircle does not have any versions that support Julia version 0.2." But it might not be worth the git hassle at this point for you to worry about this. Just good practice going forward to get most packages thinking about and explicitly recording their Julia version dependencies.

acrosby:
Alright, thanks. Hopefully only a minor issue in the short term.

milktrader:
Nice package!

PR #2998, opened 2015-07-30
Register + tag AppConf

tmlbl:


tkelman:
I don't think we support registered packages on gitlab at this time. Can you publish the repository on github?

tkelman:
Oh, it's there - https://github.com/tmlbl/AppConf.jl

Best to change the url to match that.

tmlbl:
My mistake, thought I switched remote before Pkg.register

tmlbl:
Sorry. Should be good now.

tkelman:
Yeah, looks better now. I see you added a require file at https://github.com/tmlbl/AppConf.jl/blob/master/REQUIRE, great. That won't be reflected until the next tag though.

Looks like a quite small package right now, but could potentially grow over time? Any objections to the name from anyone? AppConf seems reasonable to me, pretty understandable what it's for.

tmlbl:
Yes it is a small part of a larger app framework I am working on privately. Since I found it useful for other projects I figured it was worth breaking out to cut my teeth on maintaining a real Julia package. Thank you for your guidance here.

PR #2999, opened 2015-07-30
Bump WoodburyMatrices, Color, HDF5, Graphics, FixedPointNumbers

timholy:
Many of these are due to the change for importing operators.

@tkelman will be pleased that, for a change, I checked the julia version in the REQUIRE file :smile:.


tkelman:
missing the `julia` require on WoodburyMatrices https://github.com/timholy/WoodburyMatrices.jl/pull/3

PR #3000, opened 2015-07-30
Tag UUID v0.0.5

WestleyArgentum:


WestleyArgentum:
@IainNZ hope this works for you now!

IainNZ:
Thanks @WestleyArgentum, really appreciate that!
@johnmyleswhite this was only dependency for Benchmarks.jl, and it didn't work on 0.3. With some compat magic, Benchmarks could work on 0.3 too I think, enabling some perf regression testing between 0.3 and 0.4. Not sure if you are ideologically opposed to polluting it with Compat, or just didn't want the hassle while hacking on it.

johnmyleswhite:
I'm open to sticking some Compat stuff into Benchmarks, although my preference is usually for keeping the 0.3 and 0.4 branches totally separate rather than trying to maintain a unified codebase.

To be honest, I'm a little scared what will happen if we calculate perf numbers on 0.3 and 0.4.

IainNZ:
I'm willing to do the work, maintaining separate branches could be more work-intensive if its still under development though.

I'm more than a little scared...

johnmyleswhite:
Let me try to finish the package first and then we can coordinate. It's almost reliable now, although it's still broken on Windows (and I think will stay that way until we bump libuv).

tkelman:
> still broken on Windows (and I think will stay that way until we bump libuv).

For what? Do you have an issue / enabled appveyor for this yet?

johnmyleswhite:
`Base.time_ns()` doesn't have high enough resolution for nanosecond fast functions. That's described here: https://github.com/johnmyleswhite/Benchmarks.jl/issues/2

The idea that libuv will fix this is a speculation based on the report that newer libuv's have better time functions on Windows.

tkelman:
Oh right, I think I read that one at some point. Does look like "improved uv_timer_t precision on windows" was mentioned at https://github.com/JuliaLang/julia/pull/12266, but dunno whether that is plugged directly into `time_ns`. There's a different windows blocker on that PR, send me a heads-up if you need some testing at any point.

johnmyleswhite:
I think it is: https://github.com/JuliaLang/julia/blob/101eb6a94eaec5511636a9a0576a6c357caf5752/src/sys.c#L382

PR #3001, opened 2015-07-31
Register GtkUtilities, bump Images

timholy:
GtkUtilities is a new package that extends Gtk.jl. Currently the only functionality is rubberbanding, stolen from ImageView. In the longer term I expect to support hit-testing, etc.

Motivation for a new package, rather than just adding to Gtk: https://github.com/JuliaLang/Gtk.jl/issues/28


timholy:
Come to think of it, I probably should have separated the Images bump because of https://github.com/timholy/Images.jl/issues/322. But Travis has been _so_ slow today, it's tempting to lump these.

timholy:
OK, now Travis isn't slow.

To ensure that Images works, I'm going to merge this, but I am fine with the idea of bikeshedding GtkUtilities's name and changing it if needed.

IainNZ:
Its a perfectly fine name, pretty easy to guess what it is.

PR #3009, opened 2015-08-03
Tag Compat v0.4.11

timholy:
For @rened's `mmap` work.


tkelman:
almost would've been worth a minor bump?

timholy:
Well, lots of new functionality got added in the 0.4 series of Compat. What's special about this one?

tkelman:
True, it's fairly arbitrary, and not quite as important to be strict while we're pre-1.0. But for post-1.0 semver, any addition of functionality should bump minor.

PR #3012, opened 2015-08-03
Tag PowerSeries v0.1.13

jwmerrill:
Update for https://github.com/JuliaLang/julia/pull/12235


tkelman:
What's the earliest version of Julia you're still testing this on? If it's an 0.2 release, then drop the `-` from the `REQUIRE` file, that refers to prereleases of 0.2. If you're not testing on 0.2 any more, then it should be changed to `julia 0.3`.

jwmerrill:
Thanks for mentioning this.

If I drop Julia 0.2 support, does that have implications for the semver of my package? E.g. should I bump my package from 0.1.x to 0.2.x when I do that?

tkelman:
Hm, good question. That might make sense, especially if you want to start using features that are 0.3-only, or 0.4-only. The important thing if there are still users on 0.2 out there would be test exactly which tag is the last one that works properly there, and bump the require on all newer tags here in metadata.

PR #3024, opened 2015-08-06
Update SimJulia to be more in line with SimPy 3

BenLauwens:


StefanKarpinski:
In the future, please do wait until automated testing passes before merging.

PR #3027, opened 2015-08-06
Register Mimi

davidanthoff:
The name here is an instance of the fourth rule of the guidelines for package names: this package implements one of several possible approaches to its domain, and so is using a less systematic name.

### DIFF
--- a/Mimi/url
+++ b/Mimi/url
@@ -1,0 +1,1 @@
+git://github.com/davidanthoff/Mimi.jl.git


IainNZ:
Whats the inspiration behind the name, out of curiosity?
(Might want to tag a version?)

davidanthoff:
I think it was "Modular Integrated Modeling Interface" or something like that, but in the end we just liked the name and are using it not as an acronym but just the name as it is.

davidanthoff:
Oh, and I'm holding of with tagging a version, there are still some things I need to fix before that is ready. Soon... And this by the way has nothing to do with the dynamic programming stuff we talked about a few weeks ago, separate project.

joehuchette:
Big fan of the naming of  ``Mimi-SNEASY.jl`` and friends.

davidanthoff:
Ha, I didn't even realize that ``Mimi-SNEASY.jl`` is quite funny. But nice, that actually works well (SNEASY, DICE and RICE are models developed by other groups in different languages, so I'm prefixing their Julia ports so that we can distinguish the different code versions).

@IainNZ anything else I need to do before this can get merged?

milktrader:
+1 Mimi

tkelman:
I think the dashes aren't going to work in module names, but cross that bridge when...

davidanthoff:
Thanks!

PR #3031, opened 2015-08-07
Register RouletteWheels

jbn:


IainNZ:
You might be interested in the weighted sampling algorithms in StatsBase http://statsbasejl.readthedocs.org/en/latest/sampling.html

jbn:
Thanks! I just took a cursory look. I'll peruse it tomorrow.

milktrader:
Regardless, a very cool name!

jbn:
@IainNZ: Dynamic proportions and growable-vectors motivate my package. For example, in network construction, preferential selection proportional to degree. AFAICT, StatsBase's weight vectors are static. Unless I'm missing something, I think my package provides something otherwise not provided. (Although, I do want to imitate some of interface from StatsBase. `sample(a, wv, n; ...)` is an obviously good idea.)

milktrader:
I'll let others weigh in but looks like a good stand-alone package. Curious why the PR does not generate a `requires` file while the package includes a `REQUIRE` file with `julia 0.3` as the floor.

milktrader:
Oh yeah, you need to tag a version for the `requires` file to be generated. Tagging a version (probably 0.0.1) resolves this.

jbn:
Thanks, @milktrader. 

That's probably my fault. I've been coding in Julia for about two weeks now, and this is my first package. And, to be unnecessarily up-front, I ported this today, while terribly hung over. Reading the docs on `Pkg` was challenging. I created the REQUIRES file by hand, after reading something that made me think I should have done so. Was I wrong? 

I'll also have to add a comment about StatsBase. If you don't need dynamic proportions, I would just recommend using that.

milktrader:
Creating the `REQUIRE` file is good, and you have a reasonable Julia version floor there with `0.3`. Once you run `Pkg.tag("RouletteWheels") - it defaults to incrementing at the _patch_ level - and then `Pkg.publish()` we should be good.

jbn:
Ah. Okay. I had done `tag`, but never `publish`ed afterwards. [Corrected pull request here](https://github.com/JuliaLang/METADATA.jl/pull/3033).

milktrader:
Your other PR has been merged.

tkelman:
> And, to be unnecessarily up-front, I ported this today, while terribly hung over. Reading the docs on `Pkg` was challenging.

You're able to code while hung over? Damn, I'm jealous, I can't do anything useful in that state.

More seriously, let us know if/how the `Pkg` docs could be clarified. So far it looks like you've done things right. Though you've apparently [turned on appveyor without having an `appveyor.yml`](https://ci.appveyor.com/project/jbn/roulettewheels-jl/build/1.0.7), you can use [the one from Example.jl](https://github.com/JuliaLang/Example.jl/blob/master/appveyor.yml), just replace `Example` with `RouletteWheels`.

jbn:
Hah. Thanks, @tkelman. I can port code hung over. But, I can't reason about new things -- hence, my poor understanding of the `Pkg` docs. I'll give it a fresh look tomorrow to see if it was unclear. But, I assume my misunderstanding was related to my fuzzy mental state. 

And, thanks for pointing out the appveyor example. I'll fix that tomorrow, and properly hook up the coverage badge. 

Definitely liking the Julia community. I think I'm here to stay.

PR #3033, opened 2015-08-08
Register RouletteWheels (correction)

jbn:
@milktrader pointed out I mishandled my original pull request. This one is properly tagged. See [prior discussion here](https://github.com/JuliaLang/METADATA.jl/pull/3031).


jbn:
Also, see [Fast Proportional Selection in Python](http://jbn.github.io/fast_proportional_selection/) for motivating complexity analysis.

jbn:
How does [pkg.julialang.org](http://pkg.julialang.org/) work? A weekly pull-refresh, or something similar? I'd like to enthusiastically tweet, "I've written my first #julialang package!" But, I don't want to field, "but, wait, where is it?"

IainNZ:
Yep, weekly. But, people can install it regardless of whether it is there or not.

johnmyleswhite:
Probably worth making clear that METADATA is the only official listing of packages; pkg is a useful service built on top of the official package listing in METADATA.

jbn:
@IaiNZ Cool. My question was not so much a complaint as it was a curiosity. Quoting and reemphasizing @johnmyleswhite, "pkg *is* a useful service." So, I'd wait to announce anything, knowing that it is often the first queried service. 

But, once every week or so is not exactly high-friction. If anything, it's probably facilitates high quality code. Don't it always seem to go, that you find your bug / typo a day after you press send?

PR #3037, opened 2015-08-08
Tag LMDB v0.0.4

wildart:


tkelman:
Sweet, now we're talking. I'd normally say if you're raising the minimum julia version requirement it's maybe worth bumping your package's minor version number, but you've only had 3 tags so probably not such a big deal.

Fixed the bug in new-Pkg though?

wildart:
I fix not one but two bugs. Testing in progress.

PR #3038, opened 2015-08-08
Add SolveBio

adelq:
`Pkg.publish()` with GitHub 2FA is giving me issues, so going ahead with it manually.


jiahao:
A more descriptive readme would have helped with reviewing this PR. It took some searching to realize that this package is an API wrapper around [SolveBio.com](https://www.solvebio.com/).

tkelman:
Note that you'll need to tag a version before `Pkg.add("SolveBio")` will work.

PR #3042, opened 2015-08-09
Tag Compat v0.4.13

timholy:


stevengj:
Oh, darn it, I didn't notice that you had tagged this, then tagged 0.5, then had a problem with a conflicting tag, then tagged 0.5.1.

(Shouldn't we be incrementing the minor version when we add new Compat functions?)

PR #3063, opened 2015-08-11
Tag LightXML v0.2.0

tkelman:
now precompile-safe thanks to @stevengj


sbromberger:
:100: Thank you both.

PR #3065, opened 2015-08-11
Pull request/7aaa6d0c

kzahedi:
Added a new set of quantification, taken from publication (listed in the README.md)


mlubin:
Did you mean to release 5 different versions here? If not, could you tidy up the commit history?

kzahedi:
I'm fine with the version history. Is that the reason why I don't see any version update of the package on pkg.julialang.org?

--
Includes iPhone typos

tkelman:
The merge commits and number of redundant 0.2.x releases is mildly annoying, but it's not worth worrying about for now. Going forward it may be useful to look at a git tutorial on how to do rebase and squash commits. We'll think about changing things in 0.3.12 (and it's already changed in 0.4.0-dev) to avoid the need for this.

PR #3070, opened 2015-08-11
Tag RNGTest v0.0.4

andreasnoack:
Some minor updates maunly to avoid warnings.


tkelman:
Are you still testing this on prereleases of julia 0.2?

andreasnoack:
Of course I am :smile: . Actually, I don't think there is anything in the package that requires more a recent Julia.

tkelman:
It should at least drop the `-`. That sets a really counterintuitive example and we should get rid of it everywhere it isn't actually needed.

andreasnoack:
I've opened a new with updated version requirement

tkelman:
thanks

PR #3071, opened 2015-08-11
Tag LightGraphs v0.3.0

sbromberger:
Breaking changes to file formats and return types in persistence - but they're now all consistent (returning a dict of graphname=>graph).

Other major changes: 
- inline documentation thanks to @hayd
- precompilation
- new requires

Lots of other minor changes.


sbromberger:
Thank you, @kshyatt!!

PR #3078, opened 2015-08-12
Tag Digits v0.1.2

greenflash1357:
New version of Digits with new functions and updated Readme.


StefanKarpinski:
That's a lot of merges. Could you rebase them out?

tkelman:
Additionally, you don't need to depend on `julia 0.3-` any more, that means "prereleases of 0.3 or newer." Since 0.3 has been out for almost a year, you can safely change that to `julia 0.3` now, which means "releases of 0.3 or newer."

greenflash1357:
rebased.

@tkelman: thanks. I've merged your PR. Your changes will be part of the next release.

tkelman:
Looks good.

PR #3083, opened 2015-08-13
Pull request/2793d6fb

jpfairbanks:


jpfairbanks:
I messed up the version number the first time. v0.1.0 is the correct version.

tkelman:
Well now both versions 0.0.1 and 0.1.0 point to the same commit, but that's harmless.

PR #3087, opened 2015-08-14
Register Faker

codeneomatrix:
This package is a generator of fake information, for julia 0.4.0

### DIFF
--- a/Faker/url
+++ b/Faker/url
@@ -1,0 +1,1 @@
+git://github.com/codeneomatrix/Faker.jl.git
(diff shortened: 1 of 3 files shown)


IainNZ:
Hah, that is quite cool. I would maybe like the name DataFaker more, because I had to look at the package to understand better, but Faker is OK too.

ViralBShah:
This could be quite handy. Thanks!

milktrader:
I like Faker as a name. Cool package!

PR #3092, opened 2015-08-15
Tag Coverage v0.2.1

kshyatt:
- Switched from JuliaParser back to the base parser so we can run coverage on base Julia
- Updated tests to work with base parser
- Changed `REQUIRE` and `.travis.yml` for 0.4


tkelman:
Too late, but generally when a package increases its minimum Julia version it's a really good idea to bump at least the minor version number. That way maintenance changes or bug fixes could still be made and tagged from the old version of the package independently.

kshyatt:
Oh no! I'm sorry. Should I bump the minor version and submit another PR?

tkelman:
It's okay. If this actually needs to be fixed, say for packages that may need to get Coverage bug fixes for 0.3, I think the right approach would be to modify the requires in metadata so v0.2.0 of coverage gets marked as 0.4-only, but we then re-tag v0.2.0 (or whatever the most recent commit from master of Coverage.jl was still 0.3-compatible) as v0.1.2, and make a release-0.1 branch in Coverage.jl from there. At the moment it's a theoretical problem. We can fix it if it becomes a real problem.

PR #3093, opened 2015-08-15
Register UnicodePlots

Evizero:
Hi!

This is basically a reboot of the idea behind TextPlots.jl library, which seems abandoned.

### DIFF
--- a/UnicodePlots/url
+++ b/UnicodePlots/url
@@ -1,0 +1,1 @@
+git://github.com/Evizero/UnicodePlots.jl.git


milktrader:
Cool. Did you try contacting TextPlots.jl with PRs and such and getting no response?

Evizero:
Well, no. I posted a pull-request [here](https://github.com/sunetos/TextPlots.jl/pull/15) and that started a discussion that he might have moved on to other adventures. 

I do, however, need these plots for a project of mine, so I moved along. People suggested I fork and just maintain the original package, but messing around in the original code I realized that it could be implemented a lot better without the huge gaps in the canvas and such. So I threw everything out started from scratch without using any of the original code. Might as well create a separate package then

ViralBShah:
Is there a dependency on a julia version here? Probably at least 0.3, I am guessing. I am ok with having a new package - given the discussion and exploration with the original package.

Evizero:
I tested it on 0.3.11 and 0.4 that I compiled about 3 days ago. So I think it should be fine (?). I'm still getting to know the Julia development processes, so please point me in the right direction if I'm missing something

IainNZ:
Don't forget to tag a release and do a pkg.publish again.

PR #3102, opened 2015-08-17
Tag Media v0.1.1

one-more-minute:


IainNZ:
Might want to update the URL on this some time, looks like you moved it under Juno?

one-more-minute:
Yeah, good spot. It doesn't seem to cause any issues thanks to github's redirection, but I should really go through and update that stuff anyway.

IainNZ:
Yeah I only realized because I was trying to figure out what Media was, but couldn't find it on your profile.

PR #3109, opened 2015-08-18
Tag SimJulia v0.3.4

BenLauwens:
SimJulia has now the same features as SimPy v3 but with a more pleasant syntax, thanks to Julia;)
Documentation consists of a 10 min introduction and a more in depth topical guide. Some more examples and an API overview will follow.
This version will be used in a simulation and modelling course for reliability of "processes" (Royal Military Academy Brussels).


StefanKarpinski:
Very cool stuff!

PR #3113, opened 2015-08-19
Deprecate 3 packages

IainNZ:
Deprecate @zachallaun's three packages from METADATA. No new commits in over 2 years, and I don't believe they load even on Julia 0.2.


tkelman:
image

IainNZ:
Yes @zachallaun if you ever want to revisit these, they just need a new release and they'll be back. This change just hides them from the Julia package list, and won't show them as "available" at the REPL from 0.4 onwards.

zachallaun:
Yep, this is a good decision. Thanks for keeping track of this stuff, Iain.

PR #3123, opened 2015-08-20
Register SolveDSGE

RJDennis:


ViralBShah:
I suspect you need to tag a version like 0.0.1 or something. Also, just curious how this compares to Dynare. Don't know if there are other DSGE packages as well.

cc: @spencerlyon2

tkelman:
It's convention for the repository name to end in `.jl` for Julia packages.

spencerlyon2:
Right now there isn't a unified package providing solvers like this, so this is a helpful contribution.

That being said, I do think a fair amount of this code is duplicated elsewhere. This certainly shouldn't prevent a package from being registered, but does prompt the question about opportunity for collaboration. Some things I noticed as being duplicate behavior with a quick glance at the source of `SolveDSGE`:

- The `dlyap` and `doubling` code is implemented (under different names) in [QuantEcon/matrix_eqn.jl](https://github.com/QuantEcon/QuantEcon.jl/blob/b4391c6c59aa461f4440a214a598796e9e27cd0a/src/matrix_eqn.jl)
- The shchur reordering routines are [already in base](https://github.com/JuliaLang/julia/pull/9701) [docs](http://julia.readthedocs.org/en/latest/stdlib/linalg/#Base.ordschur


I also worry about licensing issues. As a specific example, I believe the [solution for Sims' form](https://github.com/RJDennis/SolveDSGE/blob/26dd25bcfe6ad389df8085451247a93d58377e77/src/solve_re.jl#L279-L417) is an implementation (port?) of Chris' `gensys` [matlab and R routines](http://sims.princeton.edu/yftp/gensys/). Given that it is more or less a port of the routines and I believe they are GPL, we would need permission from Chris to release them under MIT. The QuantEcon group has collaborated with Chris on this subject and obtained the permission. We just haven't found the right home for the gensys implementation we [wrote](https://github.com/spencerlyon2/Hobo.jl/blob/master/src/gensys.jl). @RJDennis, we'd be happy to collaborate on this.

RJDennis:
Thanks for the comments.  I don't think that the vast majority of these solvers are currently available in Julia.  Last year when I announced the routines on the users forum some action was spurred from the Dynare people, so I didn't try to make this an official package, but nothing really seems to have happened so I thought that it would be useful to make these routines more easily available to people.

I can imagine that there is a bit of overlap for some routines --- like the dlyap and doubling routines --- but this keeps the module self-contained so I'm not sure that it is completely a bad thing.  I also have my own givens routine when there is one in base.  All of these routines (other than gensys) are translations of ones I've written in Gauss, and back when I wrote the QZ-based solvers Gauss did not have the generalized QZ decomposition or routines to reorder it (I think it still doesn't have the latter).  I'm happy to update the QZ routines down the line to take better advantage of what is now in base.  The only routine that I did not write from scratch is the gensys function, which is a port of Chris' code (I believe I state this in the function).  I have never used gensys, but some like it so I thought to include it.  I wasn't aware that there was a GPL licence attached to it, but I'm happy to write Chris and get his permission.  I am also happy to chuck my gensys routine in the bin or replace it if there is a better one out there (and Spencer's one is no-doubt better).  Happy to collaborate.

I think/hope that there is a group of people out there that would find these routines useful.  To this end, if someone could tell me how I can rename my repository to add the .jl suffix without deleting everything and starting from scratch I'd appreciate it!

tkelman:
You can go to the github settings for the repository and there's an option to rename.

RJDennis:
A REQUIRE file has been added, the repository has been renamed to follow convention, and permission from Chris Sims to release the gensys solution procedure under an MIT license has been obtained.

ViralBShah:
I believe the `url` file in this commit also needs to be updated to reflect the name.

ViralBShah:
Also, do we need to tag a version?

tkelman:
superseded by #3157, closing

PR #3128, opened 2015-08-20
correct sha

matthieugomez:
sorry I made a typo


IainNZ:
Try using `Pkg.tag` and `Pkg.publish`, they work really quickly now!

PR #3129, opened 2015-08-20
Tag IPNets v0.2.0

sbromberger:
No breaking changes (removed a function that wasn't being used, but that's about it).


sbromberger:
Thank you!

PR #3132, opened 2015-08-21
Pull request/a1a9e64d

aaronstanton:


aaronstanton:
Adding Seismic.jl for processing and plotting of seismic data.

kshyatt:
Hi @aaronstanton. I took a look at your README, this package looks cool! But you don't need to tell people to add the dependencies - if they `Pkg.add("Seismic")`, everything in `Seismic.jl`'s `REQUIRE` will be installed if needed. Also, in your `REQUIRE`, you list specific versions of `PyPlot`, `Docile`, `Lexicon`, and `Grid`. Does `Seismic.jl` *only* work with those specific versions? If not, you can remove the specific version number besides those packages. You should keep it next to `julia`, though - keep `julia 0.3` if that's the Julia version the package will need.

aaronstanton:
Hi Katherine,

Thanks for your message, I removed the version dependencies next to PyPlot,
Docile, Lexicon, and Grid as you are correct, they aren't specifically
needed.

I have a question: since PyPlot takes a while to load I split the module
into two so users can do

using Seismic

to quickly process data without any plotting. or

using SeismicPlotting

to generate plots, or

using Seismic,SeismicPlotting

to do both processing and plotting. Is this ok that the project is
structured in this way, or do I need to split it into two separate projects
or merge them into one module?

Cheers,

Aaron

kshyatt:
I don't see anything wrong with the split. This LGTM - anyone else want to chime in?

tkelman:
You should add some unit tests and enable Travis CI, as development continues.

aaronstanton:
Thanks for the advice I enabled Travis CI with one test so far with more to follow.

tkelman:
Note that the tests are driven by a `test/runtests.jl` file, looks like you have a `test/IJulia/runtests.jl` right now. We'll merge now and when you're ready you can do `Pkg.tag("Seismic"); Pkg.publish()`

aaronstanton:
Thanks, I fixed that.

PR #3135, opened 2015-08-21
Tag Taro v0.2.0

aviks:
Ability to generate PDF files via FOP


tkelman:
What's the minimum release version of Julia this is still being tested on? Should add a `julia` line to `REQUIRE`.

PR #3138, opened 2015-08-21
Register SALSA.jl package

jumutc:


kshyatt:
In your `REQUIRE` you don't list a Julia version - does this work with `0.3` or `0.4-` (or both!)?

jumutc:
Sorry, I forgot to include this. Now ``REQUIRE`` is updated.

kshyatt:
Great! This LGTM - anyone else?

kmsquire:
Might want to tag a version, so that people can use Pkg.add("SALSA").

jumutc:
Version is added.

tkelman:
Okay we'll merge this now and the `-` will be gone for the next tag.

PR #3140, opened 2015-08-21
Tag OnlineStats v0.1.1

joshday:
Remove Convex and SCS dependencies


OnlineStats/versions/0.1.1/requires
```diff
@@ -0,0 +1,12 @@
+Docile
+Compat
+Reexport
+FactCheck
```
IainNZ: This should probably be in `test/REQUIRE`

OnlineStats/versions/0.1.1/requires
```diff
@@ -0,0 +1,12 @@
...
+
+StatsBase
+Distributions
+
+MultivariateStats
+GLM
+Dates
+Reactive
```
IainNZ: This seems commented out in your code?

OnlineStats/versions/0.1.1/requires
```diff
@@ -0,0 +1,12 @@
+Docile
```
IainNZ: Should have `julia 0.3` as first line

OnlineStats/versions/0.1.1/requires
joshday: What seems commented out?

OnlineStats/versions/0.1.1/requires
joshday: Will do.

OnlineStats/versions/0.1.1/requires
joshday: Will do.

OnlineStats/versions/0.1.1/requires
IainNZ: Reactive:
https://github.com/joshday/OnlineStats.jl/blob/42e4119366b0f85ad54e2c349ec1dca10376e2e0/src/OnlineStats.jl#L31

joshday: Ah, my collaborator's work...I'll look into that.

joshday:
Sorry for cluttering with pull requests.  See #3141.

PR #3144, opened 2015-08-22
Register IntArrays: v0.0.1

bicycle1885:
This is a package to pack integers into an array as small space as possible.
I cannot wait the release of v0.4 :smiley: 

Please see the repository for more details: https://github.com/bicycle1885/IntArrays.jl


IntArrays/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
+FactCheck
```
IainNZ: Is this a true dependency, or just a dependency for tests? e.g. `test/REQUIRE`?

bicycle1885: Wow, I didn't know that trick! It's just for tests, so I'll fix it.

bicycle1885:
Fixed.

PR #3158, opened 2015-08-24
Pull request/23f87e97

aaronstanton:
adding a tagged version


aaronstanton:
Hi Tony,

I'm having a problem that I believe is related to the fact that this
package has two modules: Seismic.jl and SeismicPlotting.jl

When I add the package: *>Pkg.add("Seismic") * followed by *>using
Seismic* everything
is fine, but the second module *>using SeismicPlotting *isn't found. Should
this module be put into another package, or is there special syntax to load
it in this case?

Cheers,

Aaron

tkelman:
Top-level `using SeismicPlotting` will only work if that's a separate package, yes.

dpsanders:
Maybe a solution (as Tony hinted) is to make a submodule so that you could write
using Seismic.SeismicPlotting?

dpsanders:
Or just using Seismic.Plotting?

aaronstanton:
That is a good solution, I will make two submodules Seismic.Processing and
Seismic.Plotting

PR #3174, opened 2015-08-25
Point to correct sha

crbinz:
Now points to the right commit


tkelman:
Where did the old commit come from? Generally it's preferable to avoid force pushes on permanent branches that other people may be pulling from, and make new tags rather than changing old ones.

crbinz:
I believe this is from some dumb mistake, like a messed up rebase or something. I don't plan on making it a habit, for sure.

PR #3183, opened 2015-08-26
Pull request/80507527 (Tag new versions of ODE.jl  and Sundials.jl)

acroy:
Tag ODE.jl v0.2.0 and Sundials.jl v0.1.3.


ViralBShah:
Can you please squash?

acroy:
What is the best way to do this? Cherry-picking the tag-commits?

WestleyArgentum:
http://stackoverflow.com/a/5201642/1615475 usually works

acroy:
@WestleyArgentum : Thanks, but there are a lot of other commits between mine and I can't really work out how to select only the relevant ones.

WestleyArgentum:
Oh, I didn't look at the dates, those tags are pretty old... would it be easy to just update your metadata repo to be in sync with the main one, delete the old git tags (both locally and remotely if they're pushed already), and then retag?

Just a thought, maybe cherry picking would work though

acroy:
Rebasing against the current master of METADATA.jl did the trick. I guess once this is merged and I after did `Pkg.update()` everything will be good?

tkelman:
Are you continuing to test either of these packages against Julia 0.2? If not, probably a good idea to change the `julia` line in `REQUIRE` to 0.3.

PR #3187, opened 2015-08-26
Metadata v2

davidavdav:
Bringing NamedArrays to 0.4 compatibility, despite a whole load of ambiguity warnings.


tkelman:
What is the minimum version of Julia this package is being tested against? You should list a julia version in the `REQUIRE` file to reflect that.

PR #3197, opened 2015-08-27
Register IJuliaPortrayals: v0.0.1

jbn:
This package adds `IPython`-style (or, scrapbook-style) portrayals of common, html-embeddable resources to `IJulia`. I opted for a package rather an a patch of `IJulia`, because I think a modular approach makes sense. Not everyone needs these portrayals. And, alternative portrayals are possible. So, it shouldn't be part of IJulia proper. However, I use them pretty constantly in my work. And, other people echo a need.

### DIFF
--- a/IJuliaPortrayals/url
+++ b/IJuliaPortrayals/url
@@ -1,0 +1,1 @@
+git://github.com/jbn/IJuliaPortrayals.jl.git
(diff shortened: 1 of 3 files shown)


jbn:
@StefanKarpinski [pointed out that JavaScript is CamelCased](https://github.com/jbn/IJuliaPortrayals.jl/issues/1). I'm closing this now. I'll increment the tag and reissue a PR in a few.

StefanKarpinski:
Since this was never merged, you could just retag, but the tools is a bit poor for that.

PR #3199, opened 2015-08-27
Pull request/eb7d2a31

jbn:
[v0.0.1](https://github.com/JuliaLang/METADATA.jl/pull/3197) didn't CamelCase JavaScript, as per the convention. v0.0.2 does so properly.


jbn:
@StefanKarpinski pointed out it was never pulled, so I could just retag. I'll do that instead, since it feels cleaner. Sorry for the PR noise, folks.

PR #3238, opened 2015-08-31
Register Conda

Luthaf:
[Conda.jl](https://github.com/Luthaf/Conda.jl) provide access to the [conda](http://conda.pydata.org/) package manager and BinDeps integration. The main points in favor of conda as a package manager are:
- It is easy to add you own software and use your own channel for software distribution;
- It can install packages as non root on Linux;
- It can work the same on all the OS, provided the binaries have been uploaded.

Fixes https://github.com/JuliaLang/BinDeps.jl/issues/151


mlubin:
Might interest @mcg1969

Luthaf:
Fixed the REQUIRE file !

mcg1969:
This looks great, folks. I'll share it with my colleagues.

jiahao:
Now to package Julia through conda...

tkelman:
> Now to package Julia through conda...

Meh. If someone really wants to do it I won't complain, but I don't see much point in this. Conda as a BinDeps provider should be useful and may help with the IJulia installation situation or other packages with Python dependencies, but for non-Python dependencies the conda ecosystem is unlikely to be compatible with much of what Julia needs.

Luthaf:
> but for non-Python dependencies the conda ecosystem is unlikely to be compatible with much of what Julia needs.

Even being Python-based at first, conda is language agnostic, and you can use it for arbitrary dependencies. It will still install a full Python distribution, but only because conda is written in Python, like Homebrew needs a Ruby installation.

mcg1969:
I agree that there would be merit in at least investigating the possibility of Conda as a parallel packaging and/or installation system. They are already doing that for R, and it works well for me. I certainly prefer it to R's own system.

tkelman:
Theoretically, sure. But almost all of the libraries that are up and available on conda right now are designed to work with Python. Meaning using the same compiler as CPython was built with, as one example. I've yet to meet someone in person who has successfully packaged anything unrelated to a python package via binstar or anaconda.org or whatever they're calling it now.

mcg1969:
Keep in mind that the reason Miles included me here is because I now work for Continuum. Anaconda is not my responsibility, but I am certain a satisfied user. So by all means, take my opinion here with a grain of salt, but if there is anything lacking about Conda that would make it prohibitive to provide package management for Julia, I'd be interested to know, and to see if it can be rectified

tkelman:
Ah, was not aware of that. Congrats.

Conda solves the "needing a compiler at install time" problem that Python had. Julia packages only have that problem on Linux, which is why I think Conda.jl as a Julia BinDeps provider (as opposed to using the Python reference implementation of conda as a Julia and/or Julia-packages installation driver, which I have little interest in pursuing) will actually be most useful on Linux as it could allow us to have a distribution-neutral binary installation mechanism.

The problem we have (from my standpoint) is more in cross-platform compile-time dependency management and automated building of binaries. Conda doesn't solve that problem at all - it requires a team to build things, and the binstar/anaconda.org tooling that's being worked on to make it a more open and user-extensible system has been very Python-centric in my limited experiments with it. So far most of the suggestions I've heard have been "try to use AppVeyor as a build platform" which is not a viable solution IMO. Since we don't have the constraint of ABI compatibility with MSVC as CPython does, I'd rather cross-compile Windows binaries as we've been doing with WinRPM, and our OSX setup with Homebrew.jl has also been working pretty well.

These are hard problems, and I'm not sure solutions built to solve one set of problems are going to be applicable to a different ecosystem where the requirements are different. It's worth experimenting with though, I might be wrong here.

mcg1969:
Lots of good info there, thanks! I've pointed the Anaconda folks here. Not sure if they will have anything to add but I am certain they will take it in ;-) The build challenges are indeed a significant issue.

tkelman:
https://github.com/JuliaLang/BinDeps.jl/issues/151 (or a planning/discussion issue on the Conda.jl repository?) would probably be a better place to continue discussion on this if anyone would like to do so, a bit more past context is there as well.

asmeurer:
Let me try to answer some of your specific issues (and please let me know if you have more)


>  Meaning using the same compiler as CPython was built with, as one example.

But this has nothing to do with conda. If using different compilers gives different issues, that will always be a problem for anyone who does it. If anything conda *solves* this problem because it lets you distribute a collection of binary packages that work together. 

> I've yet to meet someone in person who has successfully packaged anything unrelated to a python package via binstar (or anaconda.org now?).

There are plenty of non-Python things on binstar. Aside from full R support (anaconda.org/r), there are tons of C libraries, as this is one of the best use-cases for conda. Even if people are ultimately using many of these with Python, they are not Python libraries themselves, and are usable outside of conda. There is also some support for perl at https://anaconda.org/dan_blanchard (although I don't think he maintains it anymore). 

> Conda solves the "needing a compiler at install time" problem that Python had. Julia packages only have that problem on Linux, which is why I think Conda.jl as a Julia BinDeps provider (as opposed to using the Python reference implementation of conda as a Julia and/or Julia-packages installation driver, which I have little interest in pursuing) will actually be most useful on Linux as it could allow us to have a distribution-neutral binary installation mechanism.

This is a problem with any language. As soon as you start linking against C libraries, you need to install those libraries to use them. Julia, as a scientific language, assumedly needs to do this a lot. Let me ask you a simple question: how do people interface with HDF5 with Julia? Can people just use the library without going compiling and compiling HDF5, regardless of whether they are on a base OS X, Linux, or Windows machine? 

> The problem we have (from my standpoint) is more in cross-platform compile-time dependency management and automated building of binaries. 

It is true that conda lives at a higher level of abstraction than most language level package managers. Conda is 100% cross-platform, in the sense that it can run anywhere that Python can (OS X, Linux, Windows, Raspberry Pi, ...). It doesn't do anything about cross compiling, but that's because it lives at a higher level of abstraction than that. 

The benefit of conda is twofold. For users, it lets them install things without ever having to compile *anything*, and it just works.  For package maintainers, it makes it easier to provide this to users by making it possible to build the package using reproducible recipes. 

> Conda doesn't solve that problem at all - it requires a team to build things, and the binstar/anaconda.org tooling that's being worked on to make it a more open and user-extensible system has been very Python-centric in my limited experiments with it. 

It doesn't require a "team". Search GitHub for repositories called "conda recipes". 

And while it's true that most effort of specific packages that have been built has been around Python or libraries needed for Python libraries, conda was designed from the start to be completely Python agnostic, so that it could be used with any language. 

> So far most of the suggestions I've heard have been "try to use AppVeyor as a build platform" which is not a viable solution IMO. Since we don't have the constraint of ABI compatibility with MSVC as CPython does, I'd rather cross-compile Windows binaries as we've been doing with WinRPM, and our OSX setup with Homebrew.jl has also been working pretty well.

The Windows compiler you use is irrelevant. If you don't depend on the MSVC runtime, then that's great. That just makes your life as a builder that much easier. It doesn't make it any easier for your users to manage dependencies. 

If you had conda packages for Julia and Julia packages, installing any one of them would be two steps:

1. Install Anaconda/Miniconda
2. `conda install <whatever>`

to get both the package *and* Julia itself. And it would be those exact same steps on OS X, Linux, and Windows. 

This doesn't even get into other questions, like virtual environments, which conda makes trivial (what is Julia's virtual environment story?).

tkelman:
> But this has nothing to do with conda. If using different compilers gives different issues, that will always be a problem for anyone who does it. If anything conda solves this problem because it lets you distribute a collection of binary packages that work together.

True, but conda's not really built around doing separate sets of binary packages that work together as sub-groups right now. Either we work with what's there, or we'd need to build and upload a whole new set of non-MSVC binaries in the Windows case and be our own separate subset. I don't see a lot of advantage in going through conda for that - it would add a dependency on Python to something we can already do without it (or you'd need to reimplement conda in Julia, which would be tractable but a fair amount of work).

> For package maintainers, it makes it easier to provide this to users by making it possible to build the package using reproducible recipes.

> Search GitHub for repositories called "conda recipes".

I have. My doubts are because I don't think hand-writing batch files for windows builds is a satisfying solution to the "reproducible recipes" problem, especially when the package/library authors frequently don't use Windows themselves. You have to admit there's way less capability there than what a homebrew formula or a spec file provides.

> This is a problem with any language. As soon as you start linking against C libraries, you need to install those libraries to use them.

Julia's notion of "install" is vastly simpler than most other C FFI's, we don't "link against" in the classical sense. We dynamically open just the shared library and the user doesn't really have to worry about it.

> how do people interface with HDF5 with Julia?

`Pkg.add("HDF5"); using HDF5` - https://github.com/timholy/HDF5.jl

> Can people just use the library without going compiling and compiling HDF5, regardless of whether they are on a base OS X, Linux, or Windows machine?

On OS X and Windows, yes (on Linux we punt to system package managers for the HDF5 case, but for many other packages we fall back to building from source there). We download the dylib or dll and ccall into it without having to compile anything. The install-time dependency management to do that is here - https://github.com/timholy/HDF5.jl/blob/master/deps/build.jl and my feeling is that the (platform-specific) tools we already have for setting that up for a new package can be easier to get up and running than building via conda. But I would need an actual counterexample of conda being used to build and provide new binaries for a Julia package to substantiate that feeling.

> (what is Julia's virtual environment story?)

Nothing standardized yet, but I don't think this is the same issue as binary dependency management.

asmeurer:
>  it would add a dependency on Python to something we can already do without it 

As @mcg1969 pointed out to me, wouldn't anyone using Jupyter notebook need Python anyway?

> My doubts are because I don't think hand-writing batch files for windows builds is a satisfying solution to the "reproducible recipes" problem, especially when the package/library authors frequently don't use Windows themselves. You have to admit there's way less capability there than what a homebrew formula or a spec file provides.

tkelman:
> wouldn't anyone using Jupyter notebook need Python anyway?

Yes, and this is the hardest part of getting IJulia installed cross-platform when compared to other Julia packages. Our life would be easier there if Jupyter were a C library. Conda.jl might make this situation better, as I said first thing. But not everyone uses Julia through IJulia. I deliberately avoid touching Python in my day-to-day work, but I'm a little strange like that.

> If you have some way of building a Windows package (and testing it?) without having Windows, then it can very easily be used when building a conda package.

I'd refer you to my JuliaCon talk here if the video had been uploaded yet. Build via cross-compiling on the openSUSE build service, test via AppVeyor is my recommended approach and I've walked through several Julia packages here without needing to touch a Windows machine to set it all up. Unless we were to teach conda how to do cross-compilation, I don't see how I would use it for this. anaconda.org still doesn't provide automated build servers for Windows or OS X on the open source plans, right? Someone has to run the build recipe somewhere (and deal with getting compilers, cmake, etc manually installed before doing so).

> manually downloading a C library dependency and then requiring the user to set LD_LIBRARY_PATH to use

No, we automate that and there's no path modification. The package developer has to find or make a build of the library somehow, somewhere, and set up the scripts that download the library to the user's machine, but from a user's standpoint all they have to do is `Pkg.add("HDF5")` and the rest is taken care of for them.

conda does a lot of things right, and it ends up working more robustly for users once it's been set up than the systems we're using in Julia right now (BinDeps has a bunch of issues and needs a rewrite, but it usually works). I just don't think it's a silver bullet to solve the initial setup problem, or that putting Julia and Julia packages into conda would be of much benefit. Would love to be proven wrong here though.

Luthaf:
> Unless we were to teach conda how to do cross-compilation

We can. Conda packaging is as simple as 
```
take snapshot n°1 of PREFIX
build/install  <= Cross-compilation can be used here
take snapshot n°2 of PREFIX
package diff n°1 .. n°2
```

> Someone has to run the build recipe somewhere (and deal with getting compilers, cmake, etc manually installed before doing so).

That is the main issue for me. I can cross-compile for Windows, use Travis for Linux, but I still have to build OS X on my own laptop.

asmeurer:
We could add this feature to conda-build if it's wanted. I don't think it would be hard (I think the main issue would be tricking conda into installing the build dependencies from the other platform).

PR #3243, opened 2015-09-01
Pull request/027d0ae8

Kelvyn88:
Minor reviews for ODEs Package


jiahao:
In general, we strongly encourage new developers to contribute to an existing package rather than contribute a new package whose domain overlaps strongly with the existing one. In this case there is a fairly developed [ODE.jl](https://github.com/JuliaLang/ODE.jl) package. Why is there a need for a second one?

jiahao:
Also, I don't think you meant to open so many pull requests.

Kelvyn88:
Ok, I understand. Please close this pull request. I'll make a pull request to the ODE package later, oce I get the newest method. Thanks.

jiahao:
I'm sure ODE.jl could use more attention. Thanks!

Kelvyn88:
Thanks for your comments!

PR #3244, opened 2015-09-01
Register ConnectSDK

Lanzafame:


tkelman:
Looks like you have the same commit getting multiple tags here, but I think that's harmless.

Lanzafame:
Yeah I wasn't sure about going and deleting the accidental extra from METADATA.jl. I haven't had the best lucky swapping between Pkg and git commands so I thought I would just leave them.

PR #3256, opened 2015-09-02
Tag GitHub v1.0.0

WestleyArgentum:
Calling this 1.0 might be a bit of a stretch, but it's out there and usable so I'm going with it.


IainNZ:
:+1: I've been using it lately (http://iaindunning.com/blog/juliaissuecount.html), much better than hitting the API myself. Converting pkg.julialang.org to use it right now actually!

WestleyArgentum:
Awesome! That's great to hear haha

PR #3266, opened 2015-09-03
Deprecate Meddle and Morsel

IainNZ:
This PR deprecates Meddle (https://github.com/JuliaWeb/Meddle.jl) and Morsel (https://github.com/JuliaWeb/Morsel.jl) as of Julia 0.5.

Both packages have been abandoned by their creators, and have no maintainers. As a result, they are often broken (even on release Julia). Their existence and implied support is a burden on the very few people who help out with JuliaWeb packages.
I've put a note in the README encouraging new contributors to revive them, and pointing them towards alternatives like Mux.jl


IainNZ:
I'll let this sit for a day or two in case of objections.

IainNZ:
One possible issue is that Morsel still seems to receive some patches, but is also broken. So doing this would mean doing the same for Morsel.
Recent contributors to Morsel include @timholy , @StefanKarpinski , @JockLawrie, @jiahao, and @yfractal - are any of you willing to take maintenance responsibility for Morsel and Meddle, since they are a joint deal anyway?

IainNZ:
Also @malmaud, who has been making JuliaWeb work better.

malmaud:
I am not using it/willing to maintain it.

ViralBShah:
I routinely run into people using Morsel. I am not sure this is a good idea. If we are deprecating Morsel and Meddle, what are the alternatives for those who are using it?

malmaud:
One possibility I was thinking of is instead of formally deprecating them (and some other abandoned projects in umbrella organizations), we just move them out of the organization to make it clear to newcomers that they are no longer 'blessed'.

IainNZ:
@ViralBShah, that isn't a good reason to not send a signal that they shouldn't be using it. The best alternative is @one-more-minute 's https://github.com/one-more-minute/Mux.jl I think, followed by one of the users stepping up to maintain it.

In the light of day, I see that 0.4 upperbound is too aggressive probably. I think I'll change it to `0.5`, and update the README on Morsel as well.

@malmaud whynotboth.jpg? I agree - is making a `JuliaAbandoned` organization unreasonable?

IainNZ:
Alright, updated to cap out Morsel too.

JockLawrie:
Hi all,
As a data point I use Meddle in a [project](https://bitbucket.org/jocklawrie/skeleton-webapp.jl/src/ddd8f8fc23fa4a0018f7b621852f2c785498005f/docs/outline.md) I am writing to learn web development in Julia (I'm a statistician, not a web developer). I use the same project as a basis for internal projects at work. I'll replace the use of Meddle with Mux, looks pretty straightforward.

That said, while the Mux API looks easy to use I find the Meddle source easier to follow...a sequence of functions that each take in (req, res) and return (req, res) is an easy concept for a beginner to grasp. It also enables the response to be modified as it passes through the stack, rather than just at the end. Perhaps I am misunderstanding Mux due to a lack of documentation? If so I'd be happy to contribute some documentation to Mux from the beginner's perspective.

I stopped using Morsel months ago because I found the lower level libraries, namely HttpCommon, HttpServer and Meddle, quite accessible as is thanks to Julia's clean syntax. Similarly, having played with web dev in Python and Go (not in any real depth) I preferred Go's approach of having decent standard net/http libraries to Python's framework approach, at least for my purposes. I think Julia should (and is by virtue of no framework devs!) take Go's approach...only Julia has nicer syntax and is easier to follow.

Cheers

IainNZ:
Thats a good data point @JockLawrie. I'm sure @one-more-minute would appreciate any documentation or examples for Mux.jl. Focussing any efforts on improving the base packages of HttpCommon and HttpServer would definitely the best use of volunteer coding time right now, they really need to come first.

I think this also makes me confident doing this. Morsel and Meddle will still be available on 0.4, it'll just be clearly signaled in the README and, thanks to this PR, on `pkg.julialang.org`, that these are in maintenance mode (not even that really!) and that if you are starting something new, look elsewhere.
If a miracle happens and someone takes "ownership" of these packages, its as easy as tagging a new release.

one-more-minute:
Yup, more docs would be great. If you start playing around with it and have any trouble I'm more than happy to help out, too.

JockLawrie:
Will do. I probably won't get to it for a week or 2, but using the Julia web stack at work provides incentive to get it done.

PR #3268, opened 2015-09-03
Register MbedTLS, a wrapper around the C mbed library

malmaud:


tkelman:
Should probably be MbedTLS, shouldn't it? Isn't mbed a larger umbrella organization with an OS and other things?

malmaud:
Yes, name has been changed.

PR #3273, opened 2015-09-03
Register MariaDB

maartenheremans:
Wrapper around the MariaDB c connector


kshyatt:
Two questions:

1) Will this work with any version of `Docile`? You don't need to specify a version number for the package, if so. You can just have `Docile` in `REQUIRE`.
2) An expanded `README` about what MariaDB is/what the package does and how to use it might be nice?

MichaelHatherly:
Since it appears to be a 0.4-only package Docile won't be necessary and can safely be removed as a dep. If you'd like to support Julia 0.3 as well then Docile ``0.5`` or higher would be my suggestion.

Also, with the current ``@doc`` usage in the package no documentation is being stored since the ``->`` syntax is missing. You can change all the

```julia
@doc doc"""
...
"""
```
usages to bare docstrings, ie.
```julia
"""
...
"""
```

and that will work fine.

There's also a missing ``@`` on https://github.com/Dynactionize/MariaDB.jl/blob/276180433b0794c6beb5248eb27ffff9ab380ab8/src/mariadb/types.jl#L117 I think.

maartenheremans:
Thanks for the reviewing and comments.

I'll update according to the suggestions and add information to the README.md file

maartenheremans:
I updated the package according to the comments above.

kshyatt:
@maartenheremans looks like you need to rebase this PR on top of latest `metadata-v2`. Does this look ok to everyone?

ViralBShah:
Should we call this MySQL.jl, which is what will make this most discoverable? Also, we at Julia Computing are also soon about to release a similar package, with somewhat more functionality. We'll figure out how to merge efforts as soon as that happens.

StefanKarpinski:
I think Oracle has thoroughly tainted the MySQL name at this point. The open source project is now called MariaDB so I'd argue we should go along with that.

johnmyleswhite:
@StefanKarpinski: I think people have different things in mind when they talk about MySQL and MariaDB. In particular, I think people who are using MySQL are generally committed to the old GPL releases of MySQL prior to Oracle's switch to the AGPL, whereas people using MariaDB are thinking about a still active open source project that might have substantive changes in the future.

ViralBShah:
The other thing is that many people still don't seem to know about MariaDB.

maartenheremans:
@ViralBShah Let me know when you are to release your package. We'll be happy to merge the efforts.

ViralBShah:
Very shortly. Thank you. Will announce it - and then let's merge all efforts and then register a common package in METADATA. @johnmyleswhite also has some stuff on a branch, he said.

ViralBShah:
Ok, here it is. I am not announcing it on julia-users yet. There's a lot that needs to be done, APIs that are awkward and such, but at least one level of performance enhancements has been done - and we are using it regularly.

https://github.com/JuliaComputing/MySQL.jl

ViralBShah:
cc: @nkottary

maartenheremans:
I'll have one of our employees test drive this out. On first sight it does seem to be a more elaborate implementation than the one we had.

ViralBShah:
Is it ok to close this PR and do a new one once we merge the efforts?

maartenheremans:
Absolutely.

hayd:
Any reason to choose GPL over MIT for MySQL.jl. (Are you forced to use it? ... I find this is a real head scratcher: http://www.mysql.com/about/legal/licensing/foss-exception/)

maartenheremans:
@hayd That's a good remark. That is one of the first things we verify on selecting packages (didn't look into it yet on this one). GPL is just too limiting to allow usage in a commercial environment

PR #3289, opened 2015-09-05
Drop support for 0.3 from webstack master

malmaud:


quinnj:
woot
I'm getting excited

tkelman:
I would normally name branches release-#.# for the version number of the package rather than the version number of julia, but good enough

malmaud:
That does make more sense - I can just rename the new branches.

PR #3302, opened 2015-09-06
Deprecate Soundex.jl

IainNZ:
First and only commit & version July 2013, doesn't work on Julia 0.3

RIP Soundex.jl

@johnmyleswhite


johnmyleswhite:
Worth getting rid of

PR #3304, opened 2015-09-06
Deprecate Resampling.jl

IainNZ:
Also managed to include #3303, woops.

Last commit Nov 14, 2013. Package does load on 0.3, but it doesn't work.

RIP Resampling.jl

@johnmyleswhite


johnmyleswhite:
Also worth getting rid of.

PR #3305, opened 2015-09-06
Deprecate ProjectTemplate.jl

IainNZ:
PkgEval has recorded tests passing on this (although now it doesn't even try to find them), but last commit was Feb 2014, and it is unsupported.

RIP ProjectTemplates.jl

@johnmyleswhite


johnmyleswhite:
Yes, please.

PR #3306, opened 2015-09-06
Deprecate Polynomial.jl

IainNZ:
From README:

> This package has been deprecated in favour of https://github.com/Keno/Polynomials.jl (which started out as a fork of this package with the coefficient order reversed)

Been this way for ~ 1 year.

RIP Polynomial.jl

@vtjnash


milktrader:
Is there a thread about METADATA deprecations? Looks like a good project!

PR #3307, opened 2015-09-06
Deprecate RunTests

IainNZ:
Last commit > 1 year ago, last tagged version ~1.5 years ago, doesn't pass its tests on 0.3 or 0.4.

RIP RunTests.jl

@burrowsa


tkelman:
MetaTools had a dependency on this, so should probably also be deprecated.

PR #3310, opened 2015-09-06
Deprecate systemimagebuilder

rened:
Set a max ceiling of 0.3 for SystemImageBuilder - we don't need it on 0.4 any more.

### DIFF
--- a/SystemImageBuilder/versions/0.0.1/requires
+++ b/SystemImageBuilder/versions/0.0.1/requires
@@ -1,2 +1,2 @@
-julia 0.3
+julia 0.3 0.4
 FunctionalData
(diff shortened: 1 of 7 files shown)


rened:
@IainNZ could you please take a look whether this is what you meant?

IainNZ:
This might do something strange, because it means `0.3 <= julia < 0.3`, so if someone has it installed then its going to complain until they uninstall it. Normally I'd just do `julia 0.3 0.4`.

rened:
Ok, but I though it wreaks havoc on 0.4? What about `julia 0.3- 0.3`?

rened:
Ok, i'll just to `0.3 0.4`

IainNZ:
`julia 0.3 0.4` won't let it be installed on `Julia 0.4`, but still install on `0.3`

rened:
aaahh, got it.

IainNZ:
Thanks for your patience :D

rened:
no no, thanks for your help!

milktrader:
The floor/ceiling syntax is confusing because it's the down to the floor *and* including but up to the ceiling and *not* including.

PR #3316, opened 2015-09-06
Register FastGaussQuadrature

dlfivefifty:
FastGaussQuadrature is a package to calculate Gauss Legendre, Jacobi, Hermite, Radau and Lobatto  quadrature rules in O(n) operations.


tkelman:
You'll need to tag a version before this will be available via `Pkg.add`. Also probably a good idea to turn on Travis testing at some point.

dlfivefifty:
OK, I tagged a version and will merge once the tests pass.

PR #3324, opened 2015-09-08
Added lowest supported Julia version to REQUIRE

dfdx:


tkelman:
Thanks! Be sure to do so in the packages themselves too, so it'll get reflected in future tags.

PR #3326, opened 2015-09-08
Register SoftConfidenceWeighted: v0.1.0, v0.1.1

IshitaTakeshi:


milktrader:
Looks solid. Tests are wrapped in methods, which is interesting. The test methods make assertions and get called at the end of the file. 

What's up with branch is out-of-date? This PR is 10 hours old.

IshitaTakeshi:
It seems TravisCI not have Julia4.0.
I'll edit to make it compatible with Julia3.0.

milktrader:
I think the branch is out of date is related to recent deprecations, but now it's greened up. Not sure what happened earlier. Travis *does* have 0.4-dev btw. You do not need to change it to support 3.x if you'd rather not.

PR #3336, opened 2015-09-09
Tag UnicodePlots v0.0.3

Evizero:
Lots of improvements: better plotting range. automatic annotation. functions are type friendlier. ability to specify plot range

cheers


milktrader:
Nice!

PR #3338, opened 2015-09-10
Register BufferedStreams and Libz

dcjones:
Fast zlib bindings and stuff. :bouquet:


jiahao:
woot

quinnj:
swiggity swooty

quinnj:
PS, I've been using this for a couple days now and it's great.

kshyatt:
> swiggity swooty

bird

jiahao:
I think we can now start thinking about deprecating GZip.jl and Zlib.jl.

dcjones:
That's my hope. It definitely does everything Zlib.jl does, and I think everything GZip.jl does except seeking in gzip files (which isn't that useful, and not hard to add). These are also julia 0.4 only, but probably wouldn't be hard to backport.

quinnj:
I quickly looked at backporting Libz to 0.3 the other day; the main hurdle is the use of `Ref` (or `RefValue` rather) for the zstream field.
What's the best practice for Ref compat? I know it's not that much code in Base, but are there codegen reasons why we couldn't just include that code in Compat.jl? @vtjnash

tkelman:
A one-element array might be a hacky and possibly-slow way of approximating it.

Not a big fan of adding even more confusion to the namespace here, is there a reason this couldn't have been done as a major-version rewrite of one of the existing packages?

quinnj:
It's a fairly different interface than the other two packages. I think it's probably easier to just release a one, final package and have people organically adopt than try to shoehorn a complicated deprecation of an entire interface.

PR #3348, opened 2015-09-11
Tag GtkUtilities v0.0.3

timholy:
Marked as 0.4-only. Not because it couldn't work on 0.3, but because I'm not all that interested in going to the effort to support older versions of julia.


tkelman:
If past tags did support 0.3, I'd like to encourage bumping at least the minor version number of a package to make it easier to bring back 0.3 support on a branch. I suppose in this case it could probably be done on master of the package if someone wants to make it work though.

timholy:
At least in terms of registered packages, it didn't have any dependents. (As of this morning it now has one, ProfileView.)

tkelman:
I wasn't meaning just for the purposes of dependents' `REQUIRE`, but also the ability to have a non-overlapping version number if you needed to tag a branched-off 0.3-supporting version. Or an 0.3-bugfix version, say if ProfileView on 0.3 hits a bug that needs to be fixed in an 0.3-supporting version of GtkUtilities. This'll be fixable in other messier ways if it comes to it, but version numbers are cheap as stevengj has said.

timholy:
Will do that if someone adds julia 0.3 support and then we decide to abandon that later.

PR #3359, opened 2015-09-11
Register Immerse

timholy:
Immerse sits on top of Gadfly and adds some new forms of interactivity. Documentation is nonexistent at the current time, but I plan to fix that by the time I release the next tutorial video.

Let the name bikeshedding begin.


tkelman:
Shouldn't it depend on everything it imports? https://github.com/JuliaGraphics/Immerse.jl/blob/dee2115d8e0a2274610142c56c97b5f796442018/src/Immerse.jl#L3

timholy:
Good catch. Will fix.

IainNZ:
This is a many-approachs-to-a-big-problem type package, so name is fine with me as-is.

quinnj:
+1

milktrader:
I want to use this name for a method in a Liquid State Machine package :smile:

tbreloff:
Don't worry... You can always name it Drown.jl

timholy:
> I want to use this name for a method in a Liquid State Machine package 

Thank goodness for lowercase, then :smile:.

PR #3361, opened 2015-09-11
Register Plots: v0.0.1, v0.0.2

tbreloff:


tkelman:
Is this ready or necessary to put in metadata just yet? Given the namespacing and discoverability issues we still haven't solved, I think having this registered under such a generic name at such an early stage is not ideal.

sjkelly:
From an organizational perspective, I think generic names like this should generally be reserved for groups with collective desicive abilities instead of personal repos.

tbreloff:
From my perspective, its ready to be in metadata, but certainly there's a lot on the todo list. If the concern is about org vs no-org, I can move the repo over to JuliaPlot, since I was planning to do that eventually anyway. I do think the design/framework is something that deserves the name, but of course I'm biased. Let me know what you think I should do, and in the meantime I'll keep working on it.   My thinking in publishing now is that plotting is a big sore spot for new users, and we'll be getting a lot of them in the next few months because of the 0.4 release. New users are not likely to use Pkg.clone, so most won't know about the package at all.

tkelman:
Until demonstrated otherwise via maturity and usability, I think there's skepticism that having a meta-package called "Plots" that tries to provide a uniform interface to every other plotting package would do much to solve the "sore spot." You're not requiring any individual other packages, so `Pkg.add("Plots")` by itself wouldn't work.

tbreloff:
I do understand the skepticism, but until we have some sort of two-tiered
metadata as was discussed at JuliaCon and elsewhere, there's not a good way
to get enough broad usage for the necessary "maturity and usability"
without adding to metadata.   If it was moved to the JuliaPlot
organization, would the name be ok?  If everyone's against the name, do you
have any suggested alternatives?

tkelman:
We may as well let this go in for now as it is and try to deal with the namespacing/curation issue soon.

johnmyleswhite:
+1 to merging as is

milktrader:
+1

tbreloff:
Thanks for the +1's.  I've got a ton of changes that are just about ready
for a new tag (new backend UnicodePlots, supporting subplots, histograms,
and a bunch of other stuff).   Can I just Pkg.tag("Plots") and
Pkg.publish() or will that screw up the repo since this one isn't merged?
Should I wait to tag it?  Sorry I'm new to the metadata shuffle.

Evizero:
+1
I would like to use this package in my current efforts as a base plotting library (for example to provide custom plots for SVM fits that show the support vectors and margins)

tbreloff:
So... I've tagged 0.0.3 locally, and would like to tag 0.0.4.  Since this
PR hasn't been merged, should I somehow cancel it and reset my forked
metadata?  I don't want to screw anything up or create unnecessary merge
commits.  Thanks for any help... I'm still a metadata newbie.

On Sun, Sep 13, 2015 at 12:21 PM, Christof Stocker <notifications@github.com

timholy:
Sorry no one noticed! Do a `Pkg.update()` and then you can publish both the 0.0.3 and 0.0.4 tags.

tbreloff:
Thanks Tim

PR #3362, opened 2015-09-11
Tagging JSON, Iterators, and DataStructures

timholy:
These packages are causing precompilation-related failures. They just need a tagged release that contains the precompile directive, which is present on the master branch in all cases.

Unfortunately this scoops up my registration of Immerse, too, so don't merge this until that's had a little more time to collect any complaints.


milktrader:
I like the name Immerse

PR #3367, opened 2015-09-12
Tag TimeSeries v0.5.11

milktrader:
Final version that supports Julia 0.3 (and no longer supports Julia 0.4) -- Next tag will only support Julia 0.4


milktrader:
Thanks @tkelman

tkelman:
Welcome. Remember to bump at least the minor package version when you raise REQUIRE to 0.4.

milktrader:
The plan is to go to 0.6.0, is that what you mean?
 
And the REQUIRE will be:
````julia
julia 0.4 0.5-
````

tkelman:
Yeah

milktrader:
Yep, that's the plan discussed [here](https://github.com/JuliaStats/TimeSeries.jl/issues/202).

I'm finding using the Compat package a bit difficult to keep track of so I'm isolating support for different Julia's, like @brian-j-smith is doing with Mamba.jl. 

I can now also isolate other downstream packages such as Quandl, MarketTechnicals, MarketData to depend on a specific version of TimeSeries.

PR #3371, opened 2015-09-13
Tag TimeSeries v0.6.0

milktrader:
First version to support only Julia 0.4

### DIFF
--- a/TimeSeries/versions/0.6.0/requires
+++ b/TimeSeries/versions/0.6.0/requires
@@ -1,0 +1,2 @@
+julia 0.3 0.4-
+Dates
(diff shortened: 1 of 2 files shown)


tkelman:
According to REQUIRE this is still only 0.3?

milktrader:
Thanks that was a mistake.

milktrader:
Is Julia 0.4-rc1 supported with `julia 0.4` or `julia 0.4-`?

tkelman:
RC1 still requires `0.4-` I believe

milktrader:
Yeah, I think so because I get this warning in Pkg.update()

````
WARNING: julia is fixed at 0.4.0-rc1 conflicting with requirement for TimeSeries: [0.4.0,0.5.0-)
````

PR #3375, opened 2015-09-14
Tag TimeSeries v0.6.1

milktrader:
fix REQUIRE to limit 0.4 only (including release candidate 1)


tbreloff:
Technically speaking, this should be v0.7.0, no?

tkelman:
Or you can modify the existing 0.6.0 tag's requires just in metadata.

milktrader:
Version 0.5.11 is the last one supporting Julia 0.3 and 0.6.0 was supposed to be the first supporting *only* Julia 0.4

milktrader:
@tkelman I could do that but I've lost commit privileges to METADATA so tagging this is probably easier. And version 0.6.2 is coming in right behind with some nice changes to support new operators

tkelman:
Right, so since 0.6.0 had its REQUIRE messed up, that is not currently the case. If you modify the file https://github.com/milktrader/METADATA.jl/blob/pull-request/158859a3/TimeSeries/versions/0.6.0/requires with another commit in this PR to make it consistent with 0.6.1 then things will work, and Julia 0.3 users will stay on the 0.5.x set of TimeSeries.jl releases.

tkelman:
Right now v0.6.0 of TimeSeries.jl claims to support v0.3 of Julia. So users of Julia 0.3 would upgrade to 0.6.0 of TimeSeries.jl and be stuck there forever even if future releases of 0.5.x get tagged with bugfixes. Changing the requires file in metadata on the existing tag is one way to fix that - the other way would be making a new 0.7.x minor release series for Julia 0.4.

milktrader:
Ok, I can hold off with pushing up 0.6.2 and make it 0.6.1 instead. 

0.6.0 will not work with Julia 0.3 so you're right here to modify this tag instead of starting a new one ...

PR #3376, opened 2015-09-14
Update requires

milktrader:
fix wrong versions of Julia in requires

### DIFF
--- a/TimeSeries/versions/0.6.0/requires
+++ b/TimeSeries/versions/0.6.0/requires
@@ -1,2 +1,1 @@
-julia 0.3 0.4-
-Dates
+julia 0.4- 0.5-


tkelman:
ref https://github.com/JuliaStats/TimeSeries.jl/issues/202 and https://github.com/JuliaStats/TimeSeries.jl/blob/master/.travis.yml

milktrader:
No I'd like this to work *only* on 0.4-rc1 and up to but not including 0.5.

Support for 0.5 will be provided under a different series of tags.

milktrader:
The problem with this approach (simply editing the requires file in METADATA) is that the METADATA version doesn't match the repositories version. To do this right, I'll need to move a different commit over to the 0.6.0 tag on the repo, and then change the sha1 in METADATA once that's done.

Since I originally tagged 0.6.1 as the solution to a bad requires file, I'll need to fix that on the repo side as well.

milktrader:
Alright, I've deleted the repo's v0.6.1 (this was never merged into METADATA). Next I'll move a reasonable commit to v0.6.0 and go from there.

milktrader:
Closing this and opening a new one that has a fixed requires and a new sha1 to match the repositories reference to v0.6.0

milktrader:
Well, it looks like there is one PR per file change so reopening this. Both this and the PR to change the sha1 should fix the problem.

tkelman:
I think changing the sha of existing tags is to be avoided, people may have already updated. It's okay for metadata and the package to disagree on require, only metadata gets used.

milktrader:
The existing sha1 points to a commit where the REQUIRE file is incorrect
````
julia 0.3 0.4-
````
But if this isn't an issue, then I'll close this.

tkelman:
I don't think I'm explaining very clearly. The `REQUIRE` file in the package source doesn't actually get used for dependency resolution, only the content of METADATA does. The `REQUIRE` is used to initially populate the METADATA `requires` but the latter is designed to be updatable after-the-fact.

milktrader:
Ok, copy. 

There is also this (outstanding issue)[https://github.com/JuliaLang/METADATA.jl/issues/1012]

PR #3377, opened 2015-09-14
 update sha1

milktrader:
updating sha1 to match the repositories v0.6.0 tag


milktrader:
This is the issue you'd like to close without merging, right @tkelman ?

tkelman:
I don't think this one is necessary, now that 0.6.0's require has been modified to be not problematic, you could make an 0.6.1 tag that's up to date. Usually tags should be immutable if possible, just dependency metadata around them may need to change after the fact.

milktrader:
Copy, then I'll close this. We'll push 0.6.1 sometime today or tomorrow.

PR #3383, opened 2015-09-14
Tag SimJulia v0.3.7

BenLauwens:
Update docs and add some more examples


tkelman:
You have 2 duplicate tags for the same commit here. Don't think that hurts anything, but just making sure you're aware.

BenLauwens:
I am aware. The tagging of v0.3.6 is accidental ...  You can merge. Thanks! Ben

tkelman:
Okay.

The preference is to avoid merge commits for PR's here, since they should pretty much never conflict with one another.

PR #3387, opened 2015-09-15
Update CovarianceMatrices to v0.0.3

gragusa:
CovarianceMatrices: Heteroskedasticity and Autocorrelation Consistent Covariance Matrix Estimation for Julia.


CovarianceMatrices/versions/0.0.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3+
+StatsBase
```
tkelman: Dont think the + is valid here

gragusa: Straggled one hour on choosing the right REQUIRE and still picked the wrong
one :)

Should it be 0.3? How do I fix this? Close PR, delete the remote tag, fix,
retag, and do a new PR?


Giuseppe Ragusa

tkelman: You can remove the + here and remove from package require separately. No need to re tag

milktrader:
Have you considered contributing to JuliaStats/TimeModels?

gragusa:
@milktrader Do you mean merging `CovarianceMatrices` into `TimeModels` or contributing in general? If you mean the latter, I could have something to contribute. If it is the former I am not sure covariance matrices are to be used in time series context. As such, it should live on a separate package.

PR #3390, opened 2015-09-15
Register ParallelSparseMatMul

madeleineudell:


jiahao:
Amazing that this package hasn't been registered yet? Did you ever figure out the performance regressions from Julia v0.2?

ViralBShah:
How about merging this functionality into DistributedArrays?

timholy:
Since it depends on SharedArrays, not DistributedArrays, I'm not certain that makes sense.

ViralBShah:
I somehow mistakenly remembered from last year that this was distributed memory. Sorry for the noise.

PR #3399, opened 2015-09-15
Tag TextWrap v0.1.4

sjkelly:
cc @carlobaldassi 

This adds support for precompilation on 0.4.


TextWrap/versions/0.1.4/requires
```diff
@@ -0,0 +1 @@
+julia 0.2-
```
tkelman: Is this still being tested on 0.2? If so it should at least remove the `-` since that refers to prereleases.

carlobaldassi:
I think it's better if I tag it myself, so that the repository is tagged too. Will do in a short while (together with ArgParse), thanks.

TextWrap/versions/0.1.4/requires
carlobaldassi: Yes it is, and I just pushed an update to remove the trailing dash.

tkelman: great thanks

sjkelly:
:) Thanks! I forgot this wouldn't tag.

PR #3405, opened 2015-09-16
Register LibHealpix: v0.0.1

mweastwood:
This is a wrapper for the Healpix library.

Healpix defines a pixelization of the unit sphere such that equal-area pixels are arranged in rings of constant latitude. This allows an FFT to be used as part of a spherical harmonic transform.

The Healpix library itself is a mess of C, C++, Fortran and IDL. LibHealpix.jl currently wraps almost all of the C library and select functions from the C++ library (the spherical harmonic transforms `alm2map` and `map2alm`).

I have chosen the name LibHealpix because @ziotom78 is working on a reimplementation of the Healpix library in Julia, where as my package is a wrapper of the already existing library. Additionally, there is a "HEALPix" vs "Healpix" ambiguity. The original authors preferred the former spelling, but the latter seems to be increasingly common. I chose the latter because it's easier to type.

https://github.com/mweastwood/LibHealpix.jl
For more information see: http://healpix.jpl.nasa.gov/
The python wrapper: https://github.com/healpy/healpy


tkelman:
+1, lgtm

jiahao:
Thanks for the writeup. It's very clear what this new package does and why its name was chosen.

PR #3410, opened 2015-09-16
Tag Playground v0.0.2

Rory-Finnegan:


tkelman:
I'd usually bump the minor version of a package when you raise the minimum julia version requirement, just in case bug fixes are needed on an 0.3-compatible version. Here at least there haven't been all that many past versions.

Rory-Finnegan:
That's a good idea; I'll do that next time. Thanks.

PR #3416, opened 2015-09-16
Register SnoopCompile

timholy:
SnoopCompile temporarily modifies the julia compiler to make it log its activities; by parsing the log file you can generate `precompile` directives. This was already used in Gadfly and Compose to cut time-to-first-plot in ~half (https://github.com/dcjones/Gadfly.jl/pull/673 and https://github.com/dcjones/Compose.jl/pull/143).

Repo: https://github.com/timholy/SnoopCompile.jl


ihnorton:
Any reason not to put the patch in mainline? Seems innocuous enough, and would make this usable without a recompile/from nightlies.

timholy:
I guess you're right, it really should hurt anything. I'll submit it and see what folks say.

timholy:
Since 0.4 won't have the mainline patch regardless, I'm going to merge this and then address 0.5-related issues later.

PR #3384, opened 2015-09-14
ROC: Receiver Operating Characteristics calculations

davidavdav:
This is a package for computing the data for ROC curves, and also contain plotting routines. 

There is a wealth of information in ROC curves, and ROC analysis can be used to evaluate probabilistic binary classifiers for their ability to produce calibrated probabilistic predictions.  Popularly speaking, if the classifier says it is 90% sure, it should, on average, actually be 90% of the times correct, and not 91% or 89%.  This packages contains several tools: metrics (Cllr, minCllr, EER convex hull) and graphs (Normalized Bayes Error rate, Applied Probability of Error, Log-Likelihood Ratio, Detection Error Trade-off, traditional Receiver Operating Characteristics). 

There are other efforts for ROC packages, as the package documentation acknowledges, but I believe this implementation outperforms the others in working wil very large data sets (~10^7 data points) while being very robust in the very scarce data region, using the ROC convex hull interpretation.  

One can debate about the short name, but I believe ROC is known in many different disciplines as such. 

This packages is my first with fairly complete in-line documentation.  

It is currently compatible with julia 0.3 and 0.4.

PR #3417, opened 2015-09-16
Register ROCAnalysis: v0.0.1

davidavdav:
This is a package for computing the data for Receiver Operating Characteristics (ROC) analysis, and also contain plotting routines.

There is a wealth of information in ROC curves, and ROC analysis can be used to evaluate probabilistic binary classifiers for their ability to produce calibrated probabilistic predictions. Popularly speaking, if the classifier says it is 90% sure, it should, on average, actually be 90% of the times correct, and not 91% or 89%. This packages contains several tools: metrics (Cllr, minCllr, EER convex hull) and graphs (Normalized Bayes Error rate, Applied Probability of Error, Log-Likelihood Ratio, Detection Error Trade-off, traditional Receiver Operating Characteristics).

There are other efforts for ROC packages, as the package documentation acknowledges, but I believe this implementation outperforms the others in working wil very large data sets (~10^7 data points) while being very robust in the very scarce data region, using the ROC convex hull interpretation.

One can debate about the acronym as part of the name, but I believe ROC is known in many different disciplines as such.

This packages is my first with fairly complete in-line documentation.

It is currently compatible with julia 0.3 and 0.4.


davidavdav:
As suggested by @diegozea [earlier](https://github.com/JuliaLang/METADATA.jl/pull/3384#issuecomment-140774414), there is now a more illustrated description in a [jupyter notebook](http://nbviewer.jupyter.org/github/davidavdav/ROCAnalysis.jl/blob/master/ROCAnalysis.ipynb), which is also mentioned in the [online README](https://github.com/davidavdav/ROCAnalysis.jl/blob/master/README.md#synopsis).

tkelman:
The tagged commit here still had the old name - https://github.com/davidavdav/ROCAnalysis.jl/tree/e1149ea7a35cb542ba3027902a112636f4b3ac47/src

Could you replace the sha with the current master?

davidavdav:
Sorry, I don't really understand how this could have happened.  Perhaps a `git reset HEAD^^` of my own metatdata in order to remove the original `Pkg.register()`.  I finally just edited the file on github, local edits were too much out of sync.

tkelman:
Thanks. This looks like it'll work on 0.4 now but will be broken on 0.3, ref https://travis-ci.org/davidavdav/ROCAnalysis.jl/jobs/81132051 - you aren't using Compat.jl yet.

davidavdav:
OK, sorry again, my bad.  Forgot to test, didn't look at Travis.  Doing too many packages at once for the upcoming 0.4 changes, but that is not an excuse...  Thanks for the thorough checks. 

Rebased the sequence of commits.

PR #3419, opened 2015-09-16
Register Bio: v0.1.0

dcjones:
:syringe: :microscope:


jiahao:
`BioBase`?

I can't help but think of the epic `Stats` -> `StatsBase` rename that happened 1.5 years ago.

dcjones:
I'm open to that. Though I'm actually not familiar with the reason for the `Stats` to `StatsBase` rename.

Ward9250:
I'm also open to BioBase.

johnmyleswhite:
In theory, `Stats` was supposed to be a generic meta-package that would load everything: `GLM`, `DataFrames`, `Distributions`, etc.

dcjones:
After a little thought, we'd prefer to keep Bio.jl. It's possible that it will eventually transition to being a meta-package like Stats, but if so it would make more sense for us to do that gradually by splitting modules out of Bio into separate packages. If we do decide do do that, BioBase will be new package consisting of a smaller subset of Bio. So I think the option will remain open to us without having to break backwards compatibility.

jiahao:
Fine by me. I think the `BioJulia` organization has a good record of package stewardship and so if there are no further objections (say by tonight) go ahead and merge.

quinnj:
+1

PR #3429, opened 2015-09-17
CUDArt: delete v0.1.4 tag, jump to v0.2.0

timholy:
From https://github.com/JuliaGPU/CUDArt.jl/issues/26 I learned I had broken julia 0.3. So this is a minor-version bump, to leave space for the future development of the 0.3 branch.


tkelman:
Hm, what happens if you delete a tag? Would an alternative be to set unsatisfiable bounds for that tag but otherwise leave it in place?

timholy:
I don't know---I thought I'd find out :smile:. Presumably everyone's metadata just tracks the change?

timholy:
I should say I also deleted the tag from the `origin` repo, so there shouldn't be any conflict between `Pkg` and the repo.

tkelman:
If they don't have CudaRT installed or updated to that tag yet then yes. If they do have CudaRT installed and on the 0.1.4 tag, would this cause any trouble? I hope not but would want to test first.

timholy:
It looks like the 0.1.4 tag stays on the local machine. So they'd have a hard time tagging a new release. I guess it would also affect distribution of a new tag?

timholy:
I guess the safest is to make it uninstallable. `julia 9223372036854775807.9223372036854775807` in the REQUIRE file should do the trick, no?

tkelman:
That or making a new tag for 0.1.5?

timholy:
Are you saying we can just skip 0.1.4 altogether? (Including deleting metadata, as here.)

tkelman:
Deleting the metadata is what worries me a little because I don't know what the consequences would be. If the issue is that the existing tag of the package is broken on julia 0.3, couldn't we make a new 0.1.5 tag (even just a copy of 0.1.3), leaving the broken 0.1.4 in place, which users would then upgrade to? I suspect `Pkg.update` should be able to handle tag order not matching chronological history order.

carlobaldassi:
I think the safest thing to do is:

1. Make it uninstallable by editing the requirements as Tim suggested
2. Wait a few days (weeks/months/years) until we can be sure all users which installed it call `Pkg.update()` and got a different version
3. Remove the version from METADATA

tkelman:
Why delete when we could supersede instead?

carlobaldassi:
It probably won't make any practical difference in this case, but as a matter of general principle I think making the version uninstallable is safer since it will automatically install the best possible version for the user (downgrade to 0.1.3 for 0.3 users, upgrade to 0.2.0 for 0.4), and warn the user in case the version is pinned, and avoid problems if they have provided requirements which would not allow 0.1.5 to install, e.g. by `Pkg.add("CUDArt", v"0.1", v"0.1.5")`.
The deletion afterwards is not even necessary, just a matter of cleaning up.
In fact, if there are versions of packages which are known to be hopelessly buggy (as in: don't even load in any version of Julia) I'd recommend making them uninstallable in just the same way (and retiring them after a while).

In case `Pkg.tag` becomes problematic because the 0.1.4 tag was deleted from the repo, one could also do both the "retiring" of 0.1.4 and the duplication of 0.1.3 into 0.1.5. But I believe that tagging 0.1.4 on top of 0.1.3 in the origin repo should be enough to prevent any problems of that sort, even if it's not reflected in METADATA, so all in all I'd say tagging a new version 0.1.5 in METADATA is unnecessary.

PR #3431, opened 2015-09-17
Tag MFCC v0.0.2

davidavdav:
Long since this package was updated. 

 - added minimal consistency test
 - Julia 0.4 syntax
 - inclusion of higher-level feature calculation routines.


tkelman:
Needs a minimum `julia` version requirement

tkelman:
Is there much overlap in functionality here between this and https://github.com/r9y9/MelGeneralizedCepstrums.jl ?

davidavdav:
There might very well be some overlap.  Seems like r9y9 wasn't aware of MFCC. But it is also, as the package says, more general and possibly more geared towards speech analysis, as MFCC's aim simply is to produce (compatible to traditional software) features for speech.  From a quick glance I get the impression MelGeneralizedCepstrums is part of a re-coding of SPTK in Julia, which would explain the parallel deveopment. 

In general, it is very hard to compute exactly the same features in speech, simply because there are so many parameters and nitty-gritty details. 

I'll close this and submit a new "Tag MFCC v0.2..2" because that is the way the moving-the-tag-one-commit-forward worked for mw.  Sorry.

PR #3434, opened 2015-09-18
Register FMIndexes: v0.0.1

bicycle1885:
https://github.com/BioJulia/FMIndices.jl

### DIFF
--- a/FMIndexes/url
+++ b/FMIndexes/url
@@ -1,0 +1,1 @@
+git://github.com/BioJulia/FMIndexes.jl.git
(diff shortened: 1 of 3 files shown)


bicycle1885:
Wait a moment! FMIndexes.jl may be a better name.

dcjones:
Great work, Kenta!

"Indexes" and "indices" are both valid and about equally used (e.g. "indexes" occurs on github 13.6e6 times, and "indices" 14e6 times). So I think either would be fine.

bicycle1885:
Thank you :smile: 
Papers referencing to this data structure seem to prefer FM-Indexes to FM-Indices.
So now I'm feeling like to rename it.
I'm going to rename the name if either is fine.

bicycle1885:
Renamed the package and fixed the URL (it was wrong!).
I think this is ready to be merged.

PR #3441, opened 2015-09-18
Register and tag CRlibm.jl

dpsanders:
This registers the package `CRlibm.jl`, a wrapper around the `CRlibm` library for correctly-rounded
mathematical functions. See the package `README` for details.

Repo:
https://github.com/dpsanders/CRlibm.jl

Cc @simonbyrne


simonbyrne:
Great idea!

dpsanders:
Thanks!

dpsanders:
What's going on with the travis failure? I did everything via Pkg.tag and Pkg.submit

pao:
From the Travis output:

```
ERROR: assertion failed: Invalid url scheme http for package CRlibm. Should be 'git' 
 in error at error.jl:21
 in anonymous at no file:15
 in include at ./boot.jl:245
 in include_from_node1 at loading.jl:128
 in process_options at ./client.jl:285
 in _start at ./client.jl:354

while loading /home/travis/build/JuliaLang/METADATA/.test/METADATA.jl, in expression starting on line 4
```

CRlibm/url
```diff
@@ -0,0 +1 @@
+http://github.com/dpsanders/CRlibm.jl
```
pao: git://github.com/dpsanders.CRlibm.jl.git

dpsanders:
@pao I saw that, thanks, I just don't understand it... What do I need to change where?

dpsanders:
OK, I see that I need to change the URL. Since everything was automatic via `Pkg`, I just don't know where to change it!

pao:
Yeah, if you bootstrapped via Pkg, and still have the steps you followed handy--there might be a Pkg bug in there. If so then steps to reproduce should probably get reported as an issue to the main Julia repository. But you can change the URL manually in the `url` file under your `METADATA.jl` checkout (which I'm not sure what the trivial way to find is these days).

mbauman:
It looks like Pkg.register uses the `origin` remote for its url.  What does `Pkg.Git.run(`remote -v`, dir=Pkg.dir("CRlibm"))` output?  You might have some wonky remotes setup.

dpsanders:
Thanks for the suggestions. I changed the URL by hand in the Pull Request, since I don't have access to the machine that I sent the PR from right now.

I have a feeling that what happened is that I sent the PR from a machine to which I had `Pkg.clone()`d the package from GitHub. Could that be the problem? In any case, it shouldn't be too hard to automatically convert the `http` address to `git` in `Pkg`.

tkelman:
There's a bit of an unresolved issue where pkg should probably be transitioned to use https instead of git, but for now it's good to just keep things uniform.

ViralBShah:
Nice!

dpsanders:
Is there any objection to merging this?

dpsanders:
The problem with the remote URL seems to have been caused by using `hub` to clone the remote, rather  than the `https` line that the clone box on GitHub gives.

dpsanders:
Closing in favor of #3466

PR #3450, opened 2015-09-19
Tag BinDeps v0.3.17

yuyichao:
Also bump version requirement for BinDeps 0.3.16 to avoid installed by 0.3 users


BinDeps/versions/0.3.16/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.4-
```
yuyichao: Looks like I remember to add the `-` this time ....

PR #3458, opened 2015-09-19
Tag Plots v0.1.4

tbreloff:
- Fix deprecation warnings
- Add argument aliases
- Added support for Gadfly/Immerse linestyles (depends on my PR 695)
- Reworked color picking, more dependent on background color
- Added computation of foreground color based on background, and set `foreground_color` arg
- Some bug fixes and reorganization


tkelman:
Do you expect that your tests are failing on Travis?

tbreloff:
I wasn't expecting it but it makes sense... I have a test that was depending on an un-merged PR in Gadfly.  I'll try to fix tonight and re-tag.  Thanks.

PR #3463, opened 2015-09-19
Register MIDI

JoelHobson:
Hi all,

I wrote a small package for working with MIDI files. I've called it MIDI since that seems to be the most explicit description of the package's purpose, but I'm not sure if this is appropriate...


tkelman:
Neat! Do the tests pass if you enable Travis CI? Could do something a little longer like MidiFile or MidiIO but MIDI is probably fine, that acronym is pretty unambiguous.

JoelHobson:
The tests do indeed pass with Travis CI!  I'd prefer to stick with MIDI over something longer. I'm a big fan of conciseness :smile:

johnmyleswhite:
I'm up for merging this. Would be great to squash the three commits into one before doing that.

JoelHobson:
Attempted to squash the commits - I'm showing a deleted line from something I don't own, not sure how that happened. It's causing the build to fail, I'll see about fixing it.

tkelman:
When you rebase, if you check the log and it looks like it only has your single commit on top of the current tip of master, you need to force push to the `metadata-v2` branch of your fork. You're most likely doing a merge first then a non-force push, the merge is the source of the problems.

tkelman:
Sorry didn't look close enough, you apparently squashed the deletion of Hwloc's url into the same commit. You can add a second commit that puts that url file back the way it was, and squash that together to undo that change.

JoelHobson:
Thanks for that. I don't really know git beyond the basics, so it helped a lot. Looks like we're good to go now.

PR #3479, opened 2015-09-21
PairwiseListMatrices 0.1.0

diegozea:
Package for working with pairwise lists as matrices. The are efficient (in memory and time) in some situations. They are useful for me when I'm measuring the distances between ~ 88000 elements (a full matrix doesn't fit in my RAM). I believe this could be useful for the community.

Best


tkelman:
LGTM

PR #3489, opened 2015-09-21
Adjust requirements on JuliaParser 0.6.3

tkelman:
This has been reported as broken on Julia 0.3, causing Juno to stop working
ref https://groups.google.com/forum/#!topic/julia-users/ntOb9HNm0ac

### DIFF
--- a/JuliaParser/versions/0.6.3/requires
+++ b/JuliaParser/versions/0.6.3/requires
@@ -1,3 +1,3 @@
-julia 0.3-
+julia 0.4-
 FactCheck
 Compat


jakebolewski:
I do not read julia-users, why didn't someone file an issue?  Juno should pin all its dependent package version numbers to avoid this situation.

tkelman:
I don't use Juno, or know what part of JuliaParser's code it relies on. It would help if JuliaParser's tests were passing, as a start to then look at whether they cover the right use cases. I had to have a student's machine in front of me to figure out what the minimal short-term fix was.

tkelman:
There was a PR. Looks like a new tag that includes https://github.com/jakebolewski/JuliaParser.jl/pull/25 might also fix this, at which point this isn't strictly necessary any more but would be good housekeeping anyway.

jakebolewski:
Ok, I don't use Juno either.  I was hoping to wait until 0.4 to fix this stuff but people are impatient.

PR #3490, opened 2015-09-21
Tag SolveBio v0.0.2

adelq:


hayd:
Doesn't being GPL kind of make this unusable? Please consider a more flexible license e.g. MIT.

StefanKarpinski:
A GPL license for bindings to a proprietary API does seem strange, but I don't think it makes it unusable.

adelq:
Will switch to MIT license in next release.

tkelman:
People are allowed to release GPL packages and they're not "unusable," but more permissive licenses are encouraged where possible.

The package could use a bit of documentation, usage instructions, etc? Does some sort of login information need to be provided on Travis for the tests to pass?

hayd:
@adelq Thank you for reconsidering/switching. I didn't mean to sound harsh in my comment above. Thanks for contributing this package!!

---

> they're not "unusable,"

Of course people are *allowed* to release GPL packages, and sometimes that makes sense. BUT This is a proprietary API, and I think this will mean that most of the time users are going to need to "contact the author"... to pay for usage. IMO This isn't something that should be encouraged by publishing on meta (especially when brandishing the product's name).

StefanKarpinski:
I was reading up on this and thinking about it and while I don't think there's actually a legal issue with using GPL code to call a proprietary API, I think it's not a good idea. Here's my thinking...

Suppose the company decides to change their service so that instead of accessing it via and HTTP API you accessing by linking in a library, say `libsolvebio.so`. It would then be a copyright violation to redistribute this GPL package with that shared library. If you're the sole copyright owner of the GPL code, it would be fine for you to distribute it (you don't need a license to distribute your own code), but if there were other contributors so that you're not the sole copyright owner, then it would be illegal for any of you to distribute the GPL code since you would be violating each other's copyrights. You would need to get permission from every contributor to relicense the code, or you would need to start a new project from scratch under a different license.

That's a hypothetical scenario, of course, but it still gives me pause about the use of the GPL license to access proprietary services.

hayd:
Thanks for addressing the licensing issue.

Your tests appear to be failing, it would be great if you could address that! Thanks!

tkelman:
We've frozen metadata for packages that support julia 0.3. If you'd like to register a version for 0.4, feel free to open a new PR.

PR #3491, opened 2015-09-22
Disable Compat 0.7.1 and 0.7.2 for Julia 0.3

yuyichao:
Since they might break 0.3 code

Ref https://github.com/JuliaLang/Compat.jl/pull/143 and https://github.com/JuliaLang/Compat.jl/pull/144 for the fixes.

Cc. @simonster

### DIFF
--- a/Compat/versions/0.7.1/requires
+++ b/Compat/versions/0.7.1/requires
@@ -1,1 +1,1 @@
-julia 0.3
+julia 0.4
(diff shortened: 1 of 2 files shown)


tkelman:
you realize this also disables these versions on 0.4 rc's?

yuyichao:
Ooops. No..... should I fix it or is it fine to just let people use 0.7.3 instead ...

PR #3498, opened 2015-09-22
Tag HTTPClient v0.1.6

amitmurthy:
Removed Julia 0.4 warnings


tkelman:
This really should have been 0.2.0 if it's the first version of the package to drop support for julia 0.3.

PR #3502, opened 2015-09-22
Tag Blocks v0.0.7

tanmaykm:


tkelman:
I didn't notice this when originally tagged, but increasing min julia version requirement to 0.4 without changing the minor package version will make it difficult to release any future bugfixes of the package for 0.3.

tanmaykm:
Yes, thanks. Will keep this in mind in the future.

PR #3503, opened 2015-09-22
Tag HDF5 v0.5.6 and update URL

yuyichao:


yuyichao:
@timholy Seems that I cannot push the tag to HDF5?

timholy:
Hmm, you _should_ be able to. You're a committer on JuliaLang. Does your private `.git/config` need updating for the new location?

timholy:
Oh, and thanks!!

PR #3505, opened 2015-09-22
Tag MeCab v0.1.4

chezou:
I fixed for v0.4.0-rc2
Cloud you merge it?


tkelman:
package needs a minimum julia version in `REQUIRE`

chezou:
@tkelman fixed on 04230f9 . Is it OK?

tkelman:
not quite, the dash refers to prereleases so for 0.3 it should just be `julia 0.3`

also be sure to update the file in the package too so the next tag gets it automatically

chezou:
I want to support both 0.3 and 0.4. If I'll write `julia 0.3` in require, is it also supports 0.4?

tkelman:
Yes. The first number is a lower bound. If you provide a second number, it's an (exclusive) upper bound. http://docs.julialang.org/en/latest/manual/packages/#man-package-requirements

chezou:
I understand and fix it 5053fff

tkelman:
Looks good. I'll submit a PR to add the same to https://github.com/chezou/MeCab.jl/blob/master/REQUIRE.

chezou:
Thanks! I recreate v0.1.4 tag on MeCab.jl and update commit hash e2d5783

chezou:
Thanks! :raised_hands:

PR #3507, opened 2015-09-22
webstack updates

malmaud:


tkelman:
Do you have a local copy of the requests 0.3.1 tag you can push to the package so it shows up at https://github.com/JuliaWeb/Requests.jl/releases?

PR #3513, opened 2015-09-23
Register WordNet: v0.0.1

jbn:
This is a working and well-tested first draft. It fulfills only the minimal immediate requirements for my project. I'll be expanding the operations as I go (e.g. common hypernyms). The next version -- baring bug fixes -- should be exhaustive over the set of operations.


tkelman:
Looks good. Which is it, Stanford's or Princeton's?

jbn:
Princeton v3.0. 

The feature set is not deep right now. Basically, it's the stuff of a fancy thesaurus. But, everything necessary for expansion is already there. I may make the `0.0.2` API more `NLTK`-like. But, I want to play with new ideas first.

PR #3518, opened 2015-09-23
Tag Conda v0.1.6

Luthaf:
We have to thank @dhoegh for the quick development of Conda.jl!


tkelman:
Thanks to both of you, this will make the user experience for ijulia and pycall so much smoother

PR #3522, opened 2015-09-24
Register and tag CUSOLVER

kshyatt:
NVIDIA CUSOLVER is a library implementing a subset of LAPACK for dense direct solvers and sparse linear algebra on the GPU.


tkelman:
Why all caps?

kshyatt:
Consistency with the other JuliaGPU libraries.

Also NVIDIA writes it that way

tkelman:
Not in the body of their documentation? http://docs.nvidia.com/cuda/cusolver/#axzz3mdTQYSH7

kshyatt:
Oh dang, this must have changed at some point. `CUSPARSE`,`CUBLAS`, and `CUFFT` over at JuliaGPU all have the same naming scheme.

tkelman:
At least BLAS and FFT are acronyms. Sparse and Solver aren't, but renaming already-registered packages (especially changing case only) is more trouble than renaming a not-yet-registered one.

jiahao:
It looks like the capitalization changed between all caps in [CUDA 5.5](http://developer.download.nvidia.com/compute/cuda/5_5/rel/docs/CUDA_Toolkit_Release_Notes.pdf) and lower case 'cu' in [CUDA 6.0](http://developer.download.nvidia.com/compute/cuda/6_0/rel/docs/CUDA_Toolkit_Release_Notes.pdf).

I think in this case either choice of capitalization is fine since I doubt that any of these CUDA-based packages depend critically on the specific CUDA version underneath. All caps at least has the advantage of being consistent with the other `CU*` packages.

StefanKarpinski:
+1 to what @jiahao said

jakebolewski:
These kind of discussions are mildly infuriating.

tkelman:
At least now we can work on Pkg to do namespacing and make renaming registered packages less painful.

StefanKarpinski:
First step: give every package a UUID.

PR #3531, opened 2015-09-24
Register LeastSquaresOptim

matthieugomez:
Following up https://github.com/JuliaOpt/Optim.jl/issues/137


mlubin:
The diff also includes a bump for FixedEffectModels

tkelman:
What is missing from Optim.jl that motivates this being a separate package?

matthieugomez:
@mlubin Thanks. updated.
@tkelman I too would like these tools to be available in Optim. The current solver does not work for high dimensional problems though. Did you have a look at [#137](https://github.com/JuliaOpt/Optim.jl/issues/137) ?

tbreloff:
The name does sound a little misleading.  Have you considered at least renaming to something which describes the method(s) a little better?  I would expect something named LeastSquares to do a very simple linear regression and not much more.  If someone wants these techniques but hasn't yet heard of your package, how will they know to check it out?

tkelman:
Sorry I read that issue after posting. I didn't realize until clicking through that it was a link to an Optim issue. The way the link shows up (at least on my phone) makes it look like a link to a METADATA issue. There is a LinearLeastSquares package but IIRC that was a modeling mini version of Convex.jl meant mostly for teaching.

Looks like the specializations for least squares problems don't fit into Optim just yet but as long as you're thinking about interoperability and trying to keep things fairly similar then sounds okay as its own package for now. I'm especially concerned whether you can leverage auto diff and other features of Optim going forward.

matthieugomez:
@tbreloff I would google "least squares optimization" or "least squares problem", so I do like "Least Squares". The name also mirrors the names of the two main types of the package : `LinearLeastSquares` and `NonLinearLeastSquares`.

My two other packages (FixedEffectModels and SparseFactorModels) now use this package as a backend, so hopefully making these methods as a standalone  package will help other people too. The whole point of the package is to have unsigned methods, so interoperability should not be too hard going forward. In any case I'd be happy to have this package merged in `JuliaOptim` at some point (or superseded by a better one).

tkelman:
Note that if you export a `LinearLeastSquares` type from your package, then using it will conflict with the existing LinearLeastSquares.jl package. It might make more sense to have non-exported types `LeastSquares.Linear` and `LeastSquares.Nonlinear`.

matthieugomez:
I've just checked the `LinearLeastSquares` and there is no type `LinearLeastSquares` surprisingly enough.

The type allows the nice syntax
`optimize!(LinearLeastSquares(x, b, A))`

tkelman:
No, a package can't export a type with the same name as its own module. Since you're potentially exporting a type with the same name as a different package, that other package becomes unusable in combination with yours.

One solution might be to deprecate the existing LinearLeastSquares package if it won't continue to be developed (cc @davidlizeng) since most of its functionality is a subset of what Convex.jl provides. Though with un-exported names your package could still do `optimize!(LeastSquares.Linear(x, b, A))`

matthieugomez:
I see. Is it common for Julia packages to rely on non exported methods to avoid name pollution? (In R, to be registered on CRAN, a package cannot call a non exported method of some other package).

tkelman:
We don't have a formal concept of "public but not exported" so it's currently informally defined by what methods you choose to document but not export. Some packages do use un-exported methods as their public API when they would otherwise be using names and type signatures that would conflict or be ambiguous with base, such as `JSON.parse`.

matthieugomez:
I'm pinging @andreasnoack in case he has any comment about the package.

davidlizeng:
We're still using LinearLeastSquares to teach http://web.stanford.edu/class/ee103/ currently. But I agree, it might be better to deprecate it as an official package after this run of the course, and just make it unofficial.

andreasnoack:
@matthieugomez First of all, thanks for the good descriptions of your packages. It makes it easy get an overview.

My feeling is that it would be good to make the content of your package available through Julia packages but that it's not necessary with another package related to least squares. More specifically, I think your LSMR implementation belongs naturally in `IterativeSolvers.jl`. I know that you have a pr there that hasn't received much response, but I'll promise to review a LSMR pr from you. (For now, try running your `lsmr!` code with `@code_warntype`. I think some of the variables are initiated as integers and then redefined as floats afterwards).

You implementations of nonlinear least squares routines belongs in a package for nonlinear least squares. It could be LsqFit.jl, but right now there is not much there and I also think the name is bit misleading. I'd prefer something that stressed the nonlinear aspect more clearly.

In the longer term, I think it would be intersting to think of a general framework for swiching between (direct/iterative) linear solvers in a non-linear solver like the Levenberg-Marquardt. This is probably a bit harder to design correctly, but it would be cool to explore.

matthieugomez:
Bump. The package is now restricted to non least squares optimization (the least squares part was transformed in a pull request in `IterativeSolvers`). The new name is `LeastSquaresOptim`.

mlubin:
I like it. Could you explain the overlap with https://github.com/JuliaOpt/LsqFit.jl in the readme?

tkelman:
Looks like we're good to go here.

matthieugomez:
Thanks!

PR #3538, opened 2015-09-27
Tag StochasticSearch v0.1.2

phrb:


tkelman:
Raising the minimum julia version requirement from 0.3 to 0.4 would usually qualify as a breaking change, so would justify increasing the minor version number (or major if post-1.0). If there's ever a bug that needs to be fixed for julia 0.3, would have to adjust the julia requirement on the existing 0.1.0 tag to be 0.4-only then tag a new 0.0.6 version of StochasticSearch.

phrb:
Should I close this, re-tag, and make the pull request again?

tkelman:
That could also work, you could tag this as 0.2.0 instead.

phrb:
Ok. Will do, thank you!

PR #3542, opened 2015-09-27
Tag Jewel v1.0.7

one-more-minute:


tkelman:
This won't be installable until 0.4 final. Please start using and paying attention to Travis on this stuff.

And increasing Julia version requirement means you should raise at least minor version number of the package, otherwise putting out future bugfix releases for 0.3 will be much more complicated.

PR #3544, opened 2015-09-27
Register FileIO and image format packages

timholy:
FileIO has been under development for a while as a general-purpose package for handling magic bytes and file formats. The original proposal was discussed in https://github.com/JuliaLang/julia/issues/7299, but a package clearly satisfies this need.

Package authors only need to register the file extension and magic bytes with FileIO, and then `load` and `save` will "automagically" do the right thing.


timholy:
FileIO README is here: https://github.com/JuliaIO/FileIO.jl

timholy:
CC @SimonDanisch, who did a lot of the work to make this happen.

SimonDanisch:
:100:

tkelman:
ImageMagick and NRRD need min julia requirements

tkelman:
is osxnativeio only for images or are there other formats it can read?

timholy:
> is osxnativeio only for images or are there other formats it can read?

only images

tkelman:
if it'll stay that way, might not be the best name - does apple have a name for the framework or whatever library gets used?

Keno:
`QuartzIO`?

timholy:
We should put Image in there somewhere, how about `QuartzImageIO`?

SimonDanisch:
Haha, Apple seems to call the module `ImageIO`. So we could call it `OSXImageIO`.
Or we just leave it as it is and add all the other formats apple can read over time?!

Keno:
I think `QuartzImageIO` is good. That way when Apple inevitably gets rid of it and creates a new API, we can make it a new package.

timholy:
OK, this has been redone with the following changes:
- OSXNativeIO has become QuartzImageIO
- All packages have julia 0.3 minimum version numbers
- shas updated for a considerable number of new commits

This also adds one more new package, [Netpbm.jl](https://github.com/JuliaIO/Netpbm.jl) for (you guessed it) reading & writing the Netpbm formats (without ImageMagick). I won't merge this instantly, in order to give folks the chance to bikeshed that name. (The name seems pretty obvious, I guess the one alternative I could imagine would be NetpbmImages.jl.)

tkelman:
Netpbm sounds reasonable to me.

dpsanders:
Maybe I missed this discussion, but is there a reason for each format to be a separate 
package, rather than including then all in a single ImageFormats.jl package? This would seem to me to be a good option.

timholy:
I've wondered about that myself, but I doubt it would scale very well once julia can support the huge zoo of image formats (e.g., https://www.openmicroscopy.org/site/support/bio-formats5.1/supported-formats.html). Many formats are complex enough to require O(1k) lines of code.

tkelman:
If there's code reuse between similar families of formats then it would make sense to do them together, but especially for the cases where reading or writing relies on external libraries I think smaller packages make more sense to decouple things better.

timholy:
I should also say, we already do that: ImageMagick.jl supports 24 formats, Netpbm 3, QuartzImageIO 6. In many cases these are alternative ways of supporting the same format, which has been discovered to be a useful things since binary dependencies have proven to be such a thorn in the side. In such cases, it's obvious they should be separate packages.

rsrock:
I hate to bring this up, but I'm not sure `QuartzImageIO` is the best name. Quartz refers to the drawing layer, if I understand correctly (its analogous to Cairo). That will cause a problem if we want to have an actual Quartz drawing library, which I'd be interested in at some point.

The actual Apple framework used in what is now `QuartzImageIO` is `CoreGraphics`. Apple's naming here is a bit of a mess. See the discussion here: http://stackoverflow.com/questions/1877987/whats-the-difference-between-quartz-core-core-graphics-and-quartz-2d

Clearly we should ditch `OSXnative`. Other options:

OSXImageIO -- Simon had suggested this above
MacImageIO

I'm somewhat ambivalent about these two.  The all-caps "OSXI" start is a pain to read, but on the other hand, it's consistent with the rest of Julia's OS names.

In any event, it might be best to settle this quickly before anyone discovers these packages :smile:

tkelman:
Probably not worth the hassle of renaming these now that they're registered.

I may not have looked that closely at first, but is the imagine format something that only exists within your lab?

timholy:
> I may not have looked that closely at first, but is the imagine format something that only exists within your lab?

We've posted it on our website, but our website is down ATM. Gotta fix that.

SimonDanisch:
So why not just go for `OSXImageIO` ?
I think it looks better and it's consistent with the julia os names.

kmsquire:
I think a suggestion was made that when (!) Apple inevitably creates their
next API and deprecates this one, it will be easier to start with a new
package.

SimonDanisch:
Yeah but if the name doesn't represent the current name of the lib, it's still not a very good idea.
Deprecations can also be handled in one package without too much hassle!
Well, I don't care too much. I just want to tag this, really :D

rsrock:
Eh, it's probably best to leave it alone.  I guess drawing libs can be `QuartzDraw.jl` or something.

Also, one upside is that autocomplete finds "Quartz..." with only a few chars.

PR #3550, opened 2015-09-28
Register ROOT: v0.0.1

jpata:
This library wraps parts of the [ROOT package](https://root.cern.ch/) used in high-energy physics.

### DIFF
--- a/ROOT/url
+++ b/ROOT/url
@@ -1,0 +1,1 @@
+git://github.com/jpata/ROOT.jl.git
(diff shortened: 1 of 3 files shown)


aviks:
We already have a `Roots.jl` package for root finding algorithms, so this might be confusing. On the other hand, `ROOT` is a well known name/package in its domain.  So maybe that's OK?

jpata:
I realize the problem, though I would say the capital letters point to it being a wrapper rather than related to root finding. Otherwise, maybe ROOTFramework?

jiahao:
The near-collision of names is unavoidable in this case.

jpata:
thanks!

PR #3553, opened 2015-09-28
Bump ChainedVectors for v0.3 and deprecate for 0.4

tanmaykm:


tkelman:
You're really not supposed to modify the sha of an existing tag.

StefanKarpinski:
It's allowed for 0.0.0 versions, but we're moving away from using that.

PR #3567, opened 2015-09-28
Tag MIDI v0.0.2

JoelHobson:
I renamed the package to MIDI.jl a while back, but didn't rename the base file from Midi.jl to MIDI.jl, causing the package to fail on case sensitive file systems.


tkelman:
Whoops, that is the kind of thing I should have caught in review. Having Travis enabled is also a very good way of catching these.

JoelHobson:
Well crap. That's probably a find/replace artifact from when I capitalized MIDI everywhere.

PR #3570, opened 2015-09-29
Tag MIDI v0.0.3

JoelHobson:
Sigh...

Fixed all the rogue capitalizations. See https://github.com/JuliaLang/METADATA.jl/pull/3567#issuecomment-143909488 for context.


tkelman:
version numbers are cheap, at least. would be good to push the tags so they show up at https://github.com/JoelHobson/MIDI.jl/releases too

JoelHobson:
Good call, done. Thanks!

PR #3571, opened 2015-09-29
Register LifeTable: v0.0.1, v0.0.2

klpn:
This package can be used to construct lifetables out of DataFrames and perform some statistical analysis on these.


tkelman:
Looks good. Does the package pass its tests if you enable [Travis CI](https://travis-ci.org/) ?

klpn:
Yes, it [passes](https://travis-ci.org/klpn/LifeTable.jl).

dpsanders:
Too late with my comment, but following the usual guidelines wouldn't we prefer the package name `LifeTables`, with an `s` at the end?

klpn:
I started naming it as `LifeTable`, but I was actually contemplating wheter I should add an `s`. However, the [guidelines](http://docs.julialang.org/en/release-0.3/manual/packages/#guidelines-for-naming-a-package) claim that  "[p]ackages that provide most of their functionality in association with a new type should have pluralized names", and as the package is, it does not really define any new types.

klpn:
However, it could be argued that I in the future should implement e.g. the tables returned by `LifeTable` as a new type (presumably a subtype of `AbstractDataFrame`), because other functions expects a data frame with a certain structure, and I am open to change the name if that is considered.

PR #3582, opened 2015-09-29
Enforce minimum versions for new package registrations

jiahao:
- Packages must now have a julia entry in requires
- Minimum allowed Julia version is v0.3.0
- Minimum allowed package version is v0.0.1
- Also minor cleanup of SHA1 check

Grandfathered exemptions have been granted for the 334 packages whose most recent registered versions do not conform to this policy. (Future PRs for these packages that tag a new version must conform to this new policy.)

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -1,6 +1,12 @@
 const url_reg = r"^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?"
 const gh_path_reg_git=r"^/(.*)?/(.*)?.git$"
 
+const minjuliaver = v"0.3.0" #Oldest Julia version allowed to be registered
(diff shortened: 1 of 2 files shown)


ivarne:
I assume you didn't manually create that list. Could you include the code you used in a comment, so that it becomes easy to regenerate the shorter exception list when the packages gets new versions?

ivarne:
And by the way. How did I end up as an owner on the METADATA.jl repo?

jiahao:
We trust you :)

I simply wrapped the new test in a `try/catch` block and put in a `println` statement in the `catch` clause.

tkelman:
switch travis to use 0.4 for startswith

jiahao:
I've added the code to generate the grandfathered list, and also bumped the Julia version to v0.4.

tkelman:
I'll probably open a PR to sort the list

PR #3584, opened 2015-09-30
Tag Compat v0.7.5

andreasnoack:
Sorry for the noise here. My compat definitions for remoteX didn't work the first time so I'll have to bump the version again.


tkelman:
The new version doesn't work on 0.3. This is why we wait for code review, and don't merge things immediately.

StefanKarpinski:
ants

PR #3590, opened 2015-09-30
Sort the list of packages missing julia version requirements

tkelman:
Best viewed as https://github.com/JuliaLang/METADATA.jl/commit/634d46956925ab096765260abda5c41c36f2d48f?w=1

ref #3582, I think the list makes more sense sorted.


jiahao:
Thanks

PR #3591, opened 2015-09-30
Tag Thrift v0.0.2

tanmaykm:


tkelman:
this won't be installable until 0.4.0 final is ready, just fyi

tanmaykm:
Yes. We are pretty close to it, so should be fine, right?

tkelman:
Yeah, you get a warning on travis but I guess that's not fatal when it's the single package under test that has not-yet-satisfiable requirements. And I'd usually say raising the minimum julia version requirement would call for changing at least the minor version number of the package to allow the possibility of future bugfix releases for 0.3. But here you've only had one past tag, so maybe not as big a deal.

PR #3578, opened 2015-09-29
Tag Jacobi v0.2.0

pjabardo:
New version of the Package Jacobi.

Now it works on Julia 0.4 without warnings but still works with version 0.3.
Also added docstrings for the new documentation system.

PR #3594, opened 2015-09-30
Tag Jacobi v0.2.0

pjabardo:
New version of the Package Jacobi.

Now it works on Julia 0.4 without warnings but still works with version 0.3.
Also added docstrings for the new documentation system.

Added julia 0.3 to REQUIRE


tkelman:
Should #3578 be closed?

PR #3600, opened 2015-10-01
Register Julia PDK (Botched)

wildart:
A wonderful package that could publish itself.


jakebolewski:
I think all the docstrings that got removed should be added to the package.

wildart:
I moved docstring from `helpdb.jl` to `PkgDev.jl` for related functions.

hayd:
Note the error:
```
ERROR: LoadError: AssertionError: Invalid url scheme https for package PkgDev. Should be 'git' 
```
Presumably the docs for package development are going remain in the main julia lang docs?

wildart:
For the new package manager all packages would be registered with `https` protocol, `git`+`ssh` will be dropped. Because the nightly version is shipped with `libgit2` library that is built without support of `ssh` protocol.

wildart:
Messed it up. See #3628

PR #3601, opened 2015-10-01
Repository url can be `https` or `git`

wildart:


jiahao:
I forget why we choose git over https, but this change will cause Travis to fail since every currently registered package uses git now and your change tests the assert on every package.

tkelman:
I think we will have to make this change but do it all at once. And once we pull the trigger on it, Pkg.generate run from 0.3 or 0.4 would always fail metadata travis. Not sure how to best manage this.

ivarne:
We chose git over https in https://github.com/JuliaLang/METADATA.jl/pull/402, because both git and https had firewall issues, and the `insteadOf` configuration in git makes it possible to change on a per user basis.

wildart:
Let's disable protocol check.

tkelman:
No, but maybe loosen it to accept either https or git so packages registered by 0.5 don't fail?

ViralBShah:
I like having fallbacks and trying different protocols before completely giving up.

hayd:
Does the protocol have to match the METADATA? For github projects the git/https URLs can be inferred from the project url (or inferred from one another). The protocol should surely be a user-setting rather than a project one... and the default should be https.

wildart:
So, do we allow both protocols?

wildart:
A package URL is normalized by `Pkg` into `https`.

wildart:
>The protocol should surely be a user-setting rather than a project one... and the default should be https.

New libgit2-backed package manager will support only `https` protocol.

hayd:
The clone/remote url used for packages is currently `git://` even in 0.5... Maybe I'm misunderstanding you.

IMO Even if `git://` is the METADATA url for the package, https should be used (user-setting).

wildart:
There is a small caveat. In `Pkg` 0.5, we disable `ssh` protocol in `libgit2` for cross-platform reasons. Because, we host packages on github, we are allowed to use `https` or `git` protocols. But github version of `git` protocol is actually `git+ssh` (`git` itself does not support any authentication mechanism). There is no problem in using pure `git` protocol, but if repository is on github, we cannot use `git` protocol anymore because of `ssh` dependency. So, either we allow both protocols and deal with different protocols by overwriting URL in `Pkg` or enforce only `https` protocol for registered packages. Or do both and merge this PR.

tkelman:
superceded by #3752

PR #3606, opened 2015-10-01
Tagging and registering GLVisualize packages

SimonDanisch:


sjkelly:
Looks like packing needs a Julia requirement.

SimonDanisch:
Packing should have a `REQUIRE`

tkelman:
The merge commit pulled in unrelated changes. When you do a rebase, you should force push instead of merging.

SimonDanisch:
I think I need to update a few hashes.

tkelman:
It looks good now, but do double-check that all the sha's match what you want, and are consistent with the pushed tags in each repo.

SimonDanisch:
Okay hope this works now. I needed to update the sha of Packages manually and hope I didn't oversee anything there.

tkelman:
You should push the tags to the package repos so they show up under https://github.com/JuliaGeometry/Packing.jl/releases etc

SimonDanisch:
okay I did that with `git push --tag`. I can't remember ever doing that, though my previously registered and tagged packages have tags... What is this sorcery?! Is there something that went wrong only with Packing, or do I need to do this now for all the other packages?

tkelman:
You don't absolutely have to do that, but it helps for organization and tracking on github what the metadata tags correspond to. Usually `Pkg.publish` should do that for you, but it is not the most reliable thing.

SimonDanisch:
Ah I see... so I only have tags for the repos where `Pkg.publish()` succeeded, which didn't happen very often.

PR #3608, opened 2015-10-01
Tag IJulia v1.1.3

stevengj:
It is such a pain that `Pkg.publish` is broken.


tkelman:
Is it broken on both 0.4 and master? For master, does https://github.com/JuliaLang/PkgDev.jl work for you yet? On 0.4 is it a token problem?

stevengj:
I haven't tried PkgDev, but TFA has been broken for a while.

PR #3611, opened 2015-10-02
Tag Blocks v0.1.0

tanmaykm:


tkelman:
Could this be made 0.1.0 instead? That way we could tag an 0.0.8 or 0.0.9 for 0.3 bugfixes in the future, ref https://github.com/JuliaLang/METADATA.jl/pull/3502#issuecomment-144890283

hayd:
Can't you just release 0.7.1 in that case?

tanmaykm:
Yes. Will change to 0.1.0

tkelman:
Thanks!

PR #3612, opened 2015-10-02
Tag BayesNets v0.1.1

mykelk:


BayesNets/versions/0.1.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3-
```
tkelman: This will fail travis because of the trailing `-`, which means prereleases. I just opened a PR to get rid of it.

tkelman:
Assuming this is superceded by #3704

PR #3614, opened 2015-10-02
Change URL for Bebop, add new require guidelines

tkelman:
@zhemao deleted https://github.com/zhemao/Bebop.jl. Unless anyone has a clone of it sitting around that included fc755ffd3e39f7bdc67c07f8dd0fc2fd988009c9, may as well delete it. Also re-generate the lists of exception packages to reflect recent tags. (the diff looks better in split mode)

Will also add the new guidelines to the README.


MichaelHatherly:
@tkelman, I have a recent clone of all the packages, which does include that specific commit. In case it's useful I've set up https://github.com/JuliaPackageMirrors to archive registered packages on a regular basis.

tkelman:
Neat, good idea! What does the Bebop package even do, is the version that you have working well enough to move the url over to an archived/forked version?

MichaelHatherly:
> What does the Bebop package even do

Not sure, seems to be an audio-related package that got deprecated.

> is the version that you have working well enough to move the url over to an archived/forked version?

The clone I've got does load fine on 0.3, though it's complaining about a missing `libportaudio` lib. I can try to figure that out this evening though.

MichaelHatherly:
After finding the correct lib the tests in https://github.com/JuliaPackageMirrors/Bebop.jl/blob/master/test/playsin.jl do work fine, so I guess it's working as well as the original would have. Should be fine to move the url over to it.

IainNZ:
Is it worth the effort, given it can't even be installed on 0.3?

MichaelHatherly:
> Is it worth the effort, given it can't even be installed on 0.3?

For this particular package, probably not, but it could come in handy if other more well-used packages were to disappear for any reason.

tkelman:
@jiahao does this look okay to you? I mostly copied the text from your julia-users post, with a few minor edits.

jiahao:
I'm not convinced that Bebop is worth resurrecting. Otherwise lgtm with the suggestion to edit the policy wording

tkelman:
bebop might now be the highest value target for preserving, but hey why not if thstixe won srorrim e

PR #3617, opened 2015-10-02
Tag Mustache v0.0.14

jverzani:
adjust for deprecations


IainNZ:
thx!

jverzani:
Heads up though. The Requires.jl package is currently not working with RC3, so if you are using a DataFrame to iterate over you will be unhappy :(

PR #3619, opened 2015-10-02
TimeSeries, MarketData, Quandl, MarketTechnicals

milktrader:
Will close the previous PR that did not include MarketTechnicals tag.


milktrader:
Sorry for the noise [here](https://github.com/JuliaLang/METADATA.jl/pull/3618), old habits ...

PR #3624, opened 2015-10-03
Register BaseTestDeprecated

IainNZ:
This is the same as Base.Test for Julia versions < v0.5. Rather than provide convoluted deprecations for the very small number of people using the features that were removed, instead we'll provide this package.

@StefanKarpinski @tkelman @ssfrr


tkelman:
looks okay to me. plans to advertise this somewhere so those who need it can find it?

IainNZ:
I'm thinking NEWS.MD for a start, but also thinking maybe of putting an error for the old with_handler method in Base that points to it.

PR #3641, opened 2015-10-04
Tag IJulia v1.1.4

stevengj:
Fixes https://github.com/JuliaLang/IJulia.jl/issues/363#issuecomment-145315657

cc: @tkelman


tkelman:
thanks much

PR #3649, opened 2015-10-05
Tag JSON v0.5.0

stevengj:
This is for JuliaLang/JSON.jl#129, which fixes precompilation by replacing the runtime (non-REQUIRE) dependency on DataStructures with a more general mechanism.

It is a minor version bump because it replaces the `ordered` keyword in the API with a `dicttype` keyword.  In principle, this breaks backward compatibility, but I can't find any usages of `ordered` in the registered packages that require JSON.

cc: @tkelman, @IainNZ


kmsquire:
I posted this on the repo, but FYI, I have seen/heard of uses of this in personal code. A deprecation message or error indicating how users should change their code would be useful.

PR #3659, opened 2015-10-06
Tag Options v0.2.6

carlobaldassi:
Compatibility with julia 0.4.
(Yes, I did test this on julia 0.2.)


jiahao:
We'll have to write in an exception to the Travis METADATA test, which will fail.

carlobaldassi:
Ah, I forgot about the new rules. It's probably easier to bump the requirement to 0.3 at this point.

PR #3660, opened 2015-10-06
Tag Options v0.2.6 (take 2)

carlobaldassi:
Julia 0.4 compatibility, bump minimum Julia version to 0.3.


tkelman:
if you change julia version requirement, it's best to also change at least minor package version. here there are few enough people left on 0.2 that it doesn't matter too much.

PR #3661, opened 2015-10-06
Julia v0.4 updates

goedman:
Bump Stan.jl tag to 0.3.1 (Julia 0.4 only). REQUIRE updated to julia 0.4- 0.5-. Removed unused badges in README. Simplified .travis.yml to only test julia 0.4.


jiahao:
If v0.2.0 is the last version to support Julia 0.3 then the first version to support Julia 0.4 only should be v0.3.0. See item 5 of our METADATA policy https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md

goedman:
Thanks Jiahao!

PR #3667, opened 2015-10-07
Tag RobustShortestPath v0.0.2

chkwon:
made a few changes for julia v0.4


jiahao:
If v0.0.1 is the last version to support Julia 0.3 then the first version to support Julia 0.4 only should be v0.1.0. See item 5 of our METADATA policy https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md

chkwon:
OK, I'll re do it.

PR #3680, opened 2015-10-07
Deprecate Jumos

Luthaf:


milktrader:
Do you mean tag instead of deprecate?

Luthaf:
No, I tag this version in order to deprecate the package: the corresponding commit is https://github.com/Luthaf/Jumos.jl/commit/9711fc747efcb1464fd96675eeede02304a17eb6

PR #3682, opened 2015-10-08
Register Julia PDK

wildart:
Package management tools which were moved from `julia/master`


stevengj:
Need to edit the `url` file to use `git://` rather than `https://`

jakebolewski:
Closed by #3749.

PR #3684, opened 2015-10-08
Deprecate Jumos formally

IainNZ:
As I understand it, this package is now deprecated by its author, @Luthaf.


Luthaf:
There should be some docs about how to do that somewhere. And thank you for doing it!

IainNZ:
Yeah, there really should. Its mostly been an informal process so far, one day we'll figure out a better strategy. This will show a nice "deprecated" on the package listing now.

PR #3694, opened 2015-10-09
Register LIBLINEAR

innerlee:
Julia bindings for [LIBLINEAR](https://www.csie.ntu.edu.tw/~cjlin/liblinear/).

https://github.com/innerlee/LIBLINEAR.jl

version tagged.


lindahua:
Thanks, Zhizhong.

tkelman:
Would have preferred without the all-caps name. Linear isn't an acronym.

innerlee:
How about LIBLinear or LibLinear?

tkelman:
I think LibLinear would fit in best with other package naming trends, though it can be a little bit of a hassle to rename a package (especially in case only) after it's already registered.

innerlee:
Guidelines for naming a package
6. Packages that wrap external libraries or programs should be named after those libraries or programs.
CPLEX.jl wraps the CPLEX library, which can be identified easily in a web search.
MATLAB.jl provides an interface to call the MATLAB engine from within Julia.

These examples of packages wrapping external libraries all use all captions.

tkelman:
It's probably fine as is. Though at least most of the all-caps library wrapper names are actually acronyms, with the exception of MATLAB and some of the JuliaGPU libraries.

For METADATA pull requests that are registering new packages, it is a good idea in the future to leave them open for a little longer just for community naming input, since it is easier to rename a package before it's registered than after.

innerlee:
:) thanks! In fact, the name of the library itself is all-caps. 
reference:
https://www.csie.ntu.edu.tw/~cjlin/liblinear/
https://www.csie.ntu.edu.tw/~cjlin/papers/liblinear.pdf

PR #3704, opened 2015-10-10
New versions of TikzGraphs and BayesNets

mykelk:
I had a separate pull request for TikzGraphs, but Pkg.publish() after updating BayesNets created this pull request with both TikzGraphs and BayesNets. Not sure if that is okay or not.


Keno:
Yes, bumping two packages is generally fine.

PR #3706, opened 2015-10-10
Register Sparrow

rennis250:
This is a Julia interface to the ctioga2 plotting system. I really like the plots that Ctioga2 produces, but I found it frustrating to constantly export Julia data and then write a separate file for plotting. This makes the process much easier and everything can be done within the Julia environment.

### DIFF
--- a/Sparrow/url
+++ b/Sparrow/url
@@ -1,0 +1,1 @@
+git://github.com/rennis250/Sparrow.jl.git


tkelman:
Note the new guidelines at https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md - tag a version and this should be good.

rennis250:
Should I close this request then? Using Pkg.publish() after Pkg.tag() creates a new pull request.

tkelman:
You could force push the tag commit to this branch, or open a new PR, up to you.

rennis250:
I'll do a new PR since it is already there. One second.

Issue #2417, opened 2015-04-10
Namespacing


IainNZ:
Thoughts:
- METADATA is a single namespace
- This works for many other languages/platforms, including AFAIK Python, Ruby, and Node.js.
  - See https://github.com/npm/npm/issues/798 for pretty interesting discussion for NPM
  - Which mutated into this https://github.com/npm/npm/issues/5239 and https://docs.npmjs.com/misc/scope apparently
- There is something that we have that those systems don't - very strong name rules
- My theory is that relaxing name rules provide a pressure release valve
  - Can use "clever" names, e.g. for mutiple HTTP servers packages
    - We already allow this for things that more obviously need multiple packages, e.g. Convex+JuMP, or Mocha, or the higher-level web stack packages
  - Can do some manual namespacing, e.g. orgname_pkgname
    - That would be fairly un-Julian in appearance

Some options:
- Relax naming rules
- Leave METADATA as a highly curated repository, and create a new, looser registry

tkelman:
Somewhat related, teaching Pkg how to declare dependencies on non-registered packages.

IainNZ:
Indeed, and maybe even having simultanuous access to a private METADATA-like thing

dcjones:
Related issue I made a while back: https://github.com/JuliaLang/julia/issues/5670

dpsanders:
On a related note (and with the recent discussion about naming), the name "METADATA" itself does not follow the naming conventions, being too generic. As a possibly far out suggestion, how about renaming it to something like
- JuliaPackages
- RegisteredPackages
- PackageList
- _PackageList_

where the underscores replace the SHOUTING IN ALL CAPS and indicate its special nature.

As regards namespacing, the organisations sort of go some way to solving this (or could do).

bicycle1885:
An idea:

* Create a new concept: meta package
* A meta package can hold packages and other meta packages.
* The current METADATA.jl is a meta package and we leave it as is.
* HDF5.jl, Gadfly.jl, etc. are packages.

UNIX filesystem analogy:
* file => package
* directory => meta package
* root directory => METADATA.jl

Advantages:
* Organizations can create their own meta package.
  * e.g. BioJulia organization may want to create a meta package named "Bio.jl"
* Maintainers of METADATA.jl can leave package management to each organization.
  * Distributed management will be simpler and scale.
* Nesting meta packages enables creating subgroups in an organization.

I'm sorry if I'm going off the topic.

[...]

PR #3718, opened 2015-10-11
Register TinySegmenter: v0.0.1

chezou:
I create Julia version of [TinySegmenter](http://chasen.org/~taku/software/TinySegmenter/), which is compact Japanese tokenizer.


jiahao:
:japanese_castle: :fireworks:

stevengj:
Maybe it would be better to have a generic descriptive name like `TokenizeJapanese`?  You might want to include alternative tokenization algorithms in the future, or modifications to the original TinySegmenter algorithm.  Also, "TinySegmenter" by itself is not a very descriptive name to people outside the field of Japanese language processing, it could just as easily mean image segmentation or something else.

chezou:
Ah, it's a difficult problem. 😓

It is a just porting package and the name "TinySegmenter" is a well known tokenization tool in Japanese NLPer, and I won't add any other algorithms. 
I read the nameing rule of Julian way like wrapping tool is simply use the original tool, like "LibSVM.jl", I think this rule is reasonable in this case.

chezou:
More descriptive document of TinySegmenter in English is as following (written by mhagiwara who is a Japanese translator of Natural Language Proccesing in Python)
http://lilyx.net/tinysegmenter-in-python/

chezou:
Or is it [namespace problem](https://github.com/JuliaLang/METADATA.jl/issues/2417)?
Then I will close this PR until the problem will be solved.

tkelman:
It looks like there are already TinySegmenter ports in Python, Ruby, Perl, Java, all named similarly. You might have to know what you're looking for to find it, but there aren't really any chances to confuse it with anything else if you look into it a bit.

stevengj:
@tkelman, I'm not worried about people who know what they are looking for finding it.  I'm worried about:

* In the future, do we want to have multiple Japenese tokenizer packages, or do we want to encourage people to work on a single package that implements multiple algorithms as needed?  (And if someone enhances the original TinySegmenter algorithm, should they fork to a new package?)   I thought that we tend to encourage unified packages when it makes sense, in which case a more generic name would be appropriate here.

* People who *aren't* looking for a Japanese tokenizer might get confused by this name; a more descriptive name would help people to know that it's not what they want.

jiahao:
I think the name is allowable under the our current naming policy. It is a port of a well known Japanese tokenize and represents one of several possible Japanese tokenization strategies. (It implements one particular trained ML model for tokenization.) Requiring discoverability on the basis of name would have precluded the registration of dozens of packages by now. (Gadfly?)

chezou:
@jiahao Thanks for merging!

This discussion let me know better understanding for package naming in Julia :)

PR #3725, opened 2015-10-12
Tag DistributedArrays v0.1.6

andreasnoack:


DistributedArrays/versions/0.1.6/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
```
stevengj: Now that 0.4 is released, maybe it should just be 0.4?

andreasnoack:
Version requirement updated to 0.4.

PR #3726, opened 2015-10-12
Tag BDF v0.1.2

sam81:


BDF/versions/0.1.2/requires
```diff
@@ -0,0 +1,4 @@
+Compat
+Docile
+HDF5
+julia 0.4-
```
stevengj: Now that 0.4 is released, maybe it should just be 0.4?   `0.4-` means that *any* pre-release of 0.4 is allowed.

sam81: thanks for pointing that out, but the package should work fine with the pre-releases as well, so maybe better leave `0.4-` in the requirements in case somebody using the package has not yet updated to the final 0.4 release?

stevengj: `0.4-` says that the package works with *all* pre-releases, which basically means it should work with 0.3 as well.  When 0.4 hadn't been released yet, `0.4-` was the best one could do, but now it is probably better to require 0.4 proper.

sam81: OK, I didn't realize that. Should I make the change in `REQUIRE`, tag a new version of the package and then run `Pkg.publish()` again?

stevengj: Yes, I think that's the easiest way.

sam81: OK thanks, I'm having a bit of trouble tagging the new version. If I call `Pkg.tag("BDF", :patch)` I get a `fatal: tag 'v0.1.2' already exists` error. If I call `Pkg.tag("BDF", v"0.1.3")` I get a `ERROR: 0.1.3 skips over 0.1.2` error.

stevengj: I usually delete the tag (`git tag -d v0.1.2; git push origin :refs/tags/v0.1.2`), tag the new version, edit the METADATA commit (via `git commit --amend`) and do a force push (`git push -f`) to update the PR.

sam81: OK, I deleted the tag and tagged it again with git, but then my `~/.julia/v0.4/METADATA/BDF/versions/` did not have a `0.1.2/` folder. If instead I tag it from within Julia after deleting the git tag with `Pkg.tag("BDF", :patch) the `0.1.2/` folder within `~/.julia/v0.4/METADATA/BDF/versions/` is created with the right `sha1` and `requires` files. So, should I just run `Pkg.publish()` now? Thanks for the help.

IainNZ:
So things seem to have got screwed up here, but it still points to a valid commit in BDF.

@sam81, I've merged this now. You might want to remove your METADATA folder, and get a fresh clone. You might also want to manually add a git tag for your package at this SHA (i.e. `git checkout` the commit, then `git tag v0.1.2`, then `git push origin v0.1.2` should do it).

IainNZ:
The next version you tag will have the nice `julia 0.4`, it wasn't essential to derail this PR.

sam81:
Thanks.

PR #3727, opened 2015-10-12
Register MergedMethods, NumericSuffixes, PrivateModules

MichaelHatherly:
A few small packages to (potentially) register:

- ~~https://github.com/MichaelHatherly/AnonymousTypes.jl~~
- ~~https://github.com/MichaelHatherly/CompositeFunctions.jl~~ https://github.com/MichaelHatherly/MergedMethods.jl
- https://github.com/MichaelHatherly/NumericSuffixes.jl
- https://github.com/MichaelHatherly/PrivateModules.jl

I've been using these for a while locally and have found them useful, so figured they might be worth registering.


tkelman:
CompositeFunctions isn't a great name if what it does is merging methods from separate modules.

MichaelHatherly:
> CompositeFunctions isn't a great name

Agreed. Perhaps `MergedMethods` then? Or some other variant on that: `MergeMethods`, `MethodMerging`, `MethodMerge`.

JeffBezanson:
AnonymousTypes seems very similar to NamedTuples. I think the approach in NamedTuples.jl makes sense, since if the type exists only to construct instances and access fields, it is basically tuple-like.

MichaelHatherly:
> NamedTuples

Thanks for pointing that one out. Having similar packages registered would be confusing, so AnonymousTypes can be left unregistered.

MichaelHatherly:
I've updated following the suggestions that were made last week. Any other input? Should be go to merge otherwise.

MichaelHatherly:
Thanks @tkelman.

PR #3728, opened 2015-10-12
Add ImageRegistration package to METADATA

macrintr:


timholy:
This should have a ".jl" in the repository name.

timholy:
The name may also be a bit generic, given the huge zoo of image registration approaches and algorithms. This is not a criticism, just a question of whether something a bit more specific can be chosen. It looks like an interesting package!

macrintr:
Up for suggestions. We were recreating features we'd used from MATLAB's [Image Registration & Geometric Transforms](http://www.mathworks.com/products/image/features.html#image-registration-and-geometric-transformations) library - although some of those functions double count in other libraries.

timholy:
Sorry I forgot to pop back in here. I don't have any better suggestions for the name, so let's go with the one you've got.

Can you squash these commits?

I'm heading out of town for a couple days, so I'm fine if someone else merges this.

macrintr:
Thanks @timholy. Let me know if that squash was what you wanted - I was using the instructions from [Packages in the Manual](http://docs.julialang.org/en/latest/manual/packages/#man-pkg-squash).

timholy:
I forgot, squashing METADATA is harder than other cases. You'd need to be rebasing against `metadata-v2`. It's still the case that there are 11 commits when 3 would do. Can you try
```
git rebase -i origin/metadata-v2
# If you see more than 2 commits (register and tag), you may need to squash more
git push <myfork> +metadata-v2
```
where you've set up `myfork` as [your remote](https://github.com/macrintr/METADATA.jl)?

macrintr:
Thanks @timholy.

timholy:
Thanks for your patience, and for the contribution!

PR #3735, opened 2015-10-12
Tag GLMNet v0.0.5, NIfTI v0.0.5

simonster:


tkelman:
GLMNet needs a Julia version requirement. If this is the first tag of NIfTI that doesn't support julia 0.3 any more, it should be 0.1.0.

PR #3742, opened 2015-10-13
tag fixedeffectmodels v0.2.2

matthieugomez:


jiahao:
For your next tag, it would be preferred to tag `julia 0.4` rather than the current prerelease `julia 0.4-`

PR #3743, opened 2015-10-13
tag LeastSquaresOptim 0.0.2

matthieugomez:
sorry for the quick update


tkelman:
Actually we shouldn't have allowed this to be registered at all without a license. Can you change the tag to b61649302c7bb431036bd56554b7d740ec4abd67 ?

matthieugomez:
yep done.

PR #3745, opened 2015-10-13
Tag Vega v0.5.0

randyzwitch:


Vega/versions/0.5.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3
+JSON
+ColorBrewer
+Compat
+Escher
```
rohitvarkey: You don't need to have Escher in REQUIRE actually. I've added a check [here](https://github.com/johnmyleswhite/Vega.jl/blob/d0cbcbc46d1d463fdb29bce1cb2e910635356378/src/Vega.jl#L72-L74) that enables the Escher integration only if the user has Escher installed.

randyzwitch: Right, Escher isn't a requirement, but a soft dependency.

Can I modify the requires file without resubmitting this PR?

rohitvarkey: You can just add a commit and push to this branch I think.

PR #3749, opened 2015-10-13
Register and tag PkgDev

jakebolewski:


tkelman:
dup of #3682

.test/METADATA.jl
```diff
@@ -47,37 +47,57 @@ for pkg in readdir("METADATA")
...
+    host = m.captures[4]
+    if host == nothing
+        error("Invalid url $url for package $pkg. Cannot extract host")
+    end
+    path = m.captures[5]
+    if path == nothing
+        error("Invalid url $url for package $pkg. Cannot extract path")
+    end
+    scheme = m.captures[2]
+    if !(ismatch(r"git", scheme) || ismatch(r"https", scheme))
```
stevengj: I think this should be a separate PR.   We made a conscious decision at some point to standardize on `git://` URLs, so changing that shouldn't be a side-effect of registering a package.

jakebolewski: I pushed to the wrong branch.

PR #3752, opened 2015-10-13
Allow https:// pkg urls and some cleanup

jakebolewski:
With the switch to libgit2, all new registered packages will be https:// by default on v0.5 dev.


tkelman:
#3601 ?

tkelman:
This is better, nevermind, that `null` has been there all along and never actually worked since `null` is a function. https://github.com/JuliaLang/METADATA.jl/commit/f9c0832f73eb244e6f1a5433d5d07834bdb7df17

PR #3753, opened 2015-10-13
Tag SFML v0.1.0

zyedidia:


tkelman:
Whenever you're ready to say you only support post-release versions of 0.4, go ahead and drop the trailing `-` from REQUIRE.

PR #3757, opened 2015-10-14
Register NonNegLeastSquares

ahwillia:
Repository here: https://github.com/ahwillia/NonNegLeastSquares.jl 

This is my first time registering a package, so let me know if something is off.


jiahao:
You must tag a version. See https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md

ahwillia:
Sorry about that. I missed the `Pkg.tag(...)` step. Good now?

tkelman:
Does the package depend on PyCall, or are you just using that for test comparisons? If it's a test-only dependency, consider moving it from the top-level `REQUIRE` to `test/REQUIRE` then it will only be installed when you run `Pkg.test("NonNegLeastSquares")`. Also consider enabling Travis CI, ref http://docs.travis-ci.com/user/languages/julia/, which will automatically run your tests on every commit and pull request.

tkelman:
How does this compare to Optim.jl's nnls implementation? https://github.com/JuliaOpt/Optim.jl#nonnegative-least-squares

ahwillia:
@tkelman -- Thanks for the tip! I was wondering whether I could have a separate `REQUIRE` file for the test folder. I removed `PyCall` from the `require` file in METADATA... Is this correct?

I added a comparison to `Optim` in the examples directory. I get roughly a 2x speedup for a simple `A*x = b` problem. It doesn't seem like `Optim` supports solving the more general problem `A*X = B` where `X` and `B` are matrices rather than vectors. (Although it could in principle be implemented.) I expect some of the algorithms in my package to outperform by a much larger margin in certain, specialized cases -- e.g. when `A` is tall & skinny and `X` is short & fat.

I am looking into enabling Travis CI now. Should I do this before I register the package? Or does it not matter?

tkelman:
Travis isn't strictly required before registering, but it's a great sanity check for development and a reassurance to users that the package works in a clean environment and not just the author's system.

I think this is good to register as is, but might be good to get @johnmyleswhite's opinion on whether it would be a better long term goal to contribute this type of code and improvements into Optim, or have it decoupled as a specialized package.

ahwillia:
@tkelman -- Trying out Travis CI now. Do I need to do something special to my `.travis.yml` file so that PyCall will work? Right now I'm getting:

```
ERROR: PyError (:PyImport_ImportModule) <type 'exceptions.ImportError'>
ImportError('No module named scipy.optimize',)
```

I'm also happy to integrate this package into Optim or elsewhere. I'm hoping to use apply what I've done to NMF.jl in the near future, so I felt like getting something registered in the near term was desirable. But again, I'm happy to restructure things!

tkelman:
Looks like you'll need to install scipy first? ref https://github.com/travis-ci/docs-travis-ci-com/blob/gh-pages/user/apt.md, try something like
```
addons:
  apt:
    packages:
      - python-scipy
```
or maybe via `pip --user`

ahwillia:
Great -- thanks for walking me through this. I have Travis CI working now.

Do we want to wait for @johnmyleswhite or others to weigh in before registering this? I'm okay either way.

johnmyleswhite:
I think this is good to merge. There's so much uncertainty about how we should move Optim forward that we shouldn't add more in until we figure some other things out.

PR #3764, opened 2015-10-15
Register OnlineMoments

petercolberg:
Dear Julia maintainers,

[OnlineMoments.jl] iteratively calculates the [higher-order moments] of a sequence of real numbers.

I have been using this package in production since a year and finally decided to publish it for the Julia community along with unit tests and documentation. In comparison to existing online statistics packages, this package focuses specifically on the efficient online calculation of moments using immutable types. The package is tested with Julia 0.3.11 and 0.4.0.

Please note that I maintain and use a public [git server] for my free-software projects, which has been operating with >99% reliability since before the first public commit to JuliaLang/julia@a9cbc03.

[OnlineMoments.jl]: http://colberg.org/OnlineMoments.jl
[git server]: http://git.colberg.org/
[higher-order moments]: https://people.xiph.org/~tterribe/notes/homs.html


tkelman:
And what's your approach for dealing with issue reports, pull requests, or continuous integration? If you were using an entirely github-equivalent service like bitbucket or gitlab then maybe it would be okay, but the julia community uses github almost exclusively, for many reasons beyond just code hosting.

tbreloff:
Hi Peter.  Any chance you've done comparisons/benchmarks with OnlineStats?
I'm very curious to see how the implementation choices impact usage.

Thanks for the contribution... I also wish it was on Github, though.

petercolberg:
Hi Tom, please try these two examples for a performance comparison with OnlineStats:

https://gist.github.com/0b5811b625b832cde317

For 10⁸ iterations on this machine (Intel Xeon E5640 @ 2.67GHz and CentOS 6.7), I get
```
# julia test_online_stats.jl
  1.352475 seconds (5 allocations: 192 bytes)
  1.991326 seconds (5 allocations: 192 bytes)
  1.502649 seconds (5 allocations: 208 bytes)

# julia test_online_moments.jl 
  0.488406 seconds (5 allocations: 192 bytes)
  1.014307 seconds (5 allocations: 192 bytes)
  1.325919 seconds (5 allocations: 208 bytes)
```
These tests should be taken very lightly, since in practice online averaging is not performed over constants or elements of an array, but the results of expensive calculations. Using OnlineStats the calculation of four moments (1.50s) is faster than the calculation of two moments (1.99s), which is strange.  The difference in performance between OnlineStats and OnlineMoments seems insignificant. 

I chose immutable types for the accumulators for their ease of use, particularly with arrays of accumulators that support all the common array idioms (zeros, fill!, sum). I have added a `@parallel` example to the OnlineMoments documentation that shows how that extends to parallel averaging.

OnlineStats has broad functionality and significant dependencies including Plot. The OnlineMoments package does only one thing and does it well, which I am certain appeals to other Julia users as well.

johnmyleswhite:
I wasn't aware that OnlineStats depended on Plot. That seems like something that should be reconsidered. IMO, core functionality, like computing means from a stream, shouldn't have dependencies on graphics packages.

petercolberg:
@tkelman: I provide public mailing lists for each of my free-software projects, which may not be as convenient to some users, but has worked well so far. I am aware that the Julia project and community depends heavily on Github. I am always pleasantly surprised at how quickly my pull requests are integrated by Julia package maintainers (especially thanks to @timholy).

On the other hand, for the few projects in which I invest mostly my own private time, I try to follow my principles, such as self-reliance through free software; which is admittedly not easy, since the free-software alternatives are usually orders of magnitude less convenient.

johnmyleswhite:
Using a private Git server is a new thing I think we should allow for now. I'm pretty uncomfortable with the idea, but let's revisit in the future.

johnmyleswhite:
Just a quick heads up, @petercolberg, your self-signed certificate for HTTPS is rejected by a lot of browsers. Hopefully won't affect anyone trying to install your package, but it's worth keeping in mind.

petercolberg:
@johnmyleswhite, thank you for noting this. The certificate is not self-signed but issued by [CAcert], whose root certificate is not included in any major browser. Fortunately, this issue will most likely be solved in a few weeks.

The [Lets Encrypt] project, a robotic certificate authority initiated by the [Electronic Frontier Foundation], will begin issuing valid server certificates beginning [middle of November]. These certificates will be accepted by all major browsers, which is achieved with a cross signature of the Let's Encrypt intermediate root certificate through an established certificate authority.

[CAcert]: http://www.cacert.org/
[Electronic Frontier Foundation]: https://www.eff.org/
[Lets Encrypt]: https://letsencrypt.org/
[middle of November]: https://letsencrypt.org/2015/08/07/updated-lets-encrypt-launch-schedule.html

StefanKarpinski:
> I wasn't aware that OnlineStats depended on Plot. That seems like something that should be reconsidered. IMO, core functionality, like computing means from a stream, shouldn't have dependencies on graphics packages.

Definitely questionable. @tbreloff, what's the rationale for OnlineStats depending on Plots?

StefanKarpinski:
Oh sorry, I see OnlineStats is @joshday's package, for some reason I thought it was @tbreloff's.

PR #3767, opened 2015-10-16
Bump Loess and Zlib

dcjones:


tkelman:
This broke Zlib on 32 bit (I have a PR open), but speaks to the potential need for arch-specific requires, not just os-specific

PR #3769, opened 2015-10-16
Limit Color to < julia 0.4 on all releases

timholy:


timholy:
In this case, the Travis failure is reassuring because it is exactly the point of the PR.

I'll give anyone 24 hours to complain before merging this (hint: you should be using Colors, not Color).

tkelman:
There will be no merging of commits that would cause travis to fail for all other PR's!

Unsatisfiable requirements need to be fixed by also putting an upper julia version limit on all past tags that depend on Color.

timholy:
> There will be no merging of commits that would cause travis to fail for all other PR's!

I don't understand this comment. It only affects packages that use Color.

Are you worried about this?
```
julia> Pkg.dependents("Color")
6-element Array{AbstractString,1}:                                                                                                                                                                                                                                             
 "DCEMRI"                                                                                                                                                                                                                                                                      
 "MatlabCompat"                                                                                                                                                                                                                                                                
 "Terminals"                                                                                                                                                                                                                                                                   
 "Brim"                                                                                                                                                                                                                                                                        
 "GLPlot"                                                                                                                                                                                                                                                                      
 "Sparrow"
```
I can file issues with those packages.

IainNZ:
Shouldn't this be 0.5? We've never deprecated a package with the current release version before. Also, can't merge as-is of course because then all future PRs would fail...

tkelman:
You're proposing making metadata's travis fail by merging this as is, which is not acceptable. If any of those 6 packages that still require Color are currently installable on 0.4, then Color can't be deprecated on 0.4 until they've all switched. Once that's done we can rebase this to see if metadata's consistency check still fails. If it does, then this may require also adding upper bounds on some past tags of other packages.

I don't think we've ever deprecated a package with anywhere near as many dependents.

timholy:
> I don't think we've ever deprecated a package with anywhere near as many dependents.

Quite some time ago I fixed all the major packages that used to depend on it. (Gadfly, Compose, Winston, Images, PyPlot, ProfileView, Esher, ColorBrewer, etc...). It was a huge, painful transition, but we're left with just the stragglers now.

The reason to deprecate it is because RGB from Color is incompatible with RGB from Colors, and all the major packages use Colors now. It's a very difficult error to debug, and keeps coming up. (E.g.,  https://github.com/timholy/Images.jl/issues/384.) See also https://github.com/JuliaLang/Color.jl/issues/101.

I'm fine with waiting for all the stragglers to cross the line, however.

PR #3771, opened 2015-10-16
Pull request/d4eb8b44

moisespsena:


malmaud:
Can you maybe give some clarification on what these packages are and how they differ from the "non-simple" variants?

tkelman:
SimpleHttpIO and StreamReader do not have licenses.

StefanKarpinski:
I'm sure this is well-intentioned, but making a bunch of minor forks of existing packages with different names is bad for the community and the ecosystem. This just causes confusion without adding value.

StefanKarpinski:
To be clear, we would love to have your input on and help with the existing web stack, but if you don't like what's there you should make your case in issues and we can improve things there.

moisespsena:
Good day!!
(Forgive me my english)

I had no computer at home so I did not answer before.

Perhaps the nomenclature was not ideal, I confess that I thought was the best: D.
I will explain the projects:

- StreamReader.jl is a facilitator for reading Streams with reading blocks and iteration to buffer in reading about a number of bytes.
- SimpleHttpIO is a facilitator for the line reading in CRLF format, which is widely used in sockets. Base.readline The function read based on the line LF and not CRLF in the case of SimpleHttpIO, each byte (UINT8) is evaluated to find CR, LF followed by returning the entire line 'UINT8 array {}'
- SimpleHttpCommon: it is an implementation purely in "Julia" for the same purpose module "HttpCommon.jl"
- SimpleHTTPServer: it is an implementation for HTTP server based on SimpleHttpCommon.jl

Regarding the licenses, it was my forgetfulness.

Suggestions and criticisms are very welcome.

PR #3776, opened 2015-10-16
Tag ElasticFDA v0.2.1

jdtuck:


tkelman:
ref https://github.com/jdtuck/ElasticFDA.jl/commit/5e1b3e7c485bb17c039225188aef66302623e47f, is JLD a runtime dependency or test only? It should either be added to REQUIRE or test/REQUIRE.

jdtuck:
Oops, forgot to add the file, I will fix this

PR #3779, opened 2015-10-16
Tag AverageShiftedHistograms v0.2.4

joshday:


tkelman:
If you're dropping julia 0.3 support, this should be package version 0.3.0. Also good to drop the trailing dash from 0.4 if you don't need to support prereleases.

joshday:
Ah, thanks.  I should've caught those.

PR #3780, opened 2015-10-16
Tag CURAND v0.0.4

dfdx:
Support for Windows by @joaquimg


CURAND/versions/0.0.4/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
```
IainNZ: You can probably switch to `julia 0.4` now (for next version)

PR #3781, opened 2015-10-16
Tag DCEMRI 0.1.2

davidssmith:


DCEMRI/versions/0.1.2/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3-
```
tkelman: remove the trailing `-`

tkelman:
when you drop support for julia 0.3 it's highly recommended to make the first 0.4-only tag a different minor version than previous tags, but oh well.

davidssmith:
Oops, sorry.  I can do that when I drop support for 0.4 in the future. ;-)

tkelman:
It's okay, it's mostly a concern if the need arises in the future to tag a bugfix 0.3-supporting release. If your next 0.4-only tag is 0.2.0 then it'll still be doable

PR #3786, opened 2015-10-17
Register GaussQuadrature: v0.2.0, v0.3.0, v0.3.1

billmclean:
I wish to register this package at v0.3.1


tkelman:
See the guidelines at https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md

You need a `REQUIRE` file that states the minimum version of Julia supported by your package.

tkelman:
Looks good, also be sure to do the same in https://github.com/billmclean/GaussQuadrature.jl/blob/master/REQUIRE for the next tag.

PR #3787, opened 2015-10-17
Allow only dev prerelease Julia versions

jiahao:
This PR institutes a requirement that tagged Julia versions for new package version records must be one of:
1. A released version (e.g. "0.3.1") that is at least `minjuliaver` or newer, or
2. A pre-release of a development version (currently "0.5.0-*"), or
3. A pre-release of a released version whose patch number is higher than zero
(e.g. "0.3.2-"). This case is occasionally needed to tag a version before the
fix is released in a new patch version.

In particular, this requirement is designed to raise an error for packages still tagging `julia 0.4-`.

The list of grandfathered packages has also been updated.


tkelman:
I think this is a little premature. We can encourage this, but enforcing it via travis failure is a bit too soon. I suspect there are package authors out there who are pinning on a prerelease or dev version of 0.4 and may still need to do active package development in that state.

tkelman:
I think it's been long enough since 0.4.0 that we can do this now. I updated the list of packages, anyone have objections to merging this if it passes?

timholy:
Seems fine to me. It would definitely reduce the load on you, @tkelman.

mlubin:
No objections here

jiahao:
Is it also time to bump `minjuliaver = v"0.3.0"`, or just keep at 0.3?

jiahao:
Rebased without changing `minjuliaver`

tkelman:
There's zero functional difference between 0.3 and 0.3.0.

jiahao:
I meant if it should be bumped to v0.4

tkelman:
Ah. No, don't think so, as long as people are testing against it.

PR #3788, opened 2015-10-17
RFC: Add explicit requirement for registered packages to have an open source license

tkelman:
This has come up in discussion before, ref https://github.com/JuliaLang/julia/pull/11910#issuecomment-116295281. We could probably auto-enforce this with a bit of scripting, for now I'm proposing writing this down as a policy.


IainNZ:
SGTM

PR #3789, opened 2015-10-17
Tag Equations v0.2.0

jhlq:
Updated for 0.4


tkelman:
see https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md, you need to add a REQUIRE file indicating the minimum version of julia supported by the package

jhlq:
Done, do I tag a patch and republish or is there a smoother way?

jhlq:
Instinctively feel like updating the hash in the metadata clone, however the document you linked does say we "are strongly discouraged from editing METADATA.jl directly"

tkelman:
you can modify your fork to get changes into this pr. you can manually create the lower case `requires` file here which Pkg.tag would normally copy.

jhlq:
Added for the most recent version, should there be for the previous ones as well?

jhlq:
Sorry about the messy PR and wasting your time, should have looked properly at the format of other packages... Maybe a prompt for the REQUIRE file at Pkg.tag could be beneficial. Also since this gets mailed to all watchers: Love you all!

tkelman:
We automatically  populate the REQUIRE in recent Pkg.generate. Thanks for fixing.

PR #3791, opened 2015-10-17
Tag MetadataTools v0.3.0 (also includes #3785)

IainNZ:
Also includes #3785, woops... I thought it made a different branch for each operation for some reason


tkelman:
It does make a separate branch, it just doesn't go back to the upstream state after you do a tag. It doesn't do interactive rebase so how would it know if you only want a subset of commits that aren't present upstream?

PR #3796, opened 2015-10-18
Register Measures: v0.0.1

dcjones:
This is an implementation of compose-like measures to be shared among Compose and other packages by @shashi, @rohitvarkey, and me.


IainNZ:
Neat!

ssfrr:
Is there any crossover between this and @Keno's [SIUnits.jl](https://github.com/Keno/SIUnits.jl)? I see in [this discussion](https://github.com/shashi/Escher.jl/issues/10#issuecomment-88174711) that apparently it was discussed at some point but couldn't find the conversation.

AFAICT SIUnits.jl has support for non-SIUnits as well, and keeps track of them in arithmetic evaluations using `promote_rule` (and promotes to the SI units in operations between SI and non-SI). Maybe `pt` and `inch` could be added as non-SI units, and `Measures.jl` could layer on top all the bounding box and vector handling, and just re-export the length units you're using?

I also could definitely be over-generalizing without a specific problem, or missing some big hurdle that makes it more complicated, so if you've thought about it and it didn't make sense I'd definitely defer to your judgement and domain knowledge. Just wanted to throw it out there so we don't have overlapping package functionality.

dcjones:
There was this discussion a while back: https://github.com/dcjones/Gadfly.jl/issues/264

It might be worth looking at again, but at least part of the difficultly is the need in Measures to be able to invent units that don't have a fixed relationship to any SI unit.

ssfrr:
ah, thanks, that's helpful. I looked in Compose and Escher but not Gadfly :)

shashi:
Dat obvious conversion between `Length{:mm}` and the Meter: https://github.com/dcjones/Measures.jl/pull/4

PR #3798, opened 2015-10-18
Register EEG

codles:


Keno:
Wow, this looks very impressive!

codles:
Thanks. It is the work of my doctoral studies. Some of it is mature and stable, other bits are a work in progress. I was hoping it could be a help to other people.

IainNZ:
Hi @codles 
Indeed, looks great. [You'll also need to tag a version, even if its just v0.0.1](https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md), before this can be merged.

tkelman:
You'll also need to add a minimum julia version requirement to https://github.com/codles/EEG.jl/blob/master/REQUIRE

codles:
I ran `Pkg.tag("EEG")` and `Pkg.publish()` but it has opened a new pull request at #3805. Sorry if that is a mistake.

tkelman:
The new PR is fine, but you're still missing the minimum julia version requirement.

PR #3801, opened 2015-10-18
Tag LightGraphs v0.3.4

sbromberger:
This will be the final version with explicit support for Julia v0.3.


sbromberger:
Thank you!

PR #3807, opened 2015-10-19
AD.jl

abarbu:


tkelman:
There are already quite a few packages for auto diff in Julia, see juliadiff.org. How does this differ from the existing packages? The package naming guidelines also strongly discourage short acronym names.

abarbu:
There are quite a few but they all fall short in some way and aren't usable as general-purpose AD libraries. Either due to fundamental bugs in the implementation, being impractically slow, or being domain-specific.

Here's a breakdown:

DualNumbers - Only provides forward-mode AD. Forward-mode is basically a toy for large data or machine learning (runtime scales with the number of inputs which is precisely what is large in ML). Moreover, it's unusably slow. It's also incorrect (see below, suffers from the same problem).

ForwardDiff - Again only provides forward-mode. But it's faster. Unfortunately it suffers from perturbation confusion producing nonsense answers. In reverse mode the package I submitted is ~20-100x faster while being much simpler. All of this without any thought being given to optimizations of any kind or to code stability. There's a lot of room to make it faster.

ReverseDiffSource - Handles only a tiny fraction of julia, not arbitrary programs.

TaylorSeries - Is domain-specific and while it uses AD internally it's not a general-purpose AD package.

ReverseDiffSparse - Only works for small programs or where you can manually produce the AST.

ReverseDiffOverload - This package is in spirit similar but the implementation method requires that your function be overly generic meaning you can't apply it to any [Real]->Real or Real->Real function. Moreover it's limited to only providing reverse mode and it suffers from the same bug as DualNumbers and ForwardDiff. (I don't know if any of the others suffer from this bug; these are the only 3 I checked).

tkelman:
If there are bugs or solvable performance issues in existing packages, please report them. This is not going to be registered under its current name, and would generally prefer collaboration with existing packages in a similar area. METADATA is a single shared namespace, and if you aren't providing a superset of functionality of existing packages (sparse jacobians of vector-valued functions? sparse hessians?), then best to have a more specific name.

abarbu:
Ok. I closed the PR.

Fixing the bug in those packages requires far more work than writing this package which does everything the other general-purpose AD packages do and is fast. The whole idea was to provide something more unified. The sparse implementations are for very limited usecases where you input the AST not arbitrary Julia code (indeed, this is the only missing feature, and ReverseDiffSparse isn't even meant to be an end-user library).

Moreover, the existing packages have similarly vague names. But I digress.

tkelman:
Under a different name and after a bit more maturation, the package would be good to register. Julia is homoiconic, getting an AST from arbitrary code is pretty straightforward via macros. You aren't going to satisfy any constrained optimization use cases without supporting sparsity, so aiming for something that's too highly unified is a bit ambitious.

StefanKarpinski:
I agree that the biggest issue here is the name – AD is too short and too generic.

StefanKarpinski:
@abarbu, you seem to have deleted the AD repo rather than renaming it. Please don't be discouraged – we always bikeshed package names before registering them. We'd love to have your contributions to automatica differentiation (and other things), especially if it subsumes the functionality of the other packages and is fast as you say.

timholy:
I'd like to second what @StefanKarpinski said. Your (former) AD.jl sounds fantastic; moreover, while I appreciate ReverseSourceDiff.jl very much, it does have its limitations, and sometimes there is no reasonable way to incrementally move from point A to point B.

So I encourage you to try again. If you search this repository for pull requests for new registrations from me, you'll see I've grown so fond of the name-bikeshedding that happens here that I actively solicit it. Good names matter, and it can be hard to come up with them on your own; **I** know what they mean, but maybe other people won't, and input from the community can be very helpful.

scidom:
@abarbu, we are some of the contributors to ForwardDiff (@mlubin, @jrevels) and we'd like to see your AD approach. It would be great if you would like to open a new PR, and I am sure we can sort the choice of package name.

jrevels:
@abarbu I know a lot of people (including myself) would be excited for robust, general purpose reverse-mode AD in Julia. If your package provides this, I'd love to check out your code!

> Unfortunately it suffers from perturbation confusion producing nonsense answers.

I'd be really interested to hear about an example of this - ForwardDiff has tests in place to check for perturbation confusion and so far all of them pass. If you have a case where this arises, *please* open an issue with ForwardDiff so that we can address the problem.

Anyway, thanks for submitting this! As @StefanKarpinski said, name bikeshedding in METADATA is pretty much standard practice - I hope it won't dissuade you from contributing.

PR #3810, opened 2015-10-19
Tag Redis v0.0.2

jkaye2012:


jiahao:
Your `requires` file needs a like like

```
julia 0.3
```

or

```
julia 0.4
```

to specify which version of Julia you are supporting.

jkaye2012:
Thanks, fixed.

PR #3811, opened 2015-10-19
Tag LibBSON; Mongo

pzion:
v0.4.0 fixes for LibBSON + Mongo, plus dangling pointer fix


tkelman:
If these are the first tags to drop support for julia 0.3, it's best to increment the minor package version

pzion:
Makes sense, I'll open a new request!

PR #3815, opened 2015-10-20
Tag ElasticFDA v0.3.0

jdtuck:


tkelman:
Did this copy the macro out of the Cpp.jl package, or was that written from scratch? Any copied code should have proper license attribution.

jdtuck:
It is a copy of the macro, but with changes. How would you suggest I do the license attribution. I was unsure of what exactly to do here

tkelman:
were the modifications more substantial than https://github.com/timholy/Cpp.jl/pull/6/files ? given https://github.com/timholy/Cpp.jl/blob/d67a8b83e04cdc2009b93031b42cb622f88ef346/LICENSE.txt#L6, I would suggest including that full license along with a comment something to the effect of "the following macro was copied from https://github.com/timholy/Cpp.jl with the following modifications - ..."

jdtuck:
Not too much more, I will make the changes and thank you for the help. I am trying to get my  package to work with windows and have had a few hang ups

tkelman:
Good goal, I'm happy to help there. I responded to a mailing list thread a few days back about the Cpp.jl issue if I recall correctly, looking more closely it appears the package is only spawning `make` to build an example library for testing. The script at https://github.com/timholy/Cpp.jl/blob/master/deps/build.jl could probably be made more robust on Windows, just skip running `make` - maybe unless you happen to have a compiler on your path, but I wouldn't count on that.

PR #3821, opened 2015-10-20
Tag FLANN v0.0.3

wildart:
0.4 compatibility update


FLANN/versions/0.0.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3+
```
tkelman: I don't think this makes any sense as a lower bound - does Travis give you a warning about this on 0.3?

wildart: no, all clear

tkelman: ref https://github.com/JuliaLang/METADATA.jl/commit/0713919ab12f07a24498734b493aaeb60f900b4a#commitcomment-13291930 - is "a version greater than all other versions starting with 0.3" a lower bound that you actually want?

jiahao: I didn't encode a check for `0.3+`. It is a somewhat odd thing to put in `requires` though, since

```
julia> v"0.3.0" < v"0.3+" < v"0.3.1"
true
```

tkelman: Ah my mistake then, with only a single dot I guess `0.3+` is treated the same as `0.3.0+`. I was thinking it meant `0.3.x` for an arbitrarily high `x` but that's not right.

wildart: Ah, my bad, `+` is quite redundant.

wildart:
closing in favor of #3828

PR #3827, opened 2015-10-20
Register ODEInterface: v0.0.1

luchr:
2nd attempt.


tkelman:
There's clearly a lot of overlap with @helgee's [Dopri.jl](https://github.com/helgee/Dopri.jl) here, but I think we can go ahead with registering this since it provides `odex` and `radau5` wrappers which I don't see in Dopri.jl. Would be fruitful to collaborate going forward though.

luchr:
Thank you. RADAU with variable order (between 5 and 13) will come very soon. 
After that SEULEX (semi implicit extrapolated Euler method).
I wrote a email to helgee, lets see. If the overlap is problematic, I can create a branch without dopri-support (dropping dopri in that branch) to make everybody happy.

PR #3834, opened 2015-10-21
ParallelAccelerator from IntelLabs

ehsantn:
ParallelAccelerator is a compiler written in Julia. It performs additional analysis and optimization on top of the Julia compiler and tries automatically parallelize programs.


kshyatt:
You don't have a minimum version of Julia specified in `REQUIRE`, and that's why CI is failing. The first line of `REQUIRE` should be `julia 0.3` or `julia 0.4` or similar.

ehsantn:
The wrong commit was tagged. It's fixed now.

kshyatt:
Great! LGTM when CI passes.

ehsantn:
Thanks.

PR #3847, opened 2015-10-23
Tag GLFW v1.0.0-alpha.8

jayschwa:


tkelman:
What's with the alpha version number? If this is the first version to drop 0.3 support it should be a new minor or major.

jayschwa:
I tacked on the alpha last year out of curiosity to see how the system would handle it. Since the next jump is :sparkles:1.0:sparkles:, I have to [break a few things in the API first](https://github.com/JuliaGL/GLFW.jl/milestones/1.0).

tkelman:
Okay, but if you ever need to tag a version with a bugfix for 0.3, having this with only a sub-semver bump relative to the newest 0.3-supporting versions will make that difficult.

jayschwa:
Makes sense, but I don't plan on supporting 0.3 anymore.

tkelman:
Not talking about active development of new features, that's fine to do just for 0.4. Talking about bugfixes. You may not care, but there are users who are going to have to stay on 0.3 for some time, and people who need to support those users. If we tag this as is, the only way to make a new bugfix release that supports 0.3 would be to retroactively make all prior 1.0.0-alpha tags uninstallable on 0.3 and do a new 0.9 tag. You see how just calling this version 1.1.0 would be much easier?

Version numbers are cheap. Not using standard semver "to see how the system would handle it" is not a very good reason.

jayschwa:
I'll hold off for now then. Once I've done the backwards-incompatible API changes, I'll tag 1.1.0 and go up from there, leaving wiggle room for any hypothetical Julia 0.3-compatible versions.

tkelman:
That works too, thanks.

PR #3850, opened 2015-10-23
add Compat to Taro/versions/0.2.1/requires

tkelman:
ref https://github.com/aviks/Taro.jl/commit/ea056e03eec4880231a0cad1819ca78cf92064c4

defensive against the possibility that Taro's direct deps might not depend on Compat forever

cc @aviks


aviks:
Thanks @tkelman

PR #3856, opened 2015-10-23
Tag Requires v0.2.1 [git://github.com/one-more-minute/Requires.jl.git]

one-more-minute:


tkelman:
First tag to drop 0.3 support should really bump at least package minor version.

PR #3867, opened 2015-10-25
Tag Mamba v0.7.1

brian-j-smith:


kshyatt:
Hello,

Does Mamba need those very specific versions of the packages in its `REQUIRE`? Or will the package work with newer/older versions too?

brian-j-smith:
Hi @kshyatt,

A single number next to a package in the REQUIRE file represents the minimum version requirement.  Thus, Mamba will install with any versions at or later than the ones specified.  The numbers that I did specified are the required package versions at which this release of Mamba was developed and tested.

Feel free to let me know if there are other questions.

kshyatt:
Ok, sounds good. Sometimes people put numbers next to the packages in `REQUIRE` because they think it's required, but it sounds like you know what's up :). It might be nice to test against older versions at some point but this lgtm.

brian-j-smith:
Thanks.  That's a good suggestion to accommodate earlier versions of required packages.  I'll make more of an effort to do that in the future.  Best.

kshyatt:
No worries! Like I said, it would be nice, but it's not super necessary, especially if the package versions you've listed are the first ones that work with 0.4 or something.

PR #3869, opened 2015-10-26
Register Munkres: v0.1.0

c42f:
A small package for solving the optimal assignment problem

@PaulBellette


mlubin:
Is the licensing situation here okay? Was the code "derived" from the [cited](https://github.com/FugroRoames/Munkres.jl/blob/240d584662196e397f7ded92e782e6ae931a4842/src/Munkres.jl#L57) matlab and C# code? If so, what license are those codes available under?

c42f:
Should be ok: The matlab code is under the two clause BSD license; the C# code is under the MIT license.

The actual implementation superficially resembles the C# implementation to some extent - was originally derived, but has been significantly modified since then to incorporate various optimizations.

tkelman:
If the matlab code was obtained from the file exchange, they impose an additional clause in the terms of use that you can only use code from there with mathworks products. If you looked at the matlab code and ported any of it, you have to obtain it from the author (not file exchange, a web site or github would be fine) and acknowledge it in your package's license.

PaulBellette:
The Matlab code wasn't used for any part of the port, only for verification of final results and performance benchmarking which was done by running the matlab code... in matlab.

c42f:
Ugh, the way that mathworks seem to have layered terms of use on top of the BSD license leaves me unclear about how to interpret the status of code downloaded from there.  Is the interpretation you're suggesting here that the code isn't really under the BSD license, but under some weird hybrid?  That could well be true, but without a legal opinion or precedent it seems like a massive grey area.

Is the METADATA.jl pull request process generally about vetting licensing?  If so it would be worth having an FAQ about various licensing gotchas in the METADATA.jl readme (or somewhere else obvious).

tkelman:
Good, you're probably alright then.

My recommendation would be to never use the file exchange at all if you ever use any tools that are not purchased from the mathworks. If you are a paying mathworks customer, I encourage you to complain that the relevant clause of the file exchange terms of use is toxic, and it's impossible to act as if open source tools like julia, scipy, octave etc don't exist and people won't use them. Tell authors of code hosted on file exchange to just use github instead, so there are not nasty hidden anti-competitive terms of use adding restrictions on top of what seems like permissive licensing.

c42f:
It's definitely an unfortunate move by mathworks, but I'd be very surprised if it does anything more than spread FUD, rather than have any actual legal teeth.

Avoiding the file exchange completely is obviously safe, but seems a bit extreme: it seems like a terms of use violation would be orthogonal to the software license?  For argument sake, say mathworks could detect someone using file exchange downloads in octave.  What recourse do they have under the terms of service?  Simply banning the downloader from the file exchange *service* (eg, blacklist their IP), but surely any downloaded code remains under the BSD license, as released by the original author.

johnmyleswhite:
Let's avoid creating an air of speculation about legal teeth without an IP lawyer being involved. A lawsuit over File Exchange code is bad for Julia's future even if the affected parties ultimately win against MathWorks.

c42f:
Clearly any interpretation here is speculation, unless anyone happens to know a handy IP lawyer.  Even then, a legal opinion is less than a legal precedent.  Woe to any average programmer who wants to know what the law actually implies!

Regardless of all this, a clear FAQ about what will and will not be accepted into METADATA.jl would reduce uncertainty for package authors, and avoid you guys having this conversation multiple times.

jakebolewski:
There are so many implementations of the linear assignment problem I don't think this is a problem in practice.  The algorithm is from the 50's.

jakebolewski:
A Mathwork's lawsuit over some poor grad student's Julia code would really be the best marketing materials ever for Julia.

PR #3876, opened 2015-10-26
Tag CRlibm v0.2.1

dpsanders:
The package now builds quietly, instead of spewing out all the compilation steps.


dpsanders:
Tagged the wrong commit.

PR #3883, opened 2015-10-27
Tag CUDNN v0.2.1

denizyuret:
cudnn_v3 compatibility fixes.


aviks:
I'm assuming you do want to support the pre-release 0.4 versions?

denizyuret:
Yes, this should work with all Julia 0.4-
The cudnn I was referring to was the nvidia library version v3.

PR #3887, opened 2015-10-27
Tag LightGraphs v0.4.0

sbromberger:


sbromberger:
Sorry for the multiple commits today, but it's not over. We're transitioning from 0.3 to 0.4 and dropping support for 0.3, and I want to make sure people have what they need to continue using the package.

Expect at least one more tag today.

PR #3888, opened 2015-10-27
Register HSA.jl Package

rollingthunder:
I would like to register the HSA.jl package.

It implements a wrapper for the HSA runtime.
HSA, similar to OpenCL allows execution of compute kernels on an accelerator that supports the
[Heterogeneous Systems Architecture] (http://www.hsafoundation.com)

That currently means one of two AMD APU chips, either Kaveri or Carrizo.

The package also supports execution of Julia functions as kernels when running on top of a suitably modified Julia binary.


StefanKarpinski:
How about a longer name? HSA is pretty terse and hard to understand at a glance.

rollingthunder:
Well, it is the name of the library being wrapped.
That's why i chose it, in accordance with the package naming recommendations.

jakebolewski:
The HSA name is fine everyone who does GPU programming will be familiar with the acronym. I'm not going to use HeterogeneousSystemsArchitecture.jl

@rollingthunder would you be up for registering this under the JuliaGPU github group?  I'll send you an invite.

rollingthunder:
@jakebolewski Sure.

rollingthunder:
Does that mean, i have to move the repo and then re-register?

jakebolewski:
Yes

PR #3889, opened 2015-10-27
Register HSA: v0.0.1

rollingthunder:
Now under JuliaGPU:

I would like to register the HSA.jl package.

It implements a wrapper for the HSA runtime.
HSA, similar to OpenCL allows execution of compute kernels on an accelerator that supports the
[Heterogeneous Systems Architecture] (http://www.hsafoundation.com)

That currently means one of two AMD APU chips, either Kaveri or Carrizo.

The package also supports execution of Julia functions as kernels when running on top of a suitably modified Julia binary.


sjkelly:
OH SNAP!
:100:

PR #3891, opened 2015-10-28
Register Qwt.jl

tbreloff:


kshyatt:
Do you want `julia 0.4-` in `REQUIRE`, rather than `julia 0.4`?

tbreloff:
I'm not sure. Would that imply that I support versions pre-AbstractFloat,
etc?

kshyatt:
Well, you have `0.4-` now, which includes prereleases. But `0.4` release is out. My understanding was that `0.4` is preferred assuming the package works with the release (and not just the prereleases?).

tbreloff:
I see... I had it backwards... Thought you were suggesting 0.4-.  Should I
tag 0.0.2 with the new REQUIRE?

kshyatt:
You might be able to just modify the current tag and reopen? Sorry, it was my poor wording that's at fault here.

tbreloff:
I force pushed a new require file with 0.4... please let me know if that looks ok.  Thanks!

kshyatt:
This LGTM - anyone else have thoughts?

milktrader:
:+1:

tbreloff:
Thanks!

PR #3892, opened 2015-10-28
Tag Logging v0.2.0

aviks:
v0.4 deprecations


aviks:
cc: @kmsquire

Logging/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia v0.3
```
tkelman: I guess the v here hasn't been causing any issues, but it's nonstandard

aviks:
Thanks @tkelman . I've fixed that and applied the change to the source.

PR #3900, opened 2015-10-29
Tag CodeTools v0.2.0

spencerlyon2:
Just noticed that we have `julia 0.4-` in REQUIRE, should I change this to `julia 0.4`?

### DIFF
--- a/CodeTools/versions/0.2.0/requires
+++ b/CodeTools/versions/0.2.0/requires
@@ -1,0 +1,6 @@
+julia 0.4
+Lazy
+LNR
+JuliaParser
+MacroTools
+Compat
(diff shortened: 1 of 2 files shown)


spencerlyon2:
Forced pushed to the PR branch to remove the `0.4-` in favor of `0.4`

tkelman:
Unless you have a need to keep running this with a prerelease, then yes it is good to gradually drop the dashes. Do the same https://github.com/JunoLab/CodeTools.jl/blob/master/REQUIRE so it's consistent at the next tag.

spencerlyon2:
Good call. Updated there, then updated sha1 here to match

PR #3906, opened 2015-10-29
Pin Gadfly to version that works for Immerse

timholy:
See https://github.com/JuliaGraphics/Immerse.jl/issues/16.


timholy:
CC @mmagnuski, @tbreloff, @dcjones

tbreloff:
FYI this breaks Plots, since as long as I have Immerse installed I can't
get the version of Gadfly that I need which includes my Gadfly PR.  I'll
figure a way around it I guess, but I guess for now the best solution is to
remove Immerse...

PR #3911, opened 2015-10-29
Tag QuantEcon v0.3.0

spencerlyon2:
Tagging a final minor version that will support Julia 0.3 (nice that it is also QuantEcon 0.3 :; )


milktrader:
Does it also support 0.4 and 0.5? The reason I'm asking is because there is no ceiling in the `julia` REQUIRE.

spencerlyon2:
Good question. 

I'm about 5 minutes away from tagging 0.4.0 that will drop 0.3 support, but will work with 0.4 and 0.5

I did this to leave room on the 0.3.x link for backporting bugfixes and other

Does my strategy make sense

milktrader:
One thing you might consider for your 0.3.0 is to include `julia 0.3 0.4-`, which stipulates that it works *only* with Julia 0.3.x, nothing earlier and nothing later (i.e., 0.4- and beyond).  

If you have bug fixes that you'd like to add to your 0.3.0 branch you can branch from there and tag 0.3.1 if I'm not mistaken. This is the approach I'm taking but I haven't actually tagged in a branch from main yet. (i.e. proceed with caution)

spencerlyon2:
So the current code base (i.e. the one associated with this sha1) works on 0.3, 0.4, 0.5 -- but changes that are sitting on a development branch will break 0.3 compatibility.

So I think it is ok to leave this open on the upper end of the julia version constraint as the code will indeed run on versions beyond the 0.3 line.

milktrader:
Mamba.jl maintainer @brian-j-smith had been experimenting with this approach earlier this year and it seems reasonable.

brian-j-smith:
+1 on the plan to have a release series (0.3.x) for julia 0.3 compatibility.  It's the approach I use in Mamba and it works great.  If QuantEcon 0.3.0 is still compatible with julia 0.4, it seems fine to me to leave the requirement open as ``julia 0.3``.  Then, as @milktrader suggest, if a future QuantEcon 0.3.x release is no longer compatible, use ``julia 0.3 0.4-``.

PR #3917, opened 2015-10-30
Tag OIFITS v0.3.0

emmt:
This version updated to work with Julia 0.4


hayd:
I think it would be great if you could add some tests and run them on travis/appveyor.

PR #3918, opened 2015-10-30
Register StackTraces

spurll:
StackTraces.jl provides simple stack traces that are both human readable and easy to use programmatically.

### DIFF
--- a/StackTraces/url
+++ b/StackTraces/url
@@ -1,0 +1,1 @@
+git://github.com/invenia/StackTraces.jl.git


tbreloff:
Thanks Gem... this is functionality I've looked for several times in the
past.  Can you comment on your choice of license?  Why not MIT?

spurll:
Fair question. Our company does MPL2 by default, but I've chatted with several of the other devs and they've agreed that MIT makes sense in this case. I'll tag a new version with MIT.

spurll:
Closing. Will open new pull request with the new license.

PR #3920, opened 2015-10-30
Tag LNR v0.0.2 [git://github.com/one-more-minute/LNR.jl.git]

one-more-minute:


tkelman:
Dropping 0.3 support warrants a minor package version change, please make the next tag of this package 0.1.0. If this had been properly reviewed through a pull request instead of pushing straight to master when initially registered, the name would have been an issue as well - this should really be `LineNumberingReader.jl` or similar.

one-more-minute:
Can do. I don't disagree about the name but this package was registered before those standards were around, and before it was standard to do name bikeshedding. Happy to change it if there's a good way to do so.

tkelman:
> before those standards were around

Hm, so it was, I had thought the discussions in https://groups.google.com/forum/#!topic/julia-dev/Qw1eaMUKYas and formally writing them down happened earlier. Naming discussions weren't a new thing even from 2014 though, for example #1224.

> if there's a good way to do so

There isn't, and it's more trouble than it's worth at the moment.

PR #3921, opened 2015-10-30
Register StackTraces

spurll:
StackTraces.jl provides simple stack traces that are both human readable and easy to use programmatically. (Now with MIT license!)


kshyatt:
Very nice README. This LGTM - anyone else have thoughts?

ViralBShah:
Looks good to merge.

PR #3924, opened 2015-10-31
Tag SerialPorts v0.0.6

sjkelly:
A couple bug fixes on 0.3. Since I messed up the tag when this package went to 0.4 only, this stays true to semver despite 0.0.4 and 0.0.5 requiring >= Julia 0.4. The upside is that 0.3 installs now get the Conda love too!


tkelman:
Sweet, that wasn't bad at all. First tag that actually needs to change something where 0.3 support would be a burden can be 0.1.0 then.

sjkelly:
Thanks!

PR #3927, opened 2015-10-31
Tag GeoStatsImages v0.0.2

juliohm:


tkelman:
If you don't need to support pre-releases from before the 0.4.0 release, would be good to eventually drop the dash in REQUIRE.

juliohm:
@tkelman thanks for the tip.

PR #3930, opened 2015-10-31
Tag ImageQuilting v0.2.3

juliohm:


IainNZ:
Now that Julia 0.4 is out, you should change the REQUIRE file in your package to `julia 0.4` instead of `julia 0.4-`, which means "Julia 0.4-prerelease or higher".

PR #3934, opened 2015-10-31
Register Word2Vec v0.0.1

weijianzhang:
Word2Vec.jl (https://github.com/weijianzhang/Word2Vec.jl) is a Julia interface to [word2vec](https://code.google.com/p/word2vec/). 

It takes a text corpus as input and produces the word vectors as output. Training is done using the original C code, other functionalities are pure Julia.

Here is a [demo](http://nbviewer.ipython.org/github/weijianzhang/Word2Vec.jl/blob/master/examples/demo.ipynb).


IainNZ:
Looks good!

weijianzhang:
Thanks!

PR #3944, opened 2015-11-01
Tag MLKernels v0.0.4

trthatcher:


tkelman:
If this is the first tag to drop 0.3 support, it should be v0.1.0 of the package via `Pkg.tag("MLKernels", :minor)`

trthatcher:
Okay, I think it's correct this time.

PR #3946, opened 2015-11-01
Register LibArchive: v0.0.1

yuyichao:
* This is a wrapper around the libarchive library which is a c library that support reading and writing archives in a variety of formats and compressions. It might not be suitable for high performance compression of data but is a very nice library for dealing with archives. The package name is directly copied from the C library being wrapped. (Could rename to Archives.jl if anyone like that better).
* 0.4 + only (0.3 support should be possible if necessary) and is compatible with package pre-compilation.
* Test passed on Travis Linux&OSX as well as AppVeyor, on both 0.4 and nightly.
* Coverage `73%`. (A big fraction of the missing lines are functions for setting different format supports...)
* `build.jl` only support installing with package managers since libarchive seems to have some dependencies itself that I don't really want to deal with. It's packaged in all major distributions and also Homebrew and WinRPM so I think that shouldn't be a big issue.
* The error handling still need some polishing (especially for streaming to and from a julia IO object) but it could already be used to do sth useful in the current state.


yuyichao:
Thanks @tkelman for encouraging me to create this package and helping me out about OSX and Windows support. =)

malmaud:
Just a reminder that this needs a license before it can be merged.

yuyichao:
License added https://github.com/yuyichao/LibArchive.jl/blob/master/LICENSE.md

malmaud:
Alrighty then.

PR #3951, opened 2015-11-02
Register NearestNeighbors: v0.0.1

KristofferC:
I also have a question. This package is basically superior to my old similar: https://github.com/JuliaGeometry/KDTrees.jl.

How should I best deprecate it and point towards this package? Tag a final release, then tag one with a deprecation warning pointing to the new package and tell people to pin it to the final release if they still want to use it?


tkelman:
That sounds reasonable. Since that package supports 0.3 but this doesn't, should probably only put the deprecation warning in a new minor 0.4-only tag of KDTrees

PR #3952, opened 2015-11-03
[DataFramesMeta] tag new version

tshort:


tkelman:
would be good to push the tags so they show up at https://github.com/JuliaStats/DataFramesMeta.jl/releases

tshort:
Thanks for the tip, @tkelman. Done.

PR #3953, opened 2015-11-03
Tag final + deprecate KDTrees

KristofferC:
Did I do this right @tkelman?

This is the deprecation commit: https://github.com/JuliaGeometry/KDTrees.jl/commit/34f6043b1020962cce8e6fc36d0a3945beea1e6d

added a warning that said to pin to continue used and bumped to 0.4 + new minor version since the new package doesnt support 0.3.


tkelman:
Looks right.

PR #3954, opened 2015-11-03
Register IndependentRandomSequences

gajomi:
For v0.0.1 of IndependentRandomSequences. All check pass.

PR #3955, opened 2015-11-03
Register IndependentRandomSequences

gajomi:
tag v0.0.1


gajomi:
OK, so I guess I accidentally open two pull requests (#3954  is a redundant version of this).

PR #3966, opened 2015-11-05
Tag BEncode v0.2.0

joshuamiller:
Updates to remove deprecated syntax in Julia 0.4 and 0.5.


tkelman:
needs a minimum julia version requirement. looks like you aren't using Compat.jl so this will no longer work on 0.3, and should thus be marked `julia 0.4`

PR #3982, opened 2015-11-07
Tag RandomMatrices v0.2.0 [git://github.com/jiahao/RandomMatrices.jl]

jiahao:
Now runs on Julia 0.4 (given that Combinatorics has been updated)


tkelman:
push tags? https://github.com/jiahao/RandomMatrices.jl/releases

I also don't see that sha on master?

jiahao:
I dunno what happened here... will fix manually

PR #3985, opened 2015-11-08
Tag FixedSizeArrays v0.0.7

SimonDanisch:
@timholy, constructor rebuild is still problematic, so I better get this tagged ;)


timholy:
Thanks!

PR #3989, opened 2015-11-09
Register MatrixNetworks [git://github.com/nassarhuda/MatrixNetworks.j…

nassarhuda:
…l.git]

### DIFF
--- a/MatrixNetworks/url
+++ b/MatrixNetworks/url
@@ -1,0 +1,1 @@
+git://github.com/nassarhuda/MatrixNetworks.jl.git
--- a/MatrixNetworks/versions/0.0.1/requires
+++ b/MatrixNetworks/versions/0.0.1/requires
@@ -1,0 +1,4 @@
+julia 0.3-
+Compat
+Docile
+#Lint
--- a/MatrixNetworks/versions/0.0.1/requires
+++ b/MatrixNetworks/versions/0.0.1/requires
@@ -1,4 +1,4 @@
-julia 0.3-
+julia 0.3
 Compat
 Docile
 #Lint
--- a/MatrixNetworks/versions/0.0.1/sha1
+++ b/MatrixNetworks/versions/0.0.1/sha1
@@ -1,0 +1,1 @@
+8a8f89734999d7c8fe397f5ae5cfad0e4fa5c80d


tkelman:
The tagged commit does not seem to have been pushed to github? https://github.com/nassarhuda/MatrixNetworks.jl/commit/8a8f89734999d7c8fe397f5ae5cfad0e4fa5c80d

nassarhuda:
It's pushed now - but that was still an older version with a wrong REQUIRE file. Should I close this PR and resubmit a new request? 

Just to make sure I'm following the notes [here](http://docs.julialang.org/en/latest/manual/packages/#generating-the-package) 
Also PkgDev.publish() seems to take forever to fork JuliaLang/METADATA.jl so I was publishing METADATA manually

tkelman:
It's okay for the tagged commit in the package repository to have a slightly wrong REQUIRE file as long as it's fixed here in METADATA, and gets fixed in the package before the next tag. You probably should be using 0.4 rather than nightlies of Julia, PkgDev is still a little rough around the edges.

dpsanders:
It's not very clear from the README what the package does. How is this different from the previous graph packages, e.g. `Graphs.jl` and `LightGraphs.jl`?

nassarhuda:
This package consists of a collection of network algorithms.
In short, the major difference between `MatrixNetworks.jl` and packages like `LightGraphs.jl` or `Graphs.jl` is the way graphs are treated.

In `LightGraphs.jl`, graphs are created through `Graph()` and `DiGraph()` which are based on the representation of G as G = (V,E). Similar types exist in `Graphs.jl` (`EdgeList`, `AdjacencyList`, `IncidenceList`, `Graph`) - this is again based on viewing a graph G as a set of nodes and edges. Our viewpoint is different.

`MatrixNetworks` is based on the philosophy that there should be no distinction between a matrix and a network - thus the name.

For example, `d,dt,p = bfs(A,1)` computes the bfs distance from the node represented by row 1 to all other nodes of the graph with adjacency matrix A. (A can be of type `SparseMatrixCSC` or `MatrixNetwork`). I find this representation a lot easier to work with and handle. I find it more user friendly too. 

The package provides documentation with sample runs for all functions - viewable through Juila’s REPL. These sample runs come with sample data, which I think makes it easier for users to get started on `MatrixNetworks`.

I have updated the [README](https://github.com/nassarhuda/MatrixNetworks.jl) file, adding a slightly different version of what is mentioned here in addition to more examples.

dpsanders:
Thanks, with this explanation and the README it's now clear how this package is different. It looks nice! 

Note  that Lexicon is no longer required with Julia v0.4

nassarhuda:
Thanks for noting this! Will update it shortly.

tkelman:
@nassarhuda I think your latest version https://github.com/nassarhuda/MatrixNetworks.jl/commit/beb530dc677b19ab929b68e3b01c3c0fa06bc76d would be fine to register. The commit that's currently tagged here tooks like the entire history was squashed into a single commit. Maybe submit a new registering+tagging PR based on master of the package?

nassarhuda:
Okay I submitted a new PR #4187 
I'll close this - as it's outdated now.

PR #3995, opened 2015-11-09
Tag Murmur3 v0.2.0

rjagerman:
This version fixes deprecation issues with Julia v0.4


Murmur3/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
```
tkelman: Note that the trailing `-` means prereleases of julia 0.4. If you want to require any release version of 0.4, just use `julia 0.4` without a trailing `-`.

rjagerman:
Thanks for the help! :+1:  I've read the [documentation](http://docs.julialang.org/en/latest/manual/packages/#man-package-requirements) about the `REQUIRE` file again, and I'm not even sure why I put a dash there in the first place. I'll submit another PR where it's fixed.

PR #4003, opened 2015-11-09
Tag ODEInterface v0.0.5

luchr:
buildscript now supports also Windows (64bit) with mingw-w64;
updated documentations: examples for compiling with Linux, Mac and Windows to create (.so, .dylib or .dll files, respectively) in case somebody wants to do the compilation by hand with different options and/or a different compiler.


tkelman:
For Windows especially, and somewhat for Mac, it's a lot simpler for users if you upload dll's or dylib's and have the package just download them, as opposed to needing every user of the package to figure out how to install gfortran and have it on the path where julia can find it (and if you're not careful, having this kind of thing on the path can easily cause incompatibilities if you're not using the right version that some other project expects). Good start though.

luchr:
Thank you for the informations. I'm not sure, if I'm able to precompile all the files for all the different operating systems. At the moment these shared libs have as library dependencies e.g. libgfortran. If I offer precompiled libs I think/believe I'll have to statically link all this in the lib, in order to have no unmet dependencies at runtime. Or did I get something wrong and there is another way?

tkelman:
Julia also depends on and ships a libgfortran dll, so you might not need to worry about it as long as you use a compatible compiler version to what the Windows binaries of Julia are using.

luchr:
Thank you very very much for this informations. After my next todo [bvpsol](http://elib.zib.de/pub/elib/codelib/bvpsol/) [requested several times] I'll try to do this precompiling thing for Windows (as fallback, if there is no gfortran). And of course the numeric people want a build script for Intel's Fortran compiler, too ... still a long way to go

PR #4011, opened 2015-11-10
Register & Tag StatefulIterators

andrewcooke:


tkelman:
contained in #4013. you can push new commits to the same branch as an open PR to avoid opening new ones.

PR #4012, opened 2015-11-10
Tag Brim v1.1.2

tpoisot:


tkelman:
Just checking, that commit https://github.com/PoisotLab/Brim.jl/commit/d0874fac562c5d40eb3930b43fb18d737f2e8ee3 is from August, is that what you intended?

tpoisot:
Huh, weird. That should be PoisotLab/Brim.jl@cdabe0d instead.

tkelman:
Since your latest master is now 0.4-only, that should probably be 1.2.0. Should we merge this then you can do a separate tag with a minor-version increment?

tpoisot:
Sounds good. I'll do a minor-version later this week.

Tony Kelman (11/11):
> Since your latest master is now 0.4-only, that should probably be 1.2.0. Should we merge this then you can do a separate tag with a minor-version increment?
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/JuliaLang/METADATA.jl/pull/4012#issuecomment-155895217

-- 
Timothée Poisot
Professeur adjoint
Quantitative and Computational Ecology
Department of Biological Sciences
Université de Montréal

WEB      http://poisotlab.io/
TWITTER  @PoisotLab

tkelman:
Oh hah this is exactly the same commit as the previous tag https://github.com/tpoisot/METADATA.jl/blob/pull-request/be5b334a/Brim/versions/1.1.1/sha1 so merging this is more or less a no-op. Should be harmless.

PR #4014, opened 2015-11-11
Tag CSV v0.0.3

quinnj:


quinnj:
@tkelman, is just a minor version bump ok for requiring 0.4.1?

tkelman:
I think that should be fine since users can upgrade julia between patch versions and either their code will keep working or we goofed. Important part is to avoid breaking users code on 0.4.0 if they do Pkg.update, which this accomplishes.

quinnj:
Thanks for the sanity check :)

PR #4017, opened 2015-11-11
Register Chemfiles

Luthaf:
These are the Julia binding for my C++ library [Chemfiles](https://github.com/chemfiles/chemfiles).


tkelman:
Cool, you were able to get WinRPM working? As possibly the only person who's put packages up on both Conda and WinRPM I'd be interested to hear your honest unbiased opinion some time.

jiahao:
I wish I had something like this back when I still did molecular dynamics

Luthaf:
@jiahao Thank you! I hope some peoples will find this useful.

@tkelman I got WinRPM working in august. I use it for dependencies instead of Conda mainly because I do not have access to a Windows machine.

About the comparison, I found some weakness of WinRPM:
   - No doc about how to create your specfile on OSB. You find some details about what is or not acceptable for inclusion in OpenSuse, but a beginner tutorial is missing. This is particularly true for personal codes, when you are using your own, in development code and not a well established lib like netcdf.
   - The installation path, using `sysroot/long/mingw/prefix/bin` is disturbing at first. 
   - Not having network accessible on OSB took me some time to overcome, because my CMake build used external projects, and downloaded them. Plus it was not documented either.

So they are mainly issues with a lack of documentation about OpenSuse build services.

Conda on the other hand do not support cross-compilation, so you have to get a windows machine to create your packages and upload them. This make builds less deterministic, but easier to setup. The doc is not great either, but a bit more accessible.

The Julia packages are similar in their functionalities, what I am comparing here is more the OpenSuse build service and the Conda service. 

What would be super nice would be a Julia-specific build farm for all supported platforms, and hosting for the generated packages a la binstar but with Windows and OS X support.

tkelman:
Thanks, that's great feedback. I'll have to put together some tutorials and more detailed documentation on the WinRPM build and packaging process, probably early next year.

> What would be super nice would be a Julia-specific build farm for all supported platforms, and hosting for the generated packages a la binstar but with Windows and OS X support.

Some day...

Luthaf:
Would you merge this so that I can send my email to julia-users ?

PR #4018, opened 2015-11-11
Tag WriteVTK v0.2.3

jipolanco:


tkelman:
Since this is the first tag to drop support for Julia 0.3, it should really be a new minor version, 0.3.0 of the package.

jipolanco:
Done, I just created a separate PR with version 0.3.0 (#4019).

tkelman:
Great, thank you. If you don't need to support pre-releases of Julia 0.4, it may be good for cleanliness' sake to drop the trailing `-` on the julia version in REQUIRE some time before the 0.3.1 tag.

jipolanco:
RIght! I'll do that.

PR #4030, opened 2015-11-13
Register CommunityDetection

afternone:


tkelman:
Exactly what do you mean by "inspired by"? louvain-igraph is GPL licensed https://github.com/vtraag/louvain-igraph/blob/master/LICENSE so if you derived your code from that library in any direct way you must respect that license.

afternone:
@tkelman ,Thank you for comment, what should I do now?

tkelman:
Is your code derived from that existing code, or did you write it entirely on your own without referring to any pre-existing copyrighted code?

afternone:
Some code derived from louvain-igraph

tkelman:
If your code is derived from an existing project, then you must respect its license. See the sections of the GPL on derived works. Essentially your package will need to also be GPL licensed and give attribution to the code that you translated. We can't register the package currently with an MIT license since that would be violating louvain-igraph's license.

afternone:
Change the content of LICENSE.md to GPL license? Is ok now?

tkelman:
Adjust the tag here to that commit https://github.com/afternone/CommunityDetection.jl/commit/bb4c7734236d1ffa1acca96fdfc1746a875978fa

sbromberger:
@afternone just a heads up - if you're forced to use a GPL license, that means that most other packages won't be able to incorporate your code/methods due to the viral nature of the GPL (the exact same thing you're facing here). That's a real shame. Can you separate the GPL-derived code into a separate package and keep your original code MIT-licensed?

afternone:
@tkelman I'm new to github, and not familiar with the process of contributing packages. Thank you very much for your comments and guidances. I have adjusted the tag.

tkelman:
You need to specifically change the content of this file https://github.com/afternone/METADATA.jl/blob/metadata-v2/CommunityDetection/versions/0.0.1/sha1 to match

afternone:
@sbromberger  Thanks for comment, I will consider you suggestion.

afternone:
File https://github.com/afternone/METADATA.jl/blob/metadata-v2/CommunityDetection/versions/0.0.1/sha1  is updated.

tkelman:
@alternone please read http://oss-watch.ac.uk/resources/iprguide, as this is not the first time you've adapted code from other authors without the right attribution: https://github.com/afternone/GraphPlot.jl/issues/3

afternone:
ok, thanks a lot

tkelman:
We can still register this now that the licenses are right as I'm sure you meant no harm, but as @sbromberger mentioned it may be better to cleanly separate the code that you wrote on your own from the code that you derived from elsewhere.

ViralBShah:
Yes, I think we should merge it now that the license issue is addressed.

StefanKarpinski:
Thanks for the contribution, @afternone – I wish this legal stuff wasn't such a pain, but unfortunately it is.

PR #4034, opened 2015-11-14
Pull request/66402dc1

simulkade:


tkelman:
the package should specify its minimum julia version requirement. Does this still work on Julia 0.3 or is it now 0.4 only?

simulkade:
On the last version of 0.3 that I used, it worked fine. After upgrading to 0.4, I only changed parts of the code to get rid of the warnings, but I have not tested it on 0.3 again. Should I add `julia 0.3` to the `require` file?

milktrader:
Metadata has the `require` file but your package has the `REQUIRE` file. You should add `julia 0.x` to your package's `REQUIRE` file with the minimum version "x" that you know the package works on. If you don't know if the package is broken on Julia 0.3 now that you've fixed warnings, then I would not recommend putting that in as the floor version.

tkelman:
you can enable travis ci and add 0.3 testing to the list in https://github.com/simulkade/JFVM.jl/blob/master/.travis.yml to find out for sure whether or not it still works on 0.3. You can also use Compat.jl.

simulkade:
I added travis-ci, but there is a problem when it tries to import Matplotlib and mayavi via pycall. I go with `julia 0.4` until I can figure out a way to write a better yml script.

tkelman:
okay, then if you're going to drop support for julia 0.3 you should do `Pkg.tag("JFVM", :minor)` to tag as 0.1.0, a new minor series

simulkade:
Thanks a lot. I created a new pull request.

PR #4036, opened 2015-11-14
Register MinMaxFilter v0.0.1

codles:


timholy:
Nice. You probably know that Images.jl's `dilate` and `erode` do this for windows of size 3, but this is more general, right?

codles:
Actually I hadn't seen `dilate` and `erode` before. A brief look at Image.jl makes me think this package is more general. Im not sure whether its different enough to deserve an entire package. What do you think?

timholy:
I'd be happy to have it live in Images, and to give you push access so that you don't have any hassles when it comes to making changes & improvements. Or I can post a link from the README to your package. Just let me know which you prefer.

codles:
Ok Ill add it to Images.jl.
What would you like the functions to be called?

timholy:
Since julia has `extrema`, how about `extrema_filter`? But I'm also fine with the name you're currently using.

At some point I wonder if we'd want a separate `maxfilter` and `minfilter`, but that could come later.

codles:
Thanks @timholy.
I have opened up an initial [pull request](https://github.com/timholy/Images.jl/pull/413) for your consideration.
I will close this request

PR #4039, opened 2015-11-14
Pull request/07f009e3

simulkade:
possibly not compatible with julia 0.3 anymore.


jiahao:
You need to specify which minimum Julia version you support for each version you want registered.

milktrader:
We just had this discussion https://github.com/JuliaLang/METADATA.jl/pull/4034

PR #4054, opened 2015-11-16
Register Mocking: v0.0.1

omus:


Mocking/versions/0.0.1/requires
```diff
@@ -0,0 +1 @@
+julia 0.4-
```
tkelman: do you need to support prereleases?

omus: Not explicitly. I started development prior to the official release of 0.4. What do you suggest?

tkelman: It's not critical, but for some future tag it would be good to clean up and remove the trailing dash now that 0.4 releases are out and most people should be able to upgrade.

omus: I'll keep that in mind in the future. For this initial release I think I'd like to allow pre-release 0.4.

PR #4063, opened 2015-11-18
Bump Requests + HttpServer

malmaud:


tkelman:
worth dropping the trailing dashes at some point, if no one needs to support prereleases

PR #4068, opened 2015-11-19
Register ClusterDicts

amitmurthy:
Implements `GlobalDicts` which sets the same key-value pair on all participating workers and `DistributedDicts` which distributed the KV pairs over participating workers (like memcached).

Will merge in a day in the absence of any objections


ClusterDicts/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4-
```
tkelman: do you need to support prereleases?

tkelman:
Turn Travis on?

ClusterDicts/versions/0.1.0/requires
amitmurthy: copy-paste error from an existing package. Thanks.

amitmurthy:
Yes, will do.

PR #4077, opened 2015-11-22
Tag ODEInterface v0.0.6

luchr:
added a wrapper/interface for the boundary value solver `bvpsol` from "Konrad-Zuse-Zentrum fuer Informationstechnik Berlin" (ZIB):  http://elib.zib.de/pub/elib/codelib/en/bvpode.html

I negotiated the license with the persons of ZIB who are in charge of the code. That's the reason why the name of this julia package "ODEInterface" is mentioned explicitly in the license text.


tkelman:
The restriction
> You may use or modify this code for your own non commercial
> purposes for an unlimited time.

may be onerous for some users. Do you need to distribute the original Fortran sources as part of your Julia package, or could you simply download the requested solver sources that the user would like to install?

luchr:
onerous? Hm, why? I ship/distribute the original Fortran *source* code (for all solvers); at the moment the build script tries to compile/link them on the client-side. If the user has to do an additional step (like downloading or choosing what source to download) I think this will be more work for the users. 

For your/the/my plan of *binary* downloads (your recommendation for windows and perhaps MAC) I've the restriction (from ZIB) to only allow the download of my prebuilt binaries *together with* the license. I think, if I support binary/prebuilt downloads, I'll have the binary-dll inside a zip file where also the license is included.

tkelman:
Therr are commercial users of Julia, and anything that gets automatically downloaded and built by your package but would be illegal to use commercially should be *very* clearly emphasized in the package documentation, and/or upon first building or calling the relevant code.

luchr:
Ah ok. I see. Is there a typical place, where to put this *very clearly* emphasized information? `README.md`? And or output of the build script?
At the moment all the licenses are in `LICENSE.md` *and* in the license help text of every solver.

tkelman:
A clear note in the readme for the corresponding bullet of the solvers list, and the output of the build script for that solver sound like they'd be sufficient. Legal usage of the software is generally the user's responsibility, but no reason not to make it a little more obvious and clearer for commercial users to do so.

luchr:
OK. Then please forget this pull request. I'll change this and I'll make a new pull request.

luchr:
I've updated this pull request with the new version, sorry. I hope this is OK. If you really want to, I can try to do "history rewriting" and replace the former v0.0.6 tag (showing to the old commit) with the new commit hash.

tkelman:
This is quite alright, thanks for adding the clarification.

PR #4091, opened 2015-11-23
Pull request/6e120b3e

tawheeler:
Updated Smile for 0.4


tkelman:
Which is the last tag of Smile.jl that works on Julia 0.3? Would be best to separate with a new minor version for the first 0.4-only tag.

tawheeler:
Good call. Last tag was v0.1.3 that worked with Julia 0.3.
I can retag to v0.2

tkelman:
Looks like same story with Discretizers, too

PR #4097, opened 2015-11-24
Pull request/064c896b

tawheeler:
Tagged Discretizers and Smile with new minor versions.
Both are now 0.4 (0.3 not officially supported).


tkelman:
Will Discretizers 0.0.3 or Smile 0.1.4 work on Julia 0.3? If not, we can adjust the requires files locally herr in metadata to make those versions not installable on 0.3

tawheeler:
Thank you! They will not work.
I edited the Discretizers entry but am not sure how to add the requirement to Smile 0.1.4.

tkelman:
you can add a new file in this branch at `Smile/versions/0.1.4/requires` that says `julia 0.4`

tawheeler:
I am not sure how to do that. I tried `git checkout 064c896b` on METADATA, then adding requires to `Smile/versions/0.1.4`, adding and committing, but when I run `git push` I get a permission denied error. Am I going about it the right way?

tkelman:
That should work. though you might need to add your fork as a remote and push to that. Or you could create a new file in the web interface with the plus icon at https://github.com/tawheeler/METADATA.jl/tree/pull-request/064c896b/Smile/versions/0.1.4

tawheeler:
Thanks for walking me through this instead of doing it yourself.

tkelman:
Looks great, thanks!

PR #4099, opened 2015-11-25
Register BlsData: v0.0.1

micahjsmith:
Pull microeconomic data from BLS directly into Julia


ViralBShah:
:+1:

tkelman:
Would be good to spell out the BLS acronym in the one-liner repository description that will show up at http://pkg.julialang.org/ next time it gets updated.

micahjsmith:
@tkelman Updated, thanks for the suggestion

PR #4102, opened 2015-11-26
Pull request/98d540c8

yichunc:


tkelman:
Note that there are no currently registered packages with underscores in their names. The package module file is also completely empty in its current state - https://github.com/yichunc/Learn_Discrete_BayesNet.jl/blob/c9ba041e9b78bb5bc31ef85a6d3982c6e05ec752/src/Learn_Discrete_BayesNet.jl

tkelman:
The package appears to have been moved to https://github.com/sisl/LearnDiscreteBayesNets.jl/blob/master/src/Learn_Discrete_BayesNet.jl and structured like a package now, though the top-level module name and filename should be changed to `module LearnDiscreteBayesNets` and `LearnDiscreteBayesNets.jl`.

PR #4104, opened 2015-11-26
Tag Mocking v0.0.2

omus:
Should I be updating the URL to be `https://github.com/invenia/Mocking.jl.git` instead of `git://github.com/invenia/Mocking.jl.git`?


tkelman:
It's not required to switch yet but we're slowly moving in that direction

PR #4107, opened 2015-11-27
GeometryTypes Update

sjkelly:
**GeometryTypes:**
 * Adds a bunch of deprecations to GeometryTypes to free up namespaces for hyper geometries.
 * Docs are now kind of a thing!
 * Improved algorithms for triangulation and extracting points under the new `decompose(::Type, geometry)` function

**Meshes v0.2.0:**
This is a big change many months in the making. It has become supplanted by the larger ecosystem, and is now mostly a meta package. See the [README](https://github.com/JuliaGeometry/Meshes.jl/blob/master/README.md) for details. I have committed to supporting ImmutableArrays and Meshes v0.1.x for the foreseeable future since a few folks depend on them in production, and the FileIO/GeometryTypes ecosystem is still missing some features. 

**Meshing and Packing:**
Minor fixes, mostly deprecation warnings.

**Others:**
Fixes are in on all master branches, but the GL* stuff is still testing the Reactive refactor and waiting on a tag (which so far is totally awesome btw).


Meshing/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
```
tkelman: do you need to support prereleases here?

sjkelly: Nope, fixed

PR #4113, opened 2015-11-29
Tag Quaternions v0.0.5

SimonDanisch:
CC: @andreasnoack @WestleyArgentum 

Thanks,
Simon


kshyatt:
What's the deal with `julia 0.2-` in the `REQUIRE`? Does this still support 0.2 pre?

SimonDanisch:
Good question. I just recently got push access, so I don't know if this is
valid. I wouldn't be surprised, if it would actually still work with 0.2.

tkelman:
If no one is going to test, or ask for a new tag that supports 0.2, probably best to change the `REQUIRE` to 0.3 since that's the oldest that's easily testable, and tag as 0.1.0.

SimonDanisch:
closed in favor of https://github.com/JuliaLang/METADATA.jl/pull/4172

PR #4120, opened 2015-11-30
Register NoveltyColors: v0.0.1

randyzwitch:
This is a new package. I would assert that all of the necessary bikeshedding happened [here](https://groups.google.com/d/msg/julia-users/E866tK4LyEM/6aKuQWVsCQAJ), but feel free to comment on this PR as appropriate. I've made this a separate package to Colors.jl because there are no references to theoretical underpinnings to these schemes (and have made a note as such in the README), whereas `Colors.jl` feels more "theoretical" or "professional" to me.

I've tagged this as 0.0.1, as it meets the minimum requirements as a plugin for `Vega.jl`. As of right now, the only thing this package does is export a `Dict` called `ColorDict`, which is a dictionary created by running all the files in `/data` through `JSON.parsefile()`. `ColorDict` has the name of the palette as the key and values are CSS hexcodes. Compatibility with `Color.jl`, a convenience function for retrieving a palette, or other functionality can be added as people file issues.

Even if no one else ever uses this, I imagine I will add other palettes such as "Cartoon" that would contain palettes of a certain likeness (such as `The Simpsons`, `Speed Racer`, etc.), and really any other interesting palettes I come across. Maybe I'll take a trip to a home improvement store and get inspiration :)

This package will immediately become a requirement of Vega.jl (it's how I developed the code/tested its correctness), so a quick "no" would be appreciated if there are strong arguments against having such a simple package.


tkelman:
lgtm

randyzwitch:
Thanks!

PR #4132, opened 2015-12-01
Register Currencies.jl: v0.0.1

TotalVerb:
I propose a new package, [Currencies.jl](https://github.com/TotalVerb/Currencies.jl), for inclusion in the official `METADATA.jl` repository.

## Intent
`Currencies.jl`provides a clean and intuitive interface to allow use of all currencies specified in the ISO 4217 standard, as well as other currencies if necessary. It is lightweight, and basic usage is quite simple:

```julia
using Currencies
@usingcurrencies USD, EUR, JPY
println(20JPY + 10JPY - 1JPY)
```

Currencies.jl also provides some more advanced functionality, including support for baskets of currencies. In each case, care was taken so that the interface is as easy and intuitive as possible. (The rest of this pull request is a somewhat-lengthy discussion; feel free to skip that if you want.)

## Motivation
Floating-point arithmetic is convenient, but for some applications it can lead to error, usually due to using strict floating point comparisons. For the issue of currencies, this is especially pertinent. Most sources recommend the use of a fixed-point base 10 decimal for currencies. This is a question that is often asked [online](http://stackoverflow.com/questions/26485156/how-to-store-currency-values-and-format-their-output-in-julia-lang), and the solutions are often quite similar.

## Prior Art
As far as I know, two existing packages provide this functionality or similar functionality: [Ito](https://github.com/aviks/Ito.jl) and [FinancialMarkets](https://github.com/imanuelcostigan/FinancialMarkets.jl). Both of these libraries are significantly more complex than Currencies.jl, and have a different focus. Currencies.jl aims to focus exclusively on currencies. Some of its features, such as baskets, do not seem to exist in other implementations. Other features are considerably easier to use in Currencies.jl

This does not mean that Currencies.jl seeks to replace Ito.jl or FinancialMarkets.jl, as it does not offer most of the functionality that the latter two do. Currencies.jl is not a superset, nor a subset, of either of those packages. Although there is overlap, it has different goals and a different approach, and even in its present incantation it already has features that (to my knowledge) the other packages lack.

Some functions offered by `Currencies.jl` are generic, and although I have not tried, presumably they may also work with these other packages.

## Documentation
I have written documentation for all exported functions. There is currently no complete documentation for the whole package itself, but the README.md file provides an in-depth explanation (with examples) about the package's intended usage.

## Testing
I have tested the package with Julia's built-in test system. While coverage is not at 100%, it is pretty high, and the package is simple enough that there should be no major issues with basic usage. This package is already being used in practice by a (currently unreleased) project of mine. However, bugs are bound to exist, and I hope to expand the test suite gradually into the future.

## Call for Comments
This is the first time I have written a Julia package from scratch (however, I have contributed to some before). I intend to take all the comments I receive seriously and address any issues that will come up. Thank you in advance!


tkelman:
This looks great. Very thorough registration PR!

sbromberger:
:+1: on the registration PR. I think the bar has been raised significantly.

PR #4135, opened 2015-12-02
Tag LightGraphs v0.4.1

sbromberger:
This is the last version before we remove edge sets.


sbromberger:
Thank you! One more coming up.

PR #4141, opened 2015-12-02
Tag LightGraphs v0.4.2

sbromberger:
We remove edge sets in this version, causing memory allocation for dense graphs to be reduced by as much as 2/3.


dpsanders:
Does this break the API? If so, maybe it should be a minor version bump?

sbromberger:
It does not break the API at all :) We were very careful. The only possible breaking change is that `edges(g)` now returns an iterator instead of referencing a set, but this preserves `for e in edges(g)` functionality which is what we were using throughout anyway.

sbromberger:
Thank you!!!

tkelman:
Apparently this broke BayesNets and TrafficAssignment, ref http://pkg.julialang.org/pulse.html, so maybe a minor bump would have been better after all.

sbromberger:
The BayesNets and TrafficAssignment errors are because they're using a direct attribute reference (e.g., `g.edges`)instead of the accessor (`edges(g)`). This practice is discouraged, and there are better ways of doing it anyway :) I'll file PRs for them.

sbromberger:
PS: @tkelman - how did you discover the failures of those packages and specifically tie them to the LightGraphs change? (Is it something automated or was it just luck that you saw this PR and noticed the failures?)

tkelman:
By stalking https://github.com/IainNZ?tab=activity to see when the pulse gets updated, and skimming through new failure logs. Especially since I'm running PackageEvaluator a bunch myself right now while preparing the next Julia point release. Also keeping an eye on Metadata PR's so I noticed the "Does this break the API?" question here.

PR #4143, opened 2015-12-02
fix Mosek requires

mlubin:
Closes https://github.com/JuliaOpt/Mosek.jl/issues/66
CC @ulfworsoe


IainNZ:
Ahh I had a related sort of thing with #4130

PR #4147, opened 2015-12-03
Register DynMultiply: v0.0.1

LMescheder:
Simple package for automatically selecting the best multiplication order of matrices depending on their shape.


tkelman:
Good stuff. Add a short description which will show up as a summary at pkg.julialang.org?

tkelman:
And it looks like your tests fail on 32 bit: https://ci.appveyor.com/project/LMescheder/dynmultiply-jl/build/1.0.3/job/x87g5t9ages1a1yj (integer literals there are `Int32`, not `Int64`)

LMescheder:
Ok, it seems to work now. :-) thanks for your help!

PR #4151, opened 2015-12-03
Register DSGE: v0.1.0

micahjsmith:


tkelman:
Acronym package names are generally discouraged - http://docs.julialang.org/en/latest/manual/packages/#guidelines-for-naming-a-package

Also the package's `REQUIRE` file should list the minimum supported Julia version.

micahjsmith:
I have updated the `REQUIRE`, thanks.

I understand that acronym package names are generally discouraged. As discussed in the thread you link to, METADATA already has `SolveDSGE`. Also, while it may appear that `DSGE.jl` is just implementing the FRBNY model, we view it as a package that contains general code for estimating any DSGE model -- and that we provide the code to define the FRBNY model as an added benefit. We fully intend the community to use this package to implement their own models or modify/extend the FRBNY model: https://github.com/FRBNY-DSGE/DSGE.jl#editing-or-extending-a-model.

Perhaps it would be more clear to update the repository description? Something like `Solve, estimate, and forecast the FRBNY Dynamic Stochastic General Equilibrium model and others`.

mlubin:
+1 for accepting the name as it

tkelman:
A better short description field would be good.

What are the copyrights of the code that was derived from other authors? Unless there was explicit permission to relicense those under BSD3, the license needs to be clarified. If there was permission to relicense, that should be written down too.

nalimilan:
I'd say calling it DSGE.jl is fine if you're willing to accept any contributions allowing the estimation of other, possibly very different, variants of DSGE models. On the other hand, if the FRBNY model represents only a particular approach to DSGE models, so that supporting all kinds of DGSE models isn't practical, I'd choose a more specific name.

(Great work, BTW! I wish the French institutions would do the same with their models...)

mlubin:
@nalimilan, I don't think being

> willing to accept any contributions allowing the estimation of other, possibly very different, variants of DSGE models

is a reasonable condition for us to impose on any package authors. First of all, we won't be removing the package from METADATA if they reject someone's PR. Second, any contributions suddenly become the package authors responsibility to maintain. We've seen with the Graphs package that these contributions just pile up and become a hodgepodge of features to the extent that the original code barely works as intended anymore. If someone wants to step up and make everything work well together, that's great, but it's not something we can force anyone to do.

Either we take the package as it is, or we don't.

nalimilan:
@mlubin Sure, I didn't mean it as a constraint on the package authors, but as a way to think about whether the name is the best one. I just wanted to note that a package shouldn't claim a quite general name unless its goal is to implement in a relatively complete way the methods that are generally considered as corresponding to that name. Then, maintainers are the only judges of the quality of contributions and of the way they would fit in the general picture.

To take an example, it wouldn't make sense if GLM.jl maintainers decided that only the Poisson family will be supported by the package. But how they support other distributions, or what pull requests they accept, is of course up to them.

tkelman:
At least the license issue needs to be clarified.

mlubin:
Yes, no objections to making sure the licenses are in order

micahjsmith:
I can confirm that we have explicit permission from authors of derived code to release under BSD-3. We'll update the language tomorrow to reflect that.

> I just wanted to note that a package shouldn't claim a quite general name unless its goal is to implement in a relatively complete way the methods that are generally considered as corresponding to that name.

I think that it's accurate to characterize our goal as @nalimilan does here. Our source consists of general methods of solving and estimating general DSGE models. We provide the FRBNY model as one example, but provide details on how one would go about defining a completely different model. Indeed, we are writing up the widely known [Smets-Wouters model](https://www.aeaweb.org/articles.php?doi=10.1257/aer.97.3.586) for an upcoming release. We would anticipate that ultimately the `src/models/` directory will have definitions of several prominent models (in addition to `FRBNY`) and users would experiment with their own models while using the `DSGE.jl` source to solve, estimate, and ultimately forecast.

In terms of our generic methods, we have implemented so far the methods that we currently use to estimate our own model. These methods can be applied to most (any?) DSGE models. We leave room to allow users to write/specify different algorithms as well, such as [here](https://github.com/FRBNY-DSGE/DSGE.jl/blob/v0.1.0/src/estimate/optimize.jl#L30).

nalimilan:
@micahjsmith OK, that's great then, sorry for the annoyance!

micahjsmith:
Sorry for the delay here, we have updated the language in the [Acknowledgements](https://github.com/FRBNY-DSGE/DSGE.jl/blob/master/README.md#acknowledgements) section to reflect the explicit permission we have received from authors of derived code. Our legal team recommended this wording so hopefully it is sufficiently clear.

PR #4153, opened 2015-12-04
Tag Playground v0.0.4

Rory-Finnegan:
Fixed the issue with the failing tests and added support for fish shell.


kshyatt:
Do you want `julia 0.4-` in `REQUIRE`?

Rory-Finnegan:
I suppose I should probably update it at some point, but it does work for both the dev and stable versions. Can I just update it next time?

tkelman:
Yeah that's fine. Just setting a cleaner example for other people, really.

PR #4156, opened 2015-12-04
Register Nemo

wbhart:
Add tags for Nemo releases. We skip v0.1, v0.2 and v0.3 to keep the tags in line with the Nemo release tags.


wbhart:
Just to head off a potential comment, we are aware that we are supporting 0.4 prereleases. This is fine with us.

PR #4160, opened 2015-12-05
Tag Lint v0.1.71 [git://github.com/tonyhffong/Lint.jl.git]

tonyhffong:


tkelman:
Looks like 0.1.69 was the first tag to drop support for 0.3? Lint should probably move to a new minor number, 0.2.0 at some point soon. That would leave open the possibility of bugfix releases for 0.3 if that ever becomes necessary.

tkelman:
Would be good to make the next tag 0.2.0 via `Pkg.tag("Lint", :minor)`.

tonyhffong:
I haven't been following the main pkg dev for a while, would doing this in Julia 0.5 be achieving the same thing?
> PkgDev.tag( "Lint", :minor )

tkelman:
Yeah, I believe it should.

PR #4161, opened 2015-12-05
Register Miniball

ovainola:
Registering Miniball, package for calculating a smallest enclosing sphere for points in arbitrary dimensions.


tkelman:
If the "Implementation is based on the Bernd Gärtner's C++ miniball" which is GPL, then you can't distribute a derived work under the MIT license without the explicit permission of the copyright holder to do so.

ovainola:
Ok, so if I change it to GPL, I wont break any licenses?

tkelman:
Yes, that's the point of the "strong copyleft" provisions in the GPL, that derived works must themselves also be GPL licensed. Exactly what constitutes a derived work can be tricky to draw the line, but generally if you looked at the original code and designed your implementation to be largely similar in functionality, a port to a different language usually does count as a derived work from what I understand - though IANAL.

You'd also need to tag a version after resolving the license issue. Note that `0.4-` in your `REQUIRE` file means pre-releases of Julia 0.4, which is probably not necessary any more unless you actually have a need to support pre-releases. If you drop the trailing `-` and put just `julia 0.4` in your `REQUIRE` file then that means any released version of Julia 0.4.0 or newer.

ovainola:
Ok, thanks for the info, not going to argue with that. I don't wan't to go to jail or to court room :). I'll also add the tag and change the require

PR #4163, opened 2015-12-05
Register QuantumLab: v0.0.1

vonDonnerstein:
This is a new Package for julia aimed at allowing lab-like interactions with the typical elements of quantum chemical theory.


tkelman:
Hi! The repository name needs to end in .jl. Under the github settings there's an option to rename the repository.

Also in your require file, the trailing `-` in `julia 0.4-` refers to prereleases of julia and is probably not what you want.

vonDonnerstein:
Thank you for the feedback. This is my first julia package, so I'm still prone to these little mistakes. I have commited a new pull requests after correcting these: https://github.com/JuliaLang/METADATA.jl/pull/4178

PR #4175, opened 2015-12-07
Tag OAuth v0.4.0

randyzwitch:


tkelman:
looks like 0.3.0 never got pushed to https://github.com/randyzwitch/OAuth.jl/releases ?

randyzwitch:
I saw that. Pkg.publish() never works for me from the REPL, so I just manage it through git directly. I guess I missed the tag part.

PR #4176, opened 2015-12-08
Tag LightGraphs v0.5.0

sbromberger:
This has a few breaking changes but a lot of new functionality:

`stochastic_block_model` has been removed in favor of new Graph constructors that take a type of `StochasticBlockModel` which can handle various random generators

Persistence has been modularized and made consistent: graphs are now loaded using `load()` and `save()` methods with symbols indicating graph type

Static small graphs and MatrixDepot graphs have been moved into a new submodule, LightGraphs.Datasets.


sbromberger:
Thank you!!!

PR #4183, opened 2015-12-08
Register BuildExecutable: v0.0.1

dhoegh:
I would like to register the BuildExecutable package as it is the succesor to `build_executable.jl` in base.  `build_executable.jl` was removed from base in https://github.com/JuliaLang/julia/pull/14176.


kshyatt:
Do you know why your Travis builds are failing on `0.5` with Linux? I know in your README you talk about portability to Linux - can you finagle your `.travis.yml` to do the right thing?

tkelman:
Something failing to open libgit2. Doesn't look like something that'll be easy to fix right away in Travis settings, will probably need more work.

dhoegh:
I can also reproduce the error locally on Julia master on Ubuntu. But I am not able to track down where the problem is.

PR #4185, opened 2015-12-09
Tag WorldBankData v0.0.5

4gh:


tkelman:
What is the minimum julia version this package is tested against?

4gh:
Julia 0.3.12 is the lowest version I have run the tests on. I also tried it on 0.4.1  and 0.5.0-dev+1491. From the travis output I guess a julia version entry in the REQUIRE file is missing, correct?

tkelman:
Correct. `julia 0.3` would do.

4gh:
OK. I pushed the change. I have to create a new tag, right? There is no way to amend this one and point to the new sha.

tkelman:
you can actually modify the requires file ib this branch of your metadata fork to be consistent and it will be fine.

4gh:
Thanks for your help. It seems to have gotten past the travis ci check.

tkelman:
No problem. Note that the juliareleases PPA is now on Julia 0.4, so the easiest way for you to continue testing against Julia 0.3 on Travis would be switching to the new-style `language: julia`, as in https://github.com/JuliaLang/Example.jl/blob/master/.travis.yml

PR #4202, opened 2015-12-12
Tag ProjectiveDictionaryPairLearning v0.3.4

quxiaofeng:


tkelman:
Since this drops support for Julia 0.3, it should be tagged as a new minor version number of the package, 0.4.0, via `Pkg.tag("ProjectiveDictionaryPairLearning", :minor)`. Also the julia version number `0.4.0-` in `REQUIRE` refers to pre-releases of Julia 0.4 which is no longer necessary. Are you sure you need to support pre-releases?

quxiaofeng:
I will try to fix both, and send a new PR

tkelman:
closing in favor of #4203

PR #4206, opened 2015-12-13
Tagging new versions of both TikzPictures.jl and PGFPlots.jl

mykelk:
Tagging new versions of both TikzPictures.jl and PGFPlots.jl.


tkelman:
If these are droppibg support for julia 0.3, they should use a new minor version of the package

tkelman:
superseded by #4207

PR #4209, opened 2015-12-13
Pull request/6e1c59ed

tpoisot:
OK, so version `1.2.1`, with `Colors` instead of `Color`.


tkelman:
If you're dropping support for julia 0.3 with 1.2.1, it should use a new minor version.

tpoisot:
Support for Julia was actually dropped at 1.2.0

tkelman:
The require file needs to be adjusted for that to be the case.

tpoisot:
Wouldn't that be re-writing history? What's the simplest way to proceed?

tkelman:
You can modify the requires file in your fork of metadata... which you appear to have deleted.

tkelman:
We can try these tags again, but don't delete your fork until after a PR is merged, otherwise there's no way to make modifications.

PR #4224, opened 2015-12-15
Register ValueOrientedRiskManagementInsurance: v0.0.1

mkriele:
This package contains example calculations for the 2nd edition of a book on actuarial risk management. The first edition used R instead of Julia. The book is written in German (see Readme)

### DIFF
--- a/ValueOrientedRiskManagementInsurance/url
+++ b/ValueOrientedRiskManagementInsurance/url
@@ -1,0 +1,1 @@
+git://github.com/mkriele/ValueOrientedRiskManagementInsurance.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
The content of requires here should match https://github.com/mkriele/ValueOrientedRiskManagementInsurance.jl/blob/master/REQUIRE - in particular, if the package requires julia 0.4 then it should not claim to support 0.3 here.

milktrader:
Package title seems a bit of a mouthful, wondering if it could be shortened up to maybe `InsuranceRisk`?

Just a suggestion of course.

mkriele:
I did not want to use up nice, short names for a package that only contains examples for a particular book. But thanks for the suggestion!

mkriele:
Many thanks for the comment regarding the outdated REQUIRE file.  Somehow my installation has a problem and a one month old commit was published.  I also have problems correcting my forked Metadata branch. So I would like to close this  and then open a new pull request once I have completely reinstalled Julia.  In any case, I now know better what to look out for!

PR #4226, opened 2015-12-15
Pull request/f5e8b3be

mkriele:
I do not know how to remove v0.0.1. But v0.0.2 references the latest commit and the correct REQUIRE file.


tkelman:
you should at least modify the requires of the 0.0.1 tag in this branch, if it doesn't support julia 0.3

mkriele:
I will try once more to produce clean metadata. Sorry for the back and forth

tkelman:
It's okay, you can modify existing PR's by committing to the same branch. Ask if you need more specific pointers.

mkriele:
Thanks for the hint, I didn’t know that this is possible.  In any case, in the meanwhile I have closed this pull request and opened a new one, and this time it seems to me that I got it right.

PR #4233, opened 2015-12-16
put an upper bound on ReverseDiffSparse dependencies

mlubin:
ReverseDiffSparse 0.4 (currently being developed at https://github.com/mlubin/ReverseDiffSparse2.jl) will be a completely breaking release. This is still likely a month away, but getting prepared anyway.


tkelman:
do want!

PR #4240, opened 2015-12-17
Tag DSGE v0.1.1

emoszkowski:


DSGE/versions/0.1.1/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4-
```
stevengj: Presumably you mean `julia 0.4` here (= `0.4.0` and later releases).  Having `0.4-` means that you are compatible with *every* **pre**-release of 0.4, not just the released version.

emoszkowski:
We're ok with supporting prereleases of 0.4 - see our discussion in #4151 if you'd like.

PR #4241, opened 2015-12-17
Tag Cubature v1.0.10 [git://github.com/stevengj/Cubature.jl.git]

stevengj:
Fixes stevengj/Cubature.jl#12 and some deprecation warnings.


tkelman:
Do push tags to https://github.com/stevengj/Cubature.jl/releases when you get a chance. Is PkgDev.jl not doing that correctly?

PR #4250, opened 2015-12-18
Register DevTools [http://github.com/JunoLab/DevTools.jl]

MikeInnes:


tkelman:
Tag a version. And write some tests. https://github.com/JunoLab/DevTools.jl/blob/master/test/runtests.jl

tkelman:
This is also an ungodly generic package name. Would have absolutely no idea this has anything at all to do with plotting or graphics from the name `DevTools`.

one-more-minute:
"DisplayPlotsWithElectronAndAlsoAProfiler.jl"?

tkelman:
ElectronPane or ElectronWindow or ElectronView or something like that? Can this use an existing system browser, or would that have to change inside Blink.jl?

one-more-minute:
This just calls Blink to produce windows, so yeah, it would have to change inside Blink.

This package provides graphical tools for Julia developers, currently including a display for graphical output that works at the REPL as well as a profile viewer. I'm not sure how "ElectronPane" gets that across more clearly.

tkelman:
Lots of julia development is non-graphical. It's a common request to have Gadfly use a dedicated window for plots, which it looks like this does. Saying "Install DevTools" to make that possible isn't really intuitive to me. BlinkExtensions? Is there a technical reason the profiler piece and the plotting piece are together in the same package? Or that these aren't implemented in Blink itself?

one-more-minute:
> Or that these aren't implemented in Blink itself?

For the same reason that ProfileView.jl and Immerse.jl aren't implemented inside Gtk.jl?

Separating the functionality and naming them separately could work, though it sounds like it also means doubling the number of name bikesheds we have to go through.

timholy:
JunoDevGUIs?

ViralBShah:
I must admit that I found `DevTools.jl` to be a confusing name, and couldn't fathom what it had to do with displaying plots in windows, until @shashi told me that it was referring to chrome dev tools. I think ElectronPane is not great either. How about BlinkPlots and BlinkProfileView?

I really wish conditional modules work and can be precompiled, so that all this can be hidden from the users.

ViralBShah:
I think JunoDevGuis is not a good name, because the primary use case (at least my current usage) is without Juno. I love it that I can get a plot window and ProfileView from the terminal, without going through too much installation effort - and that it is cross-platform.

StefanKarpinski:
`using ChromeDevTools`?

ViralBShah:
I was thinking about `ChromeDevTools` too.

EricForgy:
As I read down the comments, I was also thinking "ChromeDevTools".

Ismael-VC:
What about `JunoDevTools.jl` ❔ Sounds nice and is very explicit to me.

kmsquire:
Bump

musm:
bump

aviks:
Closing, since it's been a couple of years. Please do a new tag if necessary.

PR #4251, opened 2015-12-19
add Queueing package

rawls238:


rawls238:
I created this PR manually instead of using ```PkgDev``` since I was unable to get ```PkgDev``` to actually work. I continually was getting that an undefvariable when attempting to invoke ```register```

tkelman:
Looks good. There's an issue and a proposed fix for that bug in PkgDev, just needs a test to be written.

Useful to push tags to github too so they show at https://github.com/rawls238/Queueing.jl/releases. Would Travis pass if you enabled it?

rawls238:
thanks @tkelman I'll watch ```PkgDev``` to find out when that fix is in. In the meantime, I've pushed a release in github and kicked off a travis build: https://travis-ci.org/rawls238/Queueing.jl/builds/97846280

PR #4263, opened 2015-12-21
Register NK: v0.1.1

glesica:
Hello, I would like to register NK.jl, a library for building evolutionary simulations using the NK family of fitness landscapes. It is currently in a useful, though primitive state. Performance and a more general API are my next projects. It supports Julia 0.4 right now, I will probably try to support the current release + the current development version going forward.


tkelman:
What does NK stand for? Generally it's preferred to use names that are more explicit rather than acronyms. Also note that if your `REQUIRE` has `julia 0.3` as a lower bound, you should add `0.3` under `julia:` in `.travis.yml` to test against that version. If you only support Julia 0.4 or higher, the lower bound should be 0.4.

glesica:
Oops, I thought I changed the lower bound. Should I manually update that in the branch, or just scrap the PR and start over?

NK literally stands for "N", and "K", the two primary parameters used in the family of models. So it really isn't an acronym in the classic sense. I realize it isn't a very descriptive name, but I couldn't think of anything better without making it something confusing or very long.

tkelman:
You can change it both here https://github.com/glesica/METADATA.jl/blob/pull-request/324ad5d0/NK/versions/0.1.0/requires and here https://github.com/glesica/NK.jl/blob/master/REQUIRE

A little on the long side isn't a big deal, tab-completion works well on package names. `NKFitnessLandscapes` maybe?

glesica:
OK, updated the lower version bound. How about `NKLandscapes`, since that is the name of the basic data structure?

tkelman:
`NKLandscapes` looks pretty unambiguously googlable to me.

glesica:
So now how do I go about changing the name? Looks like I'll need to change the directory name in the METADATA. I assume I should change my repo name as well. Anything else?

tkelman:
Rename this file https://github.com/glesica/NKLandscapes.jl/blob/master/src/NK.jl, grep and sed to replace `NK` with `NKLandscapes` anywhere it appeared in the code - at least the `module` line in the main src file and the `using` line in `runtests.jl`, maybe others? The sha1 value here should correspond to a post-rename version once it passes tests, ideally.

glesica:
OK, I think I did everything correctly. The tests pass, and I pointed the PR at a new tag that has everything renamed in the repo.

glesica:
Actually, one of the test architectures failed, but it appears to have been a bad download from AWS when it tried to fetch the OS X image. I'm re-running.

tkelman:
LGTM. Will leave open for a little while to see if anyone else has any comments, and merge tomorrow if not.

glesica:
Cool, thanks for your assistance!

quinnj:
+1. LGTM

PR #4264, opened 2015-12-21
Tag AffineTransforms v0.1.3

timholy:
Fixes for array transforms with the parameters specified by non-real numbers (as happens when using ForwardDiff etc).


tkelman:
Do you still need to support prereleases? If not, drop the trailing dash.

PR #4265, opened 2015-12-22
Register OBJs

ericjang:
New package for importing OBJs

### DIFF
--- a/OBJs/url
+++ b/OBJs/url
@@ -1,0 +1,1 @@
+git://github.com/ericjang/OBJs.jl.git


tkelman:
"OBJ" can mean a lot of things. Would "WavefrontOBJ" be more unambiguous? Also note that if you translated any code from https://github.com/syoyo/tinyobjloader, you need proper attribution according to that project's BSD license (which they should really explicitly write out in a separate license file).

ericjang:
Sure, I'll re-submit the package under "WavefrontOBJs".

I did use tinyobjloader as a reference during implementation, but it's not
an exact translation (julia handles a lot of the low-level things for free)
and there are some differences in how I am parsing certain tags ("o","g",
etc.). I'll include an attribution anyway.

tkelman:
closing in favor of https://github.com/JuliaLang/METADATA.jl/pull/4269

PR #4270, opened 2015-12-22
WavefrontOBJs package

ericjang:


ericjang:
Oops, sorry - still trying to figure out package management. need to get rid of old commit

timholy:
Comparison to https://github.com/JuliaIO/WavefrontObj.jl?

ericjang:
Ah, I didn't know that package existed. It looks like that one has all the
code commented out though? I'm hesitant to provide a patch that re-writes
everything.

- Eric

timholy:
You just might want to post an issue in that package. Assuming it's registered with FileIO, you'd also want to change the registration.

sjkelly:
There is also: https://github.com/JuliaIO/MeshIO.jl/blob/master/src/io/obj.jl Which works for me, minus 'MTL' files, which I do not use yet. Maybe @SimonDanisch can comment? Regardless, I think merging with MeshIO would be the best way forward here!

ericjang:
That's cool - I'm more confident in my MTL parser implementation than my OBJ parser. Let me know what I can do to help.

PR #4276, opened 2015-12-24
Tag BackpropNeuralNet v0.1.0

compressed:


tkelman:
If the previous tags supported julia 0.3, then this tag should use a new minor version number 0.1.0

compressed:
@tkelman agreed!

I've rebased this now.

tkelman:
the package itself should really add a top level REQUIRE file

compressed:
Oh good point, let me take care of that. Thanks.

PR #4285, opened 2015-12-27
Fall back to last passing versions of AppConf and Diversity on Julia 0.3

tkelman:
cc @tmlbl and @richardreeve - both of these tags fail on Julia 0.3. Diversity.jl is fixed on master and just needs a new tag, AppConf.jl can be fixed by https://github.com/tmlbl/AppConf.jl/pull/4. Adjusting the requires of these existing tags so these versions don't get installed by users on Julia 0.3 when they do `Pkg.update()` is good to do in the meantime, will get/keep them passing on PackageEvaluator for one thing.

### DIFF
--- a/AppConf/versions/0.1.0/requires
+++ b/AppConf/versions/0.1.0/requires
@@ -1,2 +1,2 @@
-julia 0.3
+julia 0.4
 Compat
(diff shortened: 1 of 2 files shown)


richardreeve:
This is fine by me. I'm working on getting the code running across 0.3-0.5, and @tkelman has helped with this, but there are still problems with Union{} even using Compat. I'm trying to fix this at source as I shouldn't really be using Union anyway. Then I'll bump Diversity to 0.2.5.

PR #4289, opened 2015-12-28
FaceDatasets v0.2.0

dfdx:
Ability to load only last images/shapes from CK+ dataset. 
Dropped support for Julia 0.3.


tkelman:
Should remove the 0.1.5 tag, or make it 0.4-only.

dfdx:
Ok, I couldn't find a way to delete old tag without re-tagging and republishing, so here's one more attempt: #4291

PR #4300, opened 2015-12-29
Tag NLsolve v0.6.0

spencerlyon2:
Uses ForwardDiff.jl for autodiff and fixes bug in newton


tkelman:
One of the transitive dependencies here apparently depends on BaseTestNext? Odd. https://travis-ci.org/EconForge/NLsolve.jl/jobs/99303609

If this tag can't be installed and pass tests on Julia 0.3 then it should be 0.4-only.

KristofferC:
Hmm, should we revert the BaseTestNext commit or move to only support 0.4?

spencerlyon2:
I'd be happy to bump to only 0.4 support. I don't plan on ever using 0.3 again, so I'm not too eager to continue officially supporting it.

I won't be able to make this change until tomorrow at the earliest,  but anyone else is welcome to adjust the require and finish the tag

KristofferC:
I also think there is no real point in keeping 0.3 support. There is no rush for me so if you tag whenever you have time that works great for me.

tkelman:
closing in favor of #4343

PR #4308, opened 2015-12-30
Register GridInterpolations: v0.0.1

mykelk:
Multidimensional grid interpolation in arbitrary dimensions.


timholy:
Presumably you're aware of Interpolations.jl, which has similar functionality (except for the simplex interpolation) and is considerably faster (see the benchmarks on the Interpolations.jl README).

But if your main interest is in the simplex interpolation, that (currently) is unique to GridInterpolations and would be a good reason to register this.

mykelk:
Yep, the readme points to Interpolations.jl, and I like that package a lot especially with the new API for Interpolations.jl. GridInterpolations.jl was originally developed for doing dynamic programming on very large multidimensional grids (10s of GB stored in a memory mapped file). The API is really optimized for this use case. Probably the main attraction of GridInterpolations is that it does [fast simplex interpolation](https://github.com/sisl/GridInterpolations.jl/issues/6) on a grid. Simplex interpolation is very attractive when going to higher dimensions; multilinear interpolation, of course, scales exponentially. See the [issue](https://github.com/tlycken/Interpolations.jl/issues/28) filed here discussing the connection with Interpolations.jl. I'm happy for our simplex code to be taken and integrated into Interpolations.jl (or elsewhere).

timholy:
Yes, the simplex interpolation is extremely attractive. I haven't yet looked carefully to tell whether simplex interpolation is something that fits naturally into Interpolations. I'm tempted to merge this PR unless you decide otherwise.

For your enjoyment: it only covers specific cases, but when applicable https://github.com/tlycken/Interpolations.jl/pull/79 can give you linear rather than exponential performance with multilinear interpolation.

mykelk:
Cool! I like that trick. It wouldn't really work though for our use case, though. As for integrating into Interpolations.jl, I think the syntax would be something like this:
```julia
A = rand(8,20)
knots = ([x^2 for x = 1:8], [0.2y for y = 1:20])
itp = interpolate(knots, A, Simplexed(Linear()))
itp[4,1.2]
```

Yes, I'm fine with this being merged.

PR #4312, opened 2015-12-30
Tag AppConf v0.1.1

tmlbl:
Restores 0.3 compat and added template detection


tkelman:
Great, thanks. Can be convenient to have the tags pushed to github so they show up at https://github.com/tmlbl/AppConf.jl/releases

PR #4317, opened 2015-12-31
Tag YT v0.2.1

jzuhone:


ViralBShah:
Hi @jzuhone - I just saw this and found the package name cryptic. I realize that is named after the equally cryptic Python package. Perhaps we can call this PyYT.jl or something a bit more descriptive?

YT/versions/0.2.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4-
```
tkelman: The trailing dash refers to prereleases, and is probably not what you want here.

tkelman:
Since this is the first version that drops julia 0.3 support, it should use a new minor package version.

It was registered some time ago so probably too late to worry about renaming.

PR #4332, opened 2016-01-03
Register ZIP

samoconnor:
High level interface for ZIP Archives.

Implemented using Fazlul Shahriar's [ZipFile](https://github.com/fhs/ZipFile.jl) IO module.
@fhs [suggested](https://github.com/fhs/ZipFile.jl/pull/24#issuecomment-167274810) that this kind of high level interface should go in a new package.

This module hides the heavy lifting and implementation detail of [ZipFile](https://github.com/fhs/ZipFile.jl) and exports only 3 symbols `open_zip`, `create_zip` and `unzip` (@timholy's unzip() function per fhs/ZipFile.jl#16)

### DIFF
--- a/ZIP/url
+++ b/ZIP/url
@@ -1,0 +1,1 @@
+git://github.com/samoconnor/ZIP.jl.git
--- a/ZIP/versions/0.0.1/requires
+++ b/ZIP/versions/0.0.1/requires
@@ -1,0 +1,2 @@
+julia 0.4
+ZipFile 0.2.6
--- a/ZIP/versions/0.0.1/sha1
+++ b/ZIP/versions/0.0.1/sha1
@@ -1,0 +1,1 @@
+3babcc58c1689aae4a2f8ca2324ce91a8e003da9


tkelman:
Bit confusing that we now have Gzip, Libz, Zlib, ZipFile, LibArchive, probably more? Though I think the only existing one that's pure Julia is ZipFile.

nalimilan:
Have you considered including these functions to ZipFile.jl? Anyway, the name ZIP.jl sounds too general to me, given that the package only offers convenience functions.

nalimilan:
Ah, OK, you did: https://github.com/fhs/ZipFile.jl/pull/24 @fhs suggested ZipFileUtils.jl as a name. Why not use that? I will make it more likely that potential users will find your package and understand its purpose.

samoconnor:
This package is intended to provide a clean user friendly interface for ZIP Archives.
It does not expose ZipFile.jl's interface. It could easily be changed to use another backend if a faster/smaller/shiner one became available.

ZipFileUtils.jl suggests that this is an add-on to ZipFile.jl
From the user's point of view, ZipFile.jl is just implementation detail, the user does not need to know about ZipFile.jl.

The package name is intended to be generic because the API is intended to be generic.

The Project description text that will show up in the package list is "Simple ZIP Archive Interface". I think this should make the intention clear to anyone who searches for "ZIP" on the [package list page](http://pkg.julialang.org)

fhs:
I call my package ZipFile because it avoids confusion with identifiers already used by Julia: zip and Zip. The name "ZIP" does not seem like a good choice. The user will have to remember 3 different capitalization of the word zip, ZIP being totally unrelated to the other two.

nalimilan:
Why not something like `SimpleZip`?

samoconnor:
I'm not a fan of names that are boasting about something: ZIPPro, FastZIP, LightZIP, EasyZIP etc.
I'd prefer to keep it factual.
Given that zip is an overloaded term, what about "ZIPArchive".
(PKZIP would be the obvious choice, but that's probably a trademark.)

samoconnor:
@fhs, would you reconsider merging this Into ZipFile?

fhs:
@samoconnor Ok, send a pull request to ZipFile.jl. Add it as a submodule of ZipFile.

samoconnor:
@fhs, see https://github.com/fhs/ZipFile.jl/pull/25

samoconnor:
Closing in favour of fhs/ZipFile.jl#25

PR #4338, opened 2016-01-03
Tag Currencies v0.2.0 [git://github.com/TotalVerb/Currencies.jl.git]

TotalVerb:
* Drop deprecated `Base.int` overload; use field access (`.val`) instead. I don't expect this to break any code, since `int()` was not very useful in the first place, but I bumped the minor version anyway.
* Drop dependency on `DataStructures`.
* Add a few more format options, including `:latex`, `:plain`, `:indian`, and `:local`.
* Add a few more currency symbols.
* Rewrite basket code to be leaner and more efficient (no functional change).
* Rewrite format code to be clearer and more efficient (no functional change).
* Extend ECB exchange rate cache.

Although there aren't any benchmarks, Travis build times are down for this package despite more tests, so this release comes with a performance boost both in package load time and run time.


tkelman:
:100: nice changelog

PR #4341, opened 2016-01-04
Register, tag ClassicalCiphers v0.0.1

Smaug123:
A package which implements classical encryption algorithms. Ultimately, should end up with automatic cracking facilities too. (This time I remembered to `Pkg.tag()` it.)


tkelman:
LGTM, will merge tomorrow if no other feedback.

PR #4342, opened 2016-01-04
Register AWSCore: v0.0.1

samoconnor:
https://github.com/samoconnor/AWSCore.jl


samoconnor:
currently blocked waiting for samoconnor/Retry.jl to be registered

samoconnor:
Dependancies of AWSCore.jl are now registered.

AWSCore.jl now passes tests on travis-ci using [AWS Credentials baked into the travis config file](https://github.com/samoconnor/AWSCore.jl/blob/master/.travis.yml): 
[Build Status](https://travis-ci.org/samoconnor/AWSCore.jl)


AWS*.jl service module tests won't run on travis yet because they depend on AWSCore which is not yet registered. [Build Status](https://travis-ci.org/samoconnor/AWSS3.jl)

ViralBShah:
Is it worth having a JuliaCloud organization to collect various packages related to AWS and other cloud services? There is [AWS.jl](https://github.com/JuliaParallel/AWS.jl), which should perhaps be renamed, and there is also [DynamoDB.jl](https://github.com/dls/DynamoDB.jl).

Just curious to know how these packages differ from the existing ones, and if we should do something to unify the efforts.

samoconnor:
I have tried to split my AWS stuff up into modules in the hope that parts of it can be re-used even if existing users don't want to switch over to it completely. (e.g. @dls should be able to hook DynamoDB.jl into AWSCore fairly easily) 

Here is my assessment of AWS.jl (@dls may have other comments?)...

_None of this is mean as a criticism of the efforts that the AWS.jl authors have put in. I think it is simply a matter of evolution and maturity of the codebase. When I wrote my first Tcl AWS library many years ago it spent a long time looking a lot like AWS.jl looks today._

The codebase seems to be not actively maintained and not well factored. For example:

 - [the first thing you see in src/AWS.jl](https://github.com/JuliaParallel/AWS.jl/blob/master/src/AWS.jl#L9) is a bunch of `EP_*` const string definitions for endpoints that are not used anywhere. Although this is src/AWS.jl, theses are all EC2 endpoints addresses, so it seems that maybe src/AWS.jl should be renamed to EC2.jl? Any why define const global variables for these strings that are already well defined identifiers? Why not just pass in the strings?

 - A bit further down src/AWS.jl tries to load credentials from some non-standard environment variables and then from a not standard ~/.awssecret file. This code runs at the global scope, so there is no way to control when it runs, or to re-load the credentials.

 - In the AWSEnv constructor, there is a call to attempt to load credentials from the EC2 meta data service, but these are only temporary credentials! There does not seem to be any mechanism to refresh them.

 - In src/EC2.jl there is another perplexing set of const strings that seem to define non-standard names for EC2 instance types. e.g. `VM_CPU_XL = "c1.xlarge"` and `VM_HMCC_OXL = "cr1.8xlarge`. It is not clear what the purpose of this is.

 - Next is a function that looks like a very verbose implementation of `Dates.format(t, "yyyy-mm-ddTHH:MM:SSZ")`
 
 - The `ec2_execute()` function combines request signing, HTTP request processing and error processing. There does not seem to be any retry for HTTP codes > 500 or and processing of re-directs.

 - Over in src/S3.jl there is another implementation of request signing, HTTP request processing and error handling, but again no retry logic.

 - In `ec2_types.jl`there are about 10,000 lines of types generated from a WSDL file (for SOAP), but this is a REST implementation, so the WSDL file is not really applicable. It is also unclear why all these types are required when the REST interface returns nicely formatted XLM that can trivially be used directly.

AWSCore, AWSS3 etc include the following features:

 - Interfaces for S3, EC2, IAM, Lambda, SQS, SNS.

 - Trivial addition of new interfaces due to general implementation in AWSCore (e.g. I have done simple proof-of-concept SDB and DynamoDB interfaces).

 - AWSCore has a [unified implementation of request signing](https://github.com/samoconnor/AWSCore.jl/blob/master/src/sign.jl) that works across all services and implements Amazon's latest V4 signatures. 

 - AWSCore has [generalised functions](https://github.com/samoconnor/AWSCore.jl/blob/master/src/names.jl) for constructing AWS Endpoint URLs as well as ARNs (Amazon Resource Names). These functions hide various "local knowledge" oddnesses, e.g. that S3 and SDB have no region suffix in "us-east-1".

 - AWSCore has an abstract [AWSException](https://github.com/samoconnor/AWSCore.jl/blob/master/src/AWSException.jl) type and an exception constructor that knows how to parse meaningful error messages from a range of JSON and XML formats used by various AWS services.

 - AWSCore has an [AWSCredentials](https://github.com/samoconnor/AWSCore.jl/blob/master/src/AWSCredentials.jl) type that knows how to load credentials from: standard environment variables (e.g. compatible with Lambda sandboxes); standard ~/.aws/credentials files (e.g. compatible with Amazon's own CLI tools ant Boto); EC2 meta-data (with automatic support for refreshing expired credentials).

 - AWSCore has execution environment detection utilities: `localhost_is_lambda()`, `localhost_is_ec2()`. 

 - AWSCore has a [http layer](https://github.com/samoconnor/AWSCore.jl/blob/master/src/http.jl) that supports automatic retry with exponential back-off for UVErrors and HTTP status <200 or >= 500. This layer also supports a streaming mode that allows very large files to be transferred from S3 without RAM overhead.

 - On top of the http layer is an [AWS request layer]( - AWSCore has tests that run on Travis.
) that handles: request signing, HTTP redirects (and re-signing of resulting requests); automatic refresh of expired instance credential tokens; and automatic parsing of XML and JSON responses.

 - AWSS3.jl supports: generation of pre-signed S3 URLs; upload with setting of appropriate Content-Type;  Listing and management of versioned objects; Listing of very large buckets where multiple requests are required to retrieve the whole list; 

 - AWSSQS.jl supports: queue creation/deletion; message sending and receiving; infinite iterator for processing messages; queue status functions; automatic retry logic for common errors such as AWS.SimpleQueueService.NonExistentQueue, QueueAlreadyExists , AWS.SimpleQueueService.QueueDeletedRecently, 

 - AWSSNS.jl supports: creation/deletion of topics; sending of notifications; subscription of SQS endpoints with creation of appropriate security policies.

 - AWSIAM.jl supports basic user, role and policy manipulation functions (some still a work in progress).

 - AWSEC2.jl supports: creating and deleting EC2 servers; automatic setup of instance profiles and credentials (including management of cleaning up old ones where there are clashes); MIME-Multipart UserData / Cloudint support.

 - AWSLambda.jl supports: automated creation of an EC2 server that compiles a Julia runtime and packages it into a format suitable for running in a Lambda sandbox; deployment of Julia functions to AWS Lambda; remote-call of Lambda-hosted Julia functions with serialisation/deserialisation of arguments and results.

 - AWSCore has tests that run on Travis.

samoconnor:
I've just updated the documentation for AWSLambda to include some better explained examples: https://github.com/samoconnor/AWSLambda.jl

Feedback is welcome.

ViralBShah:
Anything holding this merge?

tkelman:
It's a lot of separate individually small packages, doesn't precompilation alleviate some of the reasons you might want to split everything up? Given we don't have a great mechanism for conditional dependencies I suppose this is the next best option.

samoconnor:
I think AWS is *way* to huge of an interface to have a single package.
The list that starts [S3, SQS, SDB, SNS....] has at least another 50 items.

It is likely that the per-service packages will continue to evolve, but the core package and the packages for the core service (i.e. IAM, S3, SQS, EC2, Lambda) will stabilise more quickly.

If I'm a user who depends on S3 for critical data storage, I'm much happier to be using a version of AWSCore and AWSS3 that hasn't changed for ages. The alternative is a monolithic AWSEverything that is updated frequently to track bleeding edge niche services. I don't want to read all those release notes to figure out if something changed that effects my simple use case.

samoconnor:
thx @tkelman.

>AWS*.jl service module tests won't run on travis yet because they depend on AWSCore which is not yet registered. Build Status

Now that this is merged it may take me a little while to get travis running for each service module.

tkelman:
Could have registered in 2 or 3 batches to deal with that issue.

PR #4345, opened 2016-01-05
Tag v0.2.1

fredo-dedup:
adapting to breaking change #13612 in JuliaLang


tkelman:
Would be good to push tags to the package so they show up at https://github.com/JuliaDiff/ReverseDiffSource.jl/releases. And mention which package you're tagging in the commit message.

PR #4349, opened 2016-01-06
Update Faker 0.0.2

codeneomatrix:
Released properties
* fixed bug Cannot `convert` Char an AbstractString
* precompile
* update string -> AbstractString


Faker/versions/0.0.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4-
```
tkelman: the trailing - refers to prereleases and is not necessary any more

tkelman:
good to push to git tags so they show up at https://github.com/codeneomatrix/Faker.jl/releases

PR #4350, opened 2016-01-06
Pull request/cb000b18

c-s:
Hi,

I would like to rename the package MultidimensionalTables.jl to DataCubes.jl. The original name was too long, and the new one seems to satisfy the Julia package naming convention: it is not an abbreviation and has a clear meaning in a wiki page. Although a cube is 3 dimensional, the terminology can be used for higher dimensional arrays, which is the case for the current package.

To rename the package, I duplicated MultidimensionalTables.jl to DataCubes.jl in github, and put deprecation warning signs in MultidimensionalTables.jl, and pointed to the new package DataCubes.jl.

In addition to renaming, overall performance has been enhanced after profiling several basic functionalities such as aggregation, sort, join, update, etc.

Thanks.

Chang-Soon


kmsquire:
Seems okay to me.  You probably didn't need to re-register v0.0.1 and v0.0.2, but it doesn't hurt.

tkelman:
We should deprecate all tags under the old name by adding a julia version upper bound.

tkelman:
closing in favor of #4356

PR #4356, opened 2016-01-07
Pull request/b1d07cfe

c-s:
Hi,

I am not sure how I can update the pull request cb000b18 #4350 , so I created another pull request reflecting suggestions. I registered only v0.0.3 and did not register v0.0.1 and v0.0.2 for DataCubes.jl, and I put a julia version upper bound in the old name package.

Thanks and please let me know if these changes work.

Chang-Soon

Below is my message for the previous pull request cb000b18 #4350 
---------------------------------------------------------------------------------------------------------
I would like to rename the package MultidimensionalTables.jl to DataCubes.jl. The original name was too long, and the new one seems to satisfy the Julia package naming convention: it is not an abbreviation and has a clear meaning in a wiki page. Although a cube is 3 dimensional, the terminology can be used for higher dimensional arrays, which is the case for the current package.

To rename the package, I duplicated MultidimensionalTables.jl to DataCubes.jl in github, and put deprecation warning signs in MultidimensionalTables.jl, and pointed to the new package DataCubes.jl.

In addition to renaming, overall performance has been enhanced after profiling several basic functionalities such as aggregation, sort, join, update, etc.

Thanks.

Chang-Soon


tkelman:
You can modify an existing pull request by pushing new commits to the same branch. The upper bound on past versions of the package will only be effective if it's added on all of the existing tags.

c-s:
Thanks for the comments. I used Pkg.publish() to push changes to the new pull-request branch, but I guess I can directly use git push. I am not an expert in git, so it is not clear to me how I can add upper bound in previous tags. Do I need to check out each past tag, edit the REQUIRE file, commit the change and then move the tag to point to the new one?

tkelman:
You can modify the content of the `requires` files for the past tags in this branch of your METADATA fork: https://github.com/c-s/METADATA.jl/tree/pull-request/b1d07cfe/MultidimensionalTables/versions. No need to modify the package itself.

c-s:
That's great! I just updated the requires files and all tags now have a julia version upper bound. Thanks.

tkelman:
Looks good. If no comments or objections to the new DataCubes name, or any issues with the deprecation as done here, will merge tomorrow.

PR #4359, opened 2016-01-07
Tag ConjugatePriors v0.1.2

kmsquire:
also run travis on v0.3


tkelman:
Travis-only changes aren't really worth tagging IMO, but more trouble to undo the tag than just merge it.

kmsquire:
Good point.  I'm in cleaning mode, and tagged out of habit.  Cheers!

PR #4363, opened 2016-01-07
Add FusionDirect, a tool to detect gene fusion

sfchen:
FusionDirect is a tool to detect gene fusion directly from fastq files, it has following features:
* no alignment needed, it just reads fastq files of pair-end sequencing output
* output fusion pattern (gene and position), along with the reads support this fusion
* ultra sensitive, comparing to delly, factera or other tools
* output file is a standard fasta file, which can be used to verify fusions using blast or other tools
* very suitable for detecting fusions from cancer target sequencing data (exom seq or panel seq)


tkelman:
Does Travis CI pass if you enable it?

sfchen:
It need some big file (reference data) to run test, so I haven't added tests.
I am looking for other ways to enable testing and Travis CI.

PR #4365, opened 2016-01-07
Tag GeoStatsImages v0.0.4 [git://github.com/juliohm/GeoStatsImages.jl…

juliohm:


tkelman:
would be good to push the git tags so they show up at https://github.com/juliohm/GeoStatsImages.jl/releases

PR #4367, opened 2016-01-07
Register Lisp: v0.1.0

swadey:


StefanKarpinski:
Gerry Sussman will be very pleased.

swadey:
lol

StefanKarpinski:
His main comment on @JeffBezanson's defense was "this is great, but can I have it with Lisp syntax?"

swadey:
that's what unholy macros like this are for...

But seriously, despite its elegance, I can't imagine wanting to do math with lisp syntax.

JeffBezanson:
Nice!

I know I'm way late here but I would reconsider the name of this package. Perhaps something like `LispSyntax` or `SExpressions`.

swadey:
Makes sense.  Will do.

PR #4372, opened 2016-01-08
Tag Mosek v0.3.0

ulfworsoe:


Mosek/versions/0.3.0/sha1
```diff
@@ -0,0 +1 @@
+f64ec988f580d30ef2f3c4af5682ca58e53f4ef1
```
tkelman: looks like this commit is from november, was that intended?

ulfworsoe:
Yes and no. Pkg.publish() has crashed a lot for me because of two-factor authentication on github. This may be something old that was never correctly published. There should be a pull request from around jan 7.

PR #4373, opened 2016-01-08
Tag Blink v0.3.0

spencerlyon2:
- Adds `evalwith` function to blink.js
- Uses bindeps to set up dependencies (download and unzip electron)
- Adds support for resolving promises in `js(...; callback=true)` calls


tkelman:
Turn Travis on to at least verify installation and dependencies? This file needs to not stay empty: https://github.com/JunoLab/Blink.jl/blob/master/test/runtests.jl

PR #4376, opened 2016-01-09
Tag CUSOLVER

kshyatt:
Had to bump minor version because this package now supports `0.4`.

The dense and sparse solvers work!


tkelman:
Should the 0.0.2 tag be removed?

PR #4377, opened 2016-01-09
New Tags

r2dbg:
Updated deprecated types for v0.5


tkelman:
`julia 0.4-` means "prereleases of julia 0.4 or newer." You probably don't want to support every possible prerelease version, dropping the trailing dash and changing it to `julia 0.4` means "releases of julia 0.4 or newer."

tkelman:
It can also be useful for navigating code changes on github if the git tags are pushed so they show up under https://github.com/r2dbg/NMEA.jl/releases etc

r2dbg:
To be honest - I just ran Pkg.tag(...).  Are you saying that with the trailing dash my new tag will be used  by ony prereleases 0.4 or newer - not releases 0.4 or newer too?  The releases are more important to me ofc, so shall we close this request and I'll issue another pull request with the trailing dash removed?

tkelman:
"Prereleases or newer" includes releases. So it's not absolutely crucial to fix right away. But I suggest removing the trailing dash in each packages' `REQUIRE` file before you make your next tag. Does `Pkg.publish()` work for you? I think that should automatically push the git tags.

tkelman:
Also note that your travis files are testing against nightlies from the ppa, rather than releases. There's now a `language: julia` option you can use to test the packages on both linux and osx (and use the faster `sudo: false` docker build workers if you'd like), ref https://docs.travis-ci.com/user/languages/julia

r2dbg:
Thanks for taking the time to explain - I'll surely make these modifications for future tags. I tried Pkg.publish when I started ConfParser a year ago, but it wasn't working for me. I've since have been using Pkg.tag and pushing manually from METADATA. I'll try Pkg.publish for future commits. I think it may be worthwhile for me to read the Pkg development wiki again and make it a point to use the recommended process. ;)

PR #4378, opened 2016-01-09
Tag EEG v0.0.2

codles:


tkelman:
Given that the previous tag supported Julia 0.3 and this tag does not, this should probably use a new minor package version 0.1.0

codles:
Is this a set policy? If not, I would like to tag as is. I have a set of core functionality I would like to implement before v0.1.0.

tkelman:
It is suggested in the readme here, though not absolutely mandatory. It becomes much more messy if there is ever a need to release a bugfix version (from a different branch, say) that still supports julia 0.3 if it isn't using a different minor version. Since there has only been one past tag here that may not be such a big concern.

codles:
Thanks for the help. I consider this still a teething/bugfix release. Once I have basic functionality in place I will use the v0.1.0 increase standard as suggested.

tkelman:
No longer supporting a version of julia that the previous version supported seems like a semver breaking change to me, though technically all bets are off while in 0.x.y. Just would like to keep the 0.3 package ecosystem in a functioning state going forward, since the package manager design makes that possible without too much work.

PR #4382, opened 2016-01-10
Register InfoZip

samoconnor:
Similar to #4332.
Now uses InfoZip as the backend.

https://github.com/samoconnor/InfoZIP.jl


samoconnor:
thx Tony

PR #4384, opened 2016-01-10
Tag BiomolecularStructures v0.0.2

gp0:


tkelman:
Since the previous version supported julia 0.3 and this version doesn't, this should probably use a new minor package version 0.1.0

gp0:
Thanks, will tag it as such.

PR #4385, opened 2016-01-10
- rename Lisp to LispSyntax

swadey:


kmsquire:
Unfortunately, deleting `Lisp.jl` will currently cause problems for anyone who has installed it, so we can't do that.

(This needs to be documented somewhere.)

kmsquire:
(Closed so this isn't accidentally merged.)

kmsquire:
See, e.g., https://github.com/JuliaStats/Distance.jl/issues/25, for how to do this.

kmsquire:
See also, https://groups.google.com/d/msg/julia-users/NHeaVWMfRfo/G4C31JzWN0kJ, for a discussion of some problems when `Stats` was renamed to `StatsBase`.

(Of course, `Lisp.jl` won't have been installed by many people, and probably isn't depended on by any packages, so many of those issues might not be relevant.)

kmsquire:
Also relevant:

* https://github.com/JuliaLang/METADATA.jl/issues/347#issuecomment-35015012
* https://github.com/JuliaLang/PkgDev.jl/issues/21

swadey:
yeah, this was in response to jeff's request (here: #4367).  I figured it would be better to rename now rather than waiting until people install it.

PR #4386, opened 2016-01-10
Lisp rename to LispSyntax and tag v0.1.2

swadey:


tkelman:
Would have been good to leave the original registration PR open longer to deal with this kind of thing @StefanKarpinski

kmsquire:
I wasn't sure exactly what (if anything) would break if `Lisp.jl` was removed from METADATA, so I tested it by setting my own METADATA.jl as a branch, installing `Lisp.jl`, and merging this.  It doesn't look like it will actually cause any problems.

From past experience, there would almost certainly be problems if there were any external packages which depended on it, but since that isn't the case, this is probably okay to merge once CI passes.

tkelman:
https://github.com/JuliaLang/METADATA.jl/pull/4356 was a recent example of the conservative approach of deprecating the old name. I think this'll work but the old tags will be broken since they'll have the wrong file name in src.

swadey:
I can do something similar but do you think that's necessary given that the package has only lived in metadata for a day?

kmsquire:
It would seem that most of the people actively interested in this package already know about the name change, so the only people that would be affected would be people who

1. update/clone METADATA between when this package was registered and when this PR is merged
2. do not update after that
3. attempt to install `Lisp`

Assuming this is right, it's probably okay in this case, but I'll defer to @tkelman.

StefanKarpinski:
> Would have been good to leave the original registration PR open longer to deal with this kind of thing @StefanKarpinski

Should I have waited three days?

tkelman:
I usually wait about one day for name or other initial feedback on new packages. I would have commented on this particular packages' registration that Lisp.jl was perhaps not specific enough of a name, but you had merged the PR a few hours before I first saw it.

PR #4390, opened 2016-01-11
Tag DualNumbers v0.2.1

mlubin:
CC @timholy @jrevels


timholy:
Thanks a ton, @mlubin.

PR #4398, opened 2016-01-12
Tag ImageQuilting v0.2.5 [git://github.com/juliohm/ImageQuilting.jl.git]

juliohm:


tkelman:
Would be good to push the git tags so they show up at https://github.com/juliohm/ImageQuilting.jl/releases

PR #4403, opened 2016-01-13
Tag Measures v0.0.2

shashi:
Closes https://github.com/dcjones/Measures.jl/issues/5


tkelman:
can you also push the tag? https://github.com/dcjones/Measures.jl/releases

PR #4414, opened 2016-01-14
Update Atom Packages

one-more-minute:


tkelman:
https://github.com/JunoLab/Media.jl/commit/827b8ad8c0cc203f1edf4680ef6f079009f70a27 breaks Media on 0.3, which it still claims to support. PLEASE add tests.

PR #4415, opened 2016-01-14
Register EasyPkg: v0.0.1

oschulz:


tkelman:
Would be good to cross-reference https://github.com/JuliaLang/FactCheck.jl/pull/71 for some of the motivation here. The name is a little bit on the ["this but better"](https://github.com/JuliaLang/METADATA.jl/pull/4332#issuecomment-168828057) side but I can't think of anything it could be called that would be clearer.

tkelman:
Also note that conditionally importing FactCheck [like this](https://github.com/oschulz/EasyPkg.jl/blob/68c50cfdff0f992ac8e15f7f40dcf742a7f9f3c1/src/testing/test_exec.jl#L5) interacts poorly with precompilation. See
```jl
julia> Pkg.clone("https://github.com/oschulz/EasyPkg.jl")
INFO: Cloning EasyPkg from https://github.com/oschulz/EasyPkg.jl
INFO: Computing changes...

julia> Pkg.add("FactCheck")
INFO: Installing Compat v0.7.8
INFO: Installing FactCheck v0.4.2
INFO: Package database updated

julia> Pkg.test("EasyPkg")
INFO: Computing test dependencies for EasyPkg...
INFO: Testing EasyPkg
INFO: Running tests in "foo/foo.jl"
Some facts
2 facts verified.
INFO: Running tests in "bar.jl"
Some other facts
2 facts verified.
INFO: EasyPkg tests passed

julia> Pkg.rm("FactCheck")
INFO: Removing Compat v0.7.8
INFO: Removing FactCheck v0.4.2
INFO: Package database updated

julia> cd(Pkg.dir("EasyPkg","test"))

julia> using EasyPkg

julia> EasyPkg.run_all_tests()
INFO: Running tests in "foo/foo.jl"
ERROR: LoadError: ArgumentError: FactCheck not found in path
 in require at ./loading.jl:233
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in impl at /home/tkelman/.julia/v0.4/EasyPkg/src/testing/test_exec.jl:69
 in impl at /home/tkelman/.julia/v0.4/EasyPkg/src/testing/test_exec.jl:60
 in run_all_tests at /home/tkelman/.julia/v0.4/EasyPkg/src/testing/test_exec.jl:73
 in run_all_tests at /home/tkelman/.julia/v0.4/EasyPkg/src/testing/test_exec.jl:35
while loading /home/tkelman/.julia/v0.4/EasyPkg/test/foo/foo.jl, in expression starting on line 3
```

oschulz:
> @tkelman: Would be good to cross-reference JuliaLang/FactCheck.jl#71 for some of the motivation here. The name is a little bit on the "this but better" side but I can't think of anything it could be called that would be clearer.

It's actually completely orthogonal to FactCheck (you're right, I'll put a link to it in the README). EasyPkg ist about automatizing test execution on a file-level, it doesn't tell you how to code your tests. Though if you do it with fact check, it'll pick up the FactCheck exit code.

On the "src" side, there's the include helper (and maybe more, later on). My goal was basically to reduce the boiler plate code typically found in Julia project's "Project.jl" and "runtests.jl" files.

As for the name - I didn't come up with anything better than "EasyPkg", but I'm certainly open to suggestions.

oschulz:
> @tkelman: Also note that conditionally importing FactCheck like this interacts poorly with precompilation. See [...]

Darn! Hm, what would you suggest?

* Disable precompilation (not that much code in EasyPkg anyhow)?
* Depend on FactCheck? Though I wanted to be able to add support for other testing frameworks later on, without forcing the user to get all of them.
* Another option - Julia experts?

tkelman:
You can do the importing in an `__init__` method in a way that the result of whether it succeeds or fails doesn't get saved into the package. Though actually I think my test wasn't demonstrating what I thought it was, so ignore it. You just always test against FactCheck and the steps above bypassed `test/REQUIRE`. It looks like you depend on FactCheck via a runtime try-catch so this isn't the same issue I've seen before with conditional dependencies.

oschulz:
> @tkelman You can do the importing in an __init__ method in a way that the result of whether it succeeds or fails doesn't get saved into the package

Hm, I just tried, but Julia says:
```
ERROR: LoadError: error compiling __init__: unsupported or misplaced expression "import" in function __init__
```

oschulz:
So I guess the easiest solution for now is to just require FactCheck? Most people will have it installed anyhow, I guess, and it's not a heavy dependency.

tkelman:
I think you're actually fine since all of your conditional imports are under try catches and I don't think they're being stored in the precompiled data. BaseTestNext should mostly start replacing FactCheck now, but that's a different issue.

I think this is fine to register, unless anyone can come up with good suggestions on the name that you like. "ProjectHelpers" or something to that effect might work. When I initially saw the EasyPkg name I thought it might have to do more with the user-facing package management functions like add, update, etc.

oschulz:
> @tkelman: I think this is fine to register

Thanks! Uh, unfortunately, I think I just found an issue with the FactCheck integration. Lemme check it, before registering.

oschulz:
@tkelman: I did some more tests (and had to fix a bug, too): With precompilation, it indeed doesn't work if FactCheck is not already installed then installing EasyPkg. There's no error, but EasyPkg will just act as if FactCheck isn't there, even if it's now used by tests. So it looks like there's only two options:

* Disable precompilation for EasyPkg
* Make EasyPkg depend on FactCheck (and BaseTestNext, for julia 0.4)

Which would you recommend?

timholy:
For your `__init__` failure, wrap it in an `eval`. ProfileView.jl pulls some interesting tricks with conditional loading (because it supports two backends, browser and Gtk), perhaps it would be a useful model.

oschulz:
> So it looks like there's only two options

@tkelman: Well or I could just focus on v0.5 Base.test / BaseTestNext instead of FactCheck. I don't have a big codebase depending on FactCheck yet, and I assume EasyPkg would mostly be of interest to new projects anyhow.

Or Projects using EasyPkg would just have to call things like `FactCheck.exitstatus()` themselves, after  `run_all_tests()`. It's just that I was thinking about enabling EasyPkg run tests in parallel, some day, which would obviously require that it collects and merges the separate testing results and then produces the right exit code.

Anyhow, I wasn't aware of BaseTestNext yet - if this is the way things are going, then adding that as a requirement to EasyPkg (for Julia < 0.5) shouldn't be too bad.

oschulz:
@timholy: You're so right, of course, eval solves it by taking it out of precompilation.

Anyhow, I took the automatic FactCheck support out of v0.0.1 for now and put it in a branch, so I can play around with it and BaseTestNext a bit more before putting it on master. It can go into 0.0.2. ;-)

I rebase everything, to make it neat (and redid the v0.0.1 tag). Would you guys be fine with registering this as it is?

oschulz:
A procedural question for the future: Would it be better to complete the registration of a package first, before tagging v0.0.1, in general?

tkelman:
It's not so useful and we have not been accepting new registrations of packages that don't have any tagged versions, since you can't do `Pkg.add` if no registered tags exist. The package name is the most important thing to consider carefully when initially registering, since changing a package's name after the fact would break all existing tags and can cause serious problems if the package has been around long enough for anything else to start depending on it. Modifying git tags after the fact is a little undesirable, but not every git tag absolutely has to be registered to metadata - only the ones that are actually in good shape for users to run.

oschulz:
> @tkelman: It's not so useful and we have not been accepting new registrations of packages that don't have any tagged versions, since you can't do Pkg.add

Sure - I meant, waiting with the tag until there's a basic "Ok to register" decision, and then tag and update the pull request before merging. But you're right, of course, registration can just as well complete with a v0.0.2. :-)

oschulz:
Thanks!

PR #4420, opened 2016-01-15
VirtualArrays.jl

samuel-massinon-invenia:
https://github.com/invenia/VirtualArrays.jl


tkelman:
as @yuyichao said on the mailing list, would be a good idea to find a way to avoid the runtim eval for later versions.

PR #4422, opened 2016-01-16
Register PerceptualColourMaps: v0.1.0

peterkovesi:


peterkovesi:
I am hoping to register my package for the generation of perceptual colour maps and tools for rendering images with the colour maps

Peter Kovesi

tkelman:
Cool, nice looking readme. Do tests pass if you enable travis CI?

peterkovesi:
I have now enabled travis CI (there is so much to learn what to do in setting up a package...)
Unfortunately it only passes on the OSX release.  It appears I have an array dimension mismatch under the nightly version which I shall track down.  In the meantime I have removed nightly from my .travis.yml  is this an acceptable thing to do?
Under linux the tests fail with the message 'failed to pyimport("matplotlib")'  I don't know what to do about this. 

(Note I am in Western Australia so the time difference may delay my responses to your comments)

kmsquire:
Welcome, @peterkovesi! See https://github.com/peterkovesi/PerceptualColourMaps.jl/pull/1 for a fix to your travis linux problems.

Regarding testing on v0.5, you get to decide which packages you support.  Since this is just temporary, it may not be a big deal, but you could add `julia v0.4 v0.5` to your `REQUIRES` file until you resolve it, so that no one tries to install the package on v0.5.  I would say it's up to you, though.

(It's nice to see you here, BTW!  I recognized your name because I previously [ported your nonmaximal suppression code to Julia in `Images.jl`](https://github.com/timholy/Images.jl/blob/master/src/edge.jl#L298-L527).)

tkelman:
There's also a way to say "allowed failures" for specific entries in the build matrix. I'd have to check the Travis CI docs for exactly how to spell that in the yml.

No worries about time zones or delays. My working schedule often matches Australia more closely than California :)

peterkovesi:
Many thanks Tony and Kevin.  Tests are now passing on release for OSX and linux now but I need to do some homework on the changes that have been made in 0.5 to fix my nightly tests
Cheers

PR #4429, opened 2016-01-18
tag GeoEfficiency v0.8.3

DrKrar:
even more tests


tkelman:
yay tests. guess it's not strictly necessary to tag if you're only adding tests, but not like it costs much of anything either.

PR #4432, opened 2016-01-19
Tag LevelDB v1.0.2

tmlbl:
@jerryzhenleicai

Includes fixes for building on OS X


tkelman:
Since the previous tag supported Julia 0.3 and this tag does not, would be best to use a new minor (or major) version number.

PR #4435, opened 2016-01-19
Register CDD [git@github.com:blegat/CDD.jl.git]

blegat:


tkelman:
This could be useful, but I think Cddlib or CDoubleDescription would be a clearer name.

Also your bindeps script should download a known constant commit or tag, not latest master of a git repository that is liable to change at any time.

blegat:
Thanks for your advises. I have changed the download link.

As for the choice of name
* I hesitated a lot between Cddlib and CDD. The author calls the library cddlib instead of cdd to make a clear distinction between cdd, the nonlibrary version. While it is important to make this distinction for someone that want to use cdd in C, I think it will just be confusing for someone that just want to use CDD from Julia and don't care about history.
* I also hesitated between CDD and CDoubleDescription. I prefered CDD for 2 reasons
  * Because it is always called like that, there is even no explicit explanation of the acronym on the official website as far as I know. The same is done for example with GLPK.jl that is not called GNULinearProgrammingKit.jl.
  * I think that it may be a good idea to do a package DoubleDescription.jl that gives a common interfaces to CDD.jl, LRS.jl, Porta.jl, ... (the two later have yet to be developed). CDoubleDescription may be a bit confusing then...

Of course the naming choice is still open and I'd be glad to have your opinions, what do you think ?

tkelman:
GLPK is pretty widely known and unambiguous to look up from the acronym, cdd is quite a bit more niche and obscure.

mlubin:
Very cool to have this package. The name CDD doesn't bother me, but then again I've used it myself.
CC @joehuchette

joehuchette:
Very very cool. It also looks like it's been renamed to [CDoubleDescription.jl](https://github.com/blegat/CDoubleDescription.jl)?

mlubin:
The name CDoubleDescription.jl seems off to me... there's no such library called CDoubleDescription.

blegat:
I have renamed it CDoubleDescription as advised by @tkelman . I can rename it back to CDD or anything else once we have converged to a name.

joehuchette:
CDD doesn't pass the "can I google this?" test, so I'd vote for CDDLib.

nalimilan:
`CDD.jl` would likely work fine with Google searches. But of course Cddlib.jl would be even easier to find, as it would work better if people write `cddlib + julia`.

tkelman:
I like Cddlib or CDDlib.

> The author calls the library cddlib instead of cdd to make a clear distinction between cdd, the nonlibrary version.

Is the "nonlibrary version" cdd a command-line executable? Your package is written as `ccall`s into the `cddlib` library, right?

blegat:
Yes, the author first developed CDD which was command line and then developed cddlib whose functions can directly be called from C.
My package uses cddlib with ccalls.
However since on the [official webpage](https://www.inf.ethz.ch/personal/fukudak/cdd_home/), the author says that only cddlib is being updated, I don't think that the distinction is so relevant that we should add `lib` in the Julia wrapper.
If one sees `CDD.jl`, it is clear that it is using the maintained version of `CDD` using ccalls and the Julia user does not care about naming issues due to history.

Of course, the name `CDDLib.jl` could be justified by the fact that this is a polyhedron manipulation *library* but any julia package can be seen as a library...

Then again, if all of you agree on `CDDLib.jl` we can call it that way, it is the same for me :)

hayd:
+1 to `CDDLib.jl`. As long as you have CDD in the description it's going to be easily googleable.

mlubin:
Go for ``CDDLib.jl``, this package shouldn't sit unregistered for months

blegat:
Yes, it seems everybody agrees on `CDDLib.jl`. However `CDDLib.jl` depends on `Polyhedra.jl` so I guess I will need to register it before `CDDLib.jl` can be accepted.
`Polyhedra.jl` is a common interface for all polyhedra manipulation library such as CDD, LRS, Porta, ... (it is to polyhedra what `MathProgBase.jl` is to optimization). I also added higher level functionality in `Polyhedra.jl` which can be reimplemented for speed by libraries. For example visualization of 3 dimensional polyhedron using `GLVisualize.jl`, intersection of polyhedra...

I have made a new PR #4703 with both packages and tags (I have created a new one to avoid the huge merge that would be needed to continue with this pull request)

PR #4438, opened 2016-01-19
Tag NIDAQ v0.1.0

bjarthur:
NIDAQ.jl now supports julia v0.4 !


bjarthur:
btw...  had to Pkg.publish() manually because on Windows 7, Julia v0.4.2 i got `ERROR: could not spawn curl --version ...`

tkelman:
Since the previous version supported julia 0.3 but this version doesn't, it should use a new minor version number 0.1.0.

bjarthur:
makes sense.  change made.

PR #4440, opened 2016-01-20
Tag Reactive 0.3.0 and Interact 0.3.0

shashi:


tkelman:
Looks like this is the first 0.4 only Interact tag?

shashi:
Yes, because it depends on a Reactive that depends on 0.4. Should I bump minor version instead?

tkelman:
That leaves open the possibility of any future 0.3 bugfix releases if necessary, so I do think it's worth separating.

shashi:
Good idea. Will do soon.

shashi:
Done.

shashi:
ping @SimonDanisch

SimonDanisch:
nice! :+1:

PR #4442, opened 2016-01-20
Tag Lora v0.5.0

scidom:


Lora/versions/0.5.0/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4-
```
tkelman: would be good to get rid of the dash before your next tag if you don't need to support prereleases

scidom: Thanks @tkelman, I will make sure the dash is removed next time I push a minor upgrade.

PR #4443, opened 2016-01-21
Tag ODBC v0.3.11 and update ODBC URL

quinnj:
This is a final tag for the current version of the ODBC package. It will pave the way for the jq/remodel branch to be merged based on the DataStreams framework. This version of the package should work fine for 0.3/0.4, but will not be updated further for 0.5.


ODBC/versions/0.3.11/requires
```diff
@@ -0,0 +1,4 @@
+DataFrames v0.6.10
+Dates 
+julia 0.3 0.5-
```
tkelman: This version is not installable on currently nightly because of the trailing dash. The second number is an exclusive upper bound, and `0.5-` means the very first 0.5.0-dev commit.

ODBC/versions/0.3.11/requires
```diff
@@ -0,0 +1,4 @@
+DataFrames v0.6.10
```
tkelman: usually no leading `v` on these

ODBC/versions/0.3.11/requires
quinnj: That's what I want, no installations on 0.5. 0.3/0.4 only. Right?

ODBC/versions/0.3.11/requires
quinnj: Will this break it? Or should I just remember for next time?

tkelman: Not positive, but I think it's relatively lenient in what it accepts. Good to avoid being inconsistent though.

ODBC/versions/0.3.11/requires
tkelman: Do you want the switchover point to be last september, or the final 0.5.0 release?

PR #4449, opened 2016-01-22
Faker 0.0.3

codeneomatrix:
* Fixed bug in data_time, creditcard.jl,user_agent.jl,misc.jl and direccion.jl


Faker/versions/0.0.3/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4-
```
tkelman: the trailing dash refers to prereleases and is no longer necessary, would be good to remove before the next tag

PR #4453, opened 2016-01-23
Tag LIBLINEAR v0.0.3

innerlee:
change log:
* support windows now.

Thanks!


tkelman:
there's only a single dll there, so it's only supporting one of 32 bit or 64 bit, not both?

innerlee:
Yes you are right, I only tested on 64 bit machine. This is the dll contained in the official package. see
https://github.com/cjlin1/liblinear/tree/master/windows

PR #4457, opened 2016-01-24
Register MySQL

nkottary:


tkelman:
Needs a minimum julia version in `REQUIRE`, then a tagged package version

nkottary:
Added minimum julia version to `REQUIRE` and tagged.

PR #4460, opened 2016-01-25
Merly.jl

codeneomatrix:
Merly is a micro framework for quickly creating web applications in Julia with minimal effort.


kmsquire:
LGTM

PR #4463, opened 2016-01-25
Various Juno-related updates

one-more-minute:


tkelman:
Requires needs to move to a new minor version, given 0.2.0 supports 0.3 according to its require file.

Blink/versions/0.3.1/sha1
```diff
@@ -0,0 +1 @@
+054fb44e853b8fe4357a032cc61a5da7c6507942
```
tkelman: push to https://github.com/JunoLab/Blink.jl/releases

MacroTools/versions/0.2.1/sha1
```diff
@@ -0,0 +1 @@
+2acea4ff989e275aa6d4cc40a43d2bba7c006623
```
tkelman: push to https://github.com/one-more-minute/MacroTools.jl/releases. also, broken on 0.3. please turn travis on once you add tests.

PR #4464, opened 2016-01-25
Updates for tag Stan-j04-v0.3.2

goedman:
Bump to v”0.3.2”:
1. Introduce Homebrew as a way to install CmdStan on OSX.
2. Minor edits to README.
3. Added @osx Homebrew to REQUIRE.
4. Example of contour plot feature of Mamba.jl.


tkelman:
BinDeps should be added to REQUIRE since it's used in deps/build.jl. You're probably getting it indirectly via a dependency of Gadfly, but best to be explicit about direct dependencies.

tkelman:
Also it looks like no recent git tags have been pushed to https://github.com/goedman/Stan.jl/releases - it can be convenient to cross reference release tags from there, you may have them locally but need to do `git push --tags`

goedman:
Thanks for your help and suggestions Tony!

PR #4469, opened 2016-01-26
Packages to add

montyvesselinov:
Please add Mads, Anasol, BIGUQ, R3Function, and MPTools


kmsquire:
Hi @montyvesselinov, these probably work fine locally (can you confirm this), but a few parts of Julia proper and the script we use to test pull requests here have some github-specific parts, and don't deal well (yet) with repositories located in other places.

You have a few options:

1. make versions of your packages available on github, and register those instead.
2. check out [the test script](https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/METADATA.jl) and [`Pkg`-related files](https://github.com/JuliaLang/julia/tree/master/base/pkg), and see what needs to be done to allow non-github repositories
3. open an issue here and in the main julia repository about this

Also, since packages share a global namespace, we generally ask that package names be chosen that either clearly indicate the purpose of the package (e.g., `MySQL.jl`, `LispSyntax.jl`, etc.), or are "clever" or "creative" in some way (`Mamba.jl`, `Gadfly.jl`, etc.).

Without looking at the actual repositories, I would guess that `Anasol.jl` fits into the second class, and the others are opaque to me.

If you haven't yet, can you check out the [package naming guidelines](http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package), and let us know a little more about your packages?

Also consider submitting fewer packages simultaneously, as a problem with one of them will hold up the whole PR.

Thanks!

tkelman:
Changing to a url rather than a git-protocol-specific address would likely work without changing the tests. We have had one past package not hosted on github but in terms of interacting with issues and pull requests from the rest of the community it's much lower friction to just use github.

All packages need a minimum julia version requirement.

kmsquire:
Thanks for the clarification, Tony.

montyvesselinov:
We have the codes on github as well. I will make the changes of URL.
On Tue, Jan 26, 2016 at 11:29 AM Kevin Squire <notifications@github.com>
wrote:

> Thanks for the clarification, Tony.
>
> —
> <https://github.com/JuliaLang/METADATA.jl/pull/4469#issuecomment-175161482>
> .
>
-- 
-monty

montyvesselinov:
Resubmitted.

-monty

tkelman:
closing in favor of https://github.com/JuliaLang/METADATA.jl/pull/4487

note BigUQ and MPTools still need minimum julia version requirements

montyvesselinov:
Both modules have files REQUIRE with julia 0.4 listed in.

-monty

tkelman:
The tag needs to be updated accordingly for BigUQ.

montyvesselinov:
BIGUQ tag is fixed. sorry about this.

-monty

tkelman:
It's okay, you can push additional commits to the same branch and they will show up in the same PR.

PR #4474, opened 2016-01-27
Register PositiveFactorizations

timholy:
This package returns a positive-definite "spiritually-approximate" factorization to any input (symmetric) matrix. The target application is Newton's method, for which one can guarantee descent if the Hessian (or modified Hessian) is positive definite. As a significant bonus, performance is competitive with `cholfact` (although this currently works only for dense matrices).

See https://github.com/timholy/PositiveFactorizations.jl for more info.


kmsquire:
LGTM.

tkelman:
This is good, but general preference is to leave new package registration PR's open for at least 24 hours comment period.

PR #4475, opened 2016-01-27
Tag Parameters v0.3.0

mauro3:
Added support for `@assert` in type-body


tkelman:
if there's a travis only 0.3 failure, it might just be more reliable to drop 0.3 support in require

tkelman:
Since I was able to reproduce the travis 0.3 failure locally, ket's raise the minimum version to 0.4 here?

mauro3:
Sounds good.

PR #4476, opened 2016-01-27
Pull request/516c2fe9

Azzaare:
Publication of NetworkFlows.jl: a package to compute max-flow/min-cut, multiroute flow, MLA-flow in pure Julia.


tkelman:
Have you compared this against LightGraphs? Seems like there's a lot of overlap.

Azzaare:
I compared this 6 months ago when I started. I guess that the max-flow algorithms where not implemented yet in LightGraphs. I was planning to merge the content of this package into LightGraphs eventually ...
In a perfect world, I would try to compare max-flow from both packages in speed, and add what is either improvement or new content into LightGraphs. Let's say I will try it at some point in the following months.

tkelman:
Cool. Barring any feedback on the naming I think we can merge tomorrow.

PR #4479, opened 2016-01-27
Tag LIBLINEAR v0.1.0

innerlee:
Major change:
* Replace LINEARModel with LinearModel which 
 * can be saved (as .jld, for example)
 * reduce memory usage significantly
* Support both 32 bit & 64 bit windows


tkelman:
Just curious, why are you using the ppa (and only nightly) instead of `language: julia` in [your .travis.yml](https://github.com/innerlee/LIBLINEAR.jl/blob/master/.travis.yml)?

innerlee:
@tkelman can't remember exactly. Either copied from https://github.com/simonster/LIBSVM.jl/blob/master/.travis.yml or generated by the template.
:-) 
what's the better way?

tkelman:
The part that you have commented out
```yml
language: julia
os:
  - linux
  - osx
julia:
  - release
  - nightly
```
should be enough for your purposes.

innerlee:
Thanks! I'll try it later.

PR #4480, opened 2016-01-27
Tag VirtualArrays v0.1.1 [git://github.com/invenia/VirtualArrays.jl.git]

samuel-massinon-invenia:
Updated [VirtualArray](https://github.com/invenia/VirtualArrays.jl).


tkelman:
Can be useful to push the git tags so they show up at https://github.com/invenia/VirtualArrays.jl/releases

PR #4487, opened 2016-01-29
Pull request/f75a5349

montyvesselinov:
Please register:

MADS : Model Analysis and Decision Support
BIGUQ : Bayesian Information Gap Decision Theory for Uncertainty Quantification, Experimental Design and Decision Analysis
Anasol: Analytical solutions for groundwater contaminant transport
MPTools: Tools to evaluate expressions
R3Function: Automated storage and retrieval of results for functions calls


Mads/versions/0.1.0/requires
```diff
@@ -0,0 +1,25 @@
+julia 0.4-
```
tkelman: note that the trailing dash refers to prereleases and is no longer necessary if all your users are on 0.4.0 or newer.

tkelman:
MPTools and R3Function are somewhat unclear names, what do the MP and R3 stand for? We try to discourage acronym package names that aren't self evident or well known.

montyvesselinov:
If we rename them "ReusableFunctions.jl" and "MetaProgTools.jl" is it going
to be ok?

-monty

tkelman:
I like those names, that would be a lot clearer IMO.

montyvesselinov:
I made a new pull request of a new version with suggested changes.

-monty

tkelman:
Yep! Merged #4490.

montyvesselinov:
Thank you very much!

-monty

PR #4488, opened 2016-01-29
Register Clipper: v0.0.1

maxvoxel8:


tkelman:
Great to see this registered! (and tell Dan I say hi)

Looks like Travis might not be enabled? Would it pass?

PR #4490, opened 2016-01-29
Pull request/13b2aeb4

montyvesselinov:
Another attempt. Now it should be ok. Thank you very much!


tkelman:
Great. Sorry for the hurdles. If you don't need to support prereleases it would be good to drop the trailing dash in the `julia 0.4-` lines that a few of these packages have, but that can be done for future tags.

PR #4491, opened 2016-01-29
Three packages: ParametricModels.jl, DynamicDiscreteModels.jl, HiddenMarkovModels.jl

BenConnault:
The README here has a fairly complete description: 
https://github.com/BenConnault/DynamicDiscreteModels.jl

Thanks,

Ben


tkelman:
Cool. If you turn on Travis CI would your tests pass there? It can also be useful to push the corresponding git tags so they show up in github under https://github.com/BenConnault/HiddenMarkovModels.jl/releases etc.

BenConnault:
I was wondering about the tags, do you recommend `git push --tags`? It seems like it is not a recommended git practice.

Tests should pass but I need to figure out how to set-up Travis as I did not use `Pkg.generate`. I will do it now.

Thanks.

tkelman:
It's not strictly used by Julia's package manager but helps when navigating the code online to be able to see directly in the same repository which sha corresponds to which version tag. Yes, `git push --tags` would do the trick if you've done `Pkg.tag` locally.

For Travis CI, the basic template is
```yml
language: julia
os:
  - osx
  - linux
julia:
  - release
  - nightly
notifications:
  email: false
```

and that should be enough to get you started. I'm happy to provide pointers if you hit trouble. Example.jl has a bit more detail, and there's a page in the Travis docs about `language: julia`.

BenConnault:
Yep, I generated a new package and copied pasted the .yml file from there. But I lost my push rights to my own repo... I think Julia changed my remote from https: to git: somewhere in my attempt at publishing? Is it going to do that each time I update Metadata?

BenConnault:
Ok, tags and travis done, to I need to refork METADATA to launch the builds?

tkelman:
Hm, the git vs https remote stuff is unfortunate, we're trying to make it better over time. Are you using Julia 0.4 or nightly? We have a `setprotocol!` function now but I think that's only on nightly.

What I personally do is add an https `upstream-push` remote and just use that remote when I'm working from the command line, if `Pkg.publish` is giving me trouble.

No need to change anything about the Metadata tags if the only thing you've modified since these is Travis-related.

BenConnault:
Ok, thanks. Yes, just added the travis file and tags. Ready to be merged :)

tkelman:
The usual practice is to leave new-package registrations open for ~24 hrs of initial feedback, but from my standpoint these look good to go. Will merge tomorrow if nothing else comes up.

BenConnault:
Perfect, thanks again for your help.

Ben

matthieugomez:
Maybe a more specific naming for `ParametricModels`?

BenConnault:
What risk of confusion do you have in mind? The scope of the package is what a Google search for ["parametric models"](https://www.google.com/search?q=parametric+models) turns out and Wikipedia's ["Parametric model"](https://en.wikipedia.org/wiki/Parametric_model) page is about. From the README:

> Defines a common interface for parametric statistical models. 
> [...]
> `ParametricModels.jl` defines a single abstract type `ParametricModel`. Parametric models promise to implement three functions:
> - `calibrate!(model,parameter)`: calibrates the model at a given parameter value.
> - `simulate(model,T)`: simulates a single time-series of T observations from a calibrated model.
> - `loglikelihood(model,data,parameter)`: computes the loglikelihood of a single time-series of observations at a given parameter value.
> 
> A default `mle()` method is provided by wrapping `loglikelihood()` with a generic numerical optimization routine.

nalimilan:
Indeed, it looks like the functions in ParametricModels.jl could be included into StatsBase.jl. See https://github.com/JuliaStats/Roadmap.jl/issues/4, https://github.com/JuliaStats/Roadmap.jl/issues/14 and other issues in the same project. You could open another issue to discuss that.

BenConnault:
Thanks for the pointers. [#14](https://github.com/JuliaStats/Roadmap.jl/issues/14) is orthogonal since it deals with regressions and ParametricModels.jl covers models which are fully specified up to a Euclidian parameter. [#4](https://github.com/JuliaStats/Roadmap.jl/issues/4) is very relevant but I am afraid I have not solved the numerous design problems raised there, and I very much agree with @lindahua's 04/01/2014 comment that the absence of multiple inheritance makes it really hard to come up with a unified type system for statistical models. 

My thinking was to try and grow a particular view of what parametric models could be in a separate, lightweight package before thinking of overhauling the stats type ecosystem. I am happy to revive [#4](https://github.com/JuliaStats/Roadmap.jl/issues/4) if people feel otherwise.

nalimilan:
I think you should file a new issue, and post a reference to it in the existing one: that would help centralizing the discussion, even if you keep it as a package for now.

BenConnault:
Sure. Now done in [#20](https://github.com/JuliaStats/Roadmap.jl/issues/20).

Evizero:
`ParametricModels.jl` does seem to be a bit too general of a name. For example the package might not lend itself to be used for parametric models (in the sense that the term is used in ML) without probabilistic interpretation, such as a primal linear support vector regressor.

I am not against the name, just playing devil's advocate here

BenConnault:
Yes, this sounds like a legitimate concern, thanks for the feedback. I could fall back on `ParametricStatisticalModels.jl`.

tkelman:
So should ParametricModels be renamed then we can go ahead?

BenConnault:
Yes, I am giving a little more time for feedback at [JuliaStats/Roadmap.jl/issues/20](https://github.com/JuliaStats/Roadmap.jl/issues/20) and I will make a new PR here. I will close this one when I do.

BenConnault:
Registering new versions instead

PR #4494, opened 2016-01-31
Tag GSL v0.3.0 [git://github.com/jiahao/GSL.jl.git]

jiahao:
A bunch of fixes to GSL wrapper functions, especially to the root finding functions (thanks @AndyGreenwell)

Also drops Julia 0.3 support


tkelman:
push tags to https://github.com/jiahao/GSL.jl/releases?

PR #4495, opened 2016-01-31
Register CQL Cassandra DB Binary Driver v3

dioptre:
Works with Datastax current version:

echo "deb http://debian.datastax.com/community stable main" | sudo tee -a /etc/apt/sources.list.d/cassandra.sources.list
curl -L http://debian.datastax.com/debian/repo_key | sudo apt-key add -
sudo apt-get update
sudo apt-get install dsc30

### DIFF
--- a/CQL/url
+++ b/CQL/url
@@ -1,0 +1,1 @@
+git://github.com/dioptre/CQL.jl.git


tkelman:
Included in #4496. You can push additional commits to the same branch as an open pull request and the new commits will show up automatically, no need to create new pull requests for every commit.

PR #4496, opened 2016-01-31
Pull request/fce1c698

dioptre:
Added a tag to CQL


dioptre:
Added injection removal methods

tkelman:
We try to discourage acronym package names, since a lot of people won't know what CQL stands for. Without a description on the repository there won't be any useful information to find out on the pkg.julialang.org listing either. Should this maybe spell out Cassandra in the name?

dioptre:
Hi @tkelman 

How do people download this from within Julia? 

I've been asked by some people for it and its not available as a Pkg?

tkelman:
Currently you can do `Pkg.clone("https://github.com/dioptre/CQL.jl")`, however see the package naming guidelines http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package

dioptre:
I read the guidelines and not sure what you are implying?

tkelman:
To quote https://github.com/JuliaLang/METADATA.jl/pull/4658#issuecomment-189062403, "because this is a shared namespace, we ask for some small concessions in naming packages. [...] These are guidelines, and not set in stone, but we generally push for people to follow them."

Points 1 and 4, avoid acronyms and err on the side of clarity. If you'd really strongly prefer to register the package under the CQL.jl name then we can go ahead and do that.

It would be good to try and turn on Travis CI so your tests will automatically run on every commit and pull request.

dioptre:
It'd be nice to have CQL as Cassandra Query Language is globally known as CQL. I think it'd be weird if it wasn't called that! I'm not a big fan of TLAs either, but this one is pretty common! 

Please accept it!

Thanks
Andy

PR #4500, opened 2016-02-01
Register and tag Gunrock.jl

ranjanan:
[Gunrock](https://github.com/JuliaComputing/Gunrock.jl) is a package for graph algorithms on GPU.


tkelman:
Encouraging users to modify a version controlled file means Pkg.update will not work on this package.

ranjanan:
Oh okay. I asked the user to change `path_to_gunrock` in `consts.jl`. Would you suggest I locate it automatically, in the form of an `init()` ?

tkelman:
Or use BinDeps, or make a separate file that sets the location but is in .gitignore.

ranjanan:
Alright, should I tag another release after this fix?

tkelman:
yes you can push an 0.0.2 tag to this same branch

PR #4501, opened 2016-02-01
Pull request/c1fd365d

damiendr:


tkelman:
The package naming guidelines generally discourage acronym package names.

damiendr:
Hm, it's more or less an interface to [ISPC](ispc.github.io), which is always referred to by this acronym.

tkelman:
Fair enough. Should this try to reuse any common infrastructure from ParallelAccelerator?

PR #4505, opened 2016-02-01
Register Voronoi: v0.0.1

maxvoxel8:
Note that unlike the VoronoiDelaunay package, this one supports both lines and points as sites.


tkelman:
Ah, whoops. Can you make an 0.0.2 tag that includes https://github.com/Voxel8/Voronoi.jl/commit/69b8daca34c2276ac86b64f37d3d4715ae78700a before too many people add this? Changing only the case of a file can confuse git on case-insensitive systems.

PR #4510, opened 2016-02-02
Two packages: DynamicDiscreteModels.jl, HiddenMarkovModels.jl

BenConnault:
Updated versions of #4491  following conversation over at [JuliaStats/Roadmap.jl/issues/20](https://github.com/JuliaStats/Roadmap.jl/issues/20)


BenConnault:
erf forgot to take ParametricModels out of REQUIRE...

Before I start again though, how do I make it so that Travis goes through even though HiddenMarkovModels.jl requires DynamicDiscreteModels.jl?

tkelman:
If they're both registered within the same PR, then the Travis check here will pass - as it did for #4491. Travis for the HiddenMarkovModels.jl package's repository won't pass until DynamicDiscreteModels gets registered, so if you want to be absolutely sure everything works right away as first tagged you could do them one at at a time.

BenConnault:
I have no idea what happened here. I tagged v0.0.3 to HiddenMarkovModels.jl and opened a new PR but it seems to have added to this one instead. I don't understand the new Travis fail at all.

tkelman:
Pushing additional commits to the same existing PR is actually preferred. Looks like you might need to remove the older tags that depend on the not-yet-registered package.

BenConnault:
Are you saying I need to rewrite my git history?

tkelman:
Not necessarily, you could just push a new commit to this branch that deletes the `requires` and `sha1` files for the earlier tags via `git rm`

BenConnault:
yep figured that out trying now

PR #4515, opened 2016-02-03
Bugfix for Parser Combinator (warning: manual!)

andrewcooke:
i've created this following the instructions at http://docs.julialang.org/en/release-0.4/manual/packages/#publishing-metadata-manually because I was seeing an error with Pkg.publish().  it's the first time i've done things this way so i may have an error (sorry).


tkelman:
Are you intending for 2 tags of the same sha?

andrewcooke:
only because i messed up and accidentally tagged twice.  but now i have another fix, so i will close this and tag and try again.

PR #4520, opened 2016-02-04
Register WCS

kbarbary:
The WCS package is meant to replace the WCSLIB package. Differences:

- Uses more recent version of the underlying C library.
- New, "Julian" API (rather than mirroring the inscrutable C API).
- Requires Julia 0.4 (for Tuples).
- Removes developer dependency on Clang, which was overkill for this library.

The name change is because (1) The API has completely changed and (2) the package is no longer a direct wrapper of the C library WCSLIB; the fact that it uses WCSLIB is an implementation detail; users need not be aware of it.

While short acronyms as package names are discouraged, the acronym "WCS" has precendent in "global namespaces": The Python equivalent is called `pywcs` and the debian package C shared library is `libwcs.so`. I wager that astronomers are likely to recognize "WCS" before something like "WorldCoordinateSystem".

If this is accepted, the WCSLIB package will be deprecated and eventually removed.


tkelman:
Is it still accomplishing the same thing? Could it be an in-place major version change? Only `SloanDigitalSkySurvey` depends on `WCSLIB` at the moment and we could put upper bounds in place to avoid breaking it. Astronomers may recognize the acronym, but would anyone else when browsing the package listing?

kbarbary:
Yes, it is just accomplishing the same thing, and yes it could just be a new version.

However, `WCS` isn't any worse than `WCSLIB` for outsider recognition, and plenty of other Julia packages wrapping C/Fortran libraries have acronym names without `Lib` or `LIB`. Honestly, my motivation is extremely petty: the capital `LIB` has always looked funny to me, and I'd rather not be stuck with it for years to come! :)

tkelman:
I can see the distaste for the all-caps name. wcslib has the advantage of being unambiguously googlable, wcs in isolation barely registers the right result within the first page.

kbarbary:
You're right that wcslib is much more readily googleable than wcs. I'm not sure why this is a concern though: people browsing pkg.julialang.org have the description line to tell them what WCS means. Going the other way, domain scientists searching for "WCS in Julia" can already find the package as the first result.

I think the only reason that Mike Nolta originally called this package WCSLIB instead of WCS is that at the time I was investigating wrapping an older implementation called libwcstools. I think it's now clear that this library is the authoritative implementation of WCS and there's no need for disambiguation from wcstools. The wrapped library name is `wcs` (as in `libwcs.so`); calling the wrapper `WCS.jl` is in line with other Julia packages.

tkelman:
Fair enough.

tkelman:
Oh, and it can be useful to push the git tags so they show up at https://github.com/JuliaAstro/WCS.jl/releases

kbarbary:
Thanks! tag pushed.

PR #4527, opened 2016-02-04
Register GenericSVD

simonbyrne:
https://github.com/simonbyrne/GenericSVD.jl


tkelman:
Travis should have failed here, this needs a tagged version.

simonbyrne:
I wasn't sure how it worked: the [docs](http://docs.julialang.org/en/release-0.4/manual/packages/) seem to imply that I should tag after being registered with METADATA.

tkelman:
We should fix the docs then, because registering without tagging isn't very useful and travis explicitly fails for that. Travis did not run here, don't know why.

tkelman:
Closing and reopening as an experiment to try to get Travis to run.

PR #4530, opened 2016-02-05
Update MultiPoly

daviddelaat:


MultiPoly/versions/0.0.2/requires
```diff
@@ -0,0 +1,3 @@
+julia v0.4
```
tkelman: the v is not necessary and not typically included here

tkelman:
since the previous tag supported julia 0.3 but this tag does not, it would be a good idea to use a new minor package version via `Pkg.tag("MultiPoly", :minor)`

daviddelaat:
`Pkg.publish()` gives an error for me. I'll try again later.

daviddelaat:
Thanks for reviewing. I couldn't get `Pkg.publish()` to work for this package (it does work for a different package), but hope the manual fixes are okay.

tkelman:
Yep. Would Travis CI pass if you enabled it?

PR #4536, opened 2016-02-06
Tag ConnectSDK v0.1.6

Lanzafame:
Updated some deprecated functions.


tkelman:
Note that this tagged sha is broken on julia 0.3. See https://github.com/Lanzafame/ConnectSDK.jl/pull/1 for a fix.

tkelman:
Bump. @Lanzafame can you either make a new tag 0.1.7 that includes the fix so it still works on 0.3 as indicated by the `REQUIRE` file, or modify the sha that the 0.1.6 tag points to? Another option would be changing this tag to 0.2.0 of the package and modifying `REQUIRE` file to only support Julia 0.4 going forward, but given master appears to work on Julia 0.3 that doesn't seem necessary just yet.

stevengj:
Bump?

tkelman:
We've frozen metadata for package versions that support julia 0.3. If you'd like to make an 0.2.0 tag for julia 0.4, feel free to open a new PR.

PR #4537, opened 2016-02-06
Tag CUDNN v0.3.0

denizyuret:
This version is compatible with cuDNN3 (7.0.64), and cuDNN4 release candidate (4.0.4).


CUDNN/versions/0.3.0/requires
```diff
@@ -0,0 +1,3 @@
+Compat
+CUDArt
```
tkelman: do you still need to support prereleases?

denizyuret: What do you suggest?

tkelman: that was meant to be on the `julia 0.4-` line below

`julia 0.4-` means "prereleases of julia 0.4 or newer" which includes the first commit after release-0.3 branched off. If you only need to support release versions of julia, `julia 0.4` means "releases of julia 0.4 or newer"

denizyuret: We don't necessarily "need to" but the package does support the prerelease
versions, so I think it is fine the way it is.  I will replace with julia
0.4 next version.

PR #4544, opened 2016-02-08
Pull request/55c5facf

lruthotto:
Added Package ParSpMatVec


tkelman:
Given https://github.com/lruthotto/ParSpMatVec.jl/commit/3498a5d7650ca6bc53825be29b62ed5f2c662c46, does this actually support julia 0.3?

How does this compare to https://github.com/madeleineudell/ParallelSparseMatMul.jl?

lruthotto:
In theory yes, but I have no time to figure out the correct call to A_mul_B! we use as a fallback. For now I removed 0.3 and support 0.4 + How do I change this here?

tkelman:
either make a new tag and leave out these initial tags from metadata, or modify the lower bounds on them in metadata so they are not installable on 0.3

PR #4555, opened 2016-02-08
Adding ParSpMatVec with tagged version 0.0.1.

lruthotto:
I hope this is what you meant, @tkelman


tkelman:
Yes. You can push additional commits to the same branch as an open PR and it will update, rather than opening a new PR for every modification.

tkelman:
You need to list BinDeps in REQUIRE, not manually add it on travis.

lruthotto:
Thanks, Tony. Added BinDeps as requirement now. It's quite elegant.

BTW: Is there a similar way to add a Fortran compiler as well? Right now, this seems to be an issue when using Travis on Linux OS's.

tkelman:
Yep: https://github.com/JuliaOpt/Ipopt.jl/blob/cc7aee5256bbafb16c74e80b9b03cd29ee12088a/.travis.yml#L11-L13

tkelman:
Travis is being flaky so I'm going to close and reopen this just to force it to run, pardon the noise.

lruthotto:
Well, I don't think I could complain about the noise here. thanks for the hint and your patience. 

seems like travis testing now works on linux, too.

tkelman:
No problem. I am still curious how this compares to @madeleineudell's similar package.

timholy:
I wondered about that too. https://github.com/madeleineudell/ParallelSparseMatMul.jl

lruthotto:
I didn't do a lot of comparison to ParallelSparseMatMul to be honest. My goal here was to have something that can replace julia's default routines without changing our codes too much. For example, the methods here don't need the user to distribute the matrix or even add processors. This is purely a convenience argument. However, the code here is faster than the one in Base (even with nproc =1 on my machine).

lruthotto:
removed mumps. hope that's it!

tkelman:
No, that new commit is completely deleting the package, that's not right either. Can you rebase out the mumps-related commits?

lruthotto:
No, I don't know how to do that. Where can I learn that?

lruthotto:
OK, I think I rebased it now. Will this PR change automatically?

tkelman:
Ah sorry, had to give a lecture. Yeah this looks right now.

PR #4558, opened 2016-02-08
Tag MbedTLS v0.2.1

malmaud:


MbedTLS/versions/0.2.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4-
```
tkelman: any need to still support prereleases?

MbedTLS/versions/0.2.1/requires
```diff
@@ -0,0 +1,4 @@
...
+BinDeps
+Compat
```
tkelman: should indicate the specific required version

malmaud:
@tkelman should be g2g now

PR #4562, opened 2016-02-09
Fix sha1 for most recent tag of Redis.jl

tkelman:
cc @jkaye2012, not sure where the existing sha came from?


jkaye2012:
Hmm, this is really weird. I have no idea. Perhaps I copied one from somewhere else by mistake? Either way, thanks for taking care of this.

PR #4563, opened 2016-02-09
Tag AWS libs and deps

samoconnor:


SymDict/versions/0.0.2/sha1
```diff
@@ -0,0 +1 @@
+9d7e77294eb1e7ff259ced442c598fde39adab6b
```
tkelman: adding a badge to the readme really does not call for a new tag

PR #4564, opened 2016-02-09
Register Robotlib

baggepinnen:


tkelman:
This needs a tagged version.

tkelman:
The repository name also needs to end in .jl. It's preferred to not export overly short and generic function names from the package as it can lead to conflicts.

tkelman:
#4565

PR #4574, opened 2016-02-10
Pull request/9f8cad3a

lruthotto:


tkelman:
This commit should not be tagged directly since it is not on any branch: https://github.com/JuliaSparse/MUMPS.jl/commit/c736354ff6fa98d3dbd90ca27e8fa8cf2925b956

lruthotto:
I'm not sure I got this right: are you just talking about the requirement for 0.0.1 or also the tag for 0.0.2?

tkelman:
The tag. It looks like that is on your fork. The registered URL for this package is under JuliaSparse.

lruthotto:
OK, this one is a complicated GIT exercise for me. The JuliaSparse/MUMPS master branch is now up to date and tested. 

I would like to tag this version. I tried to just clone it locally and run Pkg.tag("MUMPS"). Unfortunately I get errors when pushing into this PR. Am I on the right track?

tkelman:
What was the error? You could do `git checkout -b newmumpstag` in your local metadata, push that to your fork and compare that to what's here. It should have the sha as https://github.com/JuliaSparse/MUMPS.jl/commit/169381d2e41d45ada4018e7f9c965fe2731a1aeb

tkelman:
Okay, is this good now? Looks okay to merge to me.

It can be useful to do `git push --tags` to the MUMPS.jl repo so the tags show up at https://github.com/JuliaSparse/MUMPS.jl/releases. Not strictly required, but convenient for cross referencing.

lruthotto:
thanks for all these hints. the tag is now pushed. I think you can merge this one.

PR #4581, opened 2016-02-11
Tag SloanDigitalSkySurvey v0.1.0

kbarbary:
Refactored WCS-related tools to use new [WCS.jl](https://github.com/JuliaAstro/WCS.jl) package.


tkelman:
Since the previous tag supported julia 0.3 but this tag does not, it should use a new minor version 0.1.0

kbarbary:
Thanks Tony; I never know whether to increment minor or patch versions for pre-1.0 packages. Now changed.

PR #4582, opened 2016-02-11
Tag Diversity v0.2.6

richardreeve:
Package no longer works in Julia v0.5 due to language updates. Fix for readstring replacing readall, other minor changes.


tkelman:
It's best to add an explicit minimum version requirement on the first tagged version of Compat.jl that introduced the features you need to use.

PR #4590, opened 2016-02-12
Register RobustPmap: v0.1.0, v0.1.1

montyvesselinov:
RobustPmap: Robust calls of pmap checking for the type of returned values


tkelman:
Would the tests pass if you enabled travis ci? This is a very tiny package right now, is there a plan to expand on its functionality instead of putting these two functions in a slightly larger utility package?

tkelman:
included in #4642

PR #4591, opened 2016-02-12
Tag Nemo v0.4.1

wbhart:


Nemo/versions/0.4.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4-
```
tkelman: the trailing `-` refers to pre-0.4.0 prereleases, do you still need to support those?

tkelman:
Do you know why your tests are failing on Travis CI? The PPA method is also not recommended, `language: julia` is the preferred method now and would allow you to also test against osx.

wbhart:
We don't need the prerelease versions any more I guess.

I have not set up Travis for Nemo. I've only had time to set it up for some
of the other projects I manage.

tkelman:
It's running, just not in a way that succeeds: https://travis-ci.org/wbhart/Nemo.jl/builds/108891604

wbhart:
Well I've never set it up. It looks to me like it is trying to test Nemo
without building it.

PR #4595, opened 2016-02-13
Register OpenDSSDirect: v0.1.0

tshort:
This is a package to interface with "OpenDSS", a power distribution system simulator.


tkelman:
Cool! Would the tests pass on appveyor?

Ref https://github.com/tshort/OpenDSSDirect.jl/blob/7bd6a2ab7cf7172e4b5fcf14ca4215a75f6cfd01/src/core.jl#L6, note that it's generally better to use `dirname(@__FILE__)` than `Pkg.dir` so the package can still work even if installed outside of the package directory.

tshort:
Thanks for the input @tkelman. I'll give appveyor a shot. I don't see why tests wouldn't work.

tshort:
Okay, I fixed the `dirname` issue that @tkelman identified and got appveyor working. That was helpful as it identified a couple of bugs.

Hopefully, ready to go...

tkelman:
Yep, looks great. Don't forget to push the git tag for 0.1.1 to https://github.com/tshort/OpenDSSDirect.jl/releases

tshort:
Thanks @tkelman, I learned something else (git tags don't get pushed automatically).

tkelman:
`Pkg.publish` is supposed to do that for you, when it's working properly anyway

PR #4596, opened 2016-02-14
Tag Elly v0.0.5

tanmaykm:
- Julia v0.5 compatibility
- Locks to allow multiple tasks communicating over same node connection


tkelman:
Just curious, what's in 0.4.3 that you need that wasn't in 0.4.0?

tanmaykm:
Only because this is not being tested with 0.4.0-2 any more.

ViralBShah:
Would the network performance issue affect Elly.jl? If so, that is a good reason to require 0.4.3.

tanmaykm:
Yes, good point. It would.

PR #4600, opened 2016-02-15
Tag Thrift v0.0.5

tanmaykm:
- new transports: `TMemoryTransport` and `TFileTransport`
- `Type` is now a reserved symbol
- utility method `enumstr`


tkelman:
Is it possible to actually run the tests on travis? All you're doing is cloning and building the package, but dince there isn't a deps/build.jl that doesn't do anything.

tanmaykm:
Yes. Travis tests were no longer working after updating to Thrift 0.9.2. See: https://github.com/tanmaykm/Thrift.jl/pull/4

I have not checked if it is possible now. Will check and do it as a separate PR.

tkelman:
OK, makes sense. How hard is getting the right thrift version via BinDeps?

tanmaykm:
Yes, it will be good to have. Will add it soon.

Also the thrift compiler is needed only for generating Julia code from IDL. We should be able to test with some pre-generated code on travis at the least.

PR #4605, opened 2016-02-16
Tag GR v0.9.16

jheinen:
Started implementing MATLAB-like functions:

- plot
- histogram
- title
- xlabel
- ylabel
- legend
- xlim
- ylim
- savefig

... to be continued :-)


jheinen:
Thanks a lot !

tkelman:
Note that it's preferable to use `dirname(@__FILE__)` rather than `Pkg.dir("GR")` to allow for the case of installing the package outside the package directory.

tbreloff:
Tony: what is best practice if you want to access a file in another
package? (I know your first response will be "don't do that"... but lets
assume it's justifiable)  Presumably the "using" and "import" logic know
where to look... can we tap into that?

tkelman:
I don't think we have a good standard solution to that yet. Some people put `using` or `import` in a try-catch, others look at `Pkg.dir` and just assume if that other package might be installed outside the package directory it's too much work to try and find it. For now I mostly discourage `Pkg.dir` when a package refers to its own location, but for referring to other packages' locations it's alright for now if that's what works.

tbreloff:
Ok thanks.  I'm changing some of my own Pkg.dir calls now... I knew that
wasn't preferable but I couldn't remember what to replace it with!

tbreloff:
FYI... one of my Pkg.dir is inside an "@eval" block, which fails with:

ERROR: MethodError: `dirname` has no method matching dirname(::Void)


So I guess that's another caveat to keep in mind.

PR #4608, opened 2016-02-17
Register HTSLIB [https://github.com/OpenGene/HTSLIB.jl.git]

zhmz90:
HTSLIB.jl is a julia wrapper for [htslib](https://github.com/samtools/htslib)


tkelman:
Needs a tagged version.

tkelman:
Use `dirname(@__FILE__)` rather than `Pkg.dir` here https://github.com/OpenGene/HTSLIB.jl/blob/064d8bc5da06fa96e178841e934bac9903153a3e/src/HTSLIB.jl#L3

Best not to clone master of a dependency, since upstream changes can cause a build failure. It's also not recommended to assume command-line git will always be present, as it will no longer be a dependency of Julia in 0.5. Best to download a fixed version of the dependency's tarball.

zhmz90:
@tkelman Thanks for your advice. I have modified my code.

tkelman:
Can you add a sentence to the description below the package title about what htslib is for? Otherwise looks good.

zhmz90:
HTSLIB.jl is a julia wrapper of [htslib](https://github.com/samtools/htslib) for accessing common high-throughput sequencing data file formats such as bam/sam files.  I have added this sentence to [the package title](https://github.com/OpenGene/HTSLIB.jl).

PR #4612, opened 2016-02-17
Tag InplaceOps v0.0.5

simonbyrne:
cc @KristofferC


KristofferC:
Looks like the `REQUIRE` file also needs updating. Should probably be `julia 0.4`.

KristofferC:
:tada:

simonbyrne:
Blegh, that was messy. We need a better way to handle `Pkg` errors.

tkelman:
Given that this is the first 0.4-only tag it should probably use a new minor package version.

PR #4613, opened 2016-02-17
Pull request/f60d35b8

baggepinnen:
Travis build works for osx but segfaults when initializing Plots backend for linux. Tests pass on my local machine on linux with same Plots backend


baggepinnen:
Second time it tried a build, it succeeded on both osx and linux

PR #4638, opened 2016-02-22
Tag Contour v0.0.9

tlycken:


tkelman:
The previous tag supports 0.3 and this does not, so it should probably be 0.1.0 of the package

tlycken:
@tkelman Hm, I hadn't noticed that - thanks for pointing it out. I'll re-tag this as 0.1.0, and then [this change](https://github.com/tlycken/Contour.jl/pull/32) will be 0.1.1.

PR #4639, opened 2016-02-23
Tag AWSLambda and deps

samoconnor:
AWSCore

 - add arn_region() function

AWSLambda

 - retry loop in create_lambda() to handle new role activation delay
 - parallel PUT lambda/code and lambda/configuraiton in update_lambda()
 - per-region role names for create_lambda_zip() and merge_lambda_zip()
 - don't use cd() in commondir() for @async-safeness
 - in create_jl_lambda() do update_lambda() instead of delete/create
   and clean up tmp S3 files
 - in create_jl_lambda_base() fix module precompilation

AWSSNS

 - make sns_subscribe_sqs() work.
 - support regions in sns_subscribe_lambda()
 - add sns_list_subscriptsion() and  sns_unsubscribe()


samoconnor:
thx tony

PR #4649, opened 2016-02-24
Tag SGP4 v0.1.0

crbinz:


tkelman:
Since the previous tag of the package supports Julia 0.3 but this tag does not, it would be better to use a new minor version and tag this as 0.1.0 of the package.

crbinz:
Makes sense to me... done.

tkelman:
don't forget to push the git tag so it shows under https://github.com/crbinz/SGP4.jl/releases

crbinz:
:+1: good call. Thanks for your help, both here and more generally with all of the METADATA.jl PRs.

PR #4653, opened 2016-02-25
Tag ImageQuilting v0.2.6 [git://github.com/juliohm/ImageQuilting.jl.git]

juliohm:


tkelman:
it would be good to push the git tags so they show up at https://github.com/juliohm/ImageQuilting.jl/releases and can be easily cross referenced

juliohm:
Thank you @tkelman, I may add the tags to GitHub at some point, thanks for pointing it out.

PR #4654, opened 2016-02-25
tagging GLVisualize next

SimonDanisch:
@sjkelly I guess I'm violating semver for GeometryTypes again?...


tkelman:
If ModernGL dropped support for 0.3 in https://github.com/JuliaGL/ModernGL.jl/commit/6d79bc38b608dc4bc906c5caffc8aec977f18afa you need to adjust the REQUIRE file and use a new minor version.

SimonDanisch:
I think I actually added back 0.3 support with that commit... but I don't even have 0.3 to test it, so whatever ;)

tkelman:
it's not like 0.3 is hard to get. does the 0.0.7 tag work correctly on 0.3?

SimonDanisch:
Yup, just tested it... Guess I need to revert this...

tkelman:
Not a huge loss to say future versions will be 0.4 only, and it will save you the hassle of having to test all future versions against 0.3.

PR #4656, opened 2016-02-25
Tag PatternDispatch v0.1.0

toivoh:
Should hopefully work with Julia 0.3. Going to try for 0.4 next.


tkelman:
would travis pass if you turned it on?

toivoh:
Not if it's running Julia 0.4, but the tests pass locally for me with Julia 0.3 so if I can get Travis to use that then it should.

But how do I actually turn on Travis for an old package like this? If I tag the next version with Julia 0.4 support soon, maybe it would be easier to turn on Travis by then?

tkelman:
Doesn't really matter how old the package is, just add the yml, go to travis-ci.org and enable it in the settings, see if it works. I'll fire across a PR with the template yml to start with.

toivoh:
Thanks!

PR #4660, opened 2016-02-26
Regster AWSES, Tag AWSLambda and deps

samoconnor:
AWSSES

 - New per request: https://github.com/samoconnor/AWSCore.jl/issues/2

AWSCore

 - Added sign_aws3() and mime_multipart() to support SES.
    (mime_multipart() is also used by AWSEC2.jl)

AWSEC2

 - Move mime_multipart() to AWSCore.jl

AWSLambda

 - Added lambda_update_configuration()

 - In python wrapper, ignore sns publish errors (seem to happen at creation
    time due to role activation delay)

 - add aws[:lambda_force_update] option to upload new lambdas regardless of
    content hash

InfoZIP

 - use Cmd( , dir = ...) to make async safe (remove use of cd())


tkelman:
New package registrations should generally be separated into different PR's from tagging existing packages.

samoconnor:
Hi @tkelman,

That sounds generally reasonable.

In this case there the changes to AWSEC2 and AWSCore are interrelated with the new AWSSES submodule, so it seems to make sense for them to all be in a single atomic commit.

PR #4661, opened 2016-02-26
Pull request/05dc307e

charlesll:
Pull request to publish Spectra:

Package name SpectraJu was changed to Spectra, package was tagged, commit for adding gfortran for Ipopt was accepted,  and everything was updated in regards of the new name.

### DIFF
--- a/SpectraJu/url
+++ b/SpectraJu/url
@@ -1,0 +1,1 @@
+git://github.com/charlesll/SpectraJu.jl.git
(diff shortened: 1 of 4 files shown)


SpectraJu/url
```diff
@@ -0,0 +1 @@
+git://github.com/charlesll/SpectraJu.jl.git
```
tkelman: this file with the old name needs to be deleted

PR #4663, opened 2016-02-26
Tag JPLEphemeris v0.3.0

helgee:


JPLEphemeris/versions/0.3.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4-
```
tkelman: the trailing dash refers to prereleases of 0.4, do you still need to support those?

helgee:
@tkelman Good catch. How do I fix this? Do I need to open a new PR?

tkelman:
No, you can just delete the trailing `-` in place at https://github.com/helgee/METADATA.jl/blob/pull-request/c30e0d65/JPLEphemeris/versions/0.3.0/requires, just be sure to make the same change in the package's `REQUIRE` file too before the next tag

helgee:
Done. Thanks!

PR #4666, opened 2016-02-26
Pull request/c6f38f82

montyvesselinov:
Updates. Switching to Lora 0.5


tkelman:
Would be cool to see Travis CI running on these packages at some point if their tests can be made to pass there.

montyvesselinov:
Not much time these days. But we are planning to do it. Can you point to a
good example ? Thank you very much!
On Fri, Feb 26, 2016 at 1:11 PM Tony Kelman <notifications@github.com>
wrote:

> Would be cool to see Travis CI running on these packages at some point if
> their tests can be made to pass there.
>
> —
> <https://github.com/JuliaLang/METADATA.jl/pull/4666#issuecomment-189462139>
> .
>
-- 
-monty

tkelman:
Great. You can see the basic template .travis.yml file in Example.jl, it may need some additional snippets to deal with system dependencies depending what is required. If you start trying it out and hit problems, feel free to ask for pointers.

montyvesselinov:
I see MADS fails the test according to
http://pkg.julialang.org/detail/Mads.html

However, I had no problems to run it at travis.

https://travis-ci.org/madsjulia/Mads.jl

I guess the problem is that we YAML from python:

using Conda
ENV["PYTHON"]=""
Conda.add("yaml")

How to add these for the test? I do not want to add this in
test/runtests.jl.


-monty

tkelman:
Is that easy to check for? If you create a deps/build.jl file that checks whether the yaml package is present, uses it if so, or adds it from conda if not. Overriding the PYTHON env var setting is probably not good though.

PR #4679, opened 2016-02-29
Pull request/1fb03976

charlesll:
Registration of Spectra.jl

Created a new branch to avoid any trouble with the old pull request.

### DIFF
--- a/Spectra/url
+++ b/Spectra/url
@@ -1,0 +1,1 @@
+git://github.com/charlesll/Spectra.jl.git
--- a/Spectra/versions/0.0.1/requires
+++ b/Spectra/versions/0.0.1/requires
@@ -2,6 +2,4 @@ julia 0.4
 PyPlot
 JuMP
 Ipopt
-Gadfly
-Cairo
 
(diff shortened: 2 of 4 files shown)


tkelman:
Depending on both pyplot and gadfly seems a bit unusual.

tkelman:
What is the license status of https://github.com/charlesll/Spectra.jl/blob/master/Dependencies/gcvspl.f?

charlesll:
gcvspl is available on the netlib website: http://www.netlib.org/gcv/gcvspl. There is nothing in the headerlines regarding a licence, so I guess it can be considered as public domain?

I will make a commit to remove dependency to gadfly, as my examples run with pyplot.

charlesll:
I found details on the gcvspl.f licence, and updated the headerlines of the gcvspl.f file accordingly.

tkelman:
Great! Spectra's a pretty generic name, but unless someone has strong objections to it I think we can register this in a few hours.

charlesll:
Great, thanks!

tkelman:
(p.s. would be a good idea to add some tests to https://github.com/charlesll/Spectra.jl/blob/master/test/runtests.jl over time so you can check that the package continues working correctly)

charlesll:
Yes, I will add them very soon. I am currently adding further functionality and examples, so I will create the tests accordingly!

PR #4682, opened 2016-02-29
Tag Mads v0.1.5

montyvesselinov:
Build changes. The code now builds successfully on Travis-CI. Can you initiate a test build at http://pkg.julialang.org/


tkelman:
Forcing the Conda installation in deps/build.jl is probably not ideal. @Luthaf is there a better way of falling back to system versions of python packages if they work?

montyvesselinov:
On Mon, Feb 29, 2016 at 7:54 AM, Tony Kelman <notifications@github.com>
wrote:

> python packages


he has many repos. Can you point me to the exact code lines? Thank you very
much!

-monty

Luthaf:
[Here](https://github.com/madsjulia/Mads.jl/blob/4680ca7b42d1cdb8ab8e702dc30a2f52f0f22d48/deps/build.jl), you force the installation of the whole Conda stuff on every users computer, just to get the yaml lib. And some peoples are complaining that Conda is already taking too much space on there computers.

> @Luthaf is there a better way of falling back to system versions of python packages if they work?

The best way I can think of is to let PyCall manage the Python version, and use it to install yaml as needed. In order to support old python version, this means using [get-pip](https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py) to fetch the lib from PyPi.

I put together a small snippet that could be used in your case (https://gist.github.com/Luthaf/368a23981c8ec095c3eb), and that might be useful for other people too.

tkelman:
Thanks for the input @Luthaf! Closing this in favor of #4688.

PR #4685, opened 2016-02-29
register SignalView

SimonDanisch:
Neat tool for visualising and interacting with Reactives signal graph.
It's based on GLVisualize, so no tests/coverage yet :(

### DIFF
--- a/SignalView/url
+++ b/SignalView/url
@@ -1,0 +1,1 @@
+git://github.com/SimonDanisch/SignalView.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
Should this have Reactive in the name if it's pretty specific to Reactive's signals?

SimonDanisch:
I did this first, but I decided that this name is more meaningful. In
theory, it should also easily work for other signal graphs ;)

PR #4687, opened 2016-03-01
Register AWSSDB (and Tag related AWS* packages)

samoconnor:


tkelman:
If you are `using SymDict`, you should `REQUIRE` it explicitly rather than assuming it will be present as a transitive dependency.

samoconnor:
ok, done : https://github.com/samoconnor/AWSSDB.jl/commit/0998d1d1032bb5ffa866138834c966fcef042b3d

PR #4689, opened 2016-03-01
Pull request/77f4aae2

baggepinnen:


tkelman:
Have the authors of the original Matlab code made their source available directly under the BSD license through some means other than on Mathworks File Exchange? Clause 2.a.iii at http://www.mathworks.com/matlabcentral/termsofuse.html?s_tid=gf_com_trm#content generally makes me hesitate and advise people not to look at or use any code from there, and instead seek out licensing and source from the authors directly.

baggepinnen:
The authors provide the following licence file, which is replicated under LICENCE.md


> Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
> 
> * Redistributions of source code must retain the above copyright
>   notice, this list of conditions and the following disclaimer.
> * Redistributions in binary form must reproduce the above copyright
>   notice, this list of conditions and the following disclaimer in
>   the documentation and/or other materials provided with the distribution
> * Neither the name of the University of Washington nor the names
>   of its contributors may be used to endorse or promote products derived
>   from this software without specific prior written permission.

baggepinnen:
It is also made available from their website
http://www.cs.washington.edu/people/postdocs/tassa/code

baggepinnen:
Just to be sure, I have now sent the original authors asking for explicit consent in publishing this version.

tkelman:
From their website you're fine.

PR #4697, opened 2016-03-01
Register Make: v0.1.0

tshort:
This package is a make-like system for data analysis. It's geared more towards data analysis than Makefile-style building. To support code and data dependencies, the package includes a way to save some state between sessions. 

https://github.com/tshort/Make.jl


tkelman:
I don't think this should be called Make if it isn't designed to read standard makefiles... maybe TaskRunner or something to that effect?

tshort:
I still like "Make". It's concise, and most people have a general idea of what "make" does. That it's concise makes it easier to use fully qualified methods. This is important because the package uses several rather commonly named methods: `file`, `task`, and `variable`.

I made the change from "Jake.jl" (@nolta's original name), so the purpose was more apparent.  

TaskRunner is a decent name, but it's wordy.

Any other opinions?

tshort:
Here are my preferences for name. I still haven't found something I like better than Make.

1. Make
2. Do
3. Run
4. Jake
5. VT
6. TX
7. TaskRunner

tkelman:
2, 3, 5, 6, and arguably 1 aren't great w.r.t. the published naming conventions. Maybe something derived from Make but not exactly the same, Maker.jl or something to that effect. If it's called Make, I'd really expect it to read and use conventional makefiles as is.

tbreloff:
I agree with @tkelman here.  If I was looking for something to run conventional makefiles I would be disappointed.  If I was looking for "makefiles for data science", I wouldn't think to even consider this package.  I would consider something more along the lines of DataDeps.jl, VarDeps.jl, etc.  `Make` sort of implies that you're building an executable (which I don't think you are).

Naming aside, I think the concept behind the package is great!

StefanKarpinski:
+1 to not calling it `Make` – that's definitely a name I would expect to be a Julia interface to `make`. I like the existence of such a package quite a lot. Doing exactly this kind of thing was a big part of what motivated me to start Julia – I was using Rakefiles for this at the time and generating R and Matlab code, and lots of CSV files.

tshort:
So, what happened to @johnmyleswhite's suggestion not to bikeshed package names? Well, I'm getting piled on, so my options seem to be:

(1.) **Use methods without module names.** 

```julia
using TaskRunner
file("in1.csv")
file("in2.csv")

file("df.csv", ["in1.csv", "in2.csv"]) do 
    ... 
end

variable("df", "df.csv") do 
    ...
end
task("default", "df.csv")
```

(2.) **Capitalize task names.** This may help them stand out more. It breaks the rule of lowercase method names, but given that these are declarative statements, maybe its okay. 

```julia
using TaskRunner
File("in1.csv")
File("in2.csv")

File("df.csv", ["in1.csv", "in2.csv"]) do 
    ... 
end

Variable("df", "df.csv") do 
    ...
end
Task("default", "df.csv")
```

(3.) **Capitalize task names prepended with something.** This may help avoid conflicts with types and shows that they are tied together. 

```julia
using TaskRunner
MFile("in1.csv")
MFile("in2.csv")

MFile("df.csv", ["in1.csv", "in2.csv"]) do 
    ... 
end

MVariable("df", "df.csv") do 
    ...
end
MTask("default", "df.csv")
```
(4.) **Use a `Make` module inside the package module.** 

```julia
using TaskRunner
Make.file("in1.csv")
Make.file("in2.csv")

Make.file("df.csv", ["in1.csv", "in2.csv"]) do 
    ... 
end

Make.variable("df", "df.csv") do 
    ...
end
Make.task("default", "df.csv")
```
(5.) **Leave the package unregistered.** 

I'm leaning toward 3 or 5, but I need to play with these more.

johnmyleswhite:
For reference, my suggestion was pretty overwhelmingly voted down by the other committers.

StefanKarpinski:
If you strongly want to use the `Make` name, that's ok and you can merge this PR, I was just registering my opinion that I think it's a confusing name. It's easier to change the name of something before people are using it than after they already are.

StefanKarpinski:
Here's an idea that's somewhat motivated by the naming issue but also motivated by the possibility of having multiple separate build DAGs in the same process. Does it make sense to instantiate a `Maker` object? Something like this:

```jl
using Maker

make = Maker.new()

make.file("foo.csv") do
    # generate foo.csv
end

# etc.
```

The annoying thing is that you want the object type to be `Maker` too :-\

Anyway, @tshort, please don't take this discussion the wrong way. Like I said, if you want to go with `Make`, go for it.

tshort:
That's an interesting idea. It's nice because it preserves what I was shooting for in terms of API. The API matters more to me than the package name. What I don't see with that approach is how one could add a new type of task beyond those originally defined: `make.newtask("a") do...`.

MikeInnes:
Just throwing in another option:
```julia
using TaskRunner
Make = TaskRunner
Make.file(...)
```
(It might be nice to have a shorthand for this, e.g. `using TaskRunner as Make`.)

tshort:
Another good option. I could also put `Make = TaskRunner` in the TaskRunner module and export it to have that alias predefined.

tshort:
I'll probably go with "Maker":

https://github.com/tshort/Maker.jl

One extra letter is fine. I do see the possible confusion if I use the name "Make". Also, I thought that the name "TaskRunner" might imply a package that deals with `Base.Tasks`.

StefanKarpinski:
Sounds good. Sorry for the hassle.

tshort:
No worries. Good input.

PR #4700, opened 2016-03-02
Register EllipticFEM.jl

gerhardtulzer:
This package provides a finite element method to solve elliptic partial differential equations.

One important fact is that the assembly of the stiffness matrix is faster than the similar code in MATLAB.

For further information please see the README.

### DIFF
--- a/EllipticFEM/url
+++ b/EllipticFEM/url
@@ -1,0 +1,1 @@
+git://github.com/gerhardtulzer/EllipticFEM.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
Would your tests pass if you enabled Travis CI? It's also useful to push the git tags to the repository so they show up under https://github.com/gerhardtulzer/EllipticFEM.jl/releases.

gerhardtulzer:
Yes, the tests pass.

I have pushed the tags to the repository.

PR #4703, opened 2016-03-03
Register CDDLib and Polyhedra

blegat:


mlubin:
Polyhedra.jl has no readme, would be helpful to have a short description of what's in the package any why it might be useful.

blegat:
You are right. I have just added a basic README. It is still not enough but I will add more later (along with proper documentation)

mlubin:
LGTM. Some might complain about the superfluous commits for CDD that could be squeezed into one "Register CDDLib" commit.

tkelman:
I wouldn't think a mathematical base Polyhedra manipulation package should depend on opengl visualizations?

mlubin:
@tkelman is that really a concern blocking registration of the package? I don't think this is the right venue to criticize how @blegat has chosen to organize his abstractions.

nalimilan:
Looking at README, this looks like a potentially optional dependency, but at the moment this isn't really possible in Julia. Anyway, that can be changed later easily. Only the name is (more or less) set in stone, so if that's OK I'd say merge this.

nalimilan:
(Though squashing into a single commit wouldn't hurt.)

blegat:
`Polyhedra.jl` indeed can works perfectly without this visualization feature. I agree that it may be better to add a separate package `PolyhedraVisualization.jl` depending on `Polyhedra.jl` and `GLVisualize.jl` adding the visualization feature.

blegat:
I have squashed into one commit for CDDLib and one for Polyhedra :)

tkelman:
Not intending to block registration, but it does get in the way of the package passing its tests easily on Travis, for example. Better raised as an issue though, that's fair.

CDDLib could use a more descriptive one-liner description field, and the git tags should be pushed so they show up under https://github.com/blegat/CDDLib.jl/releases.

PR #4704, opened 2016-03-03
Pull request/94518074

latticetower:


tkelman:
Hi! Note that the only registered Julia package that begins with a lower-case letter was registered over 3 years ago, before we established naming conventions. The package also needs to have a license file.

PR #4705, opened 2016-03-04
Tag ParserCombinator v1.7.8

andrewcooke:
having to do this manually.

various small fixes.


ParserCombinator/versions/1.7.8/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.3
+Compat 0.7.6
```
tkelman: `readstring` requires a more recent version of Compat than this

andrewcooke: thanks

PR #4709, opened 2016-03-04
Pull request/93e7db11

ehsantn:


tkelman:
Great to see something like this released! The HPAT acronym name is somewhat against the package naming guidelines, but I guess it's not confusable with much else here.

ehsantn:
Thanks. I'm aware of the guidelines but unfortunately, the full name (which describes the functionality well) is too long for a package. The conflicts seems minor to me.

PR #4710, opened 2016-03-04
Tag Blink v0.3.3

spencerlyon2:
Adds support for transpiling julia expressions containing conditionals and comparisons to javascript


tkelman:
Does this need to run `AtomShell.install()` in the yml files to pass on CI?

spencerlyon2:
Ahh yes, good call. We should add that. I'll update soon.

PR #4718, opened 2016-03-05
Register Celeste: v0.0.1, v0.0.2

jeff-regier:


tkelman:
what version of julia does celeste 0.0.1 require?

jeff-regier:
Julia 0.4, exactly (It won't work with 0.3 or 0.5). Thanks.

tkelman:
can you add `julia 0.4` to https://github.com/jeff-regier/METADATA.jl/blob/pull-request/b0875cf6/Celeste/versions/0.0.1/requires then?

jeff-regier:
I see. I think instead I'll just remove those tags, and open a new PR.

PR #4719, opened 2016-03-06
Tag updates to PGFPlots and TikzPictures

mykelk:
This adds support for optional omission of latex preamble to make embedding in existing documents easier.


tkelman:
PGFPlots' Travis tests are failing, does it require a specific minimum version of TikzPictures?

mykelk:
Yes, it requires the version I just tagged.

tkelman:
That should probably be indicated in the require file then - can be done separately in metadata, doesn't require re-tagging the package

mykelk:
I'm a little confused about what you are suggesting. For this new feature, both packages require an update. Should I change the require for pgfplots so that it specifies this new tagged version of tikzpictures so that Travis can be satisfied, then retag pgfplots, and then Pkg.publish()? Thanks for the help!

tkelman:
1. Add the required version number as a lower bound in https://github.com/mykelk/METADATA.jl/blob/pull-request/58808b2e/PGFPlots/versions/1.3.3/requires
2. After this is merged, make the corresponding change in https://github.com/sisl/PGFPlots.jl/blob/master/REQUIRE

mykelk:
Thanks! Got it. Just completed step one and am waiting for the merge before going on to your step 2.

PR #4721, opened 2016-03-07
Tag LightGraphs v0.5.2

jpfairbanks:


jpfairbanks:
I did this manually without `Pkg.tag` and `Pkg.Publish`. I followed the instructions in http://docs.julialang.org/en/release-0.4/manual/packages/#publishing-metadata-manually. So if this works, Go Docs!

tkelman:
Looks right.

PR #4728, opened 2016-03-08
Tag Boltzmann v0.4.0

dfdx:
Major refactoring with minor breaking changes.


tkelman:
Note that using the PPA in `.travis.yml` is not recommended any more, see https://docs.travis-ci.com/user/languages/julia for the more recent `language: julia` method.

dfdx:
Thanks for the tip, I will make a change in master.

PR #4733, opened 2016-03-09
Tag LibHealpix v0.1.0

mweastwood:
* Use autotools to build the Healpix dependency
* Install cfitsio from apt or homebrew if possible
* Wrap the iterative `map2alm` function for more accurate spherical harmonic transforms
* Fix an assortment of minor bugs

Link: https://github.com/mweastwood/LibHealpix.jl
There is also a brief description of LibHealpix here: #3405


tkelman:
Probably not a good idea to be running global homebrew commands in deps.jl - https://github.com/mweastwood/LibHealpix.jl/commit/4af074cd8cb9bac6ede9b9e1bd5624709edca738. Did Homebrew.jl not work?

mweastwood:
Ah that commit was because Healpix wasn't able to find cfitsio if it was installed by Homebrew.jl. It was easier for me to just install it globally than to try and wrangle the default Healpix build system (which is an absolute mess) into finding the libraries.

It turns out there are actually two ways to build Healpix and the second way using autotools is much nicer. So I could give Homebrew.jl another shot.

Is `apt-get install ...` similarly frowned on?

tkelman:
If you'd be able to find a binary source for the package that doesn't need sudo and works lots of places, then that would be preferred over relying on apt-get. Maybe Conda.jl would let you get some of these, but that needs to find a Python installation (or installs its own private one) at the moment even to download binary-only packages.

mweastwood:
Eh, I might just have to build cfitsio myself then. That seems to be what the python wrapper does. I'll be back with something better.

PR #4743, opened 2016-03-10
Register JukitLearn.jl

cstjean:
I would have liked to call it ScikitLearn.jl, but I asked for permission from the scikit-learn developers and [didn't get it](https://sourceforge.net/p/scikit-learn/mailman/scikit-learn-general/thread/CAEJt5f7XOArmC1vTdUkAM_t1qqV%2BF80UiLoyi2vYRSdjMW%2BOPA%40mail.gmail.com/#msg34907568). I'm not super fond of "JukitLearn" (though I'd like to preserve some kind of name association). Suggestions are welcome.

### DIFF
--- a/JukitLearn/url
+++ b/JukitLearn/url
@@ -1,0 +1,1 @@
+git://github.com/cstjean/JukitLearn.jl.git
--- a/JukitLearnBase/url
+++ b/JukitLearnBase/url
@@ -1,0 +1,1 @@
+git://github.com/cstjean/JukitLearnBase.jl.git


nalimilan:
Sounds like a bit silly not to call it ScikitLearn.jl. I would stress to the upstream developers that this is the convention used by all library wrappers. If they want to get credit for their work, they'd better accept you make it clear in the name of your package that it's not a new pure-Julia implementation. Julia packages all have their GitHub project to file issues, so they shouldn't be too annoyed by Julia users.

cstjean:
I did [point it out](https://sourceforge.net/p/scikit-learn/mailman/scikit-learn-general/thread/CAEJt5f41Ct1uOwYE3aGcF-CV_O-C6n6FVm_Oaov%3DYN07A8xp9Q%40mail.gmail.com/#msg34914612), source-forge just split up the thread. 

> Julia packages all have their GitHub project to file issues, so they shouldn't be too annoyed by Julia users.

That's a good point. Can we make a list of packages following that convention? I've got Pandas.jl, Stan.jl, MATLAB.jl and Bokeh.jl so far. If there's enough of them, then it should be clear that Julia users know about the convention.

cstjean:
There's a lot more of these than I expected.

AWS.jl, AWSEC2.jl, Accumulo.jl, AmplNLWriter.jl, AppleAccelerate.jl, ArcadeLearningEnvironment.jl, Arduino.jl, Atom.jl, Augur.jl

I posted it on their mailing list, and will update here when I get an answer.

PR #4744, opened 2016-03-10
Register AstroLib

giordano:
I started writing a bundle of astronomical functions, based on [IDL Astronomy User's Library](http://idlastro.gsfc.nasa.gov/homepage.html).


mlubin:
You should push a tag to the PR as well.

tkelman:
What is the license status of the code this was derived from?

giordano:
The license of IDL AstroLib is BSD-2-Clause, see http://idlastro.gsfc.nasa.gov/idlfaq.html#A14

giordano:
For the record, Python AstroLib [claims](http://www.hs.uni-hamburg.de/DE/Ins/Per/Czesla/PyA/PyA/pyaslDoc/pyasl.html#license-note) that original IDL AstroLib is in the public domain, but I don't know where they drawn this conclusion from.

For the time being, the few functions present in the AstroLib.jl share only the name with IDL AstroLib's procedures, not the algorithm, but in the future the same algorithm for some functions will be likely used.

PR #4745, opened 2016-03-10
Register and tag ConicNonlinearBridge

mlubin:
I'm open to bikeshedding on the name, but we'd like to have this package registered because it's needed by [Pajarito](https://github.com/mlubin/Pajarito.jl) which we're preparing to release.
This functionality can't go directly into the MathProgBase package because JuMP is a dependency. 
CC @madeleineudell @emreyamangil


mlubin:
Thanks!

PR #4746, opened 2016-03-10
Added Shoco.jl

ararslan:
I created a package called [Shoco](https://github.com/ararslan/Shoco.jl) which wraps the [shoco C library](http://ed-von-schleck.github.io/shoco) for string compression. Both my package and the C library are MIT-licensed.

`Pkg.publish()` wasn't working for me so I'm doing this the old-fashioned way. :smile:


tkelman:
The `@BinDeps.load_dependencies` is a somewhat older and not-very-commonly-used method, but looks good otherwise.

tkelman:
Oh, and you should probably be downloading a particular tag or sha rather than [master](https://github.com/ararslan/Shoco.jl/blob/84341371762bbd28a60d2fcaa8978408193f7914/deps/build.jl#L21) which could change at any time.

ararslan:
@tkelman Thanks for the feedback! I'll update the download. What do you suggest I use in place of `@BinDeps.load_dependencies`? I'm surprised it isn't commonly used since it's described in the README and seems pretty convenient. I'm fine with whatever though. ¯\\\_(ツ)\_/¯

tkelman:
If it works I think you can stick with it. The method that is more popular is to let BinDeps create a `deps.jl` file and load it as in https://github.com/JuliaOpt/SCS.jl/blob/b1aae45aa8dbb2c84bf073ef2b4c19a43bc95ccd/src/SCS.jl#L5-L9. That has the advantage that BinDeps isn't a runtime dependency, only a `Pkg.build`-time dependency, but with package precompilation that might not make as much difference as it used to.

ararslan:
@tkelman I made the changes you suggested and tagged a v0.0.2 GitHub release, but now I can't seem to get the version of Shoco in METADATA to update to 0.0.2 with `Pkg.update()`. Is there a step I'm missing?

PR #4749, opened 2016-03-11
Tag TimeSeries v0.7.4

GordStephen:


TimeSeries/versions/0.7.4/requires
```diff
@@ -0,0 +1 @@
+julia 0.4- 0.5-
```
tkelman: lower bound should be `0.4` now, unless you have a need to support prereleases of 0.4

milktrader: Fixed for future versions. How did this slip by me? Nice catch!

PR #4750, opened 2016-03-11
Register DiscriminantAnalysis: v0.0.1

trthatcher:
I removed the ``0.0.1`` and ``0.1.0`` tags and restarted version number... it points to the latest commit. This should be considered the first release.


nalimilan:
Looks like a nice package! Have you thought about adding a formula interface?

trthatcher:
Thanks!

I did want to provide a better interface, but I was hoping wait it out until there was more of a consensus on the API approach for statistical models (ex. JuliaStats/Roadmap.jl#14). Additionally, the package just requires Base Julia right now. I didn't want to make DataFrames a required package as I see this package as providing the model & solver layer (at least for now). However, I'm not sure how to approach conditional functionality contingent on DataFrames being installed.

nalimilan:
Indeed. Conditional modules will make this easier to handle (https://github.com/JuliaLang/julia/pull/6884). We could also move formulas out of data frames, so that models can be fitted from other data sources. It should likely go into StatsBase or an essential package like this.

Regarding the common API, I think it will only develop if package authors take the lead when they need it. Contributions welcome!

PR #4753, opened 2016-03-11
Added Control.jl to METADATA

mfalt:


tkelman:
Needs a minimum Julia version in `REQUIRE`. Also you didn't answer @jcrist's license question in https://github.com/JuliaControl/Control.jl/pull/10#issuecomment-191996147 - if any of the implementations were ported from Matlab or Octave, they are a derived work and need to be attributed as such.

mfalt:
I added `julia 0.4` to the `REQUIRE` file now. We did take care of the concerns that @jcrist voiced. I went over the code once again now and removed the only function which of the origin I thought was still doubtful (unwrap), which already was implemented twice. I updated the hash to reflect the new commit.

mfalt:
Is there anything more you think I should fix before this is pulled @tkelman?

tkelman:
Control is a pretty generic name, ControlSystems.jl might be a bit more specific. Any package that has a dependency (even indirect) on this would not be able to have an internal sub-module called Control.

mfalt:
I realize that this is up to you to decide @tkelman, but I would prefer Control.jl.
I see where you are coming from, and I could see the argument for e.g LinearControl.jl, ControlSynthesis.jl, ControlAnalysis.jl, FeedbackControl.jl or something similar but this package is not limited to any one of those areas. I seems to me that it is implicit that a Control.jl package is meant for Systems Control, (what else can you control?) and that the risk of someone naming their module Control without looking for a package like this is negligible.

tkelman:
> I seems to me that it is implicit that a Control.jl package is meant for Systems Control

Julia is a general purpose language used by people in wide-ranging fields who may have never heard of control theory, so I don't think this is the case.

> (what else can you control?)

[The control set in an experiment](https://en.wikipedia.org/wiki/Scientific_control)? [The flow of a program](http://docs.julialang.org/en/release-0.4/manual/control-flow/)? Or from a sampling on github code search:

- [GUIs](https://github.com/jverzani/JGUI.jl/blob/master/src/types.jl)
- [simulation parameters](https://github.com/Luthaf/Jumos.jl/blob/master/src/Simulations/MolecularDynamics.jl)
- [communication protocols](https://github.com/dfdx/Woody.jl/blob/master/src/slow/controller.jl)
- [quantum states](https://github.com/BBN-Q/QSimulator.jl/blob/master/src/base-types.jl)

Most of those are using `Control` as an abstract type or function name, but I think it gets the point across.

mfalt:
I see what you mean @tkelman, and I guess my statement was a bit provocative, of course there are other things to control, although I still believe that a "Control package" doesn't seem to be too broad. But I guess ultimately its not up to me do decide. However, it seems a bit restrictive to deny a name based on that it might have other uses than the one in the package, wouldn't the same apply to for example (only going from A-D):
Blocks, Brownian, Calculus, Cartesian, Clustering, Codecs, Color, Contour, Convex, Crypto, Currencies, DSP, Dates, Debug, Decimals, Digits, Distance, Distances and Distributions?
I am not contesting that ControlSystems is a slightly more descriptive name, and I wouldn't go into this trouble of discussing it if I didn't think Control was a good name and because I know that there are already people using the package for whom the name change would be a minor inconvenience.

tkelman:
It's not strictly my decision, I just spend more time reviewing metadata pr's that most other contributors.

Metadata is a shared namespacr and it's much easier to rename a package before it gets registered than after, so I'd like to get some opinions from others before deciding here.

mlubin:
:+1: for Control. If you create a solid, useful package like this one you should get the benefit of the doubt with naming.

pao:
Maybe idiosyncratic, but possibly "Controls.jl", with the "s"?

mfalt:
I think "Controls.jl" is a bit confusing, to me it sounds like we are providing a "Control" type, I would probably prefer ControlSystems.jl then.

pao:
ControlSystems sounds good, too, and probably fits best in the ecosystem.

ssfrr:
+1 for ControlSystems. When I first saw the PR for Control.jl I thought "remote control? home automation?". "ControlSystems" to me unambiguously means feedback networks, error estimation, differential equations, etc., i.e. what you learn about in engineering school.

For reference:
https://www.google.com/search?q=control
https://www.google.com/search?q="control+systems"

baggepinnen:
+1 for ControlSystems.jl. It's a well established name (except for the plurality) due to the Matlab toolbox which has been widely used in academia for a long time. I also think that ControlSystems is better than ControlSystem since many julia packages are named with the plural form of a noun.

tkelman:
closing in favor of #4785

PR #4755, opened 2016-03-11
Register and tag ScikitLearn.jl and ScikitLearnBase.jl

cstjean:
Thanks @nalimilan


tkelman:
ScikitLearn does not depend on SkikitLearnBase?

cstjean:
It does. Can I just manually edit `ScikitLearn/versions/0.0.1/requires` in "Files Changed", and update the REQUIRE in my repo?

tkelman:
Yes, you can edit https://github.com/cstjean/METADATA.jl/blob/pull-request/e6c5663c/ScikitLearn/versions/0.0.1/requires so it reflects in this PR and make the corresponding change in `REQUIRE` without re-tagging any new sha's.

tkelman:
Less critical since this can be fixed later, but where you have
```
include("Skcore.jl")
module ScikitLearn
import Skcore
```
should probably switch the order of the `include` and the `module`, otherwise you're effectively reserving both the `Skcore` and `ScikitLearn` name any time anyone imports your package.

And if you enabled Travis CI, would your tests pass there?

cstjean:
I made the edit.

> should probably switch the order of the include and the module

Noted, will do in the next version. I might get rid of Skcore altogether.

> And if you enabled Travis CI, would your tests pass there?

The tests pass locally. Can I install scikit-learn (the Python version) on Travis? How does PyPlot do it?

Thank you for the review.

tkelman:
Don't forget to also change https://github.com/cstjean/ScikitLearn.jl/blob/master/REQUIRE before the next tag.

Might be easiest to leverage Conda.jl in your `.travis.yml` to get scikit-learn installed. PyPlot sadly has no tests, but I think there are some other examples of Python-wrapper packages that you can look at. Having trouble finding a good one right this second though...

PR #4758, opened 2016-03-12
Register PyLCM

rdeits:
Registers PyLCM, which contains bindings to LCM, the Lightweight Communications and Marshalling library via PyCall. I chose the "Py" prefix because someone may eventually want to directly wrap the LCM C or C++ API without the Python overhead, so I'm leaving the LCM.jl name available.


tkelman:
It's not really recommended to use the PPA any more for Travis testing, it's not actively maintained.

PR #4761, opened 2016-03-12
Register VegaLite

fredo-dedup:
Provides bindings to the recently published Vega-Lite viz grammar


tkelman:
Is any of the code derived from the existing Vega.jl package? If so it should contain a license attribution.

Also good idea to push the git tag to https://github.com/fredo-dedup/VegaLite.jl/releases so it can be easily cross-referenced while browsing the code.

fredo-dedup:
Part of the code is indeed from the Vega.jl package (only the render.jl file). By license attribution, you mean adding `Copyright (c) 2016 John Myles White and Randy Zwitch.`  to the license file ?  Or is it more complicated ?

Should I also mention the 'University of Washington Interactive Data Lab.' (the copyright owners of Vega-Lite) ? They already have their own license file in the `asset/bower_components` directory, but I used their properties description in the package docs.

fredo-dedup:
Latest version has the license attribution to the Vega.jl authors + this version has ben tagged.

VegaLite/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+JSON
```
tkelman: you added Compat here in the more recent tag

PR #4763, opened 2016-03-13
Initial Atmosphere-Ocean repo branch

rickedanielson:


tkelman:
We generally try to discourage acronym package names, but I guess here it's pretty unambiguous. Also would be great to add some tests to https://github.com/JuliaAtmosOcean/ICOADSDict.jl/blob/master/test/runtests.jl going forward.

rickedanielson:
Agreed; the format is being updated.  Tests would ideally catch the most recent formats and fail otherwise (I'm guessing the data providers have some ideas; a few will follow progress here).

PR #4765, opened 2016-03-13
Tag ReverseDiffSparse v0.5.5

mlubin:
Fixes for some corner cases


tkelman:
should add a NaNMath version dependency for sqrt

mlubin:
updated

PR #4766, opened 2016-03-13
Register Rsvg

lobingera:


lobingera:
Travis fail is unclear, some github communication issue.

PR #4768, opened 2016-03-13
Register Swifter [https://github.com/wookay/Swifter.jl]

wookay:
This is a prototype to interact with iOS application that using Swift.
Although, Swift has a REPL system, but that's not useful than Julia REPL.
This is not to call Swift functions directly (like PyCall),
This is for Server side using Swift and Client side using Julia script/REPL.

### DIFF
--- a/Swifter/url
+++ b/Swifter/url
@@ -1,0 +1,1 @@
+git://github.com/wookay/Swifter.jl.git


wookay:
got errors much, close this pull request and create new one.

PR #4769, opened 2016-03-13
Register Swifter 

wookay:
This is a prototype to interact with iOS application that using Swift.
Although, Swift has a REPL system, but that's not useful than Julia REPL.
This is not to call Swift functions directly (like PyCall),
This is for Server side using Swift and Client side using Julia script/REPL.

https://github.com/wookay/Swifter.jl

### DIFF
--- a/Swifter/url
+++ b/Swifter/url
@@ -1,0 +1,1 @@
+git://github.com/wookay/Swifter.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
If you need to add HttpServer as a test-only dependency, the right place to do that is in a `test/REQUIRE` file, that way `Pkg.test` will work for everyone, not just on Travis.

wookay:
added test/REQUIRE and fixed travis, appveyor.
thanks

PR #4772, opened 2016-03-14
Register Tables: v0.0.1

andyferris:
Hi all,

I've been working on a package for strongly-typed data tables in Julia, which I would like to publish to METADATA.jl. The package is somewhat related to DataFrames; please see [here](https://github.com/JuliaStats/DataFrames.jl/issues/744#issuecomment-195118485) for some feedback from the JuliaStats community (@tshort, @nalimilan).

Currently, the package is called "Tables.jl" since its primary purpose is to provide the `Table` class. According to the [Julia naming guidelines](http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package) :

> 3) Packages that provide most of their functionality in association with a new type should have pluralized names.
> * DataFrames provides the DataFrame type.
> * ...

On the other hand, the community may feel the name is too broad. Both "DataTables.jl" and "TypedTables.jl" are open possibilities.

Andy


Tables/url
```diff
@@ -0,0 +1 @@
+ssh://git@github.com/FugroRoames/Tables.jl.git
```
andyferris: Is https customary here, or ssh is OK?

tkelman: https please

Tables/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BaseTestNext
```
tkelman: this is likely a test-only dependency so it should go in `test/REQUIRE`

quinnj:
I have a `Data.Table` type in the `DataStreams` package (which is more an interface than implementation), so I'd appreciate not going with `DataTables.jl`. I think `Tables.jl` is fine, though I kind of like `TypedTables.jl` better since it's a little more distinct and google-able. Your choice.

andyferris:
Thanks @tkelman, I have updated the URL, REQUIRE, test/REQUIRE, requires and relevant commit hash. Passes checks now.

andyferris:
@quinnj So `DataTables` wouldn't *strictly* clash with `Data.Table`, right? But I do understand the potential confusion.

TypedTables.jl as a name will be clear in meaning right now. It does make a clear distinction to DataFrames.jl as it currently stands. But in the future (say Julia 0.6), we will all be thinking "Of course the data table is strongly typed! Why wouldn't it be?"

Tables.jl is also OK. Any other opinions if this is to broad?

nalimilan:
> I have a Data.Table type in the DataStreams package (which is more an interface than implementation), so I'd appreciate not going with DataTables.jl. I think Tables.jl is fine, though I kind of like TypedTables.jl better since it's a little more distinct and google-able. Your choice.

@quinnj If that's an interface, could it be renamed to `AbstractTable` and rely only on implementations providing a set of methods? It would also be more convenient if `DataFrame` and `TypedTable`/`DataTable` objects were actually `AbstractTable`, without needing an additional wrapper.

@andyferris Is right that we may find it obvious that tables are "typed" in the future, so the name `TypedTables` would be a bad choice. One advantage of `DataTables` is that [it's also the name of a very popular R package](https://github.com/Rdatatable/data.table/wiki) which is known to be much more efficient than data frames. So the story repeats again in Julia -- and former R users looking for performance (which are a potential user base for @andyferris's package) would find it easily.

(As I said on the other thread, `Tables.jl` sounds really too broad to me. I wanted to register my `FreqTables.jl` package under that name, and changed my mind. Let's not make this a race for the first to claim this name. :-)

andyferris:
OK, I've done some research. Here are some packages with "Tables" in the name:

[ThermodynamcsTable](https://github.com/DANA-Laboratory/ThermodynamicsTable.jl)  (@DANA-Laboratory)
[DataCubes](https://github.com/c-s/DataCubes.jl) (previously MultidiemsionalTables) (@c-s)
[LifeTable](https://github.com/klpn/LifeTable.jl) (@klpm)
[FreqTables](https://github.com/nalimilan/FreqTables.jl)  (@nalimilan)

Interesting to note: all use of the world "table" appears (to me) to refer to the same thing we all learn about in school with columns and rows, etc, so at least the word is unambiguous (e.g. a frequency table is a certain type of table, etc).

Obviously relevant package:

[DataFrames](https://github.com/JuliaStats/DataFrames.jl) (Contributers include @johnmyleswhite, @garborg, @simonster, @tshort, I'm sure there are many many more) 

Partially relevant package:

[NamedArrays](https://github.com/davidavdav/NamedArrays.jl/) (@davidavdav)

Sorry for spamming everyone but some of you might have some feelings on naming - feel free to ignore this or add your 2 cents. My preferences in order are "Tables", "DataTables" and "TypedTables" but I bow to the community's opinion. "Tables" is most succinct, using everyday language that shouldn't confuse a high school student, while "DataTables" is the most complete description of *what* it is (and sounds similar to DataFrames), and "TypedTables" describes merely *how* it is implemented.

andyferris:
+1 for AbstractTable with a reasonably well-defined interface (at least for access). But it would have to go in Base, no?

johnmyleswhite:
I'm fine with anything, but think DataTables is slightly better than Tables, which is a little vague.

Glad to see you working on this problem. Hoping to dig in and understand your approach at some point.

andyferris:
Cool. I'll switch to DataTables with the understanding that I'm not (at this stage at least) attempting to adhere to the traditions of R's `data.table`. 

For instance, I'm not an R user at all, and I just had a quick look at the interfaces for `data.frame`, `data.table`. and `dplyr`, and the syntax for `data.table` doesn't appear very Julian. The idea around the "verb" syntax of `dplyr` (is "dplyr" meant to be  "data plier"??) looks much easier to use and implement, and the original R syntax could be reasonably faithfully reproduced with some more macros. Of course, hopefully we can squeeze some of the speed from `data.table` somehow...

PS - `data.table` says it's functions are implemented in C... maybe someday someone will write Julia wrappers for those C functions, since they are fast and well maintained... but then what would that package be called? RDataTables? A similar overlap in naming also applies to @quinnj's `DataStreams.Data.Table`.

quinnj:
I'd really prefer anything but `DataTables.jl` to avoid confusion with `DataStreams.Data.Table`. A couple other ideas that come to mind:

* DataGrids.jl
* DataCubes.jl
* DataMesh.jl
* DataSquare.jl
* RowTables.jl
* DataSets.jl

or honestly, going with something a little more *unique*, since I feel like this fits into the "unique interface of functionality" rather than a "foundational" package (i.e. more similar to Mocha.jl as opposed to SQLite.jl).

andyferris:
Hmm... interesting. We have seen some solid arguments *against* all 3 names that I originally was proposing.

I think the package needs to be findable and the name needs to be descriptive, so I'd rather have "Table" somewhere in the name. I don't really want to call it "MyTable" or "FerrisTable" or something even sillier. "RowTables" was an idea I had when I first started programming it with a different storage format, but since the package will have both "wide" and "long" formatted tables it doesn't seem generic enough (and "RowAndColumnTable" is getting a bit like hands-and-shoulders-and-knees-and-toes!). 

@nalimilan clearly feels "Tables" is too broad and I can understand the sentiment (though in his case, a frequency table is just a special type of table, in two senses - it could be a 2-column table listing value and frequency, or it could be a specific way of storing a generic data table where the "value" is a row-tuple and the frequency the number of matching rows, like a PooledDataVector of rows. Conversely to the latter, my package is also just a special way of storing a data table where each row is stored explicitly, including repeats). (Furthermore, the package currently plans to include other table types, like wide tables `DenseTable`, tables with key values `KeyTable` and `DenseKeyTable`, and so-on, and a frequency-table/pooled-data representation could be included in the future...)

I guess the "abstract" version of a table is a relation, but not everyone who needs some kind of table-like-thing to store data in will know that terminology - though RelationalTables.jl would work fine. Thoughts?

andyferris:
Or we can go full circle and do "TypedTables"

c42f:
+1 for `TypedTables`, since for my usage it's the type stability which is the main attraction of this package.  There's no problem with packages which solve a pressing problem in the current julia, but which may be unnecessary as the language matures.

andyferris:
Right, we can call it `TypedTables` for now and then we have the following possibilities for the future:

1. If the package falls out of use (e.g. `DataFrames` subsumes all of its functionality), it can be deprecated.
2. If the package remains useful in the future, it can be renamed appropriately, depending on what makes it useful.
3. If (by some fluke) it becomes very popular, we can revisit calling it `Tables`.

Everyone OK with `TypedTables`?

johnmyleswhite:
+1 to TypedTables

quinnj:
+1

andyferris:
Package renamed and checks passed. Everyone seems happy with the name.

Can we merge, please?

andyferris:
Unfortunately, I forgot to rename one file (and the test mechanism doesn't actually run Pkg.test()??) so I'll have to do a 0.0.2 now...

StefanKarpinski:
If it's not too late, I think given that this never got out there in general usage, you can just overwrite 0.0.1.

PR #4775, opened 2016-03-14
Register Pages: v0.0.1

EricForgy:


EricForgy:
Haha. I am so excited to submit my first package I keep refreshing this page hoping to see a merge :smile: 

:pray:

aviks:
In general, for new package, we now tend to keep it open for a day or two to see if anyone has comments on the package name. Given that package names are in a global shared namespace, we want to err on the side of caution here

EricForgy:
Good know. Thanks @aviks. I'm happy to answer any questions to the best of my ability.

By the way, I have a number of packages I'm working on and many depend on Pages.jl, e.g. Azure AD, Power BI, OAuth 2.0, Microsoft Graph, etc., so I am testing the waters with my first submission here and hope to contribute more shortly :muscle:

tkelman:
Pages is a rather generic name. It could use a description field and some tests (then enable Travis CI), as well.

EricForgy:
Thanks @tkelman 

I added the description field and enabled both Travis and Appveyor. I haven't added any tests other than the default yet, but the default Travis tests are passing.

image

However, the Appveyor tests are not passing for nightlies, but the error is occurring when building HttpServer and I don't think is related to Pages.

image

I agree the name "Pages" is generic, but I like it. Simple. Clean. If it is a problem, I can change it, but prefer not if at all possible.

What do you think?

Cheers

PS: Just realized my commit comment has a typo: callbacks["notify"] should be callbacks["connected"].

tbreloff:
> I agree the name "Pages" is generic, but I like it. Simple. Clean.

I like the simplicity too, but it's also not very descriptive of what's inside, and this is a pretty simple package to use such a generic name.   SimpleServers.jl?  EasyRPC.jl? PagePrototypes.jl?

Also an aside... it kinda rubs me the wrong way that it starts a server after `using Pages`... you might consider requiring something like `using Pages; serve()`.

Just my $0.02... thanks for the contribution.

EricForgy:
Thanks for the feedback @tbreloff . I see your point about the server and will change that :+1:

I will go ahead and change the name too.

This is my first PR to METADATA and I'm still not an expert with GitHub. How do I change the name? Should I just close this PR and submit a new one with a new name?

On the naming, if I change the name to something like SuperSimpleWebPages.jl and then define my module, is it ok to immediate just re-export another module called Pages? For example,

~~~julia
module SuperSimpleWebPages

module Pages

# My stuff

export myexport1, myexport2, myexport2

end

using SuperSimpleWebPages.Pages

export Pages, myexport1, myexport2, myexport3

end
~~~

This way the name of the package doesn't really matter and it wouldn't change the way I use it, e.g. I still do things like

~~~julia
julia> Pages.broadcast("say","Hello")
~~~

Would that work?

StefanKarpinski:
FWIW, I think the name `Pages` is fine and we kind of decided to stop bikeshedding package names so much.

aviks:
Pages is a good name. While it IS generic, it does not appropriate any other field of study.

EricForgy:
Thanks guys. Being on HK time makes communication difficult :)

When I get into the office, I'll look into the server thing, but I thought Blink.jl also silently launches a server after "using Blink". Maybe I am confused. Any other thoughts on that?

As for the name, any thought on my question above? If I do change the name, I'll probably just immediately export module Pages. Would I need to close this PR and submit a new one with the new name?

Thanks :)

EricForgy:
PS: Since Pages is inspired by Blink, in addition to my acknowledgement in the README, should I add Blink to the LICENSE? I'm new to all this so thanks for your patience :sweat_smile:

tkelman:
If you literally copy pasted any code, yes. If you just looked at it for inspiration but your own work is not derived from it, then you're probably okay.

EricForgy:
Thanks @tkelman. When you mention that, I remember I did copy/paste a few lines from the example in WebSockets, so added this to the LICENSE (in latest master)

hayd:
IMO `using` triggering the start is a little strange/devious.

```
julia> using Pages
Listening on 0.0.0.0:8000...
```
I wonder if `Pages.start()` or similar would be better/more explicit etc. Thoughts on that bit?

EricForgy:
Hi @hayd ,

Thanks for the feedback. @tbreloff shared a similar concern [above](https://github.com/JuliaLang/METADATA.jl/pull/4775#issuecomment-196842667).

I'm looking into it now. In particular, I am looking at Blink.jl to see how its done there.

Appreciate the thoughts :+1:

EricForgy:
Hi everyone,

I fixed it so the server doesn't start after "using Pages" and updated the README accordingly:

~~~julia
julia> using Pages

julia> Pages.start();
Listening on 0.0.0.0:8000...
~~~

Is it ok now?

tkelman:
Should probably make a new tag to reflect the latest changes. Also CI status badges on the readme can be convenient.

EricForgy:
Thanks @tkelman . I added the build status and a new tag :+1: :)

PR #4777, opened 2016-03-14
Register SpikingNetwork

YaoLuCNS:
This is a julia package for modeling biological neural network with detailed document.


tbreloff:
Hi @YaoLuCNS thanks for the contribution!  I just wanted to jump in real quick and make a few comments:

- It might be a good idea to change the name... it should end in ".jl" by convention, and it might be good to pluralize in case you'd like to use the singular as a type:  "SpikingNetworks.jl"
- I'm excited to have a working simulator for Izhikevich's models... looking forward to checking this out later.
- For plotting, I recommend looking at my package [Plots.jl](https://github.com/tbreloff/Plots.jl) and the companion package [MLPlots.jl](https://github.com/JuliaML/MLPlots.jl) which has recipes for spiking networks (https://github.com/JuliaML/MLPlots.jl/blob/master/src/OnlineAI/onlineai.jl#L73).  I can help with visualizations if you want.  There shouldn't be a need to call into matlab.

tkelman:
closing in favor of #4783

PR #4784, opened 2016-03-15
Tag NoveltyColors v0.1.0

randyzwitch:


tkelman:
don't forget to push the git tags so they show under https://github.com/randyzwitch/NoveltyColors.jl/releases

randyzwitch:
Thanks. This is one of those things that seems to work differently on every computer I develop from. Sometimes the tags push, sometimes they don't.

tkelman:
`Pkg.publish` is finicky, esp. with ssh keys. I just manually do `git push --tags` myself.

PR #4785, opened 2016-03-15
Register ControlSystems: v0.1.1, v0.1.2

mfalt:
After the discussion in https://github.com/JuliaLang/METADATA.jl/pull/4753 I changed the name of the package from Control.jl to ControlSystems.jl, as suggested.
I messed some stuff up with the tags and decided it was easier to create a new PR than trying to manually rewrite and move the old files.


mfalt:
I saw that you closed the previous PR @tkelman, thanks. Is there anything else I need to fix before this is pulled?

tkelman:
Nope. A slightly longer description field would be good, and your `.travis.yml` is using a very old convention for testing [here](https://github.com/JuliaControl/ControlSystems.jl/blob/792524a13a5f9bf7afee2a1a1d8782f6c611adf2/.travis.yml#L7-L8), but those don't need to be fixed right away.

mfalt:
Thanks, I didn't know that the the travis.yml was old, I will look into updating that when I have time. Yes we should also update the description, that will probably be part of the work on the documentation that we are planning. Thanks for your help and work on julia!

PR #4793, opened 2016-03-16
Pull request/7558af64

dysonance:
I would like to add two packages I'm working on for quantitative finance applications in Julia. Please let me know if there are any problems. Thanks :)


tkelman:
Neat. How do these compare to the JuliaQuant packages? Stratagems could use a bit more fleshing out in its readme, and both could use some tests going forward.

tkelman:
oh, and Stratagems is empty in its current state - missing some includes https://github.com/dysonance/Strategems.jl/blob/master/src/Strategems.jl

dysonance:
Thanks for the quick response! The goals are relatively similar to the JuliaQuant work, but offer a different approach that might be more amenable to some users. For example, Indicators doesn't have any dependencies, whereas MarketTechnicals requires the TimeSeries class. The current trading strategy / backtesting framework requires Reactive, and I'm looking to make something more flexible & efficient. Basically I want to make it easy to rapidly build backtests, optimize parameters, and ultimately manage live execution.

dysonance:
I am also working on a new time series package/type, which I plan to try publishing soon as well once more features are rounded out. It is similar to TimeArray, but with stylistic differences. Indexing functionality is analogous to the Array base type, there isn't any metadata attribute (to make it more lightweight), field names aren't required in the constructor (they auto-populate to mimic Excel columns if unspecified), and methods/attributes are given more concise names (for faster prototyping/experimentation at the REPL).

dysonance:
At some point I would also like to add some visualization functionality as well, but I'm kind of waiting on a standard Julia plotting library.

tbreloff:
Hi @dysonance... nice to see more interest in this space.  However, I wonder if maybe you want to hold off on registering?  It seems like you've just started work, and it's not currently usable.  Is there a reason you're registering now?

As for visualization, you should listen to my totally unbiased input (I swear!) and check out Plots.jl.

dysonance:
Hey @tbreloff, yeah you make a good point. I'll add some more functionality over the next couple weeks and resubmit.

And I have definitely looked at Plots --- it's a totally awesome (and necessary) effort in the crowded Julia graphics space, where there's like 10 libraries to choose from that are all *okay*  . . . with the exception of PyPlot, which is pretty feature-rich but kind of cheating in a way ;)

I have to imagine at some point though the Julia community is going to have to decide once and for all on a single plotting library, if for no other reason than simplicity and less confusion for new users. This is definitely going to be a subject for some serious debate over the next few years.

PR #4794, opened 2016-03-16
Tag TypedTables 0.0.2 to fix name problem

andyferris:
Sorry, I forgot to rename the main package file and some other minor things when we renamed Tables -> TypedTables. Package actually works fine now. (I thought the METADATA tests would include `Pkg.test("TypedTables")` but apparently not!)


tkelman:
No, not at the moment. That's why I encourage package authors to always enable Travis CI if the package can pass its tests there.

PR #4795, opened 2016-03-16
Tag BiomolecularStructures v0.2.2

gp0:


tkelman:
BaseTestDeprecated is probably a test-only dependency, isn't it?

PR #4799, opened 2016-03-16
Tag Swifter v0.0.2

wookay:
* Getter, Setter implementations
* Query mode : By pressing the > key  changes the repl mode between Julia and Swifter.jl

thanks

https://github.com/wookay/Swifter.jl


tkelman:
I think you should acknowledge the Cxx code you borrowed from in your license. Repl mode keys may start to become a scarce commodity.

wookay:
Thanks, Added the acknowledge of Keno's Cxx.jl project to the LICENSE.
REPL mode key could change by setting the ENV
ENV["SWIFTER_QUERY_MODE_KEY"] = ">"
Thanks for the review.

PR #4801, opened 2016-03-17
Register Rsvg

lobingera:


lobingera:
will rebase first and review require again?

Rsvg/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4-
```
malmaud: You can drop the '-' now that .4 is out

lobingera:
I somehow know and I think it's committed, but the commits are tested sequentially

PR #4805, opened 2016-03-17
Register Rsvg

lobingera:


tkelman:
could you rebase and squash this into a single commit?

tkelman:
You should also make the dependency on BinDeps (and Homebrew on osx) explicit instead of assuming they will be present as transitive dependencies.

lobingera:
@tkelman : you mean adding in the REQUIRE of the package, or the require of metadata?

tkelman:
Both. It's fine in metadata for now, if you plan on tagging another version soon then changing it just in the package is fine since it'll get picked up in the next tag.

lobingera:
note to myself, learn more git...

PR #4807, opened 2016-03-18
Register VT100

Keno:
Going through my backlog of packages that are used in a number of places, but I never actually registered. First up, we have VT100, a small package for VT100 emulation, for those of us who like writing their own terminal emulators or want to regression test rich terminal applications.


tkelman:
Should have `FixedPointNumbers` explicitly in REQUIRE instead of assuming it will be present as a transitive dep

PR #4813, opened 2016-03-19
Register Rsvg and Tag

lobingera:
should now contain the necessary actions in single commit.

### DIFF
--- a/Rsvg/url
+++ b/Rsvg/url
@@ -1,0 +1,1 @@
+git://github.com/lobingera/Rsvg.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
would be good to push the git tag so it shows up at https://github.com/lobingera/Rsvg.jl/releases

lobingera:
@tkelman, as Pkg.publish() didn't work, i did some steps manually. You mean Rsvg.jl should contain a releases file in the top level directory? (and it contains what?)

tkelman:
No, you should have some tags listed when you do `git tag`, but they haven't been pushed. `git push --tags` will do so.

lobingera:
i did, but how do i see this on github?

tkelman:
https://github.com/lobingera/Rsvg.jl/releases is nonempty now, so looks right

PR #4819, opened 2016-03-20
Tag Benchmark v0.1.1

Rory-Finnegan:


johnmyleswhite:
Thanks for submitting this. Looks like needs a Julia version.

Rory-Finnegan:
I'm guessing that should be added to Benchmark.jl and then the METADATA version should be updated accordingly? Also, I just checked and it seems to work on a relatively recent version of 0.5, so are you okay with `julia 0.5-`?

tkelman:
if you only provide one number then it's a lower bound. should use the lowest version of julia for which tests are regularly being run (is travis on?)

tkelman:
if you only provide one number then it's a lower bound. should use the lowest version of julia for which tests are regularly being run (is travis on?)

Rory-Finnegan:
@tkelman No, travis is not set up for this package.

KristofferC:
This can be closed?

Rory-Finnegan:
Sure, I forgot this PR was still open and I'm pretty sure the pkg is abandoned now.

PR #4821, opened 2016-03-21
Tag JPLEphemeris v0.3.2

helgee:


tkelman:
Looks like you have a test failure on 32 bit?

helgee:
Indeed, but I cannot reproduce the error outside of AppVeyor. Tried it both with 32bit Julia on 64bit-machine and also within a 32bit VM.

PR #4825, opened 2016-03-21
Pull request/c5ab4935

sam81:
pull request to register SDT.jl a package for computing signal detection theory (SDT) measures


tkelman:
What is the minimum version of Julia that this is tested against? Would Travis CI pass if you turned it on?

It's up to you, but IMO a non acronym name would be easier to disambiguate, clearer for potential users who aren't familiar with the field to see what the package does, and more in line with Julia package naming conventions (http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package). There's some active discussion on loosening those guidelines but it's still a somewhat divided issue.

StefanKarpinski:
To be clear: if you feel strongly about using the name `SDT` then you can go ahead with it, the guidelines are just there to help pick names that will be a bit less cryptic.

sam81:
thanks for the help. The minimum version of Julia against which it has been tested is 0.4. I'll add that to REQUIRE.

Regarding the name. I don't feel strongly about naming it `SDT`, `SignalDetectionTheory` is a bit long to type, but is the best alternative I can think of.

I'm not sure how to amend the MEATADATA pull request after I change REQUIRE and rename the package. If it's easier I'll just close this pull request and open a new one after everything is fixed.

tkelman:
You can modify the files in the `pull-request/c5ab4935` branch of your METADATA fork, or just make a new pull request, either works.

`SignalDetection` would be a bit shorter and seems to unambiguously mean the same thing? Unless it is a common acronym to always see it including "Theory" within the field.

sam81:
thanks, I'll try modifying my METADATA fork and push the changes. I'll open a new pull request if I get in trouble with that.

Most of the time it is named in full as "Signal Detection Theory", but there is at least one major book that refers to it simply as "Detection Theory", so `DetectionTheory` could be an alternative.

sam81:
closing this, I'll open a new PR after the module has been renamed

PR #4840, opened 2016-03-23
Tag DataFrames v0.6.11

tshort:


tkelman:
Please re-tag this as 0.7.0, since it's the first version to drop support for Julia 0.3.

tshort:
I know how to add a tag, but are you suggesting I should remove the 0.6.11 tag? If so, how?

tkelman:
We can leave the 0.6.11 tag as is, but if the need arises in the future to tag a new bugfix release that still support 0.3 off of a branch of DataFrames, we should leave room for it to be 0.6.12. The versions wouldn't be monotonic but that's not too big a deal.

tkelman:
Please do `Pkg.tag("DataFrames", :minor)`. It's better for someone with commit access to do it, but otherwise I'll make the change here.

nalimilan:
@tshort Tagged 0.7.0, but should we also publish it?

tkelman:
Yes, I meant metadata tag more than git tag.

nalimilan:
OK. See https://github.com/JuliaLang/METADATA.jl/pull/4855.

PR #4846, opened 2016-03-23
Tag Lora v0.5.1

scidom:
Ported accept-reject sampler (ARS), random adaptive Metropolis (RAM), HMC and SMMALA from older Lora and added support for forward and reverse mode automatic differentiation of targets.


Lora/versions/0.5.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.4-
```
tkelman: drop the trailing `-`

scidom: Thanks @tkelman, I dropped it.

tkelman: here too https://github.com/scidom/METADATA.jl/blob/pull-request/f7d14f72/Lora/versions/0.5.1/requires

scidom:
I think I changed it in metadata too now @tkelman.

PR #4848, opened 2016-03-24
Register SDWBA

ElOceanografo:
Package for modeling acoustic backscatter from zooplankton using the stochastic distorted-wave Born approximation.


StefanKarpinski:
I don't have any issue with this acronym (it's likely to be pretty unique), but I sooo want there to be a package called Zooplankton.

tkelman:
And a version should be tagged once you're absolutely set on the name, otherwise Pkg.add will not work yet.

ElOceanografo:
I considered "ZooplanktonScattering," but that would imply the package implements a bunch of other physics-based models for sound scattering I don't have time to implement right now.  Maybe at some point that package will appear and eat this one.

StefanKarpinski:
Ok, carry on :-)

ElOceanografo:
Got my versions out of sync trying to tag and re-publish.  Just opened a new PR with the tagged version, so closing this one.

PR #4852, opened 2016-03-24
Tag PlotlyJS v0.1.2

spencerlyon2:
Adds the `attr` function and fixes a display error in IJulia notebooks caused by require.js


tkelman:
Some of your tests are failing.

spencerlyon2:
Last I checked it was because the Web stack failed to install.

Everything works for me locally.

// Spencer

spencerlyon2:
Ahh that's why the Linux ones failed, but there is something else for osx. Will resolveand update. Good catch.

spencerlyon2:
Ok I have amended the tests and tags have been updated.

PR #4860, opened 2016-03-25
Tag ConjugatePriors v0.2.0

andreasnoack:
As mentioned in https://github.com/JuliaStats/ConjugatePriors.jl/pull/9#issuecomment-201294444

@tkelman Is this sufficient?


tkelman:
D'oh - thanks for this, but now that I look at it closer, master of ConjugatePriors hasn't been adjusted to work with PDMats 0.4 or newer so this won't actually fix the issue yet. I think.

andreasnoack:
I've delete the tag and will close here. When the issues have been settled I'll retag.

PR #4870, opened 2016-03-26
Register GUITestRunner

meoke:
Pull request for registering module GUITestRunner which is a graphical user interface for Julia TestRunner.


tkelman:
If version 0.0.1 does not actually work on Julia 0.3 it's better to remove that tag.

meoke:
@tkelman, thank you for your suggestion. I've removed that tag.

PR #4889, opened 2016-03-31
Register and Tag JuliaAudio packages

ssfrr:
This PR registers and tags SampledSignals.jl and RingBuffers.jl which provide underlying useful functionality for working with sampled data. It also registers 3 dependent packages - JACKAudio.jl, PortAudio.jl, and LibSndFile.jl, which provide functionality for reading/writing audio to system devices as well as files.

Everything is set up with Travis on OSX and Linux, 0.4 and nightly with the exception of PortAudio.jl, which isn't testable on Travis because they don't provide any audio devices and the underlying portaudio library doesn't provide any sort of dummy devices that could be used for testing. Perhaps in the future I'll write a little C shim that emulates the portaudio API so that the code can be automatically tested. For now I run the tests on my local machine.


tkelman:
Do the old tags of PortAudio and LibSndFile need to be registered? That will make the old versions of the packages newly installable via `Pkg.add` if anyone is still using Julia 0.3.

ssfrr:
no, those tags don't need to be registered. Both those packages are forks of AudioIO so those are leftover tags from the old package.

I did all the registering, tagging, and publishing with `Pkg`. If I want to remove some of the tags can I just remove the unwanted tag directories from the `version` directory and update the PR?

tkelman:
Yep, un-registering the tags is just a matter of removing the files from this metadata branch.

ssfrr:
OK, just pushed a new commit to the PR so now only the newest tag for each package is being registered

ssfrr:
Woot. thanks. And thanks as always for your attention to detail and general Pkg gardening.

PR #4893, opened 2016-04-01
Tag the object file suite

Keno:
Except for COFF.jl, which will come in a bit once I've had a chance to work on it again.


tkelman:
Most of these are missing some or all of: Readmes, tests, Travis

Keno:
Most of these are already in METADATA from when we didn't require those things

tkelman:
We don't absolutely require any of these things, but they're highly recommended to make sure the package is actually usable, documented and tested. Especially if these are going to be dependencies of critical infrastructure like the debugger.

Keno:
All of the above will come, but require some time investment, which I do not have available at the moment.

StefanKarpinski:
We can work on testing, docs, etc. for these packages in the very near future but it's very important that we get the packages needed to Pkg.add the debugger out now.

tkelman:
Turning on Travis takes a few seconds. It doesn't have to pass right away, but it does have a tendency to catch bugs.

PR #4894, opened 2016-04-01
Tag JuliaParser v0.7.0

Keno:
Major rewrite. From the announcement email to julia-dev:

 Impact for existing users is likely:

- May be slightly slower, though I haven't done any performance work yet, so that'll probably be resolved soon
- Minor API incompatibilities
- A significant number of internal APIs take extra arguments (only an issue if you're
  building on top of JuliaParser)
- May no longer parse your code. I tried my best to get all the syntax corner cases right, but the julia syntax is large, so it is likely I missed something. Please file any issues you encounter.
- Targets 0.5- only. Most of the code to parse older versions is still there, but maintaining it is a significant amount of work. If somebody wants to step up and maintain that part of it, I'll happily accept pull requests. Otherwise, I'll probably slowly remove the remaining bits of support for old versions.

Now for what you get in exchange.

- I've updated JuliaParser for all recent parser syntax changes and bugfixes.
- I've added a new mode that adds token-accurate location tracking
- You can use that mode to get pretty error messages:

screen shot 2016-03-31 at 9 19 30 pm


Keno:
Hold on, need to retag.

PR #4899, opened 2016-04-01
Tag PyLCM v0.0.2

rdeits:
The primary change in this version is that I have mostly figured out BinDeps and am using it to build the underlying LCM library, rather than relying on a systemwide installation.


rdeits:
Also, the `travis.yml` for this package is now a purely standard `language: julia` build, so this should pass the CI servers without any special configuration.

tkelman:
best to add an explicit dependency on Compat instead of assuming it will be present as a transitive dependency

rdeits:
Good point, thanks. I've added it for the future.

PR #4907, opened 2016-04-02
Register Restful: v0.0.4

ylxdzsw:
a package that helps developing web servers.


ylxdzsw:
Was there something went wrong? It seems that Travis failed during downloading of julia?

nalimilan:
That was a transient failure. I've restarted it and it's OK. What's the point of tagging older releases? (Not that it would create any issues, but...)

ylxdzsw:
Pkg.publish() did all the things. Should I delete the old release manually?

tkelman:
They all have the same Julia version dependency, so tagging older versions is alright. Note that if you use `URIParser` then you should add it as a direct dependency, instead of assuming it will be present as a dependency of other packages.

ylxdzsw:
@tkelman Thanks for your suggestion. I will fix it soon.

ylxdzsw:
Finally I deleted these old tags by hand. Requirements was also fixed.

tkelman:
Okay. Note as a warning that base Julia has a currently exported `rest` function and a currently unexported `Rest` type for iterators, ref https://github.com/JuliaLang/julia/blob/c45175f81abf0d05db835a9d302b1545b730476c/base/iterator.jl#L148-L152

If the latter were to get exported it would cause the package to stop working given its current name.

ylxdzsw:
OK... It seems that I have to change the name, and I did that.

PR #4910, opened 2016-04-03
add two packages

jingpengwu:
1. Watershed.jl : do watershed segmentation from affinity graph or any other graphs.
2. EMIRT : Electronic Microscopy Image Reconstruction Toolbox


tkelman:
What is the minimum julia version requirement? Needs to be listed in the require file.

tkelman:
And what is the license of the C++ code that Watershed.jl is derived from? Needs a license file.

It also helps if you push the git tags so they show up under https://github.com/seung-lab/Watershed.jl/releases

jingpengwu:
Fixed these issues, please help to check again.

EMIRT/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4-
```
tkelman: drop the trailing `-` as that means "prereleases of julia 0.4 or newer"
you most likely want `julia 0.4` which means "releases of julia 0.4.0 or newer"

jingpengwu: yes, you are right. Just fixed this. Tks.

tkelman:
Add a license file to Watershed and push the git tag to EMIRT then this should be okay.

jingpengwu:
thanks for the checking, updated.
- created tag for EMIRT
- added `LICENSE` file for both repo.

tkelman:
The git tag on EMIRT should match the metadata tagged commit 8a323631bc00d549a30760fcf37d5fe95f79f430

jingpengwu:
changed the taged commit.
Thanks.

PR #4913, opened 2016-04-04
Tag TaylorSeries v0.1.2

lbenet:
This will be the last version compatible with Julia 0.3.

### DIFF
--- a/TaylorSeries/url
+++ b/TaylorSeries/url
@@ -1,1 +1,1 @@
-git://github.com/lbenet/TaylorSeries.jl.git
+git://github.com/JuliaDiff/TaylorSeries.jl.git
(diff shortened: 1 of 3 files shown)


lbenet:
Incidentally, can you please update the url associated with the package (it is now part of JuliaDiff). 
The correct url should read: git://github.com/JuliaDiff/TaylorSeries.jl.git

tkelman:
You can add a commit that does so here, just modify https://github.com/lbenet/METADATA.jl/blob/metadata-v2/TaylorSeries/url

lbenet:
Thanks! I was not sure if I should do it.

tkelman:
I was going to say if you're supporting Julia 0.3 then you should test against it on Travis, but if you're about to drop support then it's not a big deal - as long as your tests currently pass on 0.3 which it looks like they do. Remember to change the minor version via `Pkg.tag("TaylorSeries", :minor)` for the first tag that does not support Julia 0.3 any more.

lbenet:
Thanks for the comment. I'll tag the first commit, which will precisely add precompilation and leave aside somethings needed for Julia 0.3 (Docile, etc).

PR #4915, opened 2016-04-04
Register LoggedDicts: v0.0.1

JockLawrie:


tkelman:
Would your tests pass on Travis CI if you enabled it?

JockLawrie:
Should do. They're passing on my local machine (Julia 0.4.5). How do I enable Travis CI?

tkelman:
Log in at https://travis-ci.org/, click the little plus next to "My repositories," then flip the switch for the appropriate repo. Might need to click "Sync account" near the top of the page if this was created recently.

JockLawrie:
Enabled Travis.
Naturally enough the tests failed.
Still passes on my machine.
Oddly, it failed on a function that causes no issues earlier in the test runner. The error is related to a log file. Any ideas?

tkelman:
Bad file descriptor in flush? Not sure what could cause that. Permissions or ulimits or something?

JockLawrie:
Don't know.
I tried again but same result.
The line that fails is called earlier in runtests.jl without any issue, even logging to the same file.
Perhaps the log file is being closed before the write at which the failure occurs. Perhaps Travis has a time limit on open files of only a second or 2?
Otherwise perhaps some system error, as there's nowhere in my code that does this.
Grasping at straws here.

tkelman:
Up to you whether you want to look into it. Might indicate a real bug, or be something particular about travis. They have several options (check their docs) for using sudo enabled workers or ubuntu trusty instead of the default precise. If you contact their support you can also get temporary ssh access to a worker to try to reproduce more directly.

JockLawrie:
Thanks Tony.
In the meantime I am following up the failure anyway.

tkelman:
No problem. Also see https://quay.io/repository/travisci/travis-ruby for a copy of the Docker image that Travis uses when your job runs on a container worker.

PR #4929, opened 2016-04-07
Register EscapeString: v0.0.1

JockLawrie:


tkelman:
missing the httpcommon dependency?

JockLawrie:
Already exists in REQUIRE.
Added to src/EscapeString.jl.
Also simplified the code.

tkelman:
The dependency isn't correctly reflected in the tag here, it should be listed in https://github.com/JockLawrie/METADATA.jl/blob/metadata-v2/EscapeString/versions/0.0.1/requires

JockLawrie:
Delete this PR. Will create another.

PR #4933, opened 2016-04-07
Registering with tag 0.0.2 and tests

dehann:


tkelman:
Are there existing implementations of KDE in JuliaStats or elsewhere? How does this compare?

It's also good to push the git tags so they show up at https://github.com/dehann/KernelDensityEstimate.jl/releases

dehann:
Hi @tkelman , there are other KDE implementations, two actually. I have reimplemented a KDE from two PhDs [1] in Julia. The main focus is the ability to take the product between two KDEs. The other KDE libraries don't really do this, or nearly as efficiently. Although this new library is not complete yet, and is missing some features that the other libraries have, I think in time we can look at converging them. I need to get this into the ecosystem so people can try it out before convergence can realistically happen.

I'm going to create a new pull request with 3 new Pkgs soon. They, and this KernelDensityPackage, interoperate, for graphical model inference and robot navigation.

This is the first Julia package I'm publishing, so thanks for any guidance like tag publishing. I'll look into that now. Thanks!

dehann:
pGM, = prodAppxMSGibbsS(dummy, [p;q], Union{}, Union{}, 5)
    pq = kde!(pGM)
    plotKDE([p;q;pq],c=["red";"green";"black"])

alt tag

dehann:
@tkelman please just check if I did this right, [Release tag v0.0.2](https://github.com/dehann/KernelDensityEstimate.jl/releases/tag/v0.0.2).

diegozea:
@dehann When should a user choose to use this KDE implementation over other implementations? Could you please add the answer in the README file?

dehann:
Hi @diegozea , I've added some motivation in the README as requested.

scidom:
@dehann, does your package deal with kernels of bounded support, i.e. would it enable approximating the KDE of a positive variable (the boundary condition for this example being x > 0) or is this sth you are planning to support in the future? I am only asking out of interest, obviously not because it is a registration pre-requirement.

dehann:
@scidom, yes ultimately that would be great. I have a similar requirement: for rotations on a manifold which usually wraps at [-pi,+pi). I'm using this package as a basis for non-parametric work and currently just dealing with |pi| < x cases post-hoc. I wasn't planning on tackling the bounded support right away though...

diegozea:
Thanks @dehann !

tkelman:
I think it would be beneficial to split out the plotting capability from the pure computational pieces, but that can be a future todo.

PR #4941, opened 2016-04-10
Tag Lora v0.5.3

scidom:
Enabled MCMC sampling when target and proposal have differing support, plus various minor tweaks.


scidom:
@tkelman, do you know why the version of Lora does not seem to get updated here?

http://pkg.julialang.org/?pkg=Lora&ver=0.4#Lora

Despite being in version 0.5.2, only version 0.5.0 appears in the ecosystem above.

tkelman:
That page only gets updated when @IainNZ runs packageevaluator, no one else has write access to that site. I've asked why we can't just run it on a totally automated cron job, but that hasn't happened yet.

scidom:
@tkelman, I needed to make one last change and made this PR point to the latest Lora commit. I wanted to ask if this is ok with you, or you prefer that I open a new PR in such occasions so that I know for next time.

@IainNZ would it be easy to update the Lora version in the packageevaluator?

tkelman:
While you've got an open PR, it makes sense to put additional changes that are related to the same package on the same branch. But changing sha of an existing tag is somewhat iffy, tags should usually be immutable. Now your metadata tag doesn't match the git tag at https://github.com/JuliaStats/Lora.jl/releases which is a bit confusing. Making a new tag is cheap and usually has less potential for confusion than changing an existing one.

scidom:
Thanks for the feedback, @tkelman. I will create a new tag next time. For now, to clean up the conflict, I will close this PR, will delete the tag from Lora's repo and will recreate both.

IainNZ:
@scidom I'll generate the website again

@tkelman too busy with thesis stuff, defense t-minus 30 days D:

scidom:
Good luck @IainNZ!! The website can certainly wait.

PR #4943, opened 2016-04-10
Tag Lora v0.5.3

scidom:
@tkelman I deleted the previous mess and set up a new tag and PR from scratch.


scidom:
It would also be nice if the packageevaluator records for Lora could be updated foo, though I know this is not within your (write access) reach.

PR #4945, opened 2016-04-10
Register DriftDiffusionPoissonSystems

Stivanification:


tkelman:
You'll need to tag a version so it can be installed with Pkg.add. Would Travis CI pass on the package if you enabled it?

PR #4952, opened 2016-04-11
Pull request/1a3794a7

Stivanification:


Stivanification:
I cant tag the package at the moment because the package manager says the package is dirty, even though neither git diff nor git commit -a return anything.

Stivanification:
added commit for package to be tagged.

PR #4959, opened 2016-04-12
Tag Weave v0.0.5

mpastell:
Updated (finally) for 0.4x compatibility


tkelman:
If this is the first tag to drop support for Julia 0.3, it should use a new package minor version 0.1.0.

mpastell:
Thanks, I will make another pull request with 0.1.0 tag

PR #4963, opened 2016-04-13
Register k_cores and powerlaws

johnybx:
I created two packages as my school project and I think it would be worth having them as official.


tkelman:
Note that we currently don't have any non deprecated registered packages that contain underscores in yheir names, or start with a lower case letter.

PR #4967, opened 2016-04-13
Pull request/85e89328 : after tagging version 0.0.3

cecileane:


tkelman:
Cool. Best to explicitly depend on ColorTypes (and all other imported packages) in REQUIRE rather than assuming it will be present as a transitive dependency of Gadfly.

Do you want to register the 0.0.1 tag which will be installable on Julia 0.3, or leave it out and make all registered versions 0.4-or-newer?

cecileane:
Thank you Tony! 

- Yes version 0.0.1 has its uses, we would like to register it. It has a set of plotting tools using GraphViz, and GraphViz only works with Julia 0.3. We developed an independent (and different) set of plotting tools  using Gadfly. 

- I just added ColorTypes to REQUIRE. Thanks for checking for consistency! The 0.0.3 tag does not include this new commit obviously (with the new REQUIRE file). Should we tag the package again with a new version (0.0.4) and send a new pull request? I don't know how to edit the tag v0.0.3 to refer to the latest commit, or how to send a publish request with a 0.0.4 tag but without 0.0.3.

tkelman:
I don't think any action is required on your part right now, just leave this open for any more feedback and merge tomorrow if none.

I'd usually suggest using different minor versions for the first new release that drops support for a specific version of Julia, but 0.3 is now quite old so the chances of needing to make a new 0.3-supporting bugfix release of the package are low.

PR #4974, opened 2016-04-15
Register KCores and PowerLaws

johnybx:
I changed the names of packages #4963


tkelman:
Looks good! If you tried to enable Travis CI (https://docs.travis-ci.com/user/languages/julia) for these packages would your tests pass?

johnybx:
I ran all tests when I was renaming packages so I expect them to pass. Problem could be that I have tests but not in runtests.jl file I probably should add it. Quick question : Is it enought to push changes to my repositories or I need to register packages again ?

tkelman:
You can push additional tags to this pull-request/9d688b1d branch in your metadata fork if you'd like, and the new commits will show up here.

johnybx:
I can't manage to push updates here because I used Pkg.tag() and Pkg.publish() which created another branch so I open one more pull request and close this probably.

PR #4986, opened 2016-04-18
Tag ImageQuilting v0.2.8

juliohm:


tkelman:
Please do `git push --tags` to the package repository so the tags show up under https://github.com/juliohm/ImageQuilting.jl/releases

juliohm:
@tkelman I will push the tags in the next minor release. I don't want to start the numbering on GitHub with a simple fix. Makes sense?

tkelman:
You already have METADATA tags for existing releases. Just push the corresponding git tags to the package repository for them.

juliohm:
@tkelman could you please elaborate on that? You mean that I can manually go over all tags so far in METADATA and tag my local repository before the pull request?

tkelman:
Do you use `Pkg.tag` for these? If so it should be doing that automatically, so you may have the tags locally already but not pushed, unless every single one of your metadata tags was done manually? If that's the case then shouldn't be too hard to write a little script to make git tags from the existing metadata tags.

juliohm:
I use `Pkg.tag`, but I have deleted my local repository multiple times switching from one machine to another. I have to write a script then?

tkelman:
It can wait, but it's a good idea for the sake of code organization and navigation.

juliohm:
@tkelman thanks, will try solve that in future requests.

PR #4992, opened 2016-04-18
Tag ScikitLearn/ScikitLearnBase v0.0.3

cstjean:


tkelman:
Ah, missed this first time around, please add a license file to ScikitLearnBase. And can that be meaningfully tested?

cstjean:
I just added the license file. ScikitLearnBase.jl is mostly just a bunch of function forward declaration, and it is amply tested by ScikitLearn.jl. Is there any particular reason to test it further?

tkelman:
To make sure it can be installed and loaded on its own.

cstjean:
ScikitLearnBase.jl is useless on its own. It exists so that ML packages can implement the interface without having to REQUIRE ScikitLearn.jl and all of its dependencies. For example, [this PR](https://github.com/STOR-i/GaussianProcesses.jl/pull/17). The alternative is to use `@require`, and that's not very clean. Is there another way? It works well, but people do seem a bit confused by the setup, even though MathProgBase/JuMP have a similar relationship.

tkelman:
MathProgBase has tests of its own to mock out that interface. And to be sure you don't introduce syntax that supported versions of julia can't handle.

cstjean:
Fair enough, I'll give it a shot.

PR #4995, opened 2016-04-19
Pull request/fd1311dd

dysonance:
I attempted to publish these packages a while ago, and the consensus seemed to be that it would be better to further develop them and then resubmit the pull request. I've now spent a great deal of time adding more functionality to these packages, and think that they might be ready for another look.

**Temporal** offers a time series type (`TS`) that I think has some promise. My primary focus was to make indexing extremely flexible and to add enough basic operations for a solid foundation to build on. I think it's sufficiently different from the TimeSeries package at this point to warrant its own space.

**Indicators** implements a series of technical analysis algorithms. It differs from MarketTechnicals in that it doesn't have any dependencies and will work with the Base Array type, and it also has a larger "selection" of indicators that have been implemented. If both Indicators and Temporal get merged I would plan to add functionality for TS objects as well.

Let me know any questions/comments/concerns/criticisms/insults, and I'd be happy to discuss :)


Temporal/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+1cfd96d419329f3618c4e56da609027c36411e8d
```
tkelman: this tag looks like it's from february?

tkelman:
move `using Base.Dates` inside the module, not outside

no need to re export names already exported by base, as long as you import and extend them to new types

tkelman:
Is any of your code derived from the R packages? They are both GPL.

dysonance:
@tkelman Is there a problem with the tag being from February? It must be from when I originally tagged it a while ago — is that a problem? I don't believe any of my code was taken from the R packages. As for moving Base.Dates inside, would I have to make that change and do another pull request? Apologies, still new to the whole Github game...

tkelman:
No, just seems like you've been developing since then and might want to have the latest code available when users to Pkg.add? I'd recommend making new tags after any additional changes, and either pushing the new tag commits into this same PR's branch, or making a new PR with multiple tags.

dysonance:
@tkelman How would I add the new tags and push them into this same PR's branch?

tkelman:
If you do Pkg.tag, then go to metadata and have a look at the logs for the newly added commit sha. Then checkout this branch from your fork, cherry pick the commit if needed, and push it to this branch in your fork. I'm in and out today but we can help translate this into specific commands if you need more help.

dysonance:
I recently installed a new OS on my laptop where I'm planning to do most of my development. So I'm going to clone the packages there, make the changes you requested, and try resubmitting a new pull request.

PR #4999, opened 2016-04-20
Tag AWS* and deps for v0.5 compat

samoconnor:
v0.5 compat tweaks and misc accumulated bug fixes


tkelman:
InfoZIP needs to add Compat to REQUIRE.

Possible to test AWSLambda on Travis?

samoconnor:
> InfoZIP all the packages that use Compat need to add it to their REQUIRE.

True, see #5001

> Possible to test AWSLambda on Travis?

In theory, yes.

But it could get expensive if things went wrong (e.g. EC2 boxes left running after failed tests), so I'm not currently prepared to put my company's AWS credentials into a travis config with permissions to do this.

However, I do have a Lambda-based CI system for Julia in the works. When that is ready, it will make, Travis seem really slow.

tkelman:
closing in favor of #5001

PR #5002, opened 2016-04-20
Register and Tag VisualRegressionTests

tbreloff:
VisualRegressionTests is a small package to manage comparison of image output to a static reference image.  It underlies the testing suite of Plots.jl.

When run in the REPL (and Gtk is installed) it will open a side-by-side comparison of the images, and ask whether to add/replace the reference image when the images differ by "enough".  This makes it easy to see and recover from regressions in any package that produces image output without requiring per-pixel exact matches, and thus makes it more likely that a package owner will actually maintain meaningful integration testing.  Thanks to @timholy for the original testing algorithm in Images.


tkelman:
please push the git tags so they show up under https://github.com/tbreloff/VisualRegressionTests.jl/releases

tbreloff:
Sorry about that... they are pushed now.

PR #5003, opened 2016-04-20
Tag ScikitLearnBase v0.0.3

cstjean:
Added Julia 0.3 support to pass CI on GaussianMixtures.jl


tkelman:
This might be one of the first packages to add support for earlier versions of julia. If in the future you need to drop support for 0.3 then be sure to change the package minor version for the first tag that does so.

PR #5004, opened 2016-04-21
Register FaSTLMM: v0.0.1

sens:
Package for Factored Spectrally Transformed Linear Mixed Models (FaSTLMM).  

Used in genetic association studies to fit linear mixed models where the error variance is a linear combination of the identity matrix and a positive definite matrix.

Saunak Sen


tkelman:
Would your tests pass if you enabled Travis CI? Note that the test file needs to be called `test/runtests.jl` for `Pkg.test` to work.

sens:
Sorry, I did not have `tests/runtests.jl` file.  I have added that.

sens:
Code builds on Travis CI for Julia 0.4, but not 0.5, the development version.
Should I change the REQUIRE file?

tkelman:
That's not absolutely required for under-development nightly versions, it looks like the failure is caused by one of your dependencies anyway which may get fixed on its own later. For now you might want to set an `allow_failures` for nightly Julia so you can have a green status badge, see https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

sens:
Thank you for the suggestion!  All green now. :)

PR #5007, opened 2016-04-21
Tag CPUTime v0.0.5

schmrlng:
Fix https://github.com/schmrlng/CPUTime.jl/issues/5. I wonder if github "Fix #" works across repos?


CPUTime/versions/0.0.5/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3-
```
tkelman: drop the trailing `-`

schmrlng:
Okay, I've dropped the '-'.

PR #5015, opened 2016-04-22
Register ToeplitzMatrices: v0.1.0

andreasnoack:
The documentation is work in progress but @dlfivefifty has requested that this package gets registered such that he can depend on it.


tkelman:
this appears to depend on an unregistered IterativeLinearSolvers package without declaring it in REQUIRE

dlfivefifty:
I've made a pull request to add a README

andreasnoack:
That dependency should be now be gone.

PR #5026, opened 2016-04-24
Register RequestsCache and Tag RequestsCache v0.0.1

femtotrader:


tkelman:
Best to add `URIParser` and `HttpCommon` explicitly to `REQUIRE` instead of assuming they will always be indirect dependencies of `Requests`.

femtotrader:
I've merged your PR https://github.com/femtotrader/RequestsCache.jl/pull/6
and update REQUIRE https://github.com/femtotrader/RequestsCache.jl/commit/5185c666181a1b0612eadaca14548b24268dfd8c

What should I do now ? Sorry for my newbies question.

femtotrader:
Thanks for merging but I noticed that even if https://github.com/femtotrader/RequestsCache.jl/blob/master/REQUIRE was update to add `URIParse` and `HttpCommon ` that's not the case for https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/RequestsCache/versions/0.0.1/requires
Maybe it will be fixed automatically when I will tag an other version ?

tkelman:
Yes, the next version you tag will reflect that change.

PR #5027, opened 2016-04-24
Register FLAC: v0.1.0

staticfloat:


tkelman:
New package registrations should really stay open for about 24 hrs feedback period.

staticfloat:
Ah, my bad.  Sorry about that.  Hopefully this particular package isn't too controversial, but I will remember to do so in the future.

PR #5040, opened 2016-04-26
New module: NumericIO.jl

ma-laforge:
Currently: Display numeric values using SI prefixes.

**Comment:**
This module makes use of Grisu, which I do not believe is a public feature of Julia... So I am not convinced it is technically "acceptable" to rely on it from an external module.


quinnj:
I don't think it's a deal-breaker to rely on Grisu, but note that these kind of "internal modules" in Base tend to be more subject to breaking change. You may just need to set up a few additional tests and try to watch if anything changes in Base and how it might affect you. But at this point, Grisu has been fairly stable for a while. If anything, I would imagine it gets replaced sometime in the future with some other fancy printing algorithm, but that's not anything to worry about now. +1

tkelman:

ma-laforge:
@tkelman,
I am not sure how the Travis or AppVeyor systems work yet.  I would have to look into that.

I am not even certain of the difference between the two.

tkelman:
Travis CI does Linux and OSX: https://docs.travis-ci.com/user/languages/julia
AppVeyor does Windows: http://www.appveyor.com/docs (and my lightning talk from last year's JuliaCon)

ma-laforge:
@tkelman,
To answer your question: Travis shows success on Julia:release, but failure on Julia:nightly.

tkelman:
Great. You can set allowed failures on nightly if you want that green badge and don't have time to worry about Julia nightly yet.

ma-laforge:
Thanks. I found out more about how to get the green badge on another post.

I guess pkg.julialang.org will eventually get updated at some point (as in I don't have to do anything else special).

PR #5041, opened 2016-04-26
Register PATH: v0.0.1

chkwon:
This is a simple wrapper for the PATH Solver, available at http://pages.cs.wisc.edu/~ferris/path.html. See also https://neos-server.org/neos/solvers/cp:PATH/AMPL.html.

I know that the package name could be too general. However, the PATH solver is almost the only successful solver available for solving complementarity problems. Well, like CPLEX for linear optimization problems. 

I was also considering PATHSolver.jl, but looks ugly. Any suggestions?


tkelman:
PATHSolver would be more specific.

chkwon:
@tkelman Alright, I'll change the name and resubmit the request.

PR #5042, opened 2016-04-26
Tag the object file suite

Keno:


tkelman:
Push the git tags. These could still really use tests and docs.

PR #5043, opened 2016-04-26
Registering PATHSolver v0.0.1

chkwon:
Julia wrapper for the PATH Solver, a numerical library for solving mixed complementarity problems.


tkelman:
Modifying `DYLD_LIBRARY_PATH` shouldn't usually be needed. If you enabled Travis (initially only for osx) would your tests pass there?

chkwon:
@tkelman Unfortunately, it didn't pass the test:

```
=============================[ ERROR: PATHSolver ]==============================
LoadError: Provider BinDeps.BuildProcess failed to satisfy dependency libpath47julia
while loading /Users/travis/.julia/v0.4/PATHSolver/deps/build.jl, in expression starting on line 43
================================================================================
```

https://travis-ci.org/chkwon/PATHSolver.jl/builds/126089617

tkelman:
These can be a bit tricky to debug. What happens if you try to call `Libdl.dlopen` on that library?

chkwon:
PATHSolver package needs three libraries:

libpath47.dylib
libpath47julia.dylib

chkwon:
PATHSolver package needs three libraries:

- `libpath47julia.dylib` (which basically calls functions in `libpath47.dylib`. The Julia code interfaces with this library)

- `libpath47.dylib` (which is the main library, provided by the original author, requiring `libgfortran.3.dylib`)

- `libgfortran.3.dylib`

The first thing I tried was to place all three libraries in one directory, and then see if `libpath47julia` is callable from Julia. No success, that's why I had to put them in `DYLD_LIBRARY_PATH`.

Given my [build.jl](https://github.com/chkwon/PATHSolver.jl/blob/master/deps/build.jl), where would you recommend to put `Libdl.dlopen` for which of the above three libraries? 

I think Julia can find  `libpath47julia` without `DYLD_LIBRARY_PATH`, which however cannot find the other two.

tkelman:
You might need to dlopen them in the reverse dependency order, once in your build script so BinDeps succeeds, and again under src in a module `__init__` function.

chkwon:
I did this and that with BinDeps, and no good yet. It can find gfortran well, but `libpath47` cannot find gfortran.

I tried Libdl.dlopen:

```julia
julia> Libdl.dlopen("/opt/homebrew-cask/Caskroom/julia/0.4.5/Julia-0.4.5.app/Contents/Resources/julia/lib/julia//libgfortran.3.dylib")
Ptr{Void} @0x00007fd228441020

julia> Libdl.dlopen("/Users/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47.dylib")
ERROR: could not load library "/Users/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47.dylib"
dlopen(/Users/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47.dylib, 1): Library not loaded: /usr/local/lib/libgfortran.3.dylib
  Referenced from: /Users/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47.dylib
  Reason: image not found
 in dlopen at /opt/homebrew-cask/Caskroom/julia/0.4.5/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in dlopen at libdl.jl:36
```

`libpath47` complains it cannot find `libgfortran.3.dylib`.

tkelman:
Ah right sorry, mac libraries don't like being relocated, or they make you work harder for it with install-name-tool or whatever that's called. Homebrew has some tricks for dealing with this so you don't have to mess with things by hand, but I guess this is a closed-source library? @dpo you don't happen to be familiar with the PATH library, do you?

chkwon:
PATH is a "free", closed software, I would say. It does not provide source codes, and it requires a license to use it, but a license is provided for free.

http://pages.cs.wisc.edu/~ferris/path.html
> The version that is downloadable from here (i.e. the file pathlib.zip in this directory) is free, but is limited to problems with no more than 300 variables and 2,000 nonzeros.
> The file LICENSE details how to set up a temporary license that removes the size restriction for a year. A new license string will be provided in December of each year.
>

chkwon:
@tkelman So, do you expect it will be smooth in Linux? without `LD_LIBRARY_PATH`?

tkelman:
Not sure, would have to try. I think we'll merge this tag for now, and BinDeps improvements can be ongoing work for future tags. Hopefully you can get the installation experience to be automatic eventually, but these things take some fiddling.

chkwon:
@tkelman I don't know how Homebrew works; would it be possible to create a custom formula for this package?

chkwon:
Yes, I really want the installation to be automatic. I want to basically use this package in classroom environments and if the installation is not automatic, my office hours will be crazy crowded.

tkelman:
Dominique would have a better answer on the Homebrew question than me. Or maybe @staticfloat, I think homebrew-cask is designed to help with repackaging and relocating possibly-proprietary binaries but I don't know whether Homebrew.jl can utilize cask packages.

staticfloat:
If you're talking about the `.mexmaci64` files, we don't need to resort to anything as complex as Homebrew.  We can just download the files directly with `BinDeps`, unpack them into `<package dir>/deps/usr/lib`, and you should be good to go.

tkelman:
It's a bit more complicated than that. There are several interdependent dylibs here, and they aren't properly seeing each other with just dlopen, unless you first set `DYLD_LIBRARY_PATH`. Homebrew has some mechanisms for making that no longer necessary, right?

chkwon:
In Linux (Xubuntu64), I did a quick test. I placed the following two file:

- `libpath47.so`
- `libpath47julia.so` (which calls `libpath47.so`)

in `PATHSolver.jl/deps/usr/lib`.

WITHOUT `LD_LIBRARY_PATH` setting:
```julia
julia> Libdl.dlopen("/home/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47.so")
Ptr{Void} @0x0000000002e56420

julia> Libdl.dlopen("/home/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47julia.so")
ERROR: could not load library "/home/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47julia.so"
libpath47.so: cannot open shared object file: No such file or directory
 in dlopen at ./libdl.jl:36
 in dlopen at libdl.jl:36
```

WITH `LD_LIBRARY_PATH` setting:
```julia
julia> Libdl.dlopen("/home/chkwon/.julia/v0.4/PATHSolver/deps/usr/lib/libpath47julia.so")
Ptr{Void} @0x0000000002bb8f30
```

Since I have compiled `libpath47julia.so` by myself, I'm afraid of any errors or mistakes during the compile process. I'm not really familiar with C. This is how I did:
```make
PATHDIR = ../../pathlib
PATHLIB = $(PATHDIR)/lib/linux64

# additional libs - we expect them to be required only for static builds
XLIBS = -lgfortran

CC    = gcc
CFLAGS = -I$(PATHDIR)/include -m64

all: linux64

linux64: pathjulia.c Standalone_Path.c
	$(CC) $(CFLAGS) -dynamiclib -shared -fPIC -o libpath47julia.so $^         -L$(PATHLIB) -lpath47 -lm
```

staticfloat:
I think it will be easier for us to relocate the file manually rather than use Homebrew to do it.  It will also be more straightforward.  Here's [a pull request](https://github.com/chkwon/PATHSolver.jl/pull/1) that does what I'm suggesting.

PR #5049, opened 2016-04-27
Point Showoff at JuliaPackageMirrors

Keno:
@dcjones is currently not checking his packages so there are changes accumulating that make it difficult to get those packages running on 0.5. Most of them have other people with commit access, but Showoff.jl does not. Thus, I propose temporarily moving development of that package to JuliaPackageMirrors until @dcjones returns. 

Will this work? I'm afraid of the old remote staying in people's local copies causing all sorts of ruckus. cc @tkelman @wildart

### DIFF
--- a/Showoff/url
+++ b/Showoff/url
@@ -1,1 +1,1 @@
-git://github.com/dcjones/Showoff.jl.git
+git://github.com/JuliaPackageMirrors/Showoff.jl.git


tkelman:
Your fear is justified, I'm not sure. I think we'd need to try making this change in our local metadata copies, and see if Pkg.update complains on either 0.4 or 0.5-dev.

tkelman:
Oh, and we should also check with @MichaelHatherly whether his mirroring script will properly cope with active changes happening on a JuliaPackageMirrors repo, and hopefully not force push over anything. How does `push --mirror` behave? https://github.com/JuliaPackageMirrors/mirror-updater/blob/07253a88c1ef57c37d62cebdb4f513b20bf470ab/update.jl#L63

wildart:
For new installations, it will work. If the package is already installed, it will never get this change.

Keno:
Right, which is a problem if we want to tag new versions at the new url. Can we have the Pkg code check if the url changed before attempting to pull new code?

tkelman:
If we do that then we should also start a new 0.5-only minor series to make any new tags from, otherwise anyone with the package installed on 0.4 would hit errors upon `Pkg.update` about not being able to find the new tagged commit.

Keno:
yes

wildart:
new METADATA branch?

Keno:
Not needed as long as any tagged versions don't target old julia versions.

tkelman:
Ideally the fix for changing urls (we should open an issue on this) should be backported to 0.4's Pkg if not too complicated, then future versions of the package could restore 0.4.y support for the first y release that includes the backport. I'm surprised this limitation hasn't come up yet. Github automatic redirects have usually been saving us.

MichaelHatherly:
`push --mirror` overwrites the remote I believe. We could add Showoff to the [blacklist](https://github.com/JuliaPackageMirrors/mirror-updater/blob/07253a88c1ef57c37d62cebdb4f513b20bf470ab/update.jl#L7) in the updater script to ignore it.

That said, the mirror org wasn't really meant to host any active dev, but rather just archive repos so if they happen to get deleted from their source they wouldn't be lost forever. Might it be better fork `dcjones/Showoff.jl` to something like `JuliaLang/Showoff.jl` and continue dev there?

tkelman:
A fork could work too, though JuliaLang has a few too many packages, maybe a different org or someone's personal account for now.

MichaelHatherly:
> few too many packages

Very true. Probably a user account then.

lobingera:
When you have problems finding a user accout (https://github.com/dcjones/Gadfly.jl/issues/815), an organisation could be the option.

tkelman:
We could move these things to JuliaGraphics or similar, but given Pkg won't do the right thing for everyone who has an existing installation of the package, we should fix that first before moving on this.

Keno:
I would be very interested in seeing the Pkg issue fixed, since right now every user of Gadfly has to first fix Showoff. Is anybody planning to work on the Pkg issue?

tkelman:
I was mistaken and overly paranoid here, sorry. Looks like Pkg actually will do the correct thing here, at least it did when I made this change in my local METADATA for 0.3, 0.4, and 0.5-dev.

Let's do this but instead of JuliaPackageMirrors, use JuliaGraphics instead. @SimonDanisch or @timholy could you get myself and Keno commit access to that org? Few other active people too while you're at it.

lobingera:
one last note: Actually what Showoff is doing is formatting data into strings. It's a requisite for labeling in graphs and plots but the processing part is closer to Formatting.jl and similar.

PR #5051, opened 2016-04-27
Register VSL

sunoru:


tkelman:
Is VSL always a part of MKL? I guess we have the precedent of acronyms with VML.jl already.

sunoru:
Yeah. VSL and VML are both common libraries in MKL for different use...

tkelman:
Oh, could you please push the git tag to the package so it shows up under https://github.com/sunoru/VSL.jl/releases? Thanks!

sunoru:
Ah I forgot. Done.

PR #5052, opened 2016-04-27
Temporal & Indicators (resubmission)

dysonance:
Following up on a previous [pull request](https://github.com/JuliaLang/METADATA.jl/pull/4995), I made the requested changes (as far as I can tell) and am again requesting Temporal and Indicators to be included in the Julia package ecosystem.

As I am still new to Git(Hub), feel free to comment if I have unintentionally botched something while trying to make the changes requested by @tkelman


tkelman:
These look good, the new 0.0.2 tags also look fine so I think you could revert the removal?

dysonance:
Is the reversion necessary (i.e. how does it affect the Pkg.add function in a practical sense)? Based on the package dev guide in the docs, it sounds like v0.0.1 is preferred for the initial public release?

tkelman:
So because you have the "removing extra tags" here, that means when someone does Pkg.add they would initially get your 0.0.1 version which is now a month or two old. That's fine, and the initial version number isn't too big a deal. Some people register at 0.0.2 or 0.0.3, or 0.1.0, or even 1.0.0 (though since Julia isn't 1.0 itself yet, the latter is a bit iffy), the main thing that initial version number guideline is about is to avoid tagging a "0.0.0" version which used to be allowed.

If it would be simpler for you to let this get merged first, then do the newer 0.0.2 tags in a separate pull request, we can do it that way, just say so.

dysonance:
How do you check when the tags were made? Reason I ask is since I recloned them into a new Julia setup, re-registered them (and retagged them), my intuition was that this was a "new" 0.0.1 and the current code would be what's retrieved from Pkg.add

The other thing that gives me this impression is that on he github pages themselves where this is hosted, each package now only has a 0.0.1 tag, so I thought that would apply to the most recent updates, as opposed to when the tag was initially registered

Apologies if this is beyond the scope of just getting this done. But you've been extremely helpful so far in explaining how things work so I just wanna make sure I have a handle on this package development workflow so that future changes will be far smoother :)

Temporal/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+1cfd96d419329f3618c4e56da609027c36411e8d
```
tkelman: I'm looking at this commit - https://github.com/dysonance/Temporal.jl/commit/1cfd96d419329f3618c4e56da609027c36411e8d which is from February, and similar above for Indicators

dysonance:
Okay makes sense I think.

I'm going to try to avoid creating a new pull request, but I'm picking this up on my work machine now, so here's my plan (sorry in advance if I clutter your life with yet another pull request).

- Get the updated package versions on my computer
- Tag the packages (so each will have 0.0.1 _and_ 0.0.2 tags)
- Do a `git push --tags` for each package so my Github repos are with the program
- Add my METADATA fork as a new remote on this computer
- Do another `git push dysonance metadata-v2`
- Pray that it does what I want by updating this pull request

dysonance:
Okay so... Success in that I now know how to edit pull requests. Failure in that I accidentally added a bunch of changes to other packages once I did the `push`.

Perhaps the better question initially might have been: how do I edit a pull request piecemeal (i.e. do a "line item veto" of what I don't want)? You mentioned reverting just one commit from the initial pull request, so now I need to do that with all the garbage I accidentally just dropped in here...

tkelman:
It looks like you did an interactive rebase (good) but then tried to push, got a complaint about it not being a "fast forward" so you followed what git told you to do of merging then pushed the merge commit. It's that merge commit with 2 parents that makes this look funny. Instead, try the interactive rebase again to only select the commits you want, and check with `git log` that the history looks right. Then instead of doing `git push` to this branch of your fork, do `git push --force` to the same branch. You need to tell git to ignore the fact that it isn't a "fast forward," you actually want to replace what was previously here.

dysonance:
Can you explain what you mean by "interactive rebase"?

dysonance:
I tried for several hours to clean up a massive mess that I ended up making, so I'm going to close this and just open up a new one.

tkelman:
Ah sorry, interactive rebase is the `-i` flag to `git rebase`, see https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

dysonance:
Yeah I checked that stuff quite a few times, couldn't seem to clean anything up though. Now that I'm familiar, I know that I did not in fact do an interactive rebase at any point. Although you were right that it did get rejected: I'm trying again now and I'm getting the following

```
$ git push dysonance metadata-v2
To https://github.com/dysonance/METADATA.jl.git
 ! [rejected]        metadata-v2 -> metadata-v2 (fetch first)
error: failed to push some refs to 'https://github.com/dysonance/METADATA.jl.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

So I tried doing a Pkg.update, and hit this wall:

```
julia> Pkg.update()
INFO: Updating METADATA...
INFO: Updating cache of Temporal...
ERROR: couldn't update C:\Users\jamos\.julia\v0.4\.cache\Temporal using `git remote update`
 in error at error.jl:21
```

Any ideas?

tkelman:
Add `--force` when you push (generally be careful with this though - it's almost always possible to get back anything you force push over, but can be a little messy sometimes) if you've rearranged the history at all. It's also generally a good idea to use new specific branches for each pull request, rather than working off of `metadata-v2` (or `master` for other repos) all the time.

dysonance:
Hm, I hadn't considered opening a new branch.

I tried it using the same `metadata-v2` as before but added a `--force` to the end, and it took it. So now that I think I've learned my lesson, opening a new PR now if you have any shred of patience left.

PR #5073, opened 2016-04-30
Tag Currencies v0.3.0 [git://github.com/TotalVerb/Currencies.jl.git]

TotalVerb:
- (Breaking) Removed `float` (method pun).
 - (Breaking) Corrected behaviour of `one`.
 - Deprecated `StaticBasket` and `DynamicBasket` in favour of unified `Basket` type.
 - More lenient currency promotion and conversion (support for mixed-precision arithmetic)
 - Support fld, mod, and (Julia master only) fld1 and mod1
 - Fix deprecation warning on latest Julia master
 - Added dependency on Compat
 - Updated currency symbols
 - Simplifications to method dispatch, reducing number of overloads (no functional change)

Bumped the minor version because of two breaking changes and a deprecation.


quinnj:
I just had a thought. I wonder if it'd be useful to somehow "publish" the commit message when a user ran `Pkg.update()`. For example, in the case above, if I had a previous version of `Currencies.jl` installed, when I ran `Pkg.update()` (after this was merged), it would run through my packages, updating, and when it upgraded `Currencies`, it would print out the commit message. Or perhaps only if I prefaced the commit with
```
RELEASE NOTES:
- breaking change `x`
- new feature `y`
- etc.
```

tkelman:
Cool idea, open an issue in PkgDev maybe? Though you'd want to be able to opt in or out of this, and I don't think we have a well-tested built-in pager in the REPL in case these get long.

PR #5082, opened 2016-05-01
Tag Optim v0.4.5

pkofod:


tkelman:
Since this dropped 0.3 support, it should have been Optim version 0.5.0. Can be fixed with the next tag.

pkofod:
Oh, sorry. So with the next tag, I should fix this ?

tkelman:
Yeah, do `Pkg.tag("Optim", :minor)` next time the package gets tagged. I'm about to propose adding an automatic check for this.

PR #5084, opened 2016-05-02
Tag Reactive v0.3.3

shashi:
fix https://github.com/JuliaLang/Reactive.jl/issues/95


tkelman:
Your tests appear to be hanging on 0.4 Linux? https://travis-ci.org/JuliaLang/Reactive.jl/builds/125016663

shashi:
It's weird how it only hangs on this one platform. I can't reproduce this on Julia 0.4.5... It's a corner case that hangs, which users are not supposed to encounter. Can we merge this? The issue this fixes is pretty bad to have around.

tkelman:
Sure, though you have an undeclared dependency on Compat, can you fix that first?

PR #5093, opened 2016-05-03
New version bump for ODBC

quinnj:
Please hold on merging until later. I want to craft an ANN post for julia-users as this is a major breaking change release.

### DIFF
--- a/ODBC/versions/0.4.0/requires
+++ b/ODBC/versions/0.4.0/requires
@@ -1,0 +1,7 @@
+julia 0.4
+Compat
+NullableArrays
+DataStreams
+CSV
+SQLite
+DecFP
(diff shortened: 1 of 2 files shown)


tkelman:
Don't do `using DataStreams` outside the top-level ODBC module. Are the CI failures on Mac and Windows expected?

quinnj:
I have to do `using DataStreams` at the top level in order to bring the submodule exports into scope `Data.stream!(csvsource, Data.Table)` and all that.

Yeah, the CI failures are annoying for ODBC because they rely on some weird non-julia dependencies. I have a whole test system that I've set up separately where I'm hitting my own DB instances and such, but those are obviously not up for public consumption.

tkelman:
Shouldn't you only need `using DataStreams` to be available in the ODBC module scope? Unconditionally forcing global use of a different package's exports isn't quite right.

PR #5098, opened 2016-05-03
Tag TimeSeries v0.8.0

GordStephen:
Not sure why the v0.7.4 commits are picked up here, as they were already merged in https://github.com/JuliaLang/METADATA.jl/pull/4749?


tkelman:
#4749 was never merged. The trailing dash needed to be removed.

tkelman:
The package should also not be doing `using Base.Dates` outside the top-level package module, only inside.

milktrader:
How should we proceed then? The trailing dash for 0.4 is no longer in the proposed 0.8.0 version.

PR #5099, opened 2016-05-04
Tag BiomolecularStructures v0.2.3

gp0:
bug fix release


tkelman:
Note it's generally better to use `dirname(@__FILE__)` rather than `Pkg.dir("BiomolecularStructures")` in case the package gets installed to a different location. It's also probably a good idea to declare the dependency on BinDeps and Conda in REQUIRE (since they're both used in your `deps/build.jl`) rather than assuming they will be present as indirect dependencies.

PR #5109, opened 2016-05-04
Tag MarketData v0.3.6

milktrader:


tkelman:
Oddly the 0.3.0 tag here says it supports Julia 0.3, but the rest of the 0.3 series is 0.4-only. If there's ever a need to do a bugfix that supports Julia 0.3 it may be worth modifying that tag and replicating it at the end of the 0.2.x series.

milktrader:
Hmm, I don't recall having made that oversight but thanks for the observation.

PR #5118, opened 2016-05-05
Retag SnoopCompile as v0.1.0

timholy:


SnoopCompile/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3
```
tkelman: is this right?

timholy: Prob not.

I have to give a talk in a few hours, this must wait.

PR #5121, opened 2016-05-06
Tag XGBoost v0.2.0

slundberg:
A much needed tagging of a new release.


XGBoost/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.3-
```
tkelman: drop the trailing dash

tkelman: actually this doesn't support 0.3 at all as it uses `Ref`, so it should be updating the lower bound here - https://github.com/dmlc/XGBoost.jl/pull/29

slundberg:
Thanks! We should have had 0.3 tests or explicitly changed support. Is there a simple alternative to `Ref` in 0.3? I would happy to keep support, but I think the c++ library interface changed which caused the use of `Ref` here.

Bumping the version dependency to `0.4` shouldn't be a big deal for this new version.

slundberg:
I looked into making everything work on 0.3 but it ended up taking more time than I had, so I bumped the Julia version to 0.4.

slundberg:
Thanks!

PR #5124, opened 2016-05-07
Register RecipesBase: v0.0.1

tbreloff:
RecipesBase is a super-lightweight method of defining visualization methods of custom data types.  The core use is to enable package developers to bundle "plotting routines" in their packages, without conditional dependencies or including bloated graphics packages.  It is intended to be used in conjunction with Plots, but could be used for other things as well.  See [the readme](https://github.com/JuliaPlots/RecipesBase.jl) for more info.


Evizero:
This could be a great step towards mimicking R's ability to plot almost anything, while continuing to embrace the philosophy that backend backages should have next to no overhead or dependencies!

tbreloff:
Thanks!

PR #5132, opened 2016-05-09
Tag Combinatorics v0.2.2, MLBase v0.5.3

jiahao:
Combinatorics v0.2.2 has the methods from `Base` expunged (leaving them only on the v0.3.x branch which is Julia 0.5-only)

cc: @spencerlyon2 

---

MLBase v0.5.3 is tagged to propagate fix for deprecation warnings on Julia 0.4 introduced in JuliaStats/MLBase.jl#25

Ref: JuliaStats/MLBase.jl#27

### DIFF
--- a/Combinatorics/versions/0.2.2/requires
+++ b/Combinatorics/versions/0.2.2/requires
@@ -1,0 +1,3 @@
+julia 0.4
+Polynomials
+Iterators
--- a/Combinatorics/versions/0.2.2/sha1
+++ b/Combinatorics/versions/0.2.2/sha1
@@ -1,0 +1,1 @@
+92f47e4647ebd6eb5feec06f2ea46b430a1e7c66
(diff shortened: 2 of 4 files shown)


spencerlyon2:
:+1: thanks for tagging!

tkelman:
if Compat is still used https://github.com/JuliaLang/Combinatorics.jl/blob/julia-0.4/src/Combinatorics.jl#L3, it should still be required

jpfairbanks:
The MLBase.jl commit has Compat in the requires file and looks ready to go.

tkelman:
If that were separate, or Combinatorics gets fixed here, this would be okay to merge.

jpfairbanks:
Is there a way for someone other than @jiahao to fix it? I don't have the permissions on either of these packages.

tkelman:
You can propose a metadata pr with just the mlbase part of the change

tkelman:
MLBase was done in https://github.com/JuliaLang/METADATA.jl/pull/5433. Combinatorics needs re-doing for Compat issue.

PR #5139, opened 2016-05-10
Register BlockArrays: v0.0.1

KristofferC:


tkelman:
Great, maybe we can get rid of syntactic concatenation from base now.

PR #5145, opened 2016-05-11
Register DiffEq

ChrisRackauckas:
DiffEq.jl is a package for solving various differential equations, including stochastic differential equations and partial differential equations. It differs from ODE.jl in that it looks to provide efficient and parallel code even if in some cases the functionality may not be as general. Currently there are finite element implementations of the Heat and Poisson equations, with a bunch of functionality for mesh generation, convergence testing, plotting, and animating.

### DIFF
--- a/DiffEq/url
+++ b/DiffEq/url
@@ -1,0 +1,1 @@
+git://github.com/ChrisRackauckas/DiffEq.git


tkelman:
Needs a tagged version to be installable with Pkg.add.

ChrisRackauckas:
There, I just tagged the initial version.

tkelman:
Metadata tag, which should follow semver.org. Try `Pkg.tag("DiffEq"); Pkg.publish()`

ChrisRackauckas:
That failed (presumably because Windows sucks). I think I manually pushed the tag. Check v0.0.1.

dpsanders:
Could you say something about how this relates to the various previous packages for SDEs and PDEs. 

Also, I couldn't find any usage instructions. 

Finally, how about DifferentialEquations.jl as a repository name?

tkelman:
See http://docs.julialang.org/en/release-0.4/manual/packages/#man-manual-publish

ChrisRackauckas:
DifferentialEquations.jl is fine as a repository name. As for other packages:

- Brownian.jl looks like it focuses on simulating core stochastic processes (Brownian motion, Brownian bridge, Geometric Brownian motion, etc). While I may use this in my package, there's a clear difference from my goal of solving general equations (and this doesn't have all of the PDE tools)
- SDE.jl looks dead and doesn't have any high order methods. For SDEs I will be moving my high strong/weak order methods based off of recent literature (this is one aspect of my PhD, most of these codes already written). And this package looks dead.
- StochasticEuler.jl has the same differences as SDE.jl. 
- JuliaFEM.jl is a completely different monster than my FEM code and looks to implement a large finite element analysis package for many applications. At its scale and current pace, it will not be released for awhile, while mine is already usable for linear equations. Mine focuses on a small robust set of equations and advanced methods for them, and will be used as part of the stochastic PDE solvers as well.
- PDESolver.jl, it's another FEM solver focusing on larger systems but has non-trivial installation requirements and I don't see how I could put stochastic things to it.

Another difference to mention is that my package is focusing on high performance and parallelism. I have some macros I am testing out so that way all of the internal loops can have options to switch on things like `@threads`, `@acc`, and `@parallel`. Thus also "low-memory versions" of the code which do not save every timestep etc. are important for my applications and so I am building my package from the ground up for these purposes. The other packages follow a format more of ODE.jl which tends to save/give more output to users at the expense of performance.

Also, a related point is I don't see any finite difference or spectral solvers. These are vastly more efficient when the geometry allows for their use (and no stiffness matrix assembly is required). Nor do I see SPDE solvers.

Usage instructions will be up once I find out how to do the documentation. I already posted on julia-users for help with finding out how to do documentation.

PR #5148, opened 2016-05-11
Update JuliaOpt solvers for MathProgBase 0.5

mlubin:
CC @IainNZ @joehuchette @chriscoey @tkelman @madeleineudell

Commercial solvers still need to be bumped, but this will let us test updates to JuMP and Convex.jl on CI.


tkelman:
If these don't use SDPs or updatemodel, can't they be 0.4 to 0.6?

mlubin:
I've removed ``updatemodel!`` from Clp, Cbc, and GLPKMathProgInterface. Didn't think it was worth the mental overhead to try to maintain support for 0.4 partially across solvers.

tkelman:
Things are easily going to get held back and it can be really tricky to figure out why that happens.

mlubin:
That seems inevitable anyway, we just need to be quick in tagging updates for the rest of the solvers.

tkelman:
Not just solvers, all the extensions too. I had JuMPeR or something holding back most of JuliaOpt a couple weeks ago.

mlubin:
Sure, but what are we supposed to do about that? Seems more like an informational issue about letting users know why they don't have the latest version of X package.

tkelman:
Be fast and wide-ranging in updating everything JuliaOpt-related. At least the registered ones.

IainNZ:
JuMPeR holding things back is an issue with Julia not being clear about any constraints in place on versions, not with JuMPeR or JuliaOpt. JuMPeR is doing the responsible thing, and is going to be updated this week anyway.

tkelman:
Yeah, being conservative and careful about dependency versions leads to correct behavior that's just totally un-discoverable. Something like a `Pkg[Dev].why` would be a good idea some day.

PR #5152, opened 2016-05-11
Tag Compat 0.7.16, make 0.7.15 tag 0.5-only

tkelman:
The previous tag caused some breakage on 0.4. cc @timholy @yuyichao


yuyichao:
`ZMQ.jl` needs a new tag of `Compat.jl` Can you include it too? Or if you want to merge this and I can tag one after that.

tkelman:
May as well do another, don't feel like replacing the tag.

yuyichao:
Done https://github.com/JuliaLang/METADATA.jl/pull/5153

PR #5155, opened 2016-05-12
Tag StatsBase v0.8.1

andreasnoack:


tkelman:
Why did you delete the git tag?

andreasnoack:
I was using `Pkg.publish()` and the two `METADATA` commits (`StatsFuns` and `StatsBase`) went on the same branch and therefore into the same PR. Therefore, I had to make a `METADATA` branch and set it to `origin/metadata-v2` before retagging.

tkelman:
That's fine, I'm referring to the git tag that should show up at https://github.com/JuliaStats/StatsBase.jl/releases

andreasnoack:
That was the tag made by `Pkg.publish` which I removed when I wanted to retag. I'll push the tag again.

PR #5156, opened 2016-05-12
upper bound on MathProgBase dependencies for Convex.jl

mlubin:
This will keep things working until Convex.jl is updated for MathProgBase 0.5. Only the most recent tag (0.2.0) had a safe upper bound on the dependency.
CC @madeleineudell @tkelman


tkelman:
good call

PR #5160, opened 2016-05-13
Register Rotations: v0.0.1, v0.1.0

awbsmith:
Pull request to add a rotations package for handling 3D rotation parametrizations and conversions between them.

Parametrizations included are:
rotation matrices
quaternions
angle axis representations
stereographic quaternion projections (a 3 parameter representation of a unit quaternion)
Euler angles (Tait-Bryan ordering e.g. XYZ)
proper Euler angles (proper euler ordering e.g. XYX)

### DIFF
--- a/Rotations/url
+++ b/Rotations/url
@@ -1,0 +1,1 @@
+git://github.com/FugroRoames/Rotations.jl.git
--- a/Rotations/versions/0.0.1/requires
+++ b/Rotations/versions/0.0.1/requires
@@ -1,0 +1,3 @@
+julia 0.4
+FixedSizeArrays
+Quaternions
--- a/Rotations/versions/0.0.1/sha1
+++ b/Rotations/versions/0.0.1/sha1
@@ -1,0 +1,1 @@
+fa90fd4028d56934ecb16cddcaaa314c6480cca9
--- a/Rotations/versions/0.1.0/requires
+++ b/Rotations/versions/0.1.0/requires
@@ -1,0 +1,3 @@
+julia 0.4
+FixedSizeArrays
+Quaternions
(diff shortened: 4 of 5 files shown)


timholy:
See also
https://github.com/timholy/AffineTransforms.jl
https://github.com/JuliaGeometry/Quaternions.jl

Any chance to combine efforts?

awbsmith:
I'm always happy to combine efforts. If the Julia Geometry people like the
package I'd be happy to move / transfer ownership or whatever works.

The Quaternions used in this package are imported from Quaternions.jl (for
better or for worse)

On AffineTransforms, there will be an issue where I'm using
FixedSizeArrays.Mat{3,3,T} for rotation matrices which is a bit scary and
consistent with the rest of AffineTransforms.  Do you think the author of
AffineTransforms would be interested in integrating them?

tkelman:

awbsmith:
Yes for release Julia, no for nightly Julia

awbsmith:
A note that a dependency (FixedSizeArrays.jl) isn't passing nighties

tkelman:
You can set nightly as an allowed failure in the meantime. That way it'll run so you can see the results, but it won't make the status red.

awbsmith:
Thats great thanks.  I added the build image to the readme while I was at it

awbsmith:
Thanks

timholy:
Ah, just noticed this, sorry for not responding earlier.

> Do you think the author of AffineTransforms would be interested in integrating them?

As the author of AffineTransforms, I can answer that question! I think integrating would be good. AffineTransforms was written before FixedSizeArrays existed, so it used other tricks. Assuming the two can be combined, I'd rather use FixedSizeArrays.

PR #5161, opened 2016-05-13
Register DifferentialEquations: v0.0.1

ChrisRackauckas:
See #5145.


tkelman:
The tagged sha still had the older name: https://github.com/ChrisRackauckas/DifferentialEquations.jl/tree/4aeb5e093d8b65aa4d2ef08d8712be7c98892c67

You should include your test files in `test/runtests.jl` so `Pkg.test("DifferentialEquations")` works. Then ideally you could enable Travis CI which would automatically run your tests for you on every commit and pull request.

ChrisRackauckas:
Should I just delete the tags and re-publish?

I was wondering about the tests. Some of the tests (solving stochastic PDEs) take quite a bit of computational power and time. Is that okay? (I've never done this Travis Cl thing before)

tkelman:
Best way to find out the answer to "is that okay" is to try it and see. It's a free service, the default timeouts are something like 50 minutes of VM time per job.

PR #5162, opened 2016-05-13
Register DifferentialEquations: v0.0.1

ChrisRackauckas:
See #5161 and #5145. Included the tests (again, some are long and computationally intensive. Is that okay?)


TotalVerb:
As for the tests, Travis CI has a limit to how long the build is allowed to take. It's okay to have long tests as long as Travis doesn't terminate them! Very few projects actually exceed the limit. You can turn [Travis CI](https://travis-ci.org/) on for your project, then make a small commit and see if the tests finish in time.

If Travis is terminating the tests for excessive runtime, then it might be worth keeping those tests to run locally but not include them in runtests, and instead including some simpler cases.

tkelman:
After it gets registered the tests will also run on PackageEvaluator which has a slightly stricter time limit than Travis, but usually if you fit comfortably within the time limit on Travis then it shouldn't be an issue.

ChrisRackauckas:
Hey, I scaled down the tests to something that should fit nicely in the testing time. However, I had the build fail because it could not build matplotlib. Do I need to make travis.ci install matplotlib?

ChrisRackauckas:
I have tests working and not taking too long. However, they are not working on Travis.CI due to it stalling, with the last printouts being `WARNING: could not attach metadata for @simd loop.` even though I don't have `@simd` in my repository at all. I think it may be due Atom.jl's `@progress`. 

Also, matplotlib was having build issues with it, so I just worked my tests around that.

ChrisRackauckas:
I have updated it and my tests are passing. I have made a documentation with Documenter.jl and it should be building via Travis CI very soon. I have added code coverage to the testing.

ChrisRackauckas:
Travis builds pass and it builds the documentation. However, I don't understand the code coverage percentage and why it's zero: most of the code should be contained as a part of some test.

tkelman:
Looks like it's failing on OSX Travis?

ChrisRackauckas:
It's because OSX doesn't have Imagemagick so it can use Plots.jl's `@gif` to make an animation.

tkelman:
Doesn't ImageMagick.jl work on osx? If not, maybe QuartzImageIO?

TotalVerb:
Plots seems to have the `QuartzImageIO` [dependency](https://github.com/tbreloff/Plots.jl/blob/17a90e3b001b4b385fc43105a7d63e31f67d09e3/test/REQUIRE) on OSX. Maybe that's not being built correctly?

ChrisRackauckas:
ImageMagick's problems a Plots.jl issue. I wrapped the two animation tests with `@linux` and let @tbreloff know about the animation issues off Linux.

Code coverage is working now, the tests are passing, and the documentation is building/hosting. The only minor issue is that LaTeX is not displaying in the documentation.

tkelman:
You may want to tag a newer version, but that can be a separate PR.

PR #5164, opened 2016-05-13
Tag BlockArrays v0.0.2

KristofferC:


tkelman:
That tagged commit does not exist on master? Push the git tag so it shows up at https://github.com/KristofferC/BlockArrays.jl/releases

KristofferC:
Sorry, thanks for being so observent. I was on the wrong branch when I tagged it. Should be updated now.

PR #5172, opened 2016-05-14
Register SuiteSparse

KristofferC:
In preparation for https://github.com/JuliaLang/julia/pull/16070


tkelman:
We should add a short summary description for the sake of pkg.julialang.org

KristofferC:
Could you add that?

tkelman:
I do have commit access, but not sure how exactly we should describe it.

PR #5179, opened 2016-05-14
Tag Conda v0.2.0

Luthaf:
This add a new way to deal with conda channels, as well as some [0.5 compatibility](https://github.com/Luthaf/Conda.jl/pull/31) changes.

This contains BREAKING CHANGES, thus the version bump here. In METADATA, these packages uses Conda without a version specification:

EEG (@codles); ExcelReaders (@davidanthoff); Mads (@montyvesselinov and @omalled); NetCDF (@meggart); PyPlot (@stevengj); ScikitLearn (@cstjean); SerialPorts (@sjkelly); SymPy (@jverzani)

However none of these packages seems to uses the previous channels capacities. I'm pinging the maintainers just in case, they may want to stick to a specific version of Conda, to prevent future breakage.


tkelman:
The thoroughness in notifying your dependents is much appreciated!

davidanthoff:
If I'm not using the channels thing (I don't even know what that is), things should be ok?

codles:
Thanks for the heads up

Luthaf:
> If I'm not using the channels thing (I don't even know what that is), things should be ok?

Yes! But you may want to specify a version in your REQUIRE file to prevent breakage.

davidanthoff:
Hm, should I really add a specific version in REQUIRE? I always thought best practice is to only add that in case of a known incompatibility. Otherwise we might and up in a situation where different packages require different versions of the same third package, and then users couldn't use such packages simultaneously. That seems undesirable, especially if this were to happen in a situation where there is actually no incompatibility.

PR #5180, opened 2016-05-14
Tag DifferentialEquations.jl v0.0.2

ChrisRackauckas:
Includes tests, Travis.CI integration, documentation (linked to Travis.CI), less requirements (dropped Atom requirement, added functions to check for availability), and code coverage statistics. Many code internals were also cleaned up and examples are more detailed (the tests).

Only two remaining issues are properly rendering the LaTeX in the documentation and making one of the stochastic tests more deterministic (one of them tests that it's close to the variance it should be. This means it fails every once in awhile (~1/20 runs?). I need to change this to take a random seed). However, both of these issues are internal issues which will be fixed in further commits that don't necessarily need to be in a release.


ChrisRackauckas:
Windows build failing because stevengj/PyPlot.jl#213

tkelman:
How much of your code is derived from iFEM? https://bitbucket.org/ifem/ifem looks like it's GPL licensed, which means any direct ports would also need to be GPL licensed. You can do a clean-room reimplementation based on documentation or papers, but referring directly to the code and translating it is questionable.

ChrisRackauckas:
Some internal parts dealing with meshing, boundary, and quadrature are derived from iFEM. The issue is that this started as a project for a course and I got an okay from the instructor to redevelop it Julia. [So I used his notes](http://www.math.uci.edu/~chenlong/226/Ch3FEMCode.pdf) in order to develop the codes (sometimes his notes are exactly iFEM), and other times looked at how iFEM changed things around to make it more robust. However, [I did a lot of tweaking like documented here](http://www.stochasticlifestyle.com/julia-ifem2/) for speed, and I haven't even gotten to putting into a "package form" a lot of other changes yet (multigrid implementations, etc.).

I actually don't know much about licensing. What's the difference between the MIT and GPL license for this case? I just chose MIT since I always used MIT.

tkelman:
MIT is more permissive, it has very few restrictions on use or redistribution. GPL requires that derived works also use the GPL license. If you have explicit permission from the author (is he the only author?) to relicense code derived from iFEM in your Julia package under the MIT license then you should be okay, but probably a good idea to make a note of that.

ChrisRackauckas:
Hi, I looked more into this. He is not the only other, but he is by far the main author with small contributions done by some PhD students. While some algorithms look similar, in many cases there are core differences in the execution due to either performance reasons or because it was derived from notes/ mathematical theory and not from source (which can be quite close in some cases, other cases not). 

The places that could be in question are the mesh tools and error calculation tools. For example, given his guidance from in lecture, to find boundary edge indices from a (node,elem) structure I developed the algorithm:

```julia
      totalEdge = sort(totalEdge,2)
      edgeMatrix = sparse(totalEdge[:,1],totalEdge[:,2],1)
      i,j = ind2sub(size(edgeMatrix),find(x->x==1,edgeMatrix))
      bdEdge = [i';j']'
      isBdNode = falses(N)
      isBdNode[bdEdge] = true
      bdNode = find(isBdNode)
```

as opposed to the iFEM algorithm:

```matlab
ne = nv; % number of edges in one element
Neall = length(allEdge);
matlabversion = version;
if str2double(matlabversion(end-5:end-2)) > 2012
    [edge, i2, j] = unique(allEdge,'rows','legacy'); %#ok<ASGLU>
else
    [edge, i2, j] = unique(allEdge,'rows'); %#ok<ASGLU>
end
NT = size(elem,1);
i1(j(Neall:-1:1)) = Neall:-1:1; 
i1 = i1';
bdFlag = zeros(Neall,1,'uint8');
bdEdgeidx = i1(i1==i2);
```

So in many cases language and naming is similar (due to in class guidance and working from lecture notes), the core algorithm is clearly not a derivative work (honestly, I don't even exactly know how his code works here...). And most of the examples for the mesh tools are like this.

The case where the resulting code is more similar are the error tools. These are mostly translations, but I think it's clear that these codes do not meat the creativity requirement since they are straightforward implementations of standard quadrature rules [like those from this paper](http://w3.pppl.gov/m3d/reference/277540.pdf) and relations between vertices / faces for simple geometric objects, which are just linear combinations with specific constants. Citations were added.

Therefore I think it's quite safe to say the code is not a derivative work, though it is based off of his notes. Thus in the license I added a citation to Long Chen's original technical report as requested by him.

PR #5190, opened 2016-05-16
Register Divergences: v.0.0.1

gragusa:
`Divergences` is a Julia package that makes it easy to evaluate divergence measures between two vectors. The package allows calculating the *gradient*  and the diagonal of the *Hessian* of several divergences.


gragusa:
Not tagged version. Closing and re-opening.

PR #5191, opened 2016-05-16
Register Divergences: v0.0.3

gragusa:
`Divergences` is a Julia package that makes it easy to evaluate divergence measures between two vectors. The package allows calculating the *gradient*  and the diagonal of the *Hessian* of several divergences.


Divergences/versions/0.0.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.3
```
tkelman: this package uses `eachindex`, it doesn't support 0.3: https://travis-ci.org/gragusa/Divergences.jl/jobs/130739315

gragusa:
I retagged v0.0.3 with the right Julia version requirement and manually update the sha-1 of this PR. Is it fine now?

tkelman:
you need to change the requires file here in metadata too

gragusa:
Done. (Hopefully, it is enough). 
Don't worry about closing. Github buttons are what they are!

PR #5196, opened 2016-05-16
Register PyLogging: v0.0.1

colinfang:
`PyLogging` is a simple Julia wrapper for Python `logging` module. It is a convenient tool for those who work on applications with mixed languages and want to use a unified logging system.


tkelman:
Some tests would be a good idea going forward.

PR #5205, opened 2016-05-17
Register CoordinateTransformations

andyferris:
*CoordinateTransformations* is a package for managing transformations of points to different reference frames. Transformations can be inverted, composed and or even differentiated (with respect to either the input coordinate or the parameters defining the transformation).

The package primarily provides an interface for `AbstractTransformation`s (and compositions, inversions, derivatives, etc) but it also provides some basic transformations like rotations of 2D and 3D points, Cartesian, polar, spherical and cylindrical coordinate systems, etc. We are currently building other packages on top of it, notably our re-write of [Geodesy.jl](https://github.com/JuliaGeo/Geodesy.jl/pull/15). But it could be used for other things, like 2D/3D graphics, physics, etc, and so hopefully it is useful to the general community.

Finally, we (@fugroroames) are open to moving this package (and also Rotations.jl) to @JuliaGeometry in the future, if there is interest.


timholy:
To me this hits the sweet spot in terms of generality-yet-focused. I like the idea of merging AffineTransforms into this, if it makes sense in general.

One thing that AffineTransforms does (not documented) is support coordinate transformations on "functions" specified by arrays, e.g., image rotation. Is that within the scope of `CoordinateTransformations`?

andyferris:
Thanks @timholy! Yes, the package could **definitely** use an `AffineTransformation`, and you are more than welcome to contribute or make any design suggestions. 

(Though, keep in mind that at this stage the compositions are powerful enough that `Translation ∘ Rotation` (or a general linear transformation) is pretty easy to use - in my work I've been defining functions that create composed transformations instead of adding more `AbstractTransformation` types.)

Regarding other kinds of transformations, absolutely, they are meant to be used as a part of the interface... it's definitely my intention to use it for more general kinds of transformations (you could even use it for generic (i.e. non-geometric) programming as a super-charged kind of `convert`, if you wanted). The question becomes - should we create one big package with lots of features, or one package defining the interface and have *ImageTransformations.jl* define pixel-based `AbstractTransformations`? I'm reasonably happy for both, but see for instance what we are doing with [Geodesy.jl](https://github.com/JuliaGeo/Geodesy.jl/pull/15), where another more-specialist package is based on this one.

In either case, one possible design for a pixel-based rotation (for instance) could be:

```julia
immutable ImageTransformation{T <: AbstractTransformation{Point{2},Point{2}} <: AbstractTransformation{Image, Image}
   t::T
end
```
where the point-transformation `t` gets called to transform points as a part of the process of transforming the image and `Image` is some type which contains the image data as a matrix and the physical dimensions (assuming that they're necessary/make sense, e.g. scale-factor between x and y).

andyferris:
Oh, and in case anyone is looking, the one package test which is failing is waiting upstream, for FixedSizeArrays.jl to tag (the `isapprox` bug is fixed on FixedSizeArrays master).

PR #5206, opened 2016-05-17
0.5 fixes for GLVisualize

SimonDanisch:


tkelman:
GLAbstraction should have moved to a new minor version instead of 0.0.6. Best to not continue tagging more versions as 0.0.x.

ModernGL/versions/0.0.8/requires
```diff
@@ -0,0 +1 @@
+julia 0.3
```
tkelman: should incorporate https://github.com/JuliaGL/ModernGL.jl/pull/28

SimonDanisch:
I think I'll nuke this PR and try a second time with tags at `0.1.0`. It makes sense, since the next version will add a lot of new features which might not be as stable as this tag.

tkelman:
Usually big feature additions are a reason worth modifying minor version (as is dropping support for an old version of Julia). Though as long as your leading version is 0, technically semver says you can do whatever you want.

SimonDanisch:
Okay, I updated ModernGL and GeometryTypes with the newest changes and I'll fix the release numbers with the next wave of changes ;)

PR #5209, opened 2016-05-17
Spark.jl + Kafka.jl

dfdx:
#### Spark.jl

Apache Spark is the most widely used computational engine for Hadoop. Recently it has got a lot of attention from statisticians and data scientists, thanks to bindings in Python and R. This packages aims to make similar bindings for Julia. 

#### Kafka.jl

Apache Kafka is a persistent message queue widely used in big data processing. Kafka.jl provides pure Julia implementation of Kafka protocol that should work with all latest version of servers (0.8.x-0.10.x).


tkelman:
Any chance of being able to test these on Travis CI?

dfdx:
Since these are wrappers/protocol implementations, really testing these packages is somewhat tricky task. E.g. at my current job to test Kafka-related stuff we have to set up and start a docker image with this service, which may be an uneasy task for Travis. Currently I'm considering different options to make it easier, but it may take some time and I'd like to make the packages searchable for other people as early as possible. 

So is CI integration needed for some technical reasons (like showing the status on pkg.julialang.org) or is it just a rule of etiquette? If first, I can make a simple Travis config to test the most basic functions, but if second, I'd like to come up with some more solid test suite in the next version.

tkelman:
Just best practice. Understandable that these could be tricky.

PR #5210, opened 2016-05-17
Register package RecurrenceAnalysis

heliosdrm:


tkelman:
A readme and some unit tests would be good to have. When you add tests (or before), consider enabling Travis CI which will automatically run your tests on every commit.

heliosdrm:
Done. Thank you. I disabled running the test in Travis for the nightly version (0.5), because it fails persistently due to differences in some basic methods (e.g. maximum, /, etc.). It is passing the release version.

tkelman:
It's better to make nightly an allowed failure rather than disable it completely. That way it'll run but not make your status red if it fails.

PR #5219, opened 2016-05-19
Tag PyCall v1.6.2

cstjean:
Another fix for several users. 

@stevengj


cstjean:
It seems that the latest PyCall is causing issues for a number of people (a new one popped up right after I submitted this PR). A lot of it is my fault, so mea culpa there, but process-wise, is it encouraged to tag a patch as each new bug is fixed, or is it better to hoard them and only tag when the dust settles? Is there a cost to these METADATA PRs?

stevengj:
In general, critical bug fixes are nice to tag quickly (within a day); everything else I usually wait a few days at least.

PR #5226, opened 2016-05-20
Measurements.jl:  Uncertainty propagation library

giordano:


tkelman:
Measurements is a slightly generic name, I might have chosen something a little more specific like MeasurementUncertainty myself, but this looks good.

giordano:
Thanks for your comment.  I followed the naming convention suggested in the [manual](http://docs.julialang.org/en/release-0.4/manual/packages/?highlight=plural#guidelines-for-naming-a-package): the new type defined by the package is `Measurement` and in my mind a measurement always has an uncertainty attached, thus making a name like `MeasurementUncertainty` a bit redundant.  I also thought about using `Uncertainty` for the type name, but I felt that `Measurement` was more descriptive (the objects are really measurements with uncertainties, not just uncertainties).

PR #5229, opened 2016-05-20
Pull request/42b10f85

russolsen:
Note that Transit is a cross language technology and one convention with Transit is that the libraries supporting it adopt the version number of the version of Transit it supports. That's why we are trying to set the version number to 0.8.1.


kmsquire:
What happens if you need to make an update to the bindings?  In Julia, you're not allowed to change the sha of an existing version (except maybe in extenuating circumstances), so you would have to tag a new version.

tkelman:
I agree that trying to tie the version numbers together exactly is a bit odd, unless you have differing levels of granularity between upstream and downstream. If upstream only ever tags releases as x.y.0 then bindings would be free to make multiply x.y.z releases corresponding to the same upstream library version. Or if bindings had non-semver 4-digit numbering, but I don't think Julia packaging would handle that very well.

If you enabled Travis CI would your tests pass there?

benkamphaus:
Hi @tkelman and @kmsquire -- as you suspect, there are different levels of granularity. Transit follows the major.minor.revision scheme and in this case, major.minor implies cross-language compatibility and the revision number is the per kit bug fixes, etc. and is not pinned between kits. Any changes to the [Transit format](https://github.com/cognitect/transit-format) specification will result in a different either major.minor or minor release. All bug fixes or additional commits within a transit format specification implementation for a particular language lib increment the revision number.

We can hook up project specific Travis CI if it's necessary for release. The within language tests (i.e. can Julia encode and decode values consistently itself) are less significant than the tests that require the transit-format test suite (which tests sending values over the wire between languages). We've [tested](https://github.com/russolsen/Transit.jl/tree/master/test) in multiple environments already to verify this functionality.

Let me know what steps or clarifications are required and we'll follow through and get back to you. Thanks for the review.

tkelman:
Travis CI isn't required, but it's highly recommended. Just a verification that your package installs correctly and can pass its tests on the supported versions of Julia.

russolsen:
Thanks for your help!

PR #5230, opened 2016-05-20
Register StandardizedMatrices

joshday:
https://github.com/joshday/StandardizedMatrices.jl


tkelman:
You'll need to tag a version for anyone to be able to install the package using `Pkg.add`

joshday:
I always forget something...thanks.

PR #5232, opened 2016-05-20
Tag PyAMG v0.0.3

cortner:
0.0.3 update to PyAMG

PR #5243, opened 2016-05-22
Tag PyAMG v0.0.4

cortner:
this now uses `pyimport_conda` instead of `@pyimport` so that the python dependencies are automatically installed.

cf. [this discussion](https://github.com/JuliaLang/METADATA.jl/pull/5232)


ViralBShah:
Should the python dependencies bit go into the developer/Pkg docs?

tkelman:
Should probably go in the PyCall docs.

ViralBShah:
I would think that it should certainly be linked into the pkg/dev docs, even though PyCall docs are perhaps the right place.

tkelman:
We could link to PyCall from the package documentation. People who right packages that call into Python code have to find out about it from somewhere. Since everyone who does this has to look at PyCall, the dealing with dependencies part should probably be documented there. It's also likely to change over time more rapidly than Julia's releases.

cortner:
Another suggestion: at the moment I do this in my `runtests.jl`:
```
print("check wether on travis ... ")
if haskey(ENV, "TRAVIS")
   println("yes: fixing PyCall installation")
   ENV["PYTHON"] = ""
   Pkg.build("PyCall")
else
   println("no")
end
```
this is because travis comes with a python distribution that does worth with `pyimport_conda`. No big deal, but if everybody who calls into Python needs it, then not great. Is it possible to have travis for Julia set up by default to use miniconda?

tkelman:
Good question. (I assume you mean "does[n't] work with `pyimport_conda`?) We can make pull requests to adjust things for `language: julia` on Travis, I guess the question is what downsides setting it unconditionally might have. Generally I'd recommend doing changes like that in `.travis.yml` instead of in `runtests.jl`, but the fact that you're at least checking for the `TRAVIS` environment variable means your way should be safe too.

cortner:
adding this (and how) to the manual or PyCall would be great too.

PR #5250, opened 2016-05-23
Register POMDPs.jl

etotheipluspi:
cc @mykelk


tkelman:
Acronym package names are generally discouraged, would be more discoverable for those not familiar eith the acronym if it were spelled out a bit.

Version 0.2.0 and 0.2.1 should not use the same minor version since the former says it supports julia 0.3  while the latter does not.

tbreloff:
@tkelman normally I'd agree but I bet POMDP is a more recognizable (and searchable) name than if you spelled out PartiallyObservableMarkovDecisionProcesses.jl.  Unless, of course, this package implements something else, in which case: please change the name!

mykelk:
Yep, it definitely implements partially observable Markov decision processes. I normally try to avoid acronyms, but this is one of those cases where you pretty much need to use it. The package that someone else implemented on (fully observable) Markov decision processes is registered as MDPs.jl.

kmsquire:
Also agree in this case that POMDPs.jl is the best name for someone looking for these in Julia.  I can't imagine these being confused with anything else.

tkelman:
Would be good to at least spell out the acronym in the description field that will appear on pkg.julialang.org, since "API for solving POMDPs in Julia" isn't much more information than the package name alone.

mykelk:
@tkelman good point, this has been addressed.

jiahao:
Congratulations @mykelk @etotheipluspi, you have registered the 1000th Julia package!

tkelman:
Depends how you count.

PR #5254, opened 2016-05-24
Temporal minor update

dysonance:
# Changes
- Added logical operations and `all` and `any` functions.
- Added logic for filling missing values (forward fill, backfill, and linear interpolation)

# Tentative Roadmap
- Make it faster: some operations take annoyingly long even after JIT warms up
- Add plotting? Problem is I don't want to add unnecessary dependencies and while `Plots.jl` is a baller package I don't know how whether best to make it interface with that, or to pick another package that more specifically suits the needs
- Data fetching? Some easy way to download data either from Quandl or Yahoo Finance, but that I fear may be a smidge beyond the scope

As always, let me know if any major issues with the PR :)


tkelman:
your 0.0.3 git tag isn't consistent with this 0.0.3 metadata tag, but it's not too big an issue

PR #5258, opened 2016-05-25
Tag Pardiso v0.1.0 [https://github.com/KristofferC/Pardiso.jl]

KristofferC:


tkelman:
Push the git tag so it shows up at https://github.com/JuliaSparse/Pardiso.jl/releases?

I think someone had been experimenting with installing and testing against MKL on Travis, but dunno where that left off.

KristofferC:
Done.

I could mail the Pardiso people and ask if it is ok If C.I downloads the library and then I have my license as a hidden environment variable that I then create my license file from. You think that is reasonable?

tkelman:
Ah that would be another interesting option. Yeah probably safer to ask for permission before trying that.

KristofferC:
One problem is that Paridso licenses are tied to the userid of the logged in user..

PR #5265, opened 2016-05-26
Tag DataFrames v0.7.3 [https://github.com/JuliaStats/DataFrames.jl]

nalimilan:
Fixes regressions introduced when porting to String.


DataFrames/versions/0.7.3/sha1
```diff
@@ -0,0 +1 @@
+15f4e95103d76742a0d13b1e10a7d91ee4659260
```
tkelman: you probably want to tag the latest commit on master, not this commit from the branch of https://github.com/JuliaStats/DataFrames.jl/pull/968 that won't have other recent changes

nalimilan: Rah, I always make mistake when tagging a package. Thanks for spotting it!

PR #5281, opened 2016-05-30
Register CheckImports, EllipsisNotation, and ChunkedArrays

ChrisRackauckas:
This is a series of utility packages.

1. CheckImports is a small package which has functions for finding out which packages a user has imported. An example of its usage is:

```
isimported("Atom") ? Main.Atom.progress(i/femMesh.numIters) : nothing 
```

which turns on the progress bar in Atom only if the user is using Juno/Atom. While a smaller package, I found the solution to be nontrivial enough to share it. Personally, I think this kind of functionality should be added to base, but this will do for now.

2.  EllipsisNotation implements the ellipsis notation requested in JuliaLang/julia#5405. It builds off of what @mschauer had posted (acknowledging him) and implements getting and setting for edge indices on arrays up to dimension 10. A future update could generate all the combinations using a macro, but this is a start which I've already begun to use in other packages.

3. ChunkedArrays implements the type ChunkedArray. It's well known that generating random numbers in chunked is more efficient than one-by-one. The ChunkedArray implements this in a nifty wrapper such that the user only has to call `next(chunkrand)` to get the next random number, and in the background it is generating them in chunks and managing the access (and can generate the next chunks in parallel). The tests include benchmarks which show that using ChunkedArrays can improve RNG performance in a loop by up to 2x. Tests currently fail since it requires EllipsisNotation, though tests on a Windows and Linux computer both pass.


tkelman:
You shouldn't call `Pkg.add` in `runtests.jl` here https://github.com/ChrisRackauckas/CheckImports.jl/blob/6183e0d8db18c0ffce73e1c86f9e08f09d174446/test/runtests.jl#L7.

Use `test/REQUIRE` for test-only dependencies.

tkelman:
superseded by #5283

PR #5283, opened 2016-05-30
Register CheckImports, EllipsisNotation, and ChunkedArrays

ChrisRackauckas:
Fixed issues from #5281


mschauer:
Cool, I am also using this in similar contexts as you, for example this implementation of the Euler-Maruyama-Scheme which is indifferent whether the Path is a Matrix or an Vector of FixedVectors 

https://github.com/mschauer/Bridge.jl/blob/master/src/euler.jl (first function)

quinnj:
Can't you use
```julia
isdefined(Main, :Mod)
```
to check if a module `Mod` has been imported in the current session?

ChrisRackauckas:
@quinnj  I did not know about `isdefined`:-/. I just tried it and that worked. Thanks for pointing that out. I'll make a new pull request without that package, and change the readme on that package to tell people to use isdefined. A Google search of "how to check if Julia package is imported" shows up with no good results (while if I change package to module, I get one hit with the isdefined answer), so I'll leave it on Github to lead people to the right answer.

@mschauer Yeah, I'm using it for a bunch of ODE/SDE/PDE/SPDE algorithms to make them indifferent to the way the user input them (so you can drag and drop an ordinary differential solver to a PDE via method of lines, yet use intuitive matrix discretizations for finite difference solvers). So when I see people saying you can just loop with that, I look at my application and go "I can... but that would look pretty ugly..." haha. This is definitely a case where the vectorized code is legible and the fully devectorized code is just unmaintainable.

ChrisRackauckas:
Superseded by #5284

mschauer:
> But check out using the ChunkedArrays, it took off some runtime in my SDE algorithms.
Thanks for the pointer

quinnj:
Yeah, we should probably throw another line in the docs for `isdefined`, it's not always immediately obvious that you could check for a module definition in `Main` to see if it's been imported.

PR #5296, opened 2016-06-01
Tag InspectDR v0.0.3

ma-laforge:


dpsanders:
What is the meaning of the package name? I would not have guessed at all that it was for plotting.

tkelman:
Would it be possible to get your tests running on Travis via xvfb-run?

tbreloff:
My guess on the name: "inspect doctor" in that it's a package built for
interactivity. (But I agree it's cryptic)

ma-laforge:
@tkelman:
I tried running Travis on InspectDR, but it appears that "using Gtk" did not work on the test machine.  I don't know why.  It works fine on my own Windows & linux machines.

Question:
WIll this xvfb-run thing make Gtk work?  If so, can you point me to somewhere with more info?

tkelman:
Gtk.jl itself uses `xvfb-run`, any package that would need to create windows on Linux probably has to use it on headless servers with no real display. Here's an example of using it cross-platform: https://github.com/JunoLab/Blink.jl/blob/master/.travis.yml

ma-laforge:
@dpsanders:
I decided to pick a name that was unique to avoid a package name like "PlottingPackage".  Maybe it was a bit too unique...

The primary goal of the package is to investigate your data graphically, in an interactive way.

A secondary goal was to generate legible, pleasant-looking plots adequate for publications, and something called "Design Reviews".

To keep the name succinct (instead of InvestigateAndDesignReview), I picked "InspectDR".

@tbreloff: I noticed it looked like "Inspect doctor" later on... I found that kind of funny.

ma-laforge:
@tkelman:

Thanks for the help.  Got my green badge.

PR #5297, opened 2016-06-01
Tag Spectra v0.0.2

charlesll:
Updated Spectra.jl package, with planning new functions using new requirements. Docs have been added compared to 0.0.1.

### DIFF
--- a/Spectra/versions/0.0.2/requires
+++ b/Spectra/versions/0.0.2/requires
@@ -1,0 +1,10 @@
+julia 0.4
+IJulia
+PyPlot
+JuMP
+Ipopt
+StatsBase
+Dierckx
+Distributions
+JLD
+LsqFit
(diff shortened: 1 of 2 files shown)


tkelman:
It looks like you found more specific license information for https://github.com/charlesll/Spectra.jl/blob/master/Dependencies/gcvspl.f - you should mention in the package's top-level license file that it includes Fortran code that does not allow commercial use.

charlesll:
Exactly, I found more details for the licence of gcvspl. Following your comment (thanks for noticing that!), I included a comment in LICENCE.md saying "Spectra.jl includes Fortran code that does not allow commercial use. See header of gcvspl.f in ./Dependencies for further details."

Is that the good way to do so?

tkelman:
That'll do I think.

PR #5302, opened 2016-06-02
Tag Mosek v0.3.2

ulfworsoe:


tkelman:
apitest.jl is also using Compat here. and the trailing + on version numbers means that particular version is not allowed, is that what you intended?

ulfworsoe:
I have added ``Compat`` test requirement and fixed the version range.

tkelman:
Great, thanks. Can you change the sha1 file here to point to https://github.com/JuliaOpt/Mosek.jl/commit/5aa0e0b5c472bcc45c3c8f651147d597325adbaf ? Then I'll merge.

ulfworsoe:
Done and done. Thanks for being patient.

PR #5303, opened 2016-06-02
Tag GaussianProcesses v0.1.5

fairbrot:


tkelman:
ref https://github.com/STOR-i/GaussianProcesses.jl/pull/23 - if this still claims to support Julia 0.3 in `REQUIRE`, it should really be tested

fairbrot:
Thanks for pointing this out. We are no longer supporting 0.3. I'll fix this and create a new pull request.

tkelman:
If you drop support for 0.3, be sure to tag as a new package minor version, 0.2.0

PR #5306, opened 2016-06-02
Tag TypedTables v0.0.6

andyferris:
bugfix


tkelman:
be sure to push the commit https://github.com/FugroRoames/TypedTables.jl/commit/899016651079e50c56b54db6e090427212ed88d5 to master, right now it only exists as the tag

andyferris:
Haha I just realized that one. Pushed.

I wonder if we should make `PkgDev.publish()` perform a check for this one?

tkelman:
That would not be a bad idea.

nalimilan:
Indeed, it's very common to forget pushing a tag, commits, or both, and this makes your life harder @tkelman!

tkelman:
I'll write a metadata auto-review bot sooner or later.

Issue #5307, opened 2016-06-03
Try using a pull request template?


tkelman:
Ref https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/ - this would likely be easier than writing an automatic metadata review bot, but wouldn't actually verify anything. It would just serve as a reminder when registering / tagging packages of the checklist of recommended steps (most of which Pkg[Dev].tag and Pkg[Dev].publish should be doing for you, if they're working correctly). These are things I spend a few seconds checking manually when I review/merge a metadata PR. I could propose a list of items and wording to solicit feedback if people think it's a good idea.

quinnj:
+1j

gragusa:
👍

ararslan:
I think this will become less necessary as more people adopt attobot. I'm not sure how attobot would do with the PR body prefilled. Though it may still be useful to have for the folks who don't use attobot.

tkelman:
would be useful if there were a way of propagating a template through attobot somehow

DilumAluthge:
METADATA.jl is no longer maintained, so I am closing this issue.

If this issue is still relevant, please re-open a new issue on the Julia General registry: https://github.com/JuliaRegistries/General

PR #5318, opened 2016-06-04
Register ConstraintProgramming: v0.1.0

dpsanders:
#### Http URL for the repository of your proposed package:

- https://github.com/dpsanders/ConstraintProgramming.jl

#### Proposed name of the package:

- `ConstraintProgramming.jl`

#### One-line description:

- Rigorous calculation of the feasible region for a set of real-valued inequalities using "interval separators"

#### Alternative package names:

> Give alternative package names (applicable if your package is not a wrapper of another package).

- IntervalConstraintProgramming
- FeasibleRegion
- ContinuousConstraintSatisfaction
- IntervalSeparators

#### Documentation:

> You are required to provide adequate documentation.
 Confirm whether documentation is provided in `README.md` or give a link to the separate documentation

- In `README.md` and an example Jupyter notebook

and to be linked to a continuous-integration service. Provide the relevant information.

- Travis is enabled on the repository. [Tests pass on 0.4](https://travis-ci.org/dpsanders/ConstraintProgramming.jl/builds/135312735)

- [Sigma](https://github.com/zenna/Sigma.jl) contains similar ideas (bisection of domains),
    but not in terms of separators. I believe that it does not use correctly-rounded intervals, so the results are not completely rigorous. It does not allow specifying arbitrary constraint sets.

- [JuMP](https://github.com/JuliaOpt/JuMP.jl) provided inspiration for the API design


tkelman:
There are many approaches to constraint programming, I think this should really mention Interval in the package name since that's not the only way to do it.

I like the template for new registration info though.

dpsanders:
I thought that might be the (completely reasonable) response. Despite having watched N of these PRs go by, I have to now ask what the process is for changing the name of the package. This doesn't seem to be documented?

tkelman:
Rename the repo under git settings, git mv the main src/PackageName.jl file, then git grep and find replace. We should write that down somewhere though. It's can break things to rename existing repositories, but for things that aren't registered yet it should go fine.

TotalVerb:
(Slightly unrelated) Is this an official package registration template? Many of the PRs here don't seem to use it.

tkelman:
this looks like @dpsanders proposal for the new package registration side of https://github.com/JuliaLang/METADATA.jl/issues/5307. I'd also want to add a section for new version tags.

dpsanders:
Yes, this is a proposal for the minimum information that a PR should contain when wanting to register a new package. It was easier to do this in a concrete way, with an actual example, when registering a new package and thinking about what information I felt was missing from previous registration PRs, rather than doing something more abstract.

dpsanders:
Closed in favor of #5325 with new name.

PR #5330, opened 2016-06-06
Tag Optim v0.5.0

pkofod:
Need to fix all the Optim dependencies, as this is breaking a few things.


pkofod:
I hope this is what you meant @tkelman

tkelman:
Yep this is what I meant, but wow there are a lot more of these than I thought. Did Optim change dramatically enough to break all of them?

pkofod:
Hm, I thought you just wanted me to be "safe" and apply it to all of them. Had kind of hoped I didn't have to go through them all :) Notice, a long of the changes here are for old versions, and it's very difficult for me to say what will break or won't break in versions that are not the latest (I'm not going through them all :) ). I took a look at the latest versions:

* Celeste I think they copied most of what Optim-related code they're using directly into their code, but they do use something directly from Optim. Not exactly sure if it will break or not.
* DSGE I'm not sure why they require Optim, they seem to have copied some routines _from_ Optim, but they're not actually calling the package as far as I can see.
* DynamicDiscreteModels symbols should be changed to constructors https://github.com/BenConnault/DynamicDiscreteModels.jl/blob/777f7c1c3c04c6d8e188597d9551501c4aa43b3a/src/emalgorithm.jl#L41https://github.com/BenConnault/DynamicDiscreteModels.jl/blob/777f7c1c3c04c6d8e188597d9551501c4aa43b3a/src/statisticalmodels-stopgap.jl#L38
* GaussianProcesses symbols should be changed to constructors https://github.com/STOR-i/GaussianProcesses.jl/blob/5ede34476b6053e61dd24bc206a2eb25dbdb0c26/docs/optimize_1d.jl#L16 https://github.com/STOR-i/GaussianProcesses.jl/blob/5ede34476b6053e61dd24bc206a2eb25dbdb0c26/src/optimize.jl#L12 https://github.com/STOR-i/GaussianProcesses.jl/blob/012b807bd581f32f295a5cee2f116ebdfb71c23c/README.md#L156
* KernelDensity should be fine
* KernelEstimator will break here https://github.com/panlanfeng/KernelEstimator.jl/blob/945a408930aadb0b31675fbbaedccbc422ab0db4/src/bandwidth.jl#L96 and here https://github.com/panlanfeng/KernelEstimator.jl/blob/945a408930aadb0b31675fbbaedccbc422ab0db4/src/bandwidth.jl#L115 . The explicit field accesses should be replaced by `Optim.minimizer()`
* LinearResponseVariationalBayes will break here https://github.com/rgiordan/LinearResponseVariationalBayes.jl/blob/63c6584934411908e66dfe61c6fc0f78ad35a72b/src/SubModels.jl#L193-L203
* LowRankModels will break here https://github.com/madeleineudell/LowRankModels.jl/blob/3d271dd425f2dee309414d2267fe4c5dd5ea160e/src/losses.jl#L97 but it is an easy fix
* LsqFit won't break I think (uses levenberg_marquardt)
* Mads won't break I think (uses levenberg_marquardt)
* NLSolve will not break as they only use the linesearch functionality which is not affected by this tag
* PowerLaws will break here https://github.com/johnybx/PowerLaws.jl/blob/master/src/estimate_xmin.jl#L51
* QuantEcon an example will break, but Spencer Lyon is aware
* RegERMs will break here https://github.com/JuliaStats/RegERMs.jl/blob/53edcb08ab041a639fed16dc442fb9afd2dc8af5/src/solvers/lbfgs.jl#L24 but it is a simple change
* Regression is a very old version, Optim is no longer in REQUIRE
StochasticSearch is now on 0.2.2 and does no longer depend on Optim, but older versions (the ones changed here) will break
* TimeModels seems to be already broken, and most have been for some time, as we don't return results as in https://github.com/JuliaStats/TimeModels.jl/blob/906bf811d62544e43c755d5cf2d9aa2a52115216/src/GARCH.jl#L104, although https://github.com/JuliaStats/TimeModels.jl/blob/906bf811d62544e43c755d5cf2d9aa2a52115216/src/parameter_estimation.jl#L6 should be fine
* TrafficAssignment will break, but the required change is simple (symbol -> constructor)

tkelman:
It is indeed safer to be more conservative, but having any of these packages installed, if the authors don't update and/or test against the new Optim 0.5 and tag newr versions, will mean Optim will get held back to 0.4 or earlier. Correct behavior but a bit inconvenient if any of these packages are still in active use but not frequently tagging new versions.

pkofod:
I understand. I can make PRs based on the links provided here, but beyond that I'm not sure if the problem is on Optim's side or "infrequently tagged package"'s side?

tkelman:
Mostly the latter. Might be good to write down this list of packages somewhere just for reference's sake, so you can notice when users report `Pkg.status()` what will be responsible for Optim getting held back.

pkofod:
https://github.com/JuliaOpt/Optim.jl/issues/217

mlubin:
:+1:

pkofod:
Should I go back and remove the limits f I find out that some packages work as intended with v0.5? If not, I am kind of forcing everyone to tag a version soon.

tkelman:
In any case where you personally get annoyed that you can't have that package and Optim 0.5 installed at the same time, sure.

PR #5337, opened 2016-06-06
Tag Currencies v0.3.2 [https://github.com/TotalVerb/Currencies.jl]

TotalVerb:
- Enable precompilation
- Fix deprecations on current master
- Use the correct string type on 0.4
- Drop dependency on `MacroTools` (no functional change)

Due to my error, this tag isn't a descendent of the last tag (v0.3.1). Is that OK, or will it cause problems on `Pkg.update()`?


tkelman:
what was the mistake? it should be fine. consider enabling branch protection for master.

TotalVerb:
I forgot what happened; I ended up tagging a commit that is not in master. I turned on branch protection for master so this won't happen again.

PR #5338, opened 2016-06-07
Tag MatpowerCases v0.4.0

kersulis:
I've introduced a few new types to hold power system data, along with functions that return instances of them. Now it's easier to navigate data without referring to Matpower's documentation. This is another step towards emulating Matpower's power flow functionality in Julia.


tkelman:
you have test failures on Julia 0.3, which you should either fix (likely by using the Compat package) or not claim to support julia 0.3 in your REQUIRE. https://travis-ci.org/kersulis/MatpowerCases.jl/builds/135762989

kersulis:
Thanks for the catch. I dropped support for 0.3.

PR #5341, opened 2016-06-08
Fix Optim version for NLsolve

pkofod:
ref https://github.com/JuliaOpt/Optim.jl/issues/217

It only uses linesearch, so it's not affected. I've run the tests in the package as an extra check.

The KernelEstimators will run fine as well.

### DIFF
--- a/KernelEstimator/versions/0.1.7/requires
+++ b/KernelEstimator/versions/0.1.7/requires
@@ -1,6 +1,6 @@
 julia 0.3
 Distributions
-Optim 0.0 0.5
+Optim
 StatsBase
 Cubature
 Compat
--- a/NLsolve/versions/0.7.1/requires
+++ b/NLsolve/versions/0.7.1/requires
@@ -1,5 +1,5 @@
 julia 0.4
-Optim 0.0 0.5
+Optim
 Calculus
 ForwardDiff 0.1.4 0.2
 Distances


tkelman:
Awesome, thanks for checking these things out. Hopefully having a look at how Optim gets used by other packages is a bit useful as you work on making major improvements to it?

PkgEval picked up a possible issue with Mads.jl: https://gist.github.com/4609c24a5c7b8bb8222012c9564958c6

pkofod:
> Awesome, thanks for checking these things out. Hopefully having a look at how Optim gets used by other packages is a bit useful as you work on making major improvements to it?

Yes, quite helpful actually :)

> PkgEval picked up a possible issue with Mads.jl: https://gist.github.com/4609c24a5c7b8bb8222012c9564958c6

Seems like it could be a bug in Optim. Thanks for letting me know.

pkofod:
> PkgEval picked up a possible issue with Mads.jl: https://gist.github.com/4609c24a5c7b8bb8222012c9564958c6
> 
> Seems like it could be a bug in Optim. Thanks for letting me know.

Okay, I hadn't noticed, but Mads is calling unexported helper functions from Optim. I'll make a PR to fix their usage. Should I reinstate the limit?

tkelman:
> Should I reinstate the limit?

Yes, I think so, since they had been passing on June 6th - http://pkg.julialang.org/detail/Mads.html

pkofod:
I will, and then they can change it once https://github.com/madsjulia/Mads.jl/pull/2 is merged, and they tag a new version.

tkelman:
It should just do the right thing automatically when they make a new tag, since new tags populate the metadata require file from the current package version, not the previous tag's metadata version.

PR #5343, opened 2016-06-08
Tag PyPlot v2.2.0 [git://github.com/stevengj/PyPlot.jl.git]

stevengj:
Numerous small fixes since the last tagged version.  Now requires Julia 0.4.   Much better performance in IJulia with Matplotlib 1.5 (fixes stevengj/PyPlot.jl#197); I was able to get rid of most of the monkey-patching and things should hopefully be more robust.  More automatic installation via `pyimport_conda`.


tkelman:
any progress on https://github.com/stevengj/PyPlot.jl/pull/193 ?

stevengj:
Not at the moment; thanks for reminding me.

PR #5346, opened 2016-06-08
Register DC.jl

tawheeler:
I would like to register DC.jl -a Julia package for automagical DC.js linked charts in your IJulia notebook.

The base api is extremely simple - `dc(::DataFrame)`

Advanced users still have access to what is generated.

Preemptive Concern: I understand that package names should be as general as possible. This package produces visualizations using the DC.js package, hence its name. DC.js stands for D3 and Crossfilter, and D3Crossfilter.jl sounds like a horrible package name. Please let me know if you think of something better, or if naming it DC.jl is a non-issue.

### DIFF
--- a/DC/url
+++ b/DC/url
@@ -1,0 +1,1 @@
+git://github.com/tawheeler/DC.jl.git


ViralBShah:
While I generally do not like to interfere too much package names, I think DC.jl would do better with a name that is a bit more descriptive, even though the js file is called dc.js.

tawheeler:
How would something like CrossfilterCharts.jl sound?

tkelman:
that could work, or DcJS to mirror the PlotlyJS naming?

kmsquire:
DCjs.jl?  DCJS, but that's starting to be hard to decrypt...

tbreloff:
My vote would be for DimensionalCharting or something similar.

TotalVerb:
Since DC stands for D3 and Crossfilter, DimensionalCharting would be a backronym.

quinnj:
I like CrossfilterCharts.jl, or honestly DC.jl doesn't seem bad too bad to me either.

tawheeler:
Ok, will change the package name to CrossfilterCharts.jl and submit a new PR. Looks like I should tag a version too.
Thanks!

PR #5347, opened 2016-06-09
Tag Plots 0.7.0, RecipesBase 0.0.2, and VisualRegressionTests 0.0.4

tbreloff:
This is a huge (somewhat breaking) update/improvement for the the Plots ecosystem.  See [the NEWS](https://github.com/tbreloff/Plots.jl/blob/master/NEWS.md) and the linked [summary](http://plots.readthedocs.io/en/latest/plots_v0.7/) for the full scoop.

As always, please file issues with bugs or feature requests, and swing by the gitter channel with questions.  The docs are slowly filling out, but I still have a lot to do there (help would be appreciated, though you have to understand Plots to do it ;)  I'll post on julia-users in the next week after I've had a chance to work through a few more bugs and add more documentation.


tkelman:
How breaking? If users run Pkg.update will getting this version of Plots break other packages that use it? If so, then you should add upper bounds on the Plots requirements of past versions of those packages.

tkelman:
RecipesBase is failing its tests, is that expected?

tbreloff:
Hmm. Any breakage will depend on how they use Plots. I'll check it out and
report back. Of course, one key point is that most packages should stop
depending on Plots, and switch to adding recipes to RecipesBase. I'll work
with package owners to make the shift.

As for the RecipesBase tests... Yeah I guess that's expected. I'll update
to get the green, though it's well tested in Plots.

tbreloff:
I pushed up fixes for RecipesBase's tests... they pass now.  I guess I'll re-tag, even though I only changed tests.

As for Plots breakage... there's a lot more packages depending on Plots than I realized!  I'll open an issue to track it and cc you.

ExperimentalAnalysis/versions/0.0.2/requires
```diff
@@ -1,5 +1,5 @@
 julia 0.4
 DataFrames
-Plots
+Plots 0.0 0.7
```
tkelman: does 0.0.1 also need this?

tbreloff:
Doh! I labeled the commit message wrong.  I actually updated ExperimentalAnalysis in that last commit, not EEG.  Sorry.

tbreloff:
I think this is good to merge.  I'll be on standby to fix anything unexpected that pops up.

tbreloff:
Thanks @tkelman!

PR #5351, opened 2016-06-09
Tag KrylovMethods v0.2.0

lruthotto:


tkelman:
Your test-only dependencies should be given in `test/REQUIRE`, not added on Travis. Your tests depend on an unregistered package, so it's not possible for them to pass when someone just does `Pkg.add("KrylovMethods"); Pkg.test("KrylovMethods")`.

lruthotto:
This is not a test-only requirement. Our code actually does break without this new version of KrylovMethods that features block iterative methods.

lruthotto:
I see what you mean now. Sorry. No problem adding MatrixDepot, but how to go about Linear Operators (not an official package)?

tkelman:
Either it should be registered or the tests of a registered package shouldn't depend on it (or those that do could be made optional).

ViralBShah:
Best to register LinearOperators.jl.

Cc @dpo as well.

dpo:
Let me make a few updates to LinearOperators and I'll submit it here.

dpo:
See #5766.

PR #5352, opened 2016-06-09
Tag AverageShiftedHistograms v0.5.2

joshday:
Switching away from Requires for RecipesBase to define plotting methods.


tkelman:
Interesting, looks like you have appveyor enabled but haven't added an appveyor.yml yet?

joshday:
Good catch.  I added an appveyor.yml now.  Build is passing.

PR #5353, opened 2016-06-09
Register Calc.jl (v0.0.1)

tshort:
This is an interactive RPN-style calculator for the REPL in the vein of Emacs Calc.


tkelman:
Would be good to add some tests and turn on travis, if at all possible.

tshort:
I hope to add tests, but I haven't figured out how to add tests for custom REPL modes.

tkelman:
I think the base repl tests do something similar? Might be worth looking at, or pestering Keno to write more documentation.

PR #5358, opened 2016-06-10
Tag StatsBase v0.8.2 [git://github.com/JuliaStats/StatsBase.jl.git]

andreasnoack:


tkelman:
Thanks for fixing this, the carnage on PkgEval from StatsBase and DataArrays is pretty significant.

PR #5359, opened 2016-06-10
Tag DataArrays v0.3.3 [git://github.com/JuliaStats/DataArrays.jl.git]

andreasnoack:


nalimilan:
Maybe re-tag after merging https://github.com/JuliaStats/DataArrays.jl/pull/198? Would be good to get tests to pass on 0.5.

andreasnoack:
@nalimilan Is that not covered by tests? I don't see any warnings in the test output.

nalimilan:
Ah, maybe that wasn't covered after all. Let's tag as-is then, I'll double-check that.

PR #5364, opened 2016-06-12
Register GMT

joa-quim:


tkelman:
Is Libui related / required here? If not, could you separate that out? GMT is a bit of a generic acronym, but there don't seem to be too many software projects called that.

joa-quim:
Sorry for the mix. Yes, Libui is completely separate from GMT but when I did a ``Pkg.publish()`` both came together and now I don't know how to take it out from the GMT PR.

Regarding on the generality of the GMT acronym, well [GMT ](http://gmt.soest.hawaii.edu) is ancient but still very  actively maintained software.

tkelman:
You could just revert that commit within this branch.

PR #5370, opened 2016-06-12
Register DifferentialEquations and GrowableArrays v0.0.3

ChrisRackauckas:
GrowableArrays had a small change to be more performant. Tests pass in v0.4 and v0.5.

DifferentialEquations added a lot: SDE solvers, ODE solvers, stationary Stokes equation solvers, systems of equations for all of them, switched to plots recipes, built-in animations, convergence testing methods, methods for more easily doing Monte Carlo simulations, and a whole bunch of tests. Tests don't pass on Mac because ImageMagick.jl doesn't seem to work there and so it fails at the generation of the animations, but other than that the tests are fine. This version is also closer to v0.5 compatibility: the current issues are due to NLSolve.


tkelman:
Can you push the git tags?

Should `LatexStrings` either be in `REQUIRE` and not just `test/REQUIRE`, or taken out of the main package code: https://github.com/ChrisRackauckas/DifferentialEquations.jl/blob/ba38c20099120cf9d3e210b78b5539600d21de92/src/DifferentialEquations.jl#L23 ?

PR #5357, opened 2016-06-10
Tag FixedSizeArrays v0.2.0

c42f:

PR #5373, opened 2016-06-13
Tag FixedSizeArrays v0.2.1

c42f:
A quick fix for a depwarn oversight in 0.2.0 see #5357 

cc @SimonDanisch Changes here are very minimal vs 0.2.0, I wanted to get this in quickly to fix problems with GeometryTypes tests (which in any case should be upgraded for the new API, but shouldn't just have broken without a depwarn.)

### DIFF
--- a/FixedSizeArrays/versions/0.2.1/requires
+++ b/FixedSizeArrays/versions/0.2.1/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.7.15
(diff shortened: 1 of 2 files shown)


tkelman:
can you push the git tag so it shows up under https://github.com/SimonDanisch/FixedSizeArrays.jl/releases? thanks!

c42f:
Done!

PR #5379, opened 2016-06-14
Tag Plots 0.7.1 and others; Register PlotRecipes

tbreloff:
See the [NEWS](https://github.com/tbreloff/Plots.jl/blob/master/NEWS.md)

Also small changes to RecipesBase and VisualRegressionTests.

[PlotRecipes](https://github.com/JuliaPlots/PlotRecipes.jl) is where I'm starting to collect common visualization routines (which are super easy to make now with Plots/RecipesBase).  There's not a ton in there yet, and I need to add good tests, but I think it's worthwhile to register.


tkelman:
Could you get Travis to pass if you turned it on? One of the tests in PlotRecipes looks like it uses Distributions but doesn't declare that dependency.

tbreloff:
Those were just meant to be example usages, not tests (yet). I could 1)
comment out everything and test that 'using PlotRecipes' works, 2) keep
travis off, or 3) remove the registration from the PR. Your choice. Let me
know which is best for you.

tkelman:
Or add it to `test/REQUIRE` ? I'd prefer anything other than option 2. I do like to separate new package registrations from tagging existing packages into independent PR's if they aren't too tightly coupled though, the tags can be merged with less turnaround time.

tbreloff:
>  I do like to separate new package registrations from tagging existing packages

I had the same thought and made the wrong choice... sorry about that.

I just turned on travis, updated the travis.yml to allow failures on nightly, and commented out the examples.  Tests pass.

tkelman:
> Tests pass.

Great! Want to adjust the tag/make a new one to include the `runtests.jl` change?

tbreloff:
All good now?

tbreloff:
Thanks again for the help @tkelman!

PR #5380, opened 2016-06-14
Tag JSON v0.5.2

kmsquire:
Major updates (thanks to @TotalVerb)

    Optimize and fix issues with the `Parser` module

    Overview of changes:

    `parse_string`
     • now type stable
     • allocates much less when parsing in-memory data
     • fast path for strings without escape sequences

    `parse_number`
     • now non-allocating for ints (still allocates for floats)
     • correctly reject more kinds of invalid numbers

    `parse_array`
     • slightly improve efficiency
     • correctly reject invalid array formats

    `parse_object`
     • substantially improve efficiency
     • change default key type to concrete

    `parse_simple`
     • don’t blindly accept all words of form t**e, f***e, or n**l
     • naming convention: instead of simples, code now refers to jscontants

    streaming parser
     • replace brittle streaming parser with a new kind of ParserState
     • much faster than old streaming parser and less error-prone

    miscellaneous
     • use `@inbounds` in more places for better performance
     • fix some possible unicode compatibility issues
     • add some docstrings in various places
     • factor out error messages
     • add file with constant definitions so `@compat` can be avoided
     • fix up some weird tests
     • delete code to allow certain failures (no more allowed failures)
     • fix @os_only deprecation warnings

    known issues
     • streaming parser doesn’t keep state and so can’t offer informative errors


tkelman:
This might've been worth a new minor version given how much of a rewrite it was? Ah well.

kmsquire:
I thought about that, but it still works for Julia 0.3.  Still, will keep that in mind for the future.

Actually, I could probably just retag it now...

tkelman:
I'd say (and the travis checks should now be enforcing) dropping support for a minor version of Julia requires at least a minor package version bump, but you're always allowed to advance major/minor more rapidly than that if you'd like. Rewrites, possibly-breaking changes, and new features if you're post-1.0 are all good reasons for that IMO.

PR #5381, opened 2016-06-14
Register IncrementalInference

dehann:
New graph based inference library in pure Julia, work ongoing.


dehann:
Going to tag new branch for test issue on Travis

PR #5388, opened 2016-06-15
Register GMT (updated)

joa-quim:


tkelman:
Will it be possible to get an automated build of the library going, or installed on Travis? Right now your runtests.jl doesn't actually include your tests.

joa-quim:
I know. The tests is something I don't know how to address. We have internal set of tests in GMT that some of which I ported to Julia but they rely on comparing with reference files (mainly postscript images) that live only in the source code tree. The ``tests.jl`` in test directory is just a small group of tests that do not test the figure creation. 
However the main issue is that creating an automatic building of the GMT package is not obvious. We have a Cmake solution described in the GMT page but still many users have troubles following the recipe. So, honestly I don't know how to do that.

PR #5389, opened 2016-06-15
Tag ForwardDiff v0.2.0

jrevels:
@mlubin


tkelman:
This broke several of your reverse dependencies - http://pkg.julialang.org/pulse.html (ignore the failures caused by Conda). When you're tagging a "this will break things" release, please say so. You'll need to go and add upper bounds to the packages that were using ForwardDiff 0.1 features that don't work any more.

jrevels:
Sorry about that.

> ForwardDiff 0.1 features that don't work any more.

Just to avoid confusion, the new ForwardDiff still has all the same functionality; the only totally removed function was `tensor`, which AFAIK only a single package was using (here's [how to perform that calculation now](https://github.com/JuliaDiff/ForwardDiff.jl/pull/102#issuecomment-225735459), if any onlookers are curious). Here's [a recipe list for updating code that uses ForwardDiff](http://www.juliadiff.org/ForwardDiff.jl/upgrade.html). I should probably add the `tensor` trick there, as well as explicitly state that API functions are not exported and need to be fully qualified.

tkelman:
#5400 should mitigate half of it, looks like it was only 2 packages (out of things that were previously passing on pkgeval and not obscured by other issues).

I guess `jacobian` is no longer exported (I agree with that change, though we need some mechanism of adding a deprecation period for removing an export without renaming it...), and `GradientNumber` is gone or changed somehow?

jrevels:
Yes. All the `ForwardDiffNumber` types were reduced to a single type. A summary of the changes is [here](https://github.com/JuliaDiff/ForwardDiff.jl/pull/102#issue-132282322)

I'll go ahead and make PRs to those two repositories to fix their ForwardDiff code.

tkelman:
Looks like this also broke PATHSolver but somehow that wasn't appearing on packageevaluator until now?

PR #5385, opened 2016-06-14
Tag JSON v0.6.0

kmsquire:
v0.5.2 should probably have been a minor version bump (cf: https://github.com/JuliaLang/METADATA.jl/pull/5380#issuecomment-225941996).

This sets v0.5.3 to the previous commit, and bumps master to v0.6.0

PR #5398, opened 2016-06-16
Bound JSON for packages that PkgEval identified as being broken by 0.6.0

tkelman:
ref https://github.com/JuliaLang/METADATA.jl/pull/5385, these are packages that PkgEval identified as broken by the JSON rewrite. Mostly of the form "Expected end of input," see http://pkg.julialang.org/pulse.html for more details.

cc @TotalVerb, I'm guessing you didn't expect this much breakage
cc @yeesian and @garborg on GeoJSON
plotly is already being fixed apparently
cc @malmaud on Requests
cc @ylxdzsw on Restful
cc @wookay on Swifter

Easiest would be if this can be fixed in JSON, but maybe it's just more picky now. Could easily have missed more packages than this that don't currently pass their tests on PkgEval.

### DIFF
--- a/Requests/versions/0.0.2/requires
+++ b/Requests/versions/0.0.2/requires
@@ -3,4 +3,4 @@ HttpParser
 URIParser
 GnuTLS
 Codecs
-JSON
+JSON 0.0 0.6
--- a/Swifter/versions/0.0.1/requires
+++ b/Swifter/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.4
 Requests
-JSON
+JSON 0.0 0.6
--- a/GeoJSON/versions/0.0.1/requires
+++ b/GeoJSON/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.3
-JSON
+JSON 0.0 0.6
 Compat
--- a/GeoJSON/versions/0.0.2/requires
+++ b/GeoJSON/versions/0.0.2/requires
@@ -1,3 +1,3 @@
 julia 0.3
-JSON
+JSON 0.0 0.6
 Compat
--- a/GeoJSON/versions/0.0.3/requires
+++ b/GeoJSON/versions/0.0.3/requires
@@ -1,3 +1,3 @@
 julia 0.3
-JSON
+JSON 0.0 0.6
 Compat
--- a/Plotly/versions/0.0.1/requires
+++ b/Plotly/versions/0.0.1/requires
@@ -1,2 +1,2 @@
 HTTPClient
-JSON
+JSON 0.0 0.6
--- a/Plotly/versions/0.0.2/requires
+++ b/Plotly/versions/0.0.2/requires
@@ -1,2 +1,2 @@
 HTTPClient
-JSON
+JSON 0.0 0.6
--- a/Plotly/versions/0.0.3/requires
+++ b/Plotly/versions/0.0.3/requires
@@ -1,2 +1,2 @@
 HTTPClient
-JSON
+JSON 0.0 0.6
--- a/Plotly/versions/0.1.0/requires
+++ b/Plotly/versions/0.1.0/requires
@@ -1,6 +1,6 @@
 julia 0.4
 Requests 0.3.5
-JSON
+JSON 0.0 0.6
 PlotlyJS 0.2.0
 Compat 0.7.20
 Reexport
--- a/Requests/versions/0.0.3/requires
+++ b/Requests/versions/0.0.3/requires
@@ -3,4 +3,4 @@ HttpParser
 URIParser
 GnuTLS
 Codecs
-JSON
+JSON 0.0 0.6
--- a/Requests/versions/0.0.4/requires
+++ b/Requests/versions/0.0.4/requires
@@ -3,4 +3,4 @@ HttpParser
 URIParser
 GnuTLS
 Codecs
-JSON
+JSON 0.0 0.6
--- a/Requests/versions/0.0.5/requires
+++ b/Requests/versions/0.0.5/requires
@@ -3,4 +3,4 @@ HttpParser
 URIParser
 GnuTLS
 Codecs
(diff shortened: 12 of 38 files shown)


malmaud:
As far as I can tell, JSON being sent back from httpbin.org is being rejected, which suggests that if there is new pickiness in JSON.jl, it's inappropriate.

TotalVerb:
Will send a PR to back out the expected EOF change tonight. Sorry about the breakage!

Could someone please provide an example JSON being rejected?

TotalVerb:
Expected EOF doesn't seem to be the change that broke Requests. Will investigate.

Requests/versions/0.0.2/requires
```diff
@@ -3,4 +3,4 @@ HttpParser
 URIParser
 GnuTLS
 Codecs
-JSON
+JSON 0.0 0.6
```
TotalVerb: This is a test only issue (bug in tests). Do not bound.

TotalVerb:
Going through failures now. Some are definitely downstream issues.

 - Requests: test failure is trying to parse an empty response as JSON. This is an OPTIONS query; there should be no response, and no response is not valid JSON. It's a good thing JSON is more picky here because it identified a legitimate code bug. Empty string is not considered valid JSON. Neither Python nor JavaScript will parse an empty string as JSON. Will submit a PR to Requests fixing the issue. Note that this is a test-only problem and it should not be any reason to bound the JSON version for Requests.

Swifter/versions/0.0.1/requires
```diff
@@ -1,3 +1,3 @@
 julia 0.4
 Requests
-JSON
+JSON 0.0 0.6
```
TotalVerb: Test-only problem; do not bound.

Requests/versions/0.0.2/requires
tkelman: If the tests don't pass with the newer version of JSON, as long as a more recent version of Requests gets tagged with tests fixed, I think the bound is still called for. Especially since Requests isn't making new tags that support 0.3 any more but JSON is, without bounds Requests will fail its tests from now on on 0.3.

TotalVerb:
The Restful failure is confounding me. There doesn't even seem to be any JSON parsing near the line that has an issue, even though there's clearly a JSON parse error! Can you reproduce it locally on v0.4?

Requests/versions/0.0.2/requires
TotalVerb: OK, that's reasonable then. Hopefully the PRs are merged soon.

TotalVerb:
LGTM. It wasn't my intent to break all these packages, but so far it's looking like none of the issues are hard to fix. I think no further action is needed on JSON.jl's part.

tkelman:
I can reproduce the Restful issue locally, on 0.4.5 Linux. Maybe HttpServer behaves differently on your system vs mine/pkgeval?

TotalVerb:
What line causes the test failure?

tkelman:
https://github.com/ylxdzsw/Restful.jl/blob/1560e49445a51ac12e386463b7cc59d51fa186eb/test/runtests.jl#L59

TotalVerb:
Ah I see, the problem is probably server side. Thanks, will investigate.

yeesian:
I do have a tendency to over-type, and will fix the GeoJSON breakage, don't worry about it; thanks for being on top of things!

TotalVerb:
Figured out why I couldn't reproduce `Restful` failure—my Julia 0.4 was on JSON v0.5.1. I had thought that I symlinked it, but apparently not.

The Restful failure is a legitimate problem. Restful is depending on parsing an empty string as nothing, merely so that it does not cause an error and abort the request. This is just a special case of any invalid JSON—an empty string as the request just happens to be pretty common.

There is an easy and quick fix: just swallow the exception and treat invalid JSON as "nothing" or something, deferring the problem later down the chain. I will submit this minimal fix: a more robust solution would be up to the author of Restful.

tkelman:
Thanks for looking into these. I think we need to gradually move towards a more conservative recommendation of always having upper bounds and only relaxing them after demonstrating things continue to work with newer versions, though doing so without introducing many more version conflicts or held-back situations may prove tricky.

TotalVerb:
We could use semantic versioning to infer safe upgrades, but this issue does illustrate semantic versioning's weaknesses quite well. Almost _every_ change is a breaking change, if code was:

 - relying on internal implementation details (Plotly)
 - relying on undocumented quirks (aka bugs) (Restful, Swifter)

Neither of the changes that broke those packages would have, nor should have, incremented the major version.

The only breakage here that would actually have warranted a major version is the one for GeoJSON, as a return type changed. If that change were excluded, then JSON.jl incrementing a minor version is perfectly consistent with SemVer, and yet packages would break.

tkelman:
Yeah, asking developers of widely-used packages to fix all their reverse-dependencies isn't going to be sustainable for long. I haven't felt so bad about doing it this way for now though because at least if we catch it quickly, whoever was responsible for making the (often unintentionally) breaking change is probably more recently active and paying closer attention than the downstream packages' developers.

TotalVerb:
@tkelman This did not fix PlotlyJS. I think that one needs to be bounded at 0.5.1.

tkelman:
propose a PR? If it wasn't previously passing on PkgEval then I wouldn't have noticed it.

TotalVerb:
Plotly and PlotlyJS were bounded in an incorrect way, because 0.5.3 (and 0.5.2) broke PlotlyJS, not 0.6. Furthermore, PlotlyJS was the affected package, not Plotly. I'll submit a PR with modifications.

tkelman:
0.5.3 is the same as 0.5.1?

TotalVerb:
No, it isn't. 0.5.3 is several commits in the future.

TotalVerb:
PR here: #5427.

PR #5400, opened 2016-06-16
Add upper bound on LeastSquaresOptim's ForwardDiff requirement

tkelman:
ref #5389, jacobian is now gone in 0.2.0

cc @jrevels @matthieugomez 

there's also a new failure in CoordinateTransformations.jl, cc @andyferris, but since that only depends on ForwardDiff in `test/REQUIRE` it needs a new tag to fix.

### DIFF
--- a/LeastSquaresOptim/versions/0.0.1/requires
+++ b/LeastSquaresOptim/versions/0.0.1/requires
@@ -1,2 +1,2 @@
 julia 0.4-
-ForwardDiff 0.1.1
+ForwardDiff 0.1.1 0.2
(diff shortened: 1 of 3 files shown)


tkelman:
Now that there are new tags of LeastSquaresOptim available I think this is less pressing, but still should be okay to merge this?

andyferris:
Thanks @tkelman for the heads up. @jrevels has resolved this for us in a very kind PR.

tkelman:
Great, thanks all for taking a quick look here. @andyferris do make a new tag when you can so your package goes green again on PkgEval.

PR #5413, opened 2016-06-17
Bump ProfileView and JLDArchives

timholy:


KristofferC:
Yay, working ProfileView again! I feel completely handicapped without it.

timholy:
:smile: It's been working on master for a week or so. Sorry for not tagging a new version earlier.

Fun thing about the new version: it will show info about inlined functions. If you right-click on a box, in what gets dumped to the REPL, the "parent" is the lowest one and the one that actually performed the work is the top (non-indented) one. I might change that at some point. It also seems the display in `Profile.print()` could be improved to show inlining info, but I haven't had time to tackle that.

PR #5423, opened 2016-06-18
New packages: GracePlot, SpiceData & LibPSF.

ma-laforge:


tkelman:
These all look good. Some real tests would be a good idea down the line.

ma-laforge:
Maybe you can help me there: I'm sort of new when it comes to regression testing.

**GracePlot**
 - Requires the installation of a separate application... Don't understand Julia's build system well enough to automate the installation.
 - I don't even seem to be able to build it at all on my system.
 - I guess I could test constructing some of the base objects (before the base application gets launched).

**SpiceData/LibPSF**
 - Requires input data.  I am trying real hard to avoid putting data inside my repositories to avoid slowdowns while cloning, etc.

**In general**
I'm not that familiar with what kind of tests one would do, and how deep these tests should go.

For example, should I mostly run tests of the users' programming interface? Should I test lower level constructs and types? How exhaustive should these test be?

Maybe you can point me to some projects that demonstrate well how to build good testbeds; projects that strike a good balance between good coverage and not overdoing things.  I typically find these test somewhat hard to read (Which is odd - they don't really seem to be doing much).

Then again: I suppose having *some* test is better than having none...

ma-laforge:
PS. All projects have a demo or two under the `sample/` directory that demonstrate functionnality.

kshyatt:
>Requires the installation of a separate application... Don't understand Julia's build system well enough to automate the installation.

Isn't this what https://github.com/JuliaLang/BinDeps.jl is for?

ViralBShah:
While BinDeps is nice to have, I would only put in that effort once there are a few users of the package, since it is a bit complicated to figure out. 

For tests, enabling Travis tests and having basic unit tests for the user exposed API is the minimum to do. The next step would be to enable coverage testing, and then on to exhaustive testing. Perhaps some of this is worth capturing in our package developer documentation.

ma-laforge:
> While BinDeps is nice to have, I would only put in that effort once there are a few users of the package, since it is a bit complicated to figure out.

Sounds reasonable... Not to mention I have not yet figured out how to take care of all dependencies in the base compile (which is why I suggest the `apt-get` install procedure).

>For tests, enabling Travis tests and having basic unit tests for the user exposed API is the minimum to do.

Reasonable as well:  @ViralBShah, do *you* know of a few good projects I could use to get ideas?  Maybe one clear example of simple high level tests, and another that does good, clear in-depth coverage testing.

I looked at Gadfly, as @kshyatt suggested.  Although I might be wrong, Gadfly's "tests" seem more like an exhaustive set of demonstration plots that get collected into an .html document for human review.  I suppose demonstration plots can be used to test for undesired exceptions, but it would seem a human eventually needs to validate the results.

tbreloff:
For testing of packages which output graphics, you should look at
VisualRegressionTests, which forms the basis of the (fairly extensive)
integrated visual tests for Plots.  The images are stored in a separate
repo: JuliaPlots/PlotReferenceImages

I intend to move much of the generic testing code from Plots into
VisualRegressionTests when I have time... For now you'll probably want to
copy/paste from Plots (attribution unnecessary but appreciated).

tbreloff:
And speaking of attribution, the underlying algorithm is thanks to Tim
Holy... Ive found it to be very good with minimal false negatives. Thanks
again Tim!

ma-laforge:
Thanks Tom.

I don't think I can do that with GracePlot at the moment since I can't automate the install, but I should look into it for InspectDR.

PR #5434, opened 2016-06-20
Register XMLRPC: v0.0.1

sjkelly:
This is a package to support [XMLRPC](http://xmlrpc.scripting.com/spec.html). I am currently using it for [Odoo.jl](https://github.com/sjkelly/Odoo.jl), to connect to the Odoo ERP software. The protocol is somewhat antiquated, but it may be useful for communicating with older systems.


sjkelly:
FWIW tests pass on 0.4, but fail on 0.5 due to a bug in LightXML. I am working on a patch.

tkelman:
Ah didn't read the log too carefully, was going to suggest maybe turning nightly to an allowed failure if you didn't have a fix incoming.

sjkelly:
Ah nice, I didn't know you could configure that with Travis: https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

PR #5444, opened 2016-06-21
Bump SQLite version

quinnj:


tkelman:
Is this a rewrite since the last tag 7 months ago? If so, then please use a new minor version number.

tkelman:
Nevermind, looks like the rewrite was for version 0.3. Many of the git tags are missing so it's not easy to cross-reference what changes were made when. Please recreate the git tags to match the metadata tags if you no longer have them locally.

tkelman:
This does however fail its Travis tests, even on release. Needs to be tighter about a dependency version, or is something just broken?

quinnj:
The tests are failing because it's waiting on the CSV version bump.

tkelman:
Then that needs to be listed explicitly in REQUIRE.

quinnj:
Looks like travis is happy here now too: https://travis-ci.org/JuliaDB/SQLite.jl

SQLite/versions/0.3.2/requires
```diff
@@ -1,6 +1,6 @@
 julia 0.4.1
 BinDeps
-CSV
+CSV 0.0.8
```
tkelman: do this in the package too https://github.com/JuliaDB/SQLite.jl/blob/master/REQUIRE

SQLite/versions/0.3.2/requires
```diff
@@ -0,0 +1,9 @@
+julia 0.4.1
+BinDeps
+CSV
+DataStreams
+DataFrames
```
tkelman: you're doing something that is causing this not to match https://github.com/JuliaDB/SQLite.jl/blob/1123351beaba878eaf45cbc94a0ad1149273b1ef/REQUIRE - are you not even using `Pkg[Dev].tag`?

SQLite/versions/0.3.2/requires
```diff
@@ -0,0 +1,9 @@
+julia 0.4.1
+BinDeps
+CSV 0.0.8
```
quinnj: I'm showing the right dependency here, @tkelman

tkelman: yeah this is good now, but whatever you did that resulted in metadata requires not matching the package requires, avoid making it a habit

PR #5446, opened 2016-06-21
Tag TaylorSeries v0.1.3

lbenet:


tkelman:
Aha, you might be my first victim of the recently-made-stricter versioning checks. Can you remember whether TaylorSeries 0.1.0 supported Julia 0.3 like its `REQUIRE` says it does? If so, then it would be good to use a separate minor version number for the current release series which is now 0.4-only. So `Pkg.tag("TaylorSeries", :minor)` to re-tag this as 0.2.0 should be enough.

lbenet:
Ahh, first victim!. Indeed, it is the problem you spotted. I just tagged the new *minor* version. 
Should I reopen a new PR?

tkelman:
Whatever's easier, new pr or push a new commit to this branch, if I'm predicting how my code will behave correctly it should be okay to leave the 0.1.3 tag alone as long as the 0.2.0 tag comes through correctly.

lbenet:
New PR is easier. I guess this can be closed.

tkelman:
closing in favor of #5447

lbenet:
Thanks!

PR #5449, opened 2016-06-21
Tag PlotlyJS v0.3.1

spencerlyon2:
Resolves issues with versions of JSON.jl

Also removes Colors.jl as a dependency -- we don't actually use it.


spencerlyon2:
RE: #5427 and cc @tkelman @TotalVerb

tkelman:
still references somewhere in src according to Travis? https://travis-ci.org/spencerlyon2/PlotlyJS.jl/jobs/139267982

spencerlyon2:
OK should be good to go now.

PR #5450, opened 2016-06-21
Tag DSGE v0.1.2

micahjsmith:


DSGE/versions/0.1.2/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4-
```
tkelman: The `-` refers to prereleases of Julia 0.4, do you still need to support these? We can add an exception if you do.

micahjsmith: Not anymore, I'll update the PR

PR #5453, opened 2016-06-22
Tag new versions of Gallium and its dependencies

Keno:


kshyatt:
So what I've learned here is that if I want a package developer to tag a new version, I just need them to give a JuliaCon talk?

Keno:
Yes

tkelman:
JuliaParser is failing its tests. https://travis-ci.org/JuliaLang/JuliaParser.jl/builds/139507218

tkelman:
So is ASTInterpreter. https://travis-ci.org/Keno/ASTInterpreter.jl/jobs/139507170

And Gallium. https://travis-ci.org/Keno/Gallium.jl/jobs/139507210

Tagging these one at a time would give a much better chance of actually making sure they work.

tkelman:
This actually should not have passed at all as COFF is not yet registered.

Keno:
I'll register COFF.

tkelman:
And fix your tests.

PR #5454, opened 2016-06-22
Tag DoubleDouble v0.2.0

dpsanders:
Updates for v0.4; Travis is enabled.
Minor tag since the public API for the constructor changed `double -> Double`

cc @simonbyrne @ivanslapnicar


tkelman:
Will existing versions of ArrowHead.jl need an upper bound?

dpsanders:
I guess so, yes, good point.

dpsanders:
The `Arrowhead.jl` tests passed so we (with @tkelman) merged this. Let me know if there are any problems @ivanslapnicar

ivanslapnicar:
@dpsanders thanks. Everything works fine.

PR #5455, opened 2016-06-22
Run Travis test on 0.4 and 0.5, add Julia upper bounds on some old BayesNets tags

tkelman:
otherwise packages that are 0.5-only do not get verified

I suspect this will fail because #5453 put METADATA in an invalid state on nightly. @Keno please use PkgDev.publish so your metadata changes get verified locally.

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -1,3 +1,7 @@
+if VERSION < v"0.4-"
+    startswith = beginswith
+end
+
 const url_reg = r"^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?"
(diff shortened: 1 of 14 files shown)


yuyichao:
Will this affect packages that are 0.4 only?

tkelman:
Yeah. It's also relying on PkgDev to not break. Are there any 0.4 only packages? Probably better to wait until there's an 0.5 rc for this. And keep this issue in mind when someone first wants to tag an 0.6-only package.

yuyichao:
Can we have both and make the nightly one optional before we have an rc?

tkelman:
yes but not without doubling the number of workers that get used here - maybe that's fine

PR #5458, opened 2016-06-22
Collective bump for DataStreams packages

quinnj:


ODBC/versions/0.4.1/requires
```diff
@@ -0,0 +1,9 @@
+julia 0.4.1
```
tkelman: Missing a `sha1` file. Use `Pkg[Dev].tag`.

PR #5459, opened 2016-06-22
BayesNets v1.0.0

tawheeler:
Big release for BayesNets.jl
This is a big API change - everything is better and new features are in!


tkelman:
Should add an explicit dependency on Compat rather than assuming it will be present as an indirect dependency - ref https://github.com/sisl/BayesNets.jl/blob/v1.0.0/src/BayesNets.jl#L5

PR #5464, opened 2016-06-22
Tag Swifter v0.0.7

wookay:
nice to meet everyone at JuliaCon 2016!

Updated followings
- fixed to show query result
- displaying views on the jupyter notebook for julia 0.4
- added `valueof` to use the result value
- specified the build number


Swifter/versions/0.0.7/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Requests
+JSON 0.0 0.6
```
tkelman: do you still need this upper bound or was this only a test issue?

wookay: oh, it was a test issue, just copied from 0.0.6's one.
I'll change that to be just JSON.

wookay: meanwhile, should I change all older ones (from 0.0.1 to 0.0.6) also to be just JSON?

tkelman: you haven't retroactively fixed the test in past versions, have you?

wookay: right, I haven't changed anything, so I suppose to be remaining past versions as "JSON 0.0 0.6".

tkelman: requests or some other package is still holding your version of JSON back to 0.5.4, but I checked with `Pkg.checkout("Swifter"); Pkg.checkout("JSON"); Pkg.test("Swifter")` and it all looks good.

wookay: thank you very much!

PR #5466, opened 2016-06-22
Tag Mimi v0.1.0

davidanthoff:


tkelman:
I'm guessing this failure in OptiMimi is expected? Worth adding an upper bound on OptiMimi's require for the old tags?

davidanthoff:
@jrising It looks like OptiMimi.jl doesn't work with the latest Mimi.jl version?

jrising:
I've now updated OptiMimi for the latest version of Mimi (and also of ForwardDiff), and so I'm just waiting for the METADATA pull request to be approved.

PR #5471, opened 2016-06-23
Tag GLM v0.5.3 [git://github.com/JuliaStats/GLM.jl.git]

andreasnoack:


GLM/versions/0.5.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+Distributions 0.4.6-
+StatsBase 0.8.0
+Reexport
+Compat 0.2.12-
```
tkelman: really not relying on anything newer than that?

andreasnoack: I don't think so but I'm not sure. We should probably begin to keep a changelog in `Compat.jl` such that this gets easier to reason about.

tkelman: agreed

PR #5478, opened 2016-06-24
Register ResultTypes [https://github.com/iamed2/ResultTypes.jl.git]

iamed2:
A Result type for Julia—it's like Nullables for Exceptions!

Full coverage and tests pass.


tkelman:
Nice. We could use more efficient compiler support for something like this.

PR #5483, opened 2016-06-25
Register ILESolver

EvgeneOskin:
Solver for interval linear equation systems. For inner and outer boundaries estimation.

### DIFF
--- a/ILESolver/url
+++ b/ILESolver/url
@@ -1,0 +1,1 @@
+git://github.com/EvgeneOskin/ILESolver.jl.git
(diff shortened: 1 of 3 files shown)


dpsanders:
This looks like a nice package.
Could you please add some documentation (e.g. in the README).
You will also need to add a license file; the usual option is the MIT license. 

Also, the proposed name of the package does not give much information about the contents -- how about  `IntervalLinearEquations`?

tkelman:
It's also recommended to enable Travis CI (see https://travis-ci.org and https://docs.travis-ci.com/user/languages/julia) which will automatically run your tests on every commit and pull request. Would your tests pass there?

PR #5486, opened 2016-06-25
Register IntervalLinearEquations: v0.1.0, v0.1.1, v0.1.2, v0.1.3, v0.2.0

EvgeneOskin:
Julia package to solve system of linear equations:

Changes from the previous PR:
- travis is configured;
- the readme is extended;
- the package is renamed (old name was ILESolver).

### DIFF
--- a/IntervalLinearEquations/versions/0.2.0/sha1
+++ b/IntervalLinearEquations/versions/0.2.0/sha1
@@ -1,0 +1,1 @@
+40f5d5dad342ca2e49b9e59ec64be29608f9b3bd
--- a/IntervalLinearEquations/url
+++ b/IntervalLinearEquations/url
@@ -1,0 +1,1 @@
+git://github.com/EvgeneOskin/IntervalLinearEquations.jl.git
--- a/IntervalLinearEquations/versions/0.2.0/requires
+++ b/IntervalLinearEquations/versions/0.2.0/requires
@@ -1,0 +1,3 @@
+julia 0.4
+Compat
+ValidatedNumerics


IntervalLinearEquations/versions/0.1.0/sha1
```diff
@@ -0,0 +1 @@
+e1b4093da686daee8d98175e6b8d39233b8167ef
```
tkelman: Since this tag (https://github.com/EvgeneOskin/IntervalLinearEquations.jl/tree/e1b4093da686daee8d98175e6b8d39233b8167ef/src) was under the old version, it won't work correctly. Since it's the only tag that claims to support Julia 0.3, it would also be the one any users of Julia 0.3 get if they try to do `Pkg.add("IntervalLinearEquations")`. Better to delete this tag (both files under `IntervalLinearEquations/versions/0.1.0`) I think.

IntervalLinearEquations/versions/0.2.0/sha1
```diff
@@ -0,0 +1 @@
+b73790ea43f03a44292580ddd3b11fad949afa21
```
tkelman: this commit doesn't match what was tagged as 0.2.0? https://github.com/EvgeneOskin/IntervalLinearEquations.jl/releases/tag/v0.2.0

EvgeneOskin:
Hi @tkelman, thank you for review and your notes. I've removed tags that won't work correctly. Also, fix sha to version 0.2.0.

Hope, it became better.

tkelman:
Looks good!

PR #5493, opened 2016-06-28
Tag GtkBuilderAid v0.0.4

Matt5sean3:
* Using `g_signal_query` to determine signal type information instead of guessing
* Used Compat package to account for most julia 0.5 changes
* Added example for using a widget as a canvas
* Allowed redirecting some warnings to a pipe
* Added support for functions defined shorthand
* Added automatic conversions for `Ptr{GObject}` to `GObject`
* Added convenience functions for using a widget as a canvas


GtkBuilderAid/versions/0.0.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Gtk
+Compat
```
tkelman: good practice to declare your minimum version dependency as a lower bound

tkelman:
should also declare the Cairo dependency explicitly in `REQUIRE` - https://github.com/Matt5sean3/GtkBuilderAid.jl/blob/71b061cf9df0cb6c679ab1a9545167731d4cc4d8/src/GtkBuilderAid.jl#L5

Matt5sean3:
I had thought that I'd declared the Cairo dependency but it looks like it's not there. It didn't cause test errors because the Gtk package pulls in Cairo.

As far as version compatibility, julia 0.4 is the earliest supported julia version for this package. It'll require some doing to sort out the minimum versions for the other dependencies.

I'll get that sorted out and try again in a bit.

PR #5494, opened 2016-06-28
Tag Plots 0.7.4, RecipesBase 0.0.6

tbreloff:
I had an bug with the last RecipesBase tag... sorry for the noise.


tkelman:
Looks like there's a failure on Travis release?

tbreloff:
Yeah should be fine... Tests all passed on my machine and only the boxplots
example for the GR backend didn't match on linux. It's annoying... but the
price we pay for the extensive tests.

tkelman:
Should you loosen the tolerances then? "Works on my machine" isn't publicly verifiable.

tbreloff:
Should I? Tough question. Personally I'd rather have the occasional false
negative in order to reduce false positives.

The most ideal would be test results that aren't binary: "100% passed on
OSX, 99% passed on Linux" looks a lot different than "failed".

tbreloff:
Actually, I suppose this isn't a limitation of travis, but of the method of testing.  Do FactCheck, or BaseTestNext, or anything else support logic like "if at least 95% of the tests pass, then consider it a pass"?  If not, this is something we should add.  

Most ideal would be the scenario that: 98% of tests pass, the repo gets the green banner, but I still get an email from Travis regarding the failed tests.

tkelman:
Is this an intermittent thing where you could average over multiple runs?

tbreloff:
No. It's deterministic. The nature of the problem is tricky with all the
slight differences between graphical output between platforms and backend
versions. I have to make a decision between false positives and false
negatives, and I'd rather fail a couple tests here and there than miss
actual bugs. We should build tests to find problems, we shouldn't build
tests to pass tests.

tkelman:
If it isn't a problem and is deterministic, then loosen the tolerances for the particular combination of platform and backend that need it?

tbreloff:
When I say it's deterministic I mean if you re-ran the test multiple times
it would return the same result each time. But I don't always know when a
test will fail until after it has failed. Yes I could hand tune every
parameter on every image test after a failure, then push it up again, but
that would be a hell of a lot of annoying manual work with no actual
benefit to the code. Let's be honest... most people would just remove the
tests (and we clearly shouldn't prefer that)

tkelman:
> I don't always know when a test will fail until after it has failed

Isn't this the whole point of CI, that this turnaround time is nearly immediate?

StefanKarpinski:
I think it makes sense to let @tbreloff figure out how to best test his own packages. Testing visual output is a very difficult problem and not many other plotting packages do so at all.

StefanKarpinski:
That said, would it make sense to have tighter bounds when you test locally?

tkelman:
When a user installs your package and does Pkg.test, having it pass instead of noisily fail seems worth going to the trouble of making a few targeted adjustments to tolerances. That seems far better to me than allowing an arbitrary unpredictable 5% subset of your tests to fail and still call that okay.

tbreloff:
> Testing visual output is a very difficult problem and not many other plotting packages do so at all.

:+1: 

> That said, would it make sense to have tighter bounds when you test locally?

I already do this.  When I test locally I usually use a 1% tolerance of "fuzzy image difference" which is really good at picking up actual differences.  During the automated travis testing, the tolerance is 5%, which is generally good for allowing for those cross platform differences in output.  In this case, however, there was a 15% difference in the "boxplots and violins" example for GR.  Sadly I don't have a good way to know **what** was different, as I only see the travis logs.  My choices are to skip that test in a new commit (frustrating because the Plots tests take 30-40 minutes after all the installations), increase the tolerances (which defeats the purpose of testing), or just accept the failure and move on with my life.  None of these options change the quality of the end user's experience... it's only for that little green banner.

> When a user installs your package and does Pkg.test, having it pass instead of noisily fail seems worth going to the trouble of making a few targeted adjustments to tolerances.

If the required tolerances were extremely consistent, I would agree with you.  But in practice the "correct" tolerances are a constantly moving target which is frequently out of my control (for example  if Conda switches to a new version of Matplotlib, or something similar).

I think, from my perspective, the optimal solution is to "pass" when some high percentage (95%) of tests pass, but to trigger a notification (email?) of the "soft pass" so that someone can review what happened.  This would encourage package developers to put in tests that might fail, and will generate useful information to improve the resiliency of the package.

For VisualRegressionTests.jl, it might be nice to additionally email the generated image files so that I can quickly see what failed and why.  All this is to say that I'd rather focus on automated solutions to good testing practices than spend all my time manually making sure I get a "pass".

BTW: thanks for the merge!!

PR #5500, opened 2016-06-28
Tag Compat v0.8.4 [https://github.com/JuliaLang/Compat.jl]

timholy:
Includes the `view` fix (@ranjanan)


timholy:
:clock10: :clock1030: :clock11: :clock1130:  Are we **sure** we want to keep Compat.jl under the JuliaLang banner? Esp. once julia 1.0 comes out and within 3 months the package-tagging rate goes up 10x?

tkelman:
I strongly suspect METADATA.jl won't survive in its exact current form beyond Julia 0.5. If Pkg3 actually gets completed on schedule, we'll likely be doing bidirectional mirroring between new, curated successor-to-metadata in some other org, and the existing metadata for 0.5 and earlier. This particular URL is hard-coded in a few too many places.

timholy:
Sounds good. It's not that I'm in a desperate hurry, it's just looking ahead to further growth of the julia ecosystem and realizing this is basically broken already, so needs fixing soon. 5+ hours for a version bump PR is too long, and if it gets to 24hrs the whole system blows up exponentially.

PR #5509, opened 2016-06-30
Tag Mocking v0.2.0 [git://github.com/invenia/Mocking.jl.git]

omus:
New version of Mocking.jl that works both on Julia 0.4 and 0.5.

I'd like to thank @vtjnash and @JeffBezanson for their assistance at JuliaCon 2016!


tkelman:
It seems like this is causing issues with DateParser and TimeZones on pkgeval:
```
WARNING: could not import Mocking.@mendable into TimeZones
==============================[ ERROR: TimeZones ]==============================

LoadError: LoadError: LoadError: UndefVarError: @mendable not defined
while loading /home/vagrant/.julia/v0.4/TimeZones/src/timezones/local.jl, in expression starting on line 158
while loading /home/vagrant/.julia/v0.4/TimeZones/src/TimeZones.jl, in expression starting on line 55
while loading /home/vagrant/.julia/v0.4/TimeZones/deps/build.jl, in expression starting on line 1
```

omus:
The Mocking update was a breaking change. What's the proper procedure for setting the upper bounds here? Do I just need to change the REQUIRES files for TimeZones and DateParser in METADATA?

tkelman:
Yeah, for all past tags of the reverse dependents, if they don't have any version bounds at all in the original tag, change the `Mocking` line to `Mocking 0.0 0.2` in the `requires` files for those past tags.

omus:
Ok. Just to be fully clear: I am just doing this within METADATA.jl? Updating the tags in the package repos could be messy...

tkelman:
Just METADATA. And add lower bounds to the dependent packages along with the fixes, but that's a from-now-on change rather than a retroactive thing.

PR #5512, opened 2016-07-01
Tag ApproxFun v0.2.0

dlfivefifty:
Includes many substantial changes, including fractional differential operators and partial 0.5 support.


tkelman:
It looks like this broke SingularIntegralEquations, cc @MikaelSlevinsky - http://pkg.julialang.org/detail/SingularIntegralEquations.html

Best to add upper bounds to all past tagged versions of that (and any other) dependent packages that this broke.

dlfivefifty:
I tag is in the works, should be done in the next couple weeks or so.

dlfivefifty:
#5526

tkelman:
Adding a new tag makes the latest version of the package work, but all old versions of the packages are broken because they were not accurately upper-bounding their dependencies. You should retroactively add an upper bound on the past tags of SingularIntegralEquations' dependency on ApproxFun in METADATA as well.

dlfivefifty:
I don't know how to do that.

tkelman:
Under the existing tags at https://github.com/JuliaLang/METADATA.jl/tree/metadata-v2/SingularIntegralEquations/versions, modify the `requires` files to add an upper bound of `0.2` after the existing lower bound on the ApproxFun dependencies. The upper bound is exclusive.

dlfivefifty:
how do you add an upper bound?

tkelman:
Just a second number, e.g. `ApproxFun 0.0.7 0.2`

PR #5516, opened 2016-07-01
Tag GR v0.12.0

jheinen:


jheinen:
I understand that it is better to use a fixed known-working version. So, I closed the pull request and reverted the last commit.

PR #5521, opened 2016-07-02
Setting upper bounds on Mocking dependency

omus:
Fixing https://github.com/JuliaLang/METADATA.jl/pull/5509#issuecomment-230092889


omus:
cc: @tkelman

PR #5525, opened 2016-07-03
Tag SeqMaker v0.1.0

sfchen:
This version of SeqMaker provides complete functions for next generation sequencing simulation


tkelman:
Is it possible to enable tests on Travis CI?

sfchen:
@tkelman I enabled it. But the test is failing because it takes too long to download a big file.

tkelman:
Ah, that's too bad. Is there a way to run on a smaller subset of the data?

sfchen:
Currently it need human genome reference data to test (~800M), I will consider to pack a tiny genome reference with the package.

tkelman:
This dataset is probably too big for an automated unit test, you're running out of memory on PkgEval. http://pkg.julialang.org/detail/SeqMaker.html

tkelman:
@sfchen any update on getting this to pass its tests again?
At some point I increased the amount of memory on PkgEval so you stopped hitting that problem, now you've been getting
```
INFO: checking SHA1...
ERROR: LoadError: test error in expression: test_ngs()
MethodError: `lowercase` has no method matching lowercase(::Array{UInt8,1})
 in check_sha1 at /home/vagrant/.julia/v0.4/OpenGene/src/Reference/Genome/downloader.jl:7
 in download_assembly at /home/vagrant/.julia/v0.4/OpenGene/src/Reference/Genome/downloader.jl:56
 in load_assembly at /home/vagrant/.julia/v0.4/OpenGene/src/Reference/Genome/reader.jl:11
 in make_seq at /home/vagrant/.julia/v0.4/SeqMaker/src/maker.jl:7
 in ngs at /home/vagrant/.julia/v0.4/SeqMaker/src/SeqMaker.jl:32
 in ngs at /home/vagrant/.julia/v0.4/SeqMaker/src/SeqMaker.jl:26
 in test_ngs at /home/vagrant/.julia/v0.4/SeqMaker/test/runtests.jl:6
```
(same pkg.julialang.org link as above)

sfchen:
@tkelman this issue is caused by SHA1 package not compatible with v0.5.

tkelman:
SHA.jl is compatible with Julia 0.5. It changed from returning a string to returning bytes though.

sfchen:
I will update SeqMaker.

tkelman:
The problem is in OpenGene https://github.com/OpenGene/OpenGene.jl/blob/v0.1.9/src/Reference/Genome/downloader.jl#L7 but only exercised by SeqMaker's tests.

sfchen:
`sha1()` in package `SHA` master switched to return `Array{UInt8,1}`, instead of `String`.  But its latest release in `METADATA` is still returning `String`.

tkelman:
Are you sure about that, or do you have a package installed that applies an upper bound keeping your local copy held back? What version do you see installed?

sfchen:
I am sure. I will add a work around to support both `Array{UInt8,1}` and `String`.

tkelman:
https://github.com/JuliaLang/METADATA.jl/pull/5708 was tagged months ago

PR #5526, opened 2016-07-03
Tag  BandedMatrices,FastTransforms,ApproxFun,SingularIntegralEquations

dlfivefifty:
This brings SingularIntegralEquations up to date with ApproxFun.  However, the packages all have changes associated with sub/slice -> view, hence needed retagging.


tkelman:
Might be good to tag the dependency packages first and the bigger ones in a followup PR so Travis can verify things using the minimum tag versions.

dlfivefifty:
I agree.  It didn't click when I tagged everything that it would break all the Travis builds.

PR #5528, opened 2016-07-03
Tag NullableArrays 0.0.4 and 0.0.5

nalimilan:
@davidagold had tagged 0.0.4, but for some reason it wasn't published. Preserve it in case it was used by someone.


davidagold:
That was an oversight on my part. Thanks.

PR #5534, opened 2016-07-03
tag GLVisualize

SimonDanisch:


tkelman:
What's causing the `LoadError: UndefVarError: renderlist not defined` on GLVisualize? https://travis-ci.org/JuliaGL/GLVisualize.jl/jobs/142031136

SimonDanisch:
not really sure, the build script is a mess, since it never gave me useful information anyways... 
But this should not occur after tagging. I've made sure locally and master looks fine:
https://github.com/JuliaGL/GLAbstraction.jl/blob/master/src/GLCamera.jl#L501
https://github.com/JuliaGL/GLWindow.jl/blob/master/src/screen.jl#L358

tkelman:
If your packages don't work with existing tags then they need to be stricter about lower bounds

SimonDanisch:
This is more about the ancient build script in GLVisualize... I'll update
it tomorrow!

SimonDanisch:
But being more considerate with version numbers doesn't sound like a bad
idea, either...

PR #5545, opened 2016-07-04
Tag ControlSystems v0.1.4

mfalt:


tkelman:
Ref https://github.com/JuliaControl/ControlSystems.jl/commit/a2c7bbf2b5ed908bef7e33bc7faa2d549b0a1a34#commitcomment-18118869, you should not be installing an unregistered package in your tests. If you need a external resource for data, you can download it, but you shouldn't install it as a package. You should also be using a specific version of external resources rather than whatever happens to be on master.

mfalt:
I agree that this isn't the nicest solution, but I wanted to find a solution that allow for testing of the plots while having minimal extra code in the main package (ControlSystems.jl), so I thought mirroring what is done in the tests in [Plots.jl](https://github.com/tbreloff/Plots.jl/blob/master/test/travis_commands.jl) would be a decent solution.
I assume that we could be "allowed" to keep resources (images) in a git repository, like ControlExamplePlots, but without the package structure and the convenient Pkg command? But that we should move the downloaded julia code to the tests? I guess this should be possible now that VisualRegressionTests.jl is registered, but I'm not sure it was before.
I also agree that a better solution would be to download a specific version of the images.

I assume this is some kind of policy that you have, so I won't try to argue with it, but I wonder what the point of disallowing it is, when pretty much the same thing can be done in a less convenient way?

tkelman:
It's more an issue of where it happens, and in what form the resources get installed. If the clone command is located in `deps/build.jl`, then it occurs whenever the package is installed or updated or `Pkg.build` is run. And if you're cloning master into the package directory, then any time the user does `Pkg.update` from then on they will pull whatever happens to be on master, which is not good for stability or reproducibility.

Generally a package should declare its runtime dependencies on registered packages in `REQUIRE` (or test-only dependencies in `test/REQUIRE`), and a registered package shouldn't have dependencies on unregistered packages if it needs to install and use them as Julia packages themselves. If you're just using it as test data, then it doesn't need to be installed as a Julia package in the user's global package directory in a way that `Pkg.update` will arbitrarily pull master.

mfalt:
So let me make sure if I get this right before doing any changes:
* I can keep the images in ControlExamplePlots.jl but remove everything else
* I should move all the julia code from ControlExamplePlots.jl to ControlSystems.jl
* I should download a specific version of ControlExamplePlots.jl, using git commands, from a file in `deps/build.jl`? But how do I do this only when the user cares about running tests?

tkelman:
You could do the downloading in `test/runtests.jl` if it's actually only needed for tests and not functionality of the package. Unless you really need the downloaded resource code to be a git repository, you could also download it as an archive off of github. You could also just register ControlExamplePlots then list it in `test/REQUIRE`, that might be easier?

tbreloff:
Please don't register a package in METADATA which just holds reference
images for testing.  I think the best long term options are:

1) Fix Pkg/REQUIRE so that  we can have non-registered dependencies (is
this hard, or just unwanted?  or is it fixed in a new version?)
2) Improve VisualRegressionTests to automate best practices of
storing/retrieving reference images.

I'll start an issue in VRT to discuss best features for a solution to #2.

In the short term, we have to choose between a bunch of imperfect, sub-par
choices, and I don't think it's clear that doing a Pkg.clone vs downloading
an archive vs something else is a straightforward choice.  There are cons
for every pro.  Lets figure out the best long term choice, and then try to
make that available in the short term!

tkelman:
> Fix Pkg/REQUIRE so that we can have non-registered dependencies (is this hard, or just unwanted?

This will happen in the revamp of Pkg. There will also be namespaces / tiers of Metadata, and the "official" curated tier won't allow dependencies outside of that tier. If it's just data, download it locally as a resource. If it's both data and associated Julia code that doesn't actually fit directly as part of the package that wants to use it for testing, what is wrong with registering it?

mfalt:
I am glad that you found this thread @tbreloff, I fully agree that registering a package that is only used for testing for one other specific package is not a good solution. I also agree that the current options are far from optimal.

tkelman:
> registering a package that is only used for testing for one other specific package is not a reasonable solution

Seems absolutely fine to me. What's the issue exactly?

TotalVerb:
I think part of the issue is that Git repositories do not scale well, and every additional package makes the problem worse.

tkelman:
Sure, Pkg and METADATA need a redesign, we all know that already. 1038 packages isn't going to be noticeably worse than 1037 though.

mfalt:
Would it be possible to pull this before we are able to decide and and implement a solution @tkelman? We would make sure that it is fixed relatively soon. We have critical bugs that were fixed since the last release (all plotting in the package is broken with the latest version of `Plots.jl` without the fixes in this release), and as I can see the changes to the tests won't be able to hurt any user (even if they decide to run the tests themselves).

tkelman:
Not until you restructure to avoid using Pkg operations in your tests. We don't allow registered packages to depend on unregistered ones.

ViralBShah:
I would prefer to get this merged and perhaps comment out the relevant tests for now. The images could be hosted at a URL and pulled during tests.

tkelman:
commenting it out would be okay, downloading a non-git archive would be better. doing Pkg.clone of some other package during the tests is harmful and we shouldn't merge a version that does that, as it means anyone who runs the tests is now using master of some unvetted, unregistered code.

StefanKarpinski:
What about checking out a specific version? SHA1 still isn't broken so that should be safe.

tkelman:
as a git clone, if you'd really like to do that instead of as an archive, sure. but it should be a local resource inside the package, not checked out as a julia package at the same level as all other packages - unless you want to register it, which again is the easiest way to move forward here.

tbreloff:
Until there's a better solution (which I'd desperately like help with over at VisualRegressionTests), I think you should replace [this code](https://github.com/JuliaControl/ControlSystems.jl/blob/master/test/runtests.jl#L4-L9) with a simple `using ControlExamplePlots` call.  **Let it error**.  If the user installs the package themselves then they can test it successfully.  On travis you can add the package yourself.

This solution is not ideal, I know, but it won't auto-clone anything the user didn't explicitly want, and it will still work if you want to test it.

tbreloff:
I'd be happy to move VRT over to JuliaCI if you think that might help in any way.

tkelman:
That's not necessary. Your approach sounds fine, if a failing package is preferable to avoid registering the test package.

tbreloff:
I wonder if we should consider a more complete solution to hosting test files.  Aside from images, I could see wanting to perform tests on binary data, or HDF5 files, or Feather data, etc.  It would be really great to have a julia solution for reproducible input data (which didn't also pollute the package directories).  Any thoughts on this?

tkelman:
I agree that would be useful but don't have any suggestions at this time.

ViralBShah:
@mfalt I really would like to get this merged. What would be the easiest thing from your perspective for now?

I prefer leaning towards a better experience for package managers and their users for now, and doing the right thing soon thereafter - but not holding things up.

ViralBShah:
@tbreloff How about github's large files or just using s3? Download as part of the tests and use as necessary. We do this in our perf tests for various data files.

tkelman:
This particular commit, with the usage of Pkg.clone, is effectively working around the registration process and should not be merged as is. Any number of minor modifications suggested above would be fine.

TotalVerb:
I agree that it is not appropriate at all for Pkg.test to install an unregistered package.

mfalt:
All right, so I don't have any time to do any long term fixes right now, so I think I will just uncomment the `Pkg.checkout` line. Could I be allowed to add the checkout to the travis file at least, that way we can still track the code coverage which would otherwise fall by at least 20%? The travis file will not be run by any user, and is not part of any julia code so I dont see why this would be a problem.
This obviously feels like a weird workaround, but I think it reflects the fact that the rule of not using a non-registered package is quite arbitrary in the sense that it is the only way you are not allowed to fetch and run code/data, although it is the most convenient.
I fell that registering the test images as a package is completely against how I would like the package repository to work and I don't want to contribute to such cluttering, I will instead wait for convergence upon a better solution.

tkelman:
Yes, go ahead and do whatever you'd like inside your `.travis.yml` file.

tkelman:
> not using a non-registered package is quite arbitrary in the sense that it is the only way you are not allowed to fetch and run code/data, although it is the most convenient.

The point is if you're installing it as a Julia package, then it needs to be registered if a registered package is going to depend on it. Unregistered packages can depend on registered and/or other unregistered packages, but registered ones can't depend on unregistered packages. This is explicitly checked for here if you had tried to use the conventional `REQUIRE` dependency mechanism instead of manually calling `Pkg.clone`.

You can clone or download the data locally inside ControlSystems.jl if it's just a testing resource. If "unregistered package" is the most convenient form for you to have things in, then this is the wrong repository to be having this conversation in. It's a far more minor distinction in terms of the amount of code you need to change - just shell out to git instead of using Pkg, that's all you need to do.

ViralBShah:
Great that we have a short term solution and a plan to do something long term. @mfalt Really excited to see the progress on ControlSystems. Not a user myself, but looks like it is shaping up to be a complete package. hope this hasn't inconvenienced users too much.

mfalt:
I made a very minor fix and removed the Pkg.clone command, so it should be possible to register a new version now. Not sure what the best option is, should I:
- 1. Simply change the sha1 hash in this pull request to the new commit and
  - 1.a, leave the tag, as is, on the ControlSystems.jl repo (thus one commit behind for version compared to METADATA.jl for 0.1.4)
  - 1.b, or remove the release and tag from from the ControlSystems.jl repo and re-push the tag for the new commit, or
- 2. Close this PR and tag a new version 0.1.5 (because 0.1.4.1 is not possible right?), leaving 0.1.4 out of METADATA.jl but in ControlSystems.jl

tkelman:
Either 1.b. or 2 would be preferable to 1.a. Whichever is easier for you.

mfalt:
Perfect, thanks for the a quick reply, I chose 1.b, and I am ready to have this pulled.
Also, thanks for taking the time to maintain the METADATA.jl repo and and making sure it works great!

tkelman:
for a future tag, couldn't you have it so `"test_plots"` doesn't run if the unregistered package isn't installed? that would at least give you some coverage and hopefully passing tests on packageevaluator etc.

mfalt:
Well, I guess I could. It makes sense that if a user wants to test it, it could be useful to see the result of the rest of the tests, but I don't think not passing tests is a big problem, especially for the plots which will fail regularly when the visuals of Plots.jl and PyPlot changes, it is more of an indicator that it is time to reverify the reference images. However the packageevaluator is a good point, but then it would still not certify the that the plots are working, and what is the value of passing tests if everything is not working anyway? I assume it would not properly install PyPlot anyway?

tkelman:
packageevaluator should be able to install pyplot. you may need to add the package to a list that need xvfb for tests to work headless.

I hope you're also planning long-term on eventually separating the plotting from the computation, since you don't need the former to do the the latter.

mfalt:
Great, I'm just used to PyPlot failing to install properly without specific commands like `ENV["PYTHON"] = ""; Pkg.clone("PyPlot"); Pkg.build("PyPlot")`, but its nice if it works.

What do you mean by splitting the computation from the plots? We could probably add a few test to functions that are only run in from within the plot functions, I agree with that. But do you also mean that we should separate the main code to have functions that first do all the computations and then all the plotting? This seems like its would clutter the code a lot, but there might be places where grid lines etc could be calculated in a different function which could be tested separately, is that what you are referring to?

tkelman:
If you just want to do design or synthesis calculations or simulations without any plots, then you don't technically need to have plotting packages installed. I guess with the most recent RecipesBase system, you can define plotting recipes with minimal dependencies and have the plotting done separately, but I'm not certain exactly how it all fits together.

tbreloff:
You could define recipes without having a dependency on Plots, but it does
make sense to test those recipes in the package they are created. In the
short term I think you should do what Tony is saying and only do the plot
tests if the reference images are installed. In the long term I hope the
visual tests are standardized and easy to set up so that everyone does it
without thinking about how.

PR #5547, opened 2016-07-05
Update ToeplitzMatrices url

andreasnoack:
Moved to an organization.


tkelman:
You'll probably need to update your badges and double-check that CI is all still running.

PR #5552, opened 2016-07-05
Tag new TerminalExtensions versions

Keno:
One for the last good version on 0.4, one for current master on 0.5


tkelman:
Could really use tests.

PR #5563, opened 2016-07-07
Register Extern: v0.0.1

ararslan:
I made a thing and I think it works, maybe other people will use it. 🎷🐴


ararslan:
Closed by #5567. 🔐

tkelman:
Looks like you don't have Travis CI enabled? Otherwise this looks fine, though note it has a lot in common with Requires.jl - particularly that it will interact poorly with precompilation, if the state of which packages are present changes between precompilation and usage.

ararslan:
Not sure what you mean about Travis... I have the badge on the repo and things look okay [here](https://travis-ci.org/ararslan/Extern.jl).

I wasn't aware of Requires! It appears to serve precisely the same purpose. I had thought about the issue you mentioned but wasn't sure how to get around it. Should I just ditch this then?

tkelman:
Whoops, was looking at the branch status page and didn't realize that you had done [ci skip] on your most recent commit, that's why it looked like it wasn't enabled.

Your package is a bit safer and doesn't do the same hacking of internals, so it's less bad than Requires. No harm registering this, but if anyone wants to start using it they should be aware of the precompilation issue. Something like an `extern` keyword (that would properly tie into the precompilation machinery) is planned, so maybe your package can be a proving ground for how to go about it.

PR #5567, opened 2016-07-07
Register Extern v0.0.1, tag StatsBase v0.8.3

ararslan:
I *finally* got `Pkg.publish()` to work! 🎉 However, it doesn't seem to care that I already have an outstanding PR for Extern (#5563) so I'll close that one in favor of this.

PR #5570, opened 2016-07-07
Tag StatsBase v0.8.3

ararslan:
Hooray, I did the thing. Supersedes #5567.


tkelman:
Did this break GLM.jl?  http://pkg.julialang.org/detail/GLM.html
```
ERROR: LoadError: test error in expression: R²(lm1) == R2(lm1)
MethodError: `R²` has no method matching R²(::DataFrames.DataFrameRegressionModel{GLM.LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredQR{Float64}},Float64})
Closest candidates are:
  R²(::StatsBase.StatisticalModel, !Matched::Symbol)
 in anonymous at test.jl:90
 in do_test at test.jl:50
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:320
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
 in anonymous at test.jl:90
 in do_test at test.jl:50
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:320
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
while loading /home/vagrant/.julia/v0.4/GLM/test/runtests.jl, in expression starting on line 24
```

ararslan:
Uhhhhh... well `R²` was deprecated in favor of `r²`. Could that have anything to do with it?

ararslan:
Looks like it might be because `DataFrameRegressionModel` doesn't inherit from `LinearModel`, and the one argument version of `r2` is for `LinearModel`s only and comes from GLM, not StatsBase.

tkelman:
How was the deprecation done? Was GLM previously extending a new method of the old name, but that broke now that it's deprecated? Seems like that should be possible?

ararslan:
The deprecation was done by replacing the definitions in StatsBase with the new versions and using the deprecation macro for the old names. GLM was indeed extending the deprecated functions with `LinearModel`s. I started a PR, JuliaStats/GLM.jl#146, to update the functions but that doesn't change the error.

ararslan:
Ohhhh I see what's happening. `DataFrames` is also overloading the deprecated functions. PR time.

tkelman:
Should we also add upper bounds to the past tags?

ararslan:
Not sure what you mean

tkelman:
Since this new version of StatsBase broke the tests of existing versions of GLM, we can modify the requires files of the GLM tags to reflect that this is an upper bound.

ararslan:
I've modified REQUIRE in my GLM and DataFrames PRs. How would I change it to an upper bound for all tags?

tkelman:
Modify the requires files in metadata for those versions.

ararslan:
How do I specify an upper bound in the requires in METADATA? Do I just make StatsBase say StatsBase 0.8.3 instead of whatever version is listed? Sorry for my ignorance, just not that familiar with how METADATA works.

tkelman:
Happy to explain. If there is one number, it's a lower bound. If there are no numbers, there are no bounds. So to add an upper bound, replace `StatsBase` with `StatsBase 0.0.0 0.8.3`, or if there is already a lower bound then just add the 0.8.3 upper bound. Upper bounds are exclusive.

ararslan:
Oh, okay, perfect. Thanks so much for the explanation! I'll do that for GLM and DataFrames and submit a PR on METADATA.

tkelman:
This package interaction is a bit complicated, so it may actually be the dependency from DataFrames on  StatsBase that is the issue causing GLM's tests to fail? If there is a new DataFrames tag after your PR gets merged, would it fix old GLM tags, or are PR's needed in both?

ararslan:
The failing `r2` test on GLM is caused by DataFrames it seems; GLM is actually testing a method specific to DataFrames. If my DataFrames PR gets merged and a new version is tagged, then I update the test/REQUIRE my GLM PR to use the new DataFrames version, that should fix everything.

ararslan:
"Should" being the operative word there.

tkelman:
Heh. With new PR's and tags of everything we can get it all passing again. As a second order thing I'm wondering which combinations of bounds can keep the old package tags also passing.

ararslan:
Well, at least passing on 0.4. The 0.5 failures all around seem to be from `Base.OneTo`.

tkelman:
The OneTo stuff is a recently inflicted mess that hopefully will get rectified in base one way or another.

PR #5573, opened 2016-07-08
Register PlotUtils

tbreloff:
PlotUtils is where I'm starting to collect utilities and helper functions which are generic enough to be shared by plotting/graphics packages.  I obviously care about using them for Plots, and so I need this to be registered if I am to add that dependency.

I have started to move some of the generic functionality in Gadfly here, starting with the `optimize_ticks` functionality (as suggested by @timholy... thanks again).  ref: https://github.com/JuliaPlots/PlotUtils.jl/issues/1

Also, I have just completed stripping out (and redesigning) most color-related functionality from Plots into PlotUtils, hence the need to register now.


tkelman:
Assuming you plan on adding docs of some kind?

tbreloff:
Yes someday.  But hopefully you'll consider merging before I add them.
Right now I don't expect this to be user-facing... I just need to register
the package so I can add it to the REQUIRE of Plots.

tbreloff:
Thanks @tkelman!

PR #5581, opened 2016-07-10
Tag GLM v0.5.4

ararslan:
@tkelman Step 2 of ~~world domination~~ fixing the 0.4 tests.


GLM/versions/0.5.4/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+Distributions 0.4.6-
+StatsBase 0.8.3
+Reexport
+Compat 0.2.12-
```
tkelman: that's getting a little long in the tooth, really not relying on anything newer?

ararslan: Huh. I guess not, since things seem to work, but I guess it would be easier to tell for sure if the `Base.OneTo` thing wasn't causing 0.5 to fail regardless of whatever else happens. I'll look for some 0.5 test logs from before `OneTo`.

I can bump Compat anyway to be on the safe side. That would require tagging another new version, yeah?

ararslan: Looks like the only current uses of Compat in GLM aren't even needed anymore. They're for 0.3-0.4 compatibility but GLM now requires 0.4 anyway. The 0.5 tests were passing a month ago and 15 days ago some model didn't converge on Travis and caused the tests to fail.

tkelman: Ah, fair enough. May as well leave it since it's almost guaranteed that some of the other dependencies need it, and GLM may need it again in the future at which point the minimum version can be raised to be accurate for whatever the new change ends up being.

PR #5582, opened 2016-07-11
Set StatsBase 0.8.3 as the upper bound on old versions

ararslan:
The third and final step in our master plan, @tkelman! Did I do it right?

### DIFF
--- a/DataFrames/versions/0.0.0/requires
+++ b/DataFrames/versions/0.0.0/requires
@@ -1,2 +1,2 @@
 Options
-StatsBase
+StatsBase 0.0.0 0.8.3
--- a/DataFrames/versions/0.1.0/requires
+++ b/DataFrames/versions/0.1.0/requires
@@ -1,3 +1,3 @@
 julia 0.1- 0.2-
 Options
-StatsBase
+StatsBase 0.0.0 0.8.3
--- a/DataFrames/versions/0.2.0/requires
+++ b/DataFrames/versions/0.2.0/requires
@@ -1,4 +1,4 @@
 julia 0.2-
 GZip
 Options
-StatsBase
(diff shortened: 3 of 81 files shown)


ararslan:
I only modified lines that matched the regex `^StatsBase [\d.]+$`.

tkelman:
This kind of thing needs to be done on all the old versions.

ararslan:
Oh, okay. What do `version-` and `version+` mean in the requires files? I assumed that syntax was from a previous version of METADATA. For ones that just say `StatsBase` can I set the lower bound at 0.0.0?

tkelman:
See the documentation on version number literals. `-` means prerelease, `+` means "last possible version" or something similar - basically neither are especially useful when referring to other packages

ararslan:
Gotcha, thanks!

ararslan:
I'm still getting the hang of what does and does not correspond directly with Julia things in requirement files. I appreciate your helpfulness and patience.

tkelman:
All good. So has every tag of DataFrames always depended on some version of statsbase? Looks like some of the early GLM tags did not.

From past times doing this, the resolver can do unexpected things sometimes. Instead of downgrading to a working StatsBase fot example, it might downgrade GLM to one of the very old versions that didn't depend on StatsBase to avoid the conflict. We can avoid that by adding a fake dependency on StatsBase (lb 0.0.0, ub 0.8.3) even to those old GLM versions.

ararslan:
Surprisingly, it looks like all versions of DataFrames since 0.0.0 have depended on StatsBase. I did not expect that!

I'll add some fake StatsBase stuff to the old GLMs momentarily.

ararslan:
Are we good now?

PR #5586, opened 2016-07-11
Tag GaussianProcesses v0.3.0

nemethis:


tkelman:
Be aware that your `@glue` macro, like most similar approaches to conditional dependencies at the moment, does not interact well with precompilation. If the packages present at load time do not match the packages present at precompile time, then you will get either incorrect functionality or an error.

PR #5590, opened 2016-07-11
Tag DataArrays v0.3.6

ararslan:
The one and only @tkelman has bestowed upon us a fix for DataArrays which should solve all of our OneTo woes.


tkelman:
I'm going to use my administrator powers and merge this without waiting for travis so it gets into tonight's pkgeval run

ararslan:
Cool. Fingers crossed that stats packages are no longer on fire and broken tomorrow.

tkelman:
With the premature rmath deletion they probably will be, but for new reasons. Never a dull day.

ararslan:
Could we temporarily revert the base rmath deletion until packages can catch up?

PR #5613, opened 2016-07-14
Set a maximum version of StatsFuns for Distributions

ararslan:
StatsFuns 0.3.0 introduces breaking changes to all current Distributions versions.

### DIFF
--- a/Distributions/versions/0.0.0/requires
+++ b/Distributions/versions/0.0.0/requires
@@ -1,1 +1,2 @@
 julia 0.1- 0.2-
+StatsFuns 0.0.0 0.3.0
--- a/Distributions/versions/0.0.0/requires
+++ b/Distributions/versions/0.0.0/requires
@@ -1,2 +1,1 @@
 julia 0.1- 0.2-
-StatsFuns 0.0.0 0.3.0
--- a/Distributions/versions/0.1.0/requires
+++ b/Distributions/versions/0.1.0/requires
@@ -1,2 +1,1 @@
-julia 0.2-
-StatsFuns 0.0.0 0.3.0
+julia 0.2- 0.3
--- a/Distributions/versions/0.1.0/requires
+++ b/Distributions/versions/0.1.0/requires
@@ -1,1 +1,2 @@
 julia 0.2-
+StatsFuns 0.0.0 0.3.0
--- a/Distributions/versions/0.1.1/requires
+++ b/Distributions/versions/0.1.1/requires
@@ -1,2 +1,1 @@
-julia 0.2-
-StatsFuns 0.0.0 0.3.0
+julia 0.2- 0.3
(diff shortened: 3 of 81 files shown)


tkelman:
I think we may need to add an artificial `StatsFuns 0.0.0 0.3.0` dependency to even earlier Distributions.jl version that supports `julia 0.3` or newer, otherwise the conflict may result in downgrading Distributions.jl to something from before it depended on StatsFuns.

ararslan:
Ah right, should have remembered that from last time. Will do.

tkelman:
I'm a little hesitant on what to do about any versions of Distributions that don't have a minimum Julia version, or have one that's pre-0.3.0. We should probably test this and see whether Distributions could ever get downgraded to anything REALLY old. Hopefully not, but we probably want to wait until after a Distributions tag that *does* support the new StatsFuns version before merging this.

ararslan:
A Distributions tag is in the works, ref https://github.com/JuliaStats/Distributions.jl/pull/512.

tkelman:
If that's going to take a while, we should maybe raise the julia minimum for statsfuns 0.3.0 temporarily to 0.5-.

For the very old Distributions versions with no julia dependency or before 0.3.0 so adding an extra statsfuns dependency would not work, we should add a julia 0.3 upper bound.

ararslan:
I'm hoping it won't be long. If someone says it's okay to merge my current Distributions PR then I'll go ahead and do that, tag a version, and update this. Hoping to get this done ASAP today.

ararslan:
Ref #5624.

andreasnoack:
What is required to resolve the `StatsFuns` situation?

tkelman:
we could either do fixes that say they're 0.5-only at first until distributions and this change are ready, or we could do the new tags in the same pr that loosens the statsfuns julia requirement back to allowing 0.4

ararslan:
We might have to upper bounds for StatsBase too.

ararslan:
Distributions versions 0.2.2-0.3.0 had no Julia version requirement but 0.2.1 requires Julia 0.2-, so I set 0.3 as an upper bound on Julia for those versions.

Distributions versions with no dependency on StatsFuns got a line that sets bounds at 0.0.0, 0.3.0. Existing StatsFuns dependencies got 0.3.0 as an upper bound if none existed, or were left as-is otherwise.

Is this what you had in mind, @tkelman?

ararslan:
I won't have much time to help 🔥💧🚒  for the next several hours but I'll try to check in occasionally if I can.

ararslan:
In the meantime I've added you two (@tkelman and @andreasnoack) as collaborators on my METADATA fork, you can make any changes to this PR if need be.

tkelman:
ok I'll adjust this when I get in front of a desk then

ararslan:
Thanks @tkelman!

PR #5615, opened 2016-07-14
Register IntPoint

mpf:
IntPoint.jl: A pure Julia conic programming solver. Includes, LP, SOCP, and SDP.

### DIFF
--- a/IntPoint/url
+++ b/IntPoint/url
@@ -1,0 +1,1 @@
+git://github.com/MPF-Optimization-Laboratory/IntPoint.jl.git


tkelman:
Hi @mpf! Very good to see a pure Julia solver.
It's a pretty generic package name though. Could this mention Conic in the name to make clear the problem classes it works on?

tkelman:
How closely is this based on cvxopt? Note that cvxopt is GPLv3 licensed https://github.com/cvxopt/cvxopt/blob/master/LICENSE so if this referred to the code at all it can't really be licensed under MIT.

tkelman:
You also need to mention all of the `using` or `import` dependencies in `REQUIRE`, and the test-only dependencies in `test/REQUIRE`. I suspect if you enabled Travis CI (which is very highly recommended) this package would fail there until that was fixed.

gabgoh:
hey tkelman! 

all the code has been written from scratch, though the implementation follows this document http://www.ee.ucla.edu/~vandenbe/publications/coneprog.pdf. I dont know the legalities of this, though it's not a line by line port.

I'm still doing the pull requests for one more dependency, will update that once it is successful. Will also enable traviscl

tkelman:
That document looks like it's all math and no code from a quick skim, so I suspect you're alright. To make things clearer you may want to rephrase "based on cvxopt" to "based on the algorithm used by cvxopt" and change the link to point to that paper. I think "ConicIntPoint" or "ConicInteriorPoint" or "IntPointConic" would be a more specific package name. Someone may write an algorithm similar to Ipopt and call it NonlinearInteriorPoint.jl, wouldn't want it to be too easy to confuse this package with that.

kmsquire:
From someone who knows just enough about optimization to hurt myself, +1 to `ConicInteriorPoint`.  (Really, I understand what ConicInteriorPoint means immediately when I see it, but `IntPoint` is totally obtuse to me...)

tkelman:
> just enough about optimization to hurt myself

:laughing: I'm going to have to use that somewhere.

dpsanders:
As far as I can see, cvxopt has a GPL license, so this package will also need to be GPL licensed. +1 to the name ConicInteriorPoint

mpf:
Agreed: `IntPoint` is a bit silly.  Any objection to `ConicIP` (a slightly shorter version of the proposed `ConicInteriorPoint`)?

tkelman:
I usually favor spelling things out rather than using acronyms, especially if it's pure julia code without an existing library.

gabgoh:
second attempt https://github.com/JuliaLang/METADATA.jl/pull/5837

mlubin:
Sorry I missed this discussion the first time around!

PR #5616, opened 2016-07-15
Register SymWoodburymatrices and BlockMatrices (IntPoint Dependencies)

gabgoh:
SymWoodbury.jl (Package for Symmetric Woodbury Matrices) and BlockMatrices.jl (Package for Block Diagonal Matrices)

### DIFF
--- a/BlockMatrices/url
+++ b/BlockMatrices/url
@@ -1,0 +1,1 @@
+git://github.com/MPF-Optimization-Laboratory/BlockMatrices.jl.git
--- a/SymWoodburyMatrices/url
+++ b/SymWoodburyMatrices/url
@@ -1,0 +1,1 @@
+git://github.com/MPF-Optimization-Laboratory/SymWoodburyMatrices.jl.git


tkelman:
Need to tag a version for these. Would it be possible to contribute SymWoodburyMatrices to the existing WoodburyMatrices package instead? How does your BlockMatrices package differ from github.com/KristofferC/BlockArrays.jl?

KristofferC:
If the package is only for block diagonal matrices shouldn't the word diagonal be in the package name?

gabgoh:
thanks for the input. yes, my package is for block diagonal matrices while blockarrays is more generic. I'm having second thoughts about releasing it as a standalone package, perhaps it should be just part of IntPoint, perhaps that would reduce the clutter? what do you think?

SymWoodburyMatrices however can be combined with WoodburyMatrices (they both perform the same function for the most part, but the latter is optimized for symmtric matrices). I'll put aside a few hours to doing that.

timholy:
This can be closed, right? (I will tag a new version of WoodburyMatrices.)

aviks:
@gabgoh are you happy closing this PR?

gabgoh:
closing this PR, all dependency issues have been resolved.

PR #5622, opened 2016-07-15
Register BioMedQuery

mirestrepo:


kmsquire:
Hi @mirestrepo, we now require that you tag a version when registering a package.  Cheers!

mirestrepo:
Hi @kmsquire, thanks for the response. I did, which corresponds to the next Pull Request #5623 . I was following the documentation which suggests to register, publish, tag then publish. I should probably had just register, tag and publish only at the end. If you prefer we can close this pull request and look at #5623. Thanks

kmsquire:
Thanks!  Yes, let's close this. The documentation needs to be updated to reflect current practice.

PR #5623, opened 2016-07-15
[WIP] Register BioMedQuery v0.0.1

mirestrepo:
Register v0.0.1


BioMedQuery/versions/0.0.1/requires
```diff
@@ -0,0 +1,11 @@
...
+ArgParse
+SQLite
+Gumbo
+Requests
+DataStreams
+LightXML
+Requests
+HttpCommon
+BaseTestNext
+ZipFile
```
ararslan: It'd probably be a good idea to set minimum versions for all of the package dependencies.

mirestrepo:
@ararslan I though about doing so, but got discourage because I don't see many of the packages in the repository doing that. Are there advantages to not setting minimum versions for all packages?

ararslan:
I don't know of any advantages to not setting them. The advantage of setting minimum versions is that it's easier to track down when certain versions of packages don't play nicely with each other or with a certain version of Julia.

tkelman:
Not setting a minimum version of any packages is effectively saying "I need some release version of this package but I don't care which." In reality you've probably been testing against the latest release versions of your dependencies over the time period you've been writing the package. So if there are any package API's from your dependencies that you think might be new or have changed over time, then it can be a good idea to look up when those were first instituted in the dependency package, since you know your package can't possibly use any earlier versions of the dependency than that.

tkelman:
It looks like your `test/runtests.jl` script is doing `Pkg.clone("https://github.com/bcbi/XMLconvert.jl.git")`. That's not really recommended practice, as it installs an unregistered package on the user's machine if they were to do `Pkg.add("BioMedQuery"); Pkg.test("BioMedQuery")`. What we can do instead would be to first register the XMLconvert.jl package, then once that is merged to METADATA you can add `XMLconvert` to a `test/REQUIRE` file in BioMedQuery.jl.

mirestrepo:
@tkelman thank you for the feedback. Yes, you're correct that I need the XMLconver.jl package. I wasn't aware I should only use registered packages - so I'll go ahead and work on registering that one. I'll also add the minimum versions on my dependencies. Is it best to close this issue and open a new one later with the updates?

tkelman:
Whatever's easier for you - close this and reopen a new one later, or leave this open but in a WIP state until you can update it.

mirestrepo:
Submitted request to register the dependency XMLconvert.jl #5643

tkelman:
@mirestrepo now that XMLconvert is registered, just let us know when you'd like to move forward here. Has BioMedQuery been updated yet to add XMLconvert to its REQUIRE file?

mirestrepo:
@tkelman I opened a new request #5806 with the suggested changes. I'll close this one.

PR #5625, opened 2016-07-15
Tag StatsBase v0.9.0 [git://github.com/JuliaStats/StatsBase.jl.git]

andreasnoack:


ararslan:
Thanks for doing this!

ararslan:
StatsBase is now broken because of https://github.com/JuliaLang/METADATA.jl/pull/5624. StatsBase lists StatsFuns 0.3.0 as a requirement but StatsFuns 0.3.0 has been changed to require Julia 0.5, so now StatsBase fails on 0.4.

tkelman:
it should just mean that this version won't be installable

ararslan:
Okay. It's failing on master, didn't know whether that would affect other things.

Travis log on master for reference: https://travis-ci.org/JuliaStats/StatsBase.jl/jobs/145101793

tkelman:
only devs or people who've done Pkg.checkout should ever hit that

tkelman:
the metadata check here might fail though, not sure yet

ararslan:
```
    StatsBase v0.9.0 – no valid versions exist for package Compat
```
??? Compat 0.8.4 is specified in the requires and is the most recent release

tkelman:
I suspect the error message is wrong and confused. Probably StatsFuns is the actual culprit. Worth digging into and seeing if we can make that message more accurate

ararslan:
The only version oddity I can see is that StatsFuns 0.3.0 is Julia 0.5 only and StatsBase requires StatsFuns 0.3.0 but allows Julia 0.4. Beyond that, I'm stumped as to what would cause that kind of failure regarding versions.

PR #5632, opened 2016-07-18
Revert "Make StatsFuns 0.3.0 require Julia 0.5- temporarily"

tkelman:
Reverts JuliaLang/METADATA.jl#5624

once we do https://github.com/JuliaLang/METADATA.jl/pull/5613 then this should be safe again. @ararslan or @andreasnoack correct me if I'm wrong.


ararslan:
Doesn't sound wrong at all to me. 👍

PR #5634, opened 2016-07-18
Tag Distributions v0.10.0 [git://github.com/JuliaStats/Distributions.jl.git]

andreasnoack:
First post Rmath release


tkelman:
It looks like the changes in this tag broke the following downstream packages on Julia 0.4, so we should add upper bounds on the Distributions dependency of their past tags:
- GaussianMixtures, cc @davidavdav http://pkg.julialang.org/detail/GaussianMixtures.html
- IndependentRandomSequences, cc @gajomi http://pkg.julialang.org/detail/IndependentRandomSequences.html
- Mamba, cc @brian-j-smith http://pkg.julialang.org/detail/Mamba.html

davidavdav:
Yes, thanks for noticing (BTW how do you do that? do you have reverse-dependency lists?), I got [#19](https://github.com/davidavdav/GaussianMixtures.jl/issues/19) but got the Distributions dependency fixed in [9f84448](https://github.com/davidavdav/GaussianMixtures.jl/commit/9f8444845882ae325281d9fd0d724e708cabfcc0), still working on 0.5, mostly waiting on dependencies.

tkelman:
http://pkg.julialang.org/pulse.html

We can adjust the dependencies of the past tags here in metadata so users don't see the issue.

brian-j-smith:
Thanks for noting the effect of Distributions 0.10.0 on Mamba.  I've pushed a new version at #5655 to fix the issue.

PR #5638, opened 2016-07-19
LightGraphs: Add 0.9.0 as an upper bound on StatsBase requirement

ararslan:
0.9.0 introduces breaking changes.

Ref https://github.com/JuliaGraphs/LightGraphs.jl/pull/383#issuecomment-233534888

### DIFF
--- a/LightGraphs/versions/0.1.0/requires
+++ b/LightGraphs/versions/0.1.0/requires
@@ -1,5 +1,5 @@
 julia 0.3
 Compat 0.2.6
 GZip 0.2
-StatsBase 0.6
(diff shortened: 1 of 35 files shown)


tkelman:
Cool. What about 0.1.11 - 0.1.14 ?

ararslan:
No dependency on StatsBase. Should I add a fake one?

tkelman:
yeah I think so, how odd that it was removed then came back?

ararslan:
Yeah, pretty weird. I'll set the lower bound as whatever the last lower bound was before the dependency mysteriously disappeared.

PR #5642, opened 2016-07-19
Tag Rmath v0.1.2

tkelman:
adds windows support


Rmath/versions/0.1.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BinDeps
```
ararslan: No minimum version?! 😜 Could use 0.3.22 (newest)

tkelman: well, older should work fine. not sure how much older, would have to check.

PR #5646, opened 2016-07-20
Register Cxx

Keno:
A humble package to interface Julia with C++.


tkelman:
License should acknowledge any code copied from base.

Keno:
There isn't any.

tkelman:
makefiles.

You also need a REQUIRE file in the package itself.

Cxx/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+192a3fd1b7696ceb4e4ff565cbf5b9725c881cad
```
tkelman: this commit isn't pushed to a branch?

Keno: It's the one that added the REQUIRE file ;).

Keno:
https://github.com/Keno/Cxx.jl/commit/f6bc213b335e3c87a6dd2a08a0846813a259abfe for LICENSE adjustments, though I should note that I consider the parts of the Makefiles fair use, and nothing in that commit should be construed to the contrary.

PR #5647, opened 2016-07-20
Register and tag MappedArrays

timholy:
A new package that provides a "view" `M` of an array `A` so that `M[i] == f(A[i])` for a specified (but arbitrary) function `f`.


timholy:
Any concerns about the name or anything else, or OK to merge?

tkelman:
I leave new package registrations open for 24 hours of initial comments unless specifically requested otherwise, but this looks good.

PR #5651, opened 2016-07-20

dysonance:
**Temporal**: added find methods and a not operator for the TS type to act similarly to an Array

**Indicators**: added functionality for finding local maxima and minima (with an added example in the readme) as a first step into trendline/support/resistance algorithm development.


Temporal/versions/0.0.4/sha1
```diff
@@ -1 +0,0 @@
-619f268f172f795766df40bcdf783e160b62432d
```
tkelman: you shouldn't delete existing versions like this

dysonance: i hadn't attempted to do this -- i imagine it has to do with how for a while i was doing the git work manually because i've only been able to get `Pkg.publish` to work recently. is there a way to roll back this change?

tkelman: probably easiest to just manually put them back as they were in https://github.com/dysonance/METADATA.jl/tree/pull-request/4c65f4ca/Temporal/versions

dysonance: @tkelman I tried doing that, but as expected, `Pkg.update` didn't like it. Is there another way of doing this?

dysonance: @tkelman I think I may have figured it out. Closing this PR and opening a new one (god willing, that one will be good to go)

PR #5672, opened 2016-07-22
Initial version of CBOR.jl

saurvs:


CBOR/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+5eef91dbb3ea8e025546352ebfd499d9c0577c1a
```
tkelman: this commit should be on a branch, not just a tag https://github.com/saurvs/CBOR.jl/commit/5eef91dbb3ea8e025546352ebfd499d9c0577c1a

PR #5673, opened 2016-07-22
Tag XSim v0.1.0

reworkhow:


XSim/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+DataFrames
+Distributions
+JWAS
```
tkelman: does this depend on the latest version?

tkelman:
you should consider adding some real tests to https://github.com/reworkhow/XSim.jl/blob/v0.1.0/test/runtests.jl

XSim/versions/0.1.0/requires
reworkhow: It needs latest JWAS

reworkhow:
real tests have been added

XSim/versions/0.1.0/requires
tkelman: then you should add a minimum version requirement here

XSim/versions/0.1.0/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.4
 DataFrames
 Distributions
-JWAS
+JWAS 0.1-
```
tkelman: This is good now, but just so you're aware, the trailing `-` means "any prerelease of JWAS 0.1.0 or newer."

PR #5677, opened 2016-07-22
Register and tag Dtree

kpamnany:
Used by Celeste.


tkelman:
If you enabled Travis CI would the package work there? Could you run the bench file you've written as part of runtests.jl?

dpsanders:
Searching for `dtree` on Google does not seem to find the package that this is a wrapper for. Is there a better option for the package name?

dpsanders:
In fact, there seems to be no information about the original package in the repository, nor in the LICENSE file.

kpamnany:
@tkelman: The microbenchmark is essentially an MPI application and has to be started with ``mpiexec`` on a few nodes/cores. I wasn't able to work out how to do that. I'll play with it some more when I get a chance, but this code is pretty well tested already.

@dpsanders: The ``README.md`` file in the repository cites the paper that describes this implementation. If you search for ``dtree`` together with my name on Google Scholar, this paper comes up.

tkelman:
> this code is pretty well tested already.

Would be good to hook that up into a standard unit test suite so that's easy to verify. You should be able to get one or multiple of the open-source mpi implementations running on Travis.

PR #5683, opened 2016-07-23
ValidatedNumerics ForwardDiff bounds

tkelman:
This is basically the same as PR https://github.com/JuliaLang/METADATA.jl/pull/5682 but for ValidatedNumerics instead of Roots, for the same reason (the usage of ForwardDiff in the two packages is almost identical...). A new tag of ValidatedNumerics that includes https://github.com/dpsanders/ValidatedNumerics.jl/pull/148 will make this not visible to users, so would be best if that could be done before merging this. cc @dpsanders @lbenet


dpsanders:
See https://github.com/JuliaLang/METADATA.jl/pull/5691 for a new tag of ValidatedNumerics.

PR #5685, opened 2016-07-23
Tag Playground v0.0.6

Rory-Finnegan:


Rory-Finnegan:
I'm aware of this CI issue, but I dropped support for 0.3 a couple of releases ago and I don't intend to fix it now. Perhaps in the future I'll just remove the releases that supported 0.3.

tkelman:
No, releases should not be removed once published. Just rename this tag to 0.1.0.

PR #5686, opened 2016-07-23
Pull request/a24a7f74

montyvesselinov:
Many fixes and improvements


tkelman:
Does this require a minimum version of Optim or any of your other dependencies?

You should consider adding some tests to RobustPmap.

PR #5688, opened 2016-07-24
Tag TikzPictures v0.3.3

mykelk:
Makes TikzPictures compatible with TexLive 2016 while maintaining compatibility with earlier versions.


tkelman:
could really use some tests in https://github.com/sisl/TikzPictures.jl/blob/master/test/runtests.jl

PR #5693, opened 2016-07-24
Register Paths

Rory-Finnegan:
A pkg for working with filesystem Paths as types. I've found this help to better utilize multiple dispatch (ie: dispatch based on path or string rather than always checking in an if statement).

Some highlights include:

Path creation:
```
julia> Path("~/repos/Paths.jl/")
Paths.PosixPath(("~","repos","Paths.jl",""))
```
or
```
julia> p"~/repos/Paths.jl/"
Paths.PosixPath(("~","repos","Paths.jl",""))
```

Human readable file status info:
```
julia> stat(p"README.md")
Status(
  device = 16777220,
  inode = 48428965,
  mode = -rw-r--r--,
  nlink = 1,
  uid = 501,
  gid = 20,
  rdev = 0,
  size = 1880 (1.8K),
  blksize = 4096 (4.0K),
  blocks = 8,
  mtime = 2016-02-16T00:49:27,
  ctime = 2016-02-16T00:49:27,
)
```

Working with permissions:
```
julia> m = mode(p"README.md")
-rw-r--r--

julia> m - readable(:ALL)
--w-------

julia> m + executable(:ALL)
-rwxr-xr-x

julia> chmod(p"README.md", "+x")

julia> mode(p"README.md")
-rwxr-xr-x

julia> chmod(p"README.md", m)

julia> m = mode(p"README.md")
-rw-r--r--

julia> chmod(p"README.md", user=(READ+WRITE+EXEC), group=(READ+WRITE), other=READ)

julia> mode(p"README.md")
-rwxrw-r--

```

Reading and writing directly to file paths:
```
julia> write(p"testfile", "foobar")
6

julia> read(p"testfile")
"foobar"
```


dpsanders:
This looks nice, but the package name is too generic IMO. (What came to my mind was paths in the sense of drawing.) How about `FilePaths.jl`?

Rory-Finnegan:
@dpsanders I agree that Path is kind of a general term, but I think in most programming languages the default meaning is filesystem paths. Since this is a pkg and not part of the base language I could be convinced of switching it to `FSPaths.jl` (it's less clean, but more descriptive).

tkelman:
Base deprecated using the FS abbreviation, it's a bit jargon-y

Rory-Finnegan:
In that case, the only other pkg name I'd feel comfortable with is FileSystem.jl. This would cover both the current features (ie: Paths, File attributes and permissions) and future features like (`mount`, `unmount`, `df`, `du`, etc).

tkelman:
Then you'd be very nearly overlapping with the base module name. Would mounting and other utilities be in the same package?

Rory-Finnegan:
Probably. I don't really know where else I'd put them. I just liked `Paths.jl` cause the Path type would be the dominant type in the package even with the other utilities.

Is almost overlapping with a module in base a big problem? I can't really think of any better names.

1. `Paths.jl` - too generic
2. `FilePaths.jl` - seems awkward to me (makes it sound like it only works for files and not directories for some reason).
3. `FSPaths.jl` - FS is too jargon-y
4. `FileSystem.jl` - almost overlaps with `Filesystem` in base. If it helps it could be 'FileSystems.jl`, since the future utilities would be looking at more than one filesystem.

tkelman:
differing only in case could be an issue on case insensitive operating systems

Rory-Finnegan:
Would `FileSystems.jl` work though?

tkelman:
My preference would be for FilePaths, then Paths, and FileSystems least. Being too close in naming to the base module, even with the added s, would also be confusing imo.

Rory-Finnegan:
Alright, 2 to 1, if nobody else speaks up between now and tomorrow I'll close this PR, change the pkg name to `FilePaths` and re-register.

StefanKarpinski:
Btw, this is a really interesting package. I've often contemplated a more type-based approach to path and file APIs but just never got around to experimenting with it. I believe that Racket has this and I've been told that it's very useful. What have you found the main benefits to be?

Rory-Finnegan:
1) So, the main benefit I've been finding in julia is that I can dispatch on a Path type. A good general example would be parsing JSON. You could have 2 `load` methods rather than `load` and `loads` like in python (1 that takes a string and the other takes a path). 

2) I've also been finding that it helps with code readability, so I might update Playground.jl to use Paths rather than Strings.

3) The methods for manipulating paths are a little nicer now ie: `norm(::Path)` rather than `normpath(::String)`

PR #5694, opened 2016-07-25
Upgrade HTTP2 and HPack to v0.0.2

sorpaas:
This adds channel interface, HttpCommon integration, lots of bug fixes and refactoring.


tkelman:
Does HTTP2 pass its tests locally? When you tag interdependent packages it's usually better to do thrm one at a time so you can verify the tests still pass on travis.

PR #5697, opened 2016-07-25
Tag LightGraphs 0.6.0

sbromberger:
I don't know why it's tagging 0.5.1 as well, but 0.6.0 is the latest.


tkelman:
Can you revert the 0.5.1 commit here if it's not needed? Also note that if you're using the new syntax in JuMP, you need a minimum version requirement.

sbromberger:
I don't know why it was included - I used `Pkg.tag("LightGraphs", v"0.6.0")` and then `Pkg.publish()`.

As far as new JuMP syntax, I don't think we are. In any case we're probably going to move the (few) functions that rely on JuMP into their own package relatively soon. This was just to get rid of Requires.jl which was causing problems on 0.5.

tkelman:
you had the commit on your local branch but didn't tag it before. shouldn't tag it now as it would be broken w.r.t. statsbase

sbromberger:
OK, but 0.5.1 was a long time ago so I don't know why it's showing up now. In any case, how do I fix this?

tkelman:
non camel case getobjectivevalue is new jump syntax, needs a minimum version

tkelman:
either revert the commit in this branch, or make a new branch that has just the 0.6.0 commit cherry picked

sbromberger:
I'll just tag 0.6.1 with the JuMP version in it. Closing this out.

PR #5702, opened 2016-07-25
Register FilePaths

Rory-Finnegan:
Renamed Paths.jl to FilePaths.jl as discussed in PR #5693


Rory-Finnegan:
Do folks have any other concerns about this package or can it be merged?

PR #5707, opened 2016-07-26

dysonance:
Attempting to fix issues associated with previous [pull request](https://github.com/JuliaLang/METADATA.jl/pull/5651).

**Temporal**: added find methods and a not operator for the TS type to act similarly to an Array

**Indicators**: added functionality for finding local maxima and minima (with an added example in the readme) as a first step into trendline/support/resistance algorithm development.


Temporal/versions/0.0.5/sha1
```diff
@@ -0,0 +1 @@
+619f268f172f795766df40bcdf783e160b62432d
```
tkelman: this commit is the 0.0.4 git tag, not 0.0.5?

Temporal/versions/0.0.5/sha1
```diff
@@ -1 +1 @@
-619f268f172f795766df40bcdf783e160b62432d
+ce3889e1a2edeee071aa810a024a56d258809075
```
dysonance: @tkelman Apologies, I see the difference now. Having put this commit into the PR, have I worked out the kinks now?

(Apologies again for my ignorance, the git learning curve is beginning to make vim look like a piece of cake...)

tkelman: this should be fine

Indicators/versions/0.0.3/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Temporal
```
tkelman: is there a specific minimum version requirement that it assumes?

dysonance: It assumes Julia 0.4, as I have not tested it on previous versions. I don't necessarily think it would have problems with previous versions, as it is all mostly loops, but as I haven't tested it on older versions it's hard to say for sure. Problem?

tkelman: I meant does it assume a minimum version of your other package Temporal.jl, would anything break if you were to somehow use older versions of Temporal along with newer versions of Indicators?

dysonance: No that shouldn't be a problem as far as I know. It just adds methods for the TS type, basically just using the constructor. So older versions should be no problem.

PR #5708, opened 2016-07-26
Tag SHA v0.2.0

staticfloat:


staticfloat:
Note that this change to `SHA.jl` has a small incompatibility; the return type of hashing functions such as `sha256()` has changed from strings to arrays of bytes, use the `bytes2hex` function to convert the bytes to a string if that is desired.  See [this patch to `BinDeps.jl`](https://github.com/JuliaLang/BinDeps.jl/commit/64c591da0cabafc2e319b2ef0dedc18a42d34c95) for an easy compatibility layer to allow your code to work with both at once.

tkelman:
This broke FunctionalDataUtils http://pkg.julialang.org/detail/FunctionalDataUtils.html, we should upper-bound its past versions

PR #5711, opened 2016-07-26
Tag Nemo v0.5.0 [git://github.com/Nemocas/Nemo.jl.git]

wbhart:


tkelman:
This broke Hecke.jl http://pkg.julialang.org/detail/Hecke.html - what we can do to fix this is adjust the version dependency of existing Hecke.jl tags to set Nemo 0.5 as a maximum (exclusive) version.

PR #5717, opened 2016-07-26
Register LGmatching

sbromberger:
LGmatching is the first of several efforts to reduce the dependencies of the core LightGraphs.jl package. In this case, JuMP was required for a single matching function.

### DIFF
--- a/LGmatching/url
+++ b/LGmatching/url
@@ -1,0 +1,1 @@
+git://github.com/JuliaGraphs/LGmatching.jl.git


tkelman:
You need to tag a version - should this have a more specific name, GraphMatching or something? There isn't much context around what "LG" means here.

sbromberger:
This is specific to LightGraphs and will not work with other packages or standalone. We have made the decision internally to use LG as a prefix for all dependent packages. There will be more of these.

tbreloff:
IMO it should be something like LightGraphsMatching.jl if it's only meant
to separate LightGraphs-specific functionality.

sbromberger:
I'm really unwilling to go through a bikeshedding exercise, and I don't think we even need this registered with METADATA, so I'll close this out and keep it repo-only for folks who want it.

CarloLucibello:
I don't like the name LGmatching as well

PR #5720, opened 2016-07-27
Tag Hiccup v0.0.3

ranjanan:
Fixes https://github.com/JunoLab/Hiccup.jl/issues/17

### DIFF
--- a/Hiccup/versions/0.0.3/requires
+++ b/Hiccup/versions/0.0.3/requires
@@ -1,0 +1,4 @@
+julia 0.3
+Lazy
+MacroTools
+Compat 0.8.2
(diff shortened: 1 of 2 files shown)


Hiccup/versions/0.0.3/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.3
```
tkelman: is this actually still tested on 0.3?

ranjanan: Yep.

TotalVerb: Yes

PR #5722, opened 2016-07-27
Add bounds for dependent packages of SHA and Nemo

tkelman:
To account for breaking changes in those packages, cc @staticfloat and @wbhart


staticfloat:
Thanks, @tkelman

PR #5723, opened 2016-07-27
Register packages that simplify custom indices

timholy:
CustomUnitRanges allows packages to easily "define" a custom `AbstractUnitRange` unique to their package, and thus control the behavior of `similar` and related functions.

EndpointRanges implements indexing like `a[ibegin+3:iend-2]`, and thus simplifies working with subsets of arrays in situations where indexing may not start with 1.


timholy:
That travis test completed so quickly, it almost makes me worried that the flu has overtaken 50% of the julia community.

tkelman:
EndpointRanges.jl doesn't work on 0.4, it shouldn't claim to support it http://pkg.julialang.org/logs/EndpointRanges_0.4.log

PR #5724, opened 2016-07-27
Register GroupSlices

AndyGreenwell:
Moving JuliaLang [PR #15503](https://github.com/JuliaLang/julia/pull/15503) into a package.

### DIFF
--- a/GroupSlices/url
+++ b/GroupSlices/url
@@ -1,0 +1,1 @@
+git://github.com/AndyGreenwell/GroupSlices.jl.git


tkelman:
You'll need to tag a version

AndyGreenwell:
If the `Pkg.tag` operation needs to occur before the `Pkg.publish` operation, then we should change the order in which these items are discussed in the [docs](http://docs.julialang.org/en/release-0.4/manual/packages/#tagging-and-publishing-your-package)

tkelman:
Yes.

AndyGreenwell:
Closing this PR.  Will open anew with a tag.

PR #5729, opened 2016-07-28
Tag Measures v0.0.3

shashi:


tkelman:
It's better if you leave package tags open for correctness review rather than immediately merging your own.

tkelman:
The fact that you didn't list Compat in REQUIRE means this package in fact will not work on its own.

PR #5735, opened 2016-07-28
Tag GtkBuilderAid v0.0.6

Matt5sean3:
* Adds precompilation
* Allows connecting signals to already created `GtkBuilder` objects
* Improved documentation
* Improved multiple dispatch support


Matt5sean3:
Tests are completely failing on GtkBuilderAid's tagged commit, I'll tag and publish again after sorting out those problems.

PR #5737, opened 2016-07-28
Bump AxisAlgorithms and place upper bound on julia version for prior releases

timholy:
Includes 0.5 fix by @carlobaldassi, and implements a recommendation to put an upper bound on the previous releases.


tkelman:
you probably don't want the trailing dash in an upper bound. that excludes the very first commit of 0.5-dev.

timholy:
What would you suggest instead? I don't really feel the need to start doing a `git-bisect` to figure out where the first breakage occurred. What's wrong with being conservative?

tkelman:
I would leave off the dash, or do rc0 if you want it to take effect immediately.

timholy:
OK now?

tkelman:
removing the dash on the lower bound for existing tags would make the package no longer installable on prereleases of 0.4 where it might have been working? probably doesn't affect anyone, but I'm usually more targeted about touching lower bounds

timholy:
(You're annoyingly good at this...what would we ever do without you? :stuck_out_tongue_winking_eye:) How about this?

tkelman:
When things quiet down post-release I'll try to write a kelman-bot to automate most of the things I check for in review here. This version looks great now.

timholy:
Seems like it should be possible to harness PkgEval for doing julia version recommendations.

tkelman:
I (or someone else) will need to implement the easy parts of https://github.com/JuliaCI/PackageEvaluator.jl/issues/131 first to make that feasible, but yes.

PR #5738, opened 2016-07-28
Pull request/cc84bca0

spencerlyon2:
Ref: https://github.com/spencerlyon2/CompEcon.jl/issues/26

### DIFF
--- a/CompEcon/versions/0.1.0/sha1
+++ b/CompEcon/versions/0.1.0/sha1
@@ -1,0 +1,1 @@
+c2f91cfd4e43c6a9768d15b49d3e0b2da22e02ac
--- a/CompEcon/url
+++ b/CompEcon/url
@@ -1,0 +1,1 @@
+git://github.com/spencerlyon2/CompEcon.jl.git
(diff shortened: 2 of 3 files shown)


tkelman:
Does this need the matlab files to be checked into the repo? Could it be a submodule (if upstream uses git?) or link and not hurt anything?

spencerlyon2:
Hey @tkelman sorry for the delay, I'm just seeing this message. It probably doesn't need matlab files in the repo. I've removed them. and updated this PR. 

@davidanthoff when I removed the matlab files I did a force push to the remote, git will complain about the remote and your local having diverged. It's probably best to just delete your local copy and re-clone.

CompEcon/versions/0.1.0/sha1
```diff
@@ -0,0 +1 @@
+c2f91cfd4e43c6a9768d15b49d3e0b2da22e02ac
```
tkelman: can you push the git tag for this? https://github.com/spencerlyon2/CompEcon.jl/releases

spencerlyon2: Yep should be done now.

PR #5739, opened 2016-07-28
Tag WCS v0.1.2

kbarbary:


tkelman:
can you push the git tag so it shows up under https://github.com/JuliaAstro/WCS.jl/releases ?

kbarbary:
Done. (I usually wait to push the tag, in case there is a problem with the METADATA PR...)

tkelman:
Oh okay. As a note, these WCS packages almost always time out on PackageEvaluator due to the flaky ftp download. Could we set up a more reliable mirror for them?

kbarbary:
Recommendations? cache.julialang.org? Add tarball to releases on WCS.jl repo?

tkelman:
Yeah we can add it to the whitelist here https://github.com/staticfloat/cache.julialang.org/blob/master/cache.py - though I'm not positive whether that will work perfectly with ftp, may need to try and see.

PR #5742, opened 2016-07-29
Register StaticArrays

andyferris:
This is a package for making and using statically sized arrays for Julia 0.5.

Compared to the existing (and quite capable) *FixedSizeArrays* package, we have that `StaticArray` is an `AbstractArray`, which simplifies compatibility. The code is also simpler and runs slightly faster compared to *FixedSizeArrays*, because this was designed from the ground up to use the Julia 0.5 features.

FYI: @c42f @SimonDanisch 

EDIT: It also has built-in higher-dimensional and mutable containers, and can use pointers to interact with BLAS and LAPACK where appropriate.


c42f:
I am looking forward to kicking the tires on this one, yet another reason to promptly move to 0.5 :-)

PR #5749, opened 2016-07-29
Bump Reactive, HDF5, and Grid; deprecate Grid

timholy:


timholy:
OK, @shashi? It has been a little while since Reactive has been tagged.

shashi:
Thanks for doing this :+1:

tkelman:
@timholy this is the cause of https://github.com/JuliaStats/KernelDensity.jl/issues/36. Basically any package where every tagged version depends on Grid is now effectively deprecated as well, and any package where only recent tags depend on Grid will end up installing really old versions to avoid it.

- ElasticFDA only has tagged versions that depend on Grid, cc @jdtuck
- GridInterpolations only has tagged versions that depend on Grid, cc @mykelk
- KernelDensity's recent versions depend on Grid, cc @simonbyrne
- PerceptualColourMaps only has tagged versions that depend on Grid, cc @peterkovesi
- Seismic's recent versions depend on Grid, cc @aaronstanton
- SloanDigitalSkySurvey only has tagged versions that depend on Grid, cc @rgiordan

Not many people should be using 0.6-dev yet, but if these package authors (or any of their users) do, they may be in for a surprise here. We should probably make it a policy that when you deprecate a package, you notify the authors of packages that depend on it as a courtesy.

timholy:
Good point, sorry about that folks: I was thinking more about users than packages, and I was definitely remiss in failing to notify you.

If people really do want to continue using Grid, we can undeprecate it. Someone else has to maintain it, though: I think Interpolations is better in every respect, so I don't intend to put any work into keeping Grid going.

peterkovesi:
I have no problem with using Interpolations for PerceptualColourMaps.  It needs a little maintenance and tagging anyway.  I am unable to do this for the next couple of weeks because  I am travelling with minimal computing access  but that work can wait.

rgiordan:
We have wrapped the functionality from SloanDigitalSkySurvey into the main
project, and AFAIK nobody else is either.  I think I will delete it.  I'm
fine with deprecating Grid.  Thanks for reaching out!

Ryan

tlnagy:
> We should probably make it a policy that when you deprecate a package, you notify the authors of packages that depend on it as a courtesy.

Is it possible to do this automatically? Like have travis fire off some emails when a package deprecation affects other packages? Can't think of a good way of making this work atm

tkelman:
@rgiordan please do not delete the package, but we can deprecate it as was done here if it's no longer useful and nothing depends on it.

tkelman:
> Is it possible to do this automatically?

Could certainly write some scripts to automate the process of what I did above.
1. grep metadata (or use Pkg.dependents) to find the dependents of the to-be-deprecated package
2. check the version numbers of those dependents to see whether all or some tagged versions depend on the to-be-deprecated package - if only some, check whether or not that includes the most recent tags (if not, Pkg.dependents might have left it out but I'll have to check the source to be sure)
3. check the url and query the github API (assuming the package is hosted on github, all but a handful currently are) to find the most frequent contributors who should be notified
4. pre-populate the message content to post in the PR that proposes deprecation (and/or file issues in the individual dependent packages)

jdtuck:
Thanks for the heads up, I think we need to automate this somehow. I can move to interpolations and will make the changes soon to be ahead of the curve for ElasticFDA.jl. As Julia user base grows this could become more of a problem.

PR #5751, opened 2016-07-29
ApproxFun 0.2.2

dlfivefifty:
• Adds ApproxFun 0.2.2 to fix compatibility with BandedMatrices 0.0.6
• Add upper bounds to old versions of ApproxFun and SingularIntegralEquations


SingularIntegralEquations/versions/0.1.0/requires
```diff
@@ -1,2 +1,2 @@
 julia 0.4
-ApproxFun 0.1
+ApproxFun 0.1 0.2-
```
tkelman: so no prereleases of 0.2 at all? the upper bound is exclusive

dlfivefifty: Yes that's intentional.

PR #5758, opened 2016-07-30
Register BGZFStreams: v0.0.1

bicycle1885:
This registers https://github.com/BioJulia/BGZFStreams.jl.


bicycle1885:
Ping @tkelman. Can you merge this one?

tkelman:
This looks fine, was waiting the usual 24-hour waiting period. Do you have something that depends on this you need to get moving on sooner?

dpsanders:
I can't see anywhere where the acronym "BGZF" is spelled out.

bicycle1885:
Ah, I didn't know 24-hour waiting period. I don't hurry but I will create a pull request that depends on this package this week.

bicycle1885:
@dpsanders Yes, I think "BGZF" is an acronym, but I can't find an official declaration about it in the specs (https://samtools.github.io/hts-specs/SAMv1.pdf, chapter 4).

tkelman:
I'm guessing, but probably "block gzip format" ?

dpsanders:
Seems so: http://blastedbio.blogspot.com/2011/11/bgzf-blocked-bigger-better-gzip.html

bicycle1885:
Yes, highly probable. I think this acronym is very common and other libraries in other languages use this name:
* Perl: [Compress::BGZF](https://metacpan.org/pod/Compress::BGZF)
* Python: [Bio.bgzf](http://biopython.org/DIST/docs/api/Bio.bgzf-module.html)
* Haskell: [bgzf](https://hackage.haskell.org/package/bgzf)
* Ruby: [bio-bgzf](https://rubygems.org/gems/bio-bgzf/versions/0.2.1)
* C: [bgzf.c](https://github.com/lh3/samtools/blob/master/bgzf.c)

One exception is Java; its common library uses `BlockCompressedInputStream` and `BlockCompressedOutputStream` as a class name, but these are parts of a large package and not package name itself: https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/samtools/util/BlockCompressedInputStream.html.

dpsanders:
It just shows exactly why acronyms are so useless if no-one actually knows what they mean!

bicycle1885:
What I meant is "BGZF" is a proper noun and we cannot find a non-acronym official name of it. So, I don't think naming it as  `BlockedGNUZipFormatStreams` or something just by guess is a good idea.

ViralBShah:
I think acronyms are only meaningful in their own fields. This sounds uncommon enough that it is unlikely to clash. If anything, rather than the full name, I would make it BioBGZFStreams.jl

tkelman:
This isn't a bioinformatics specific format though, just more commonly used in that field as a consequence of what it does, right?

bicycle1885:
@tkelman Exactly. At first, I developed it as a submodule of Bio.jl but I thought it is not specific to biology, so I separated it as an independent package.

PR #5761, opened 2016-07-31
Tag DifferentialEquations.jl v0.1 and GrowableArrays v0.0.5

ChrisRackauckas:
GrowableArrays.jl got a small bugfix.

DifferentialEquations.jl is having its first minor release. The documentation is very complete and IJulia notebooks were added as well. The dependencies are way down and a method for using conditional dependencies was used to wrap all of the functionality of ODEInterface.jl and ODE.jl. In addition, many API improvements via the array interface and controlling the IO make the package much more user friendly. Also, there are some major additions in terms of solvers, including an order 14 Runge-Kutta solver.


tkelman:
Do your tests depend on having master or other branches checked out in dependent packages? That isn't what users are going to get from `Pkg.add` or `Pkg.update`.

Doing conditional dependencies in a try-catch is not going to work well in combination with precompilation. If the set of available dependencies at precompile-time does not match the available dependencies at runtime, you will either get reduced functionality or errors.

ChrisRackauckas:
DifferentialEquatrions.jl currently needs GrowableArrays.jl checked out to pass. That's this bug fix and why it's being tagged concurrently.

I've been testing this method of conditional dependencies on 3 development computers (on different OS's) and it seems to work fine. And it doesn't give Travis/Appvoyer a problem. It's the same method Plots.jl uses. I hear there can be an issue if you're trying to use a macro from one of these packages, but all of these packages are conditional dependencies only for the functions. If there's reduced functionality, then all of the functions I am using are still within what's usable so it works. And I know from testing that, if it errors then you get a dialog box saying to install a package, and once you install the package it will work and you can pick up from where you left off. So I have not had an issue with this at all (and Plots.jl seems to work fine!).

tkelman:
You haven't tested the actual issue. Install the package without its optional dependencies present and precompile it. Then install one of the conditional dependencies and try to use your package, it won't trigger recompilation and the optional features will most likely not work. That's the less bad failure mode.

The bad failure mode is if you have a package thay depends on both this and one of your optional dependencies. Install DifferentialEquations first, then install the other package that depends on both. Load the package that depends on both so both get precompiled. Then remove the package that depends on both, so the conditional dependency should get removed, but DifferentialEquations won't. If you then try to load DifferentialEquations, it will not load and the error will be very difficult to fix.

If Plots does things this way, then it's susceptible to the same failure mode. Unless you disallow precompilation or tie into it, conditional dependencies are going to have this problem.

ChrisRackauckas:
Continued on the other request.

PR #5765, opened 2016-07-31
Tag DifferentialEquations.jl v0.1.1, GrowableArrays v0.0.5

ChrisRackauckas:
Continuing #5761. Additional examples notebooks and other niceties were added to complement the v0.1 release of DifferentialEquations.jl and its upcoming announcement.


ChrisRackauckas:
@tkelman but it works. If I make a clean install, setup DifferentialEquations, try to use the external solver, get the error telling you to installing it, and then install the external solver, the functionality now works. Indeed it doesn't precompile again, but it doesn't error or give wrong output in any way: it simply works, and is robust.

I can give you a gist showing that is really the case (it gets messy because DifferentialEquations v0.0.3 had a lot more dependencies). Maybe there is something better I should do to help it precompile again, which we should discuss as something to add in the next version. However, for a release: I've tested that if you call the external solver, error, and then install it, at least on Windows and Linux, v0.4.6 and v0.5, it works. The fact that it works robustly across operating systems and versions, and not just for me but also the widely used Plots.jl, makes me think of it as not an issue which should block a new release. 

Rather a new release could improve this feature later (and if there ever truly is a canonical way of handling conditional dependencies, I will switch over to that ASAP).

ChrisRackauckas:
https://gist.github.com/ChrisRackauckas/d05ffcfe3bd9bd99cc02e84bbeb56bbe

tkelman:
that's the harmless failure mode. try the more complicated version where you precompile including the dependency but then remove it before usage. you also may be getting lucky that precompilation may be giving up on code in a try catch, which would not always be guaranteed behavior.

ChrisRackauckas:
When I did what you just described, [I still got harmless behavior where it still works after removal](https://gist.github.com/ChrisRackauckas/9e75bbedec3df725851da6b54e9d6be9). Can you give me a concrete example I don't get lucky? 

How do you suggest I go forward? Do you want me to just add a NLsolve and ForwardDiff as dependencies again? Is there a way to trigger precompilation each time I run the conditional dependency routine? This is a rare issue for small parts of the package.

tkelman:
```
julia> Pkg.rm("ODE")
INFO: Package database updated
```
it looks like something else depended on it and it didn't get removed. better to test within a temporary JULIA_PKGDIR.

> Is there a way to trigger precompilation each time I run the conditional dependency routine? 

Calling Base.compilecache or deleting the existing .ji file might work. But not if you already have the module loaded.

> How do you suggest I go forward?

I think you're okay for now, just warning you that these systems are going to be brittle until there is standardized support in the language. Unconditional dependencies are more reliable right now. If you want them to be optional, one way of achieving that is by splitting functionality into separate packages that can declare unconditional dependencies on everything it uses.

ChrisRackauckas:
> I think you're okay for now, just warning you that these systems are going to be brittle until there is >standardized support in the language. Unconditional dependencies are more reliable right now. If you >want them to be optional, one way of achieving that is by splitting functionality into separate packages >that can declare unconditional dependencies on everything it uses.

I understand. I plan to use any official system when it comes out (I heard there may be one coming from Stefan's JuliaCon talk?). Until then, this seemed like my only option and since Plots.jl has been really stable for me, I thought it would at least work well here. I don't think I truly have an option to implement the interfaces the same way as JuMP. Maybe if things really take off I could centralize and get a MathProgBase-type API, but that seems a ways off (and I'd hope to have a more official solution by then anyways).

If I hear about issues, I'll tag a version which adds dependencies. Adding back ForwardDiff and NLsolve is possible (though they are used by so few methods it seems very unnecessary to keep them around). I made sure to highlight in the documentation all of the things going on with the conditional dependencies. The ODE.jl wrapper only works for their current dev branch (as stated in the docs), so I see that as an experimental feature that few will use (but will be important for building benchmarks between the libraries). I will never have ODEInterface as a dependency because its reliance on gfortran would turn my package into a no-go on most Windows systems. And those 4 are the only conditional dependencies.

Thank you for letting me know about what issues I may encounter. I will likely be useful.

PR #5766, opened 2016-07-31
Register LinearOperators: v0.1.0

dpo:


tkelman:
needs a minimum julia version requirement

dpo:
Seems like something `Pkg.publish()` could tell me before creating the pull request. I can't figure out how to update this pull request cleanly, so I'll just open another.

tkelman:
Good point, perhaps we should make `PkgDev.publish` run a few of the metadata policy checks from https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/METADATA.jl locally.

dpo:
Some sort of package audit would save you some effort down the line. It might even involve running `Lint.jl` on the source files.

tkelman:
I'm absolutely planning on writing something for that. It may be part of Lint.jl, it may be separate.

PR #5768, opened 2016-08-01
Register LinearOperators: v0.1.0

dpo:


stevengj:
Very interesting!   Have you considered contributing the L-BFGS implementation to https://github.com/JuliaLang/IterativeSolvers.jl instead?

GPLv3 is a perfectly good free/open-source license.  However, a lot of the Julia ecosystem prefers the more-permissive MIT license (which is GPL-compatible, of course), if you don't have strong feelings about the license choice.

stevengj:
You might consider making `dtype` a type parameter rather than a field of `LinearOperator`, so that you can dispatch on it.

Also, calling `op.prod` in `(*)(op :: AbstractLinearOperator, ...)` is not very Julian; it would be more normal for subtypes to simply overload the `*` operator.   Note also that you probably want to take `AbstractVector` arguments, not `Vector`, so that your functions work with any type of array.

dpo:
Thanks. Limited-memory quasi-Newton approximations are operators that are useful beyond iterative solvers, especially in optimization (see, e.g., https://github.com/JuliaSmoothOptimizers/Optimize.jl/blob/master/src/solver/lbfgs.jl).

I'll certainly rethink the license later but wanted to get this PR going so #5351 can move along as well.

And yes, parametric types are on the list of things to do: https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/issues/13

PR #5769, opened 2016-08-01
Tag GenericSVD v0.0.2

simonbyrne:
https://github.com/simonbyrne/GenericSVD.jl


GenericSVD/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should have a minimum version for `view`

simonbyrne: Is there an easy way to find it?

tkelman: blame on the Compat.jl README works well

simonbyrne: How can I convert that to a version number?

tkelman: if you click on the commit, and the git tag for the Compat.jl release matched the metadata tagged sha and was pushed, you should get a little icon under the commit message that tells you which tags include the commit

PR #5774, opened 2016-08-01
Tag Shannon v0.3.1

kzahedi:


Shannon/versions/0.3.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3
```
tkelman: Travis might be a little slow to start up today, but if https://github.com/kzahedi/Shannon.jl/pull/4 fails its tests on Julia 0.3 then this line should be changed (and so should the minor package version number).

tkelman:
closing in favor of #5780

PR #5782, opened 2016-08-01
Register TensorFlow

malmaud:


tkelman:
cc @benmoran, note that https://github.com/benmoran/TensorFlow.jl has existed for some time - not yet registered, but he may have been planning to do so.

ViralBShah:
That one hasn't seen updates in a long time.

tkelman:
4 months isn't all that long. Anyway, hope you could combine efforts if needed. Given this uses the C interface and the other one uses PyCall, perhaps the Python API connections could be exposed in a PyTensorFlow.jl package and any pure-Julia interfaces not dependent on PyCall could be contributed here.

PR #5783, opened 2016-08-02
Tag GeoJSON v0.1.0 [https://github.com/JuliaGeo/GeoJSON.jl]

visr:
Fix deprecation warnings on v0.5, drop support for v0.3, and fixes a bug.

Had to do this manually so please let me know if I did something wrong.


tkelman:
> drop support for v0.3

This should use a new package minor version, and be 0.1.0 via `Pkg.tag("GeoJSON", :minor)` (or manually, if you remember to also push the git tag)

visr:
Ok thanks I tagged a minor version. Should I get rid of the first commit? And remove the v0.0.5 git tag from GeoJSON.jl?

tkelman:
That might be a little cleaner. It should be harmless if you don't feel like cleaning it up though, since anyone still on 0.3 won't see the 0.0.5 tag, and anyone on 0.4 or newer should go to 0.1.0 since it's newer.

visr:
Might as well do it properly. I think I'm done now.

PR #5785, opened 2016-08-02
Register AlgebraicNumbers [origin/master]

anj1:


AlgebraicNumbers/url
```diff
@@ -0,0 +1 @@
+origin/master
```
tkelman: this will need to be an actual repository url - and you'll need to have a version tagged for it to be installable via `Pkg.add`

PR #5788, opened 2016-08-02
Tag Temporal v0.0.6

dysonance:
- `setindex!` methods added so data can be changed on the fly
- Adding ability to download time series data into the TS type from Quandl and Yahoo Finance
    - This means adding a dependency on `Requests`
    - The reason I don't just leave this to the `Quandl` package is that it currently supports the `TimeSeries` package and not `Temporal`, and I wanted to allow easier integration (like how `quantmod` in `R` allows easy downloading of ticker symbols)


stevengj:
Probably this should have been a minor version bump (to v0.1) rather than a patch bump, since this adds new features (and a new dependency), not just bugfixes?

dysonance:
@stevengj I had planned on making the jump to v0.1 upon adding plotting/graphics functionality for the TS type, which will incorporate the final external dependency for the foreseeable future (as far as I know). That is probably the next major item on the todo list right now. Is that a fair plan?

stevengj:
Version numbers are cheap, and the point of the semver spec is to indicate compatibility, not whether the changes are small or large.

But I suppose that can wait until your next version bump.

PR #5791, opened 2016-08-02
Tag TimeSeries v0.8.4

milktrader:
Now you can customize how missing values are shown. This is non-breaking and defaults to the current behavior. 

Only changes REPL output, not underlying values.


TimeSeries/versions/0.8.4/requires
```diff
@@ -0,0 +1 @@
+julia 0.4 0.5-
```
stevengj: Shouldn't this be just `julia 0.4`, i.e. 0.4 or later?

stevengj:
Since this is a new backwards-compatible feature (e.g. you added a new constructor), shouldn't it be a minor version bump and not a patch?

stevengj:
(Similarly, it looks like some of your previous patch releases added features...)

Technically, the [semver spec](http://semver.org/) allows you to ignore its rules for 0.x releases, but it still seems like a good idea to bump the minor version for new backwards-compatible functionality.

TimeSeries/versions/0.8.4/requires
milktrader: I haven't started testing against 0.5 yet, though there is an open issue to start doing so. The reason is because I suspect that there will be quite a bit of breakage with indexing that will need to be resolved

stevengj: Ah, okay.

milktrader:
This release does not add a new constructor, it only gives users to option to change how the REPL output is displayed for `NaN` placeholders. So previously this was the only output for missing values

````julia
julia> lag(cl, padding=true)
500x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31

Close
2000-01-03 | NaN
2000-01-04 | 111.94
2000-01-05 | 102.5
2000-01-06 | 104.0
⋮
2001-12-26 | 21.36
2001-12-27 | 21.49
2001-12-28 | 22.07
2001-12-31 | 22.43
````
And now you can control the output to get something like this ...

````julia
julia> lag(cl, padding=true)
500x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31

Close
2000-01-03 | NA
2000-01-04 | 111.94
2000-01-05 | 102.5
2000-01-06 | 104.0
⋮
2001-12-26 | 21.36
2001-12-27 | 21.49
2001-12-28 | 22.07
2001-12-31 | 22.43
````
If you wanted you could fill the missing slot with whatever you choose.

So per the server spec it's not a bug fix, but not sure it rises to an added feature.

milktrader:
Here is the [documentation](http://timeseriesjl.readthedocs.io/en/latest/dotfile.html#missing)

stevengj:
Your documentation is a little unclear.  How do you select e.g. `BLACKHOLE`?   Is this a new constant?

The basic question is this: if you write code using the new feature (e.g. referencing the `BLACKHOLE` constant), will it still work with the previous version?  If the answer is "no" then it should be a minor version bump.

stevengj:
(A new constructor was added in JuliaStats/TimeSeries.jl#269, it seems, so 0.8.3 should also have been a minor version bump, not a patch, compared to 0.8.2.)

milktrader:
Oh, I see what you mean by a new constructor, I thought you were referencing this version.

milktrader:
I can clear up the documentation. The MISSING paragraph was the last one on the page explaining the use of the .timeseriesrc file. 

`A dot file named .timeseriesrc sets two variables that control how TimeArrays are displayed.`

tkelman:
As @stevengj likes to say, version numbers are cheap, so might be good to be more mindful of semver going forward, even though you're technically allowed to do anything during 0.x.y.

milktrader:
We might add documentation about complying with semver. `Pkg.tag` docs don't even mention it.

PR #5795, opened 2016-08-02
Tag LogParser v0.3.0

randyzwitch:
I intentionally made this `julia 0.5` instead of `julia 0.5-`, as I don't want to have to do another PR just to switch it in a few weeks.

### DIFF
--- a/LogParser/versions/0.3.0/requires
+++ b/LogParser/versions/0.3.0/requires
@@ -1,0 +1,4 @@
+julia 0.5
+DataFrames
+GZip
+Compat
--- a/LogParser/versions/0.3.0/requires
+++ b/LogParser/versions/0.3.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5.0-rc0
 DataFrames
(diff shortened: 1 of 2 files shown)


tkelman:
you could make it `0.5.0-rc0` and it will be installable now, otherwise this version won't be installable yet

randyzwitch:
If I put  0.5.0-rc0, it will also allow for 0.5 when it comes out?

tkelman:
yes, the first version is an inclusive lower bound, the second version (if present) is an exclusive upper bound

tkelman:
You're running Travis and AppVeyor against 0.4 still, are you intending for this to be 0.5-only?

tkelman:
And looks like Gzip is a test-only dependency so could go in `test/REQUIRE`, and you aren't using Compat at all at the moment, but better to have more dependencies listed than you need than the other way around.

randyzwitch:
Thanks for the careful eyes Tony, I'll clean this up and keep in mind for next submissions

PR #5802, opened 2016-08-03
Register ULID: v0.1.0

ararslan:
For some reason I couldn't get `Pkg.publish()` (0.4) nor `PkgDev.publish()` (0.5) to work for this. Maybe because I specified the Julia requirement as 0.5? I can install and use it on 0.5 though. Any idea why that could be, @tkelman?


tkelman:
> `julia 0.5`

This won't be installable via `Pkg.add` until the final 0.5.0 release is out.

Is ULID a standard acronym? A bit more of a spelled-out name might be more discoverable for users to see at a glance what the package is for.

tkelman:
And don't forget to push the git tag to https://github.com/ararslan/ULID.jl/releases - need to do it manually if Pkg[Dev].publish isn't working.

ararslan:
Oh yeah. I'll make the bound 0.5- then and push the tag. Thanks!

ULID is the official acronym (it's meant to be a replacement for UUID). Not sure what else I'd call it, plus the name matches that of the initial implementation.

ULID/versions/0.1.0/requires
```diff
@@ -1 +1 @@
-julia 0.5
+julia 0.5-
```
tkelman: may want to do this here too https://github.com/ararslan/ULID.jl/blob/master/REQUIRE if you plan on making another tag before 0.5.0 final is done

ararslan: I was just about to mention this. Yeah, I'll do that. Thanks again!

ararslan:
Does this 👀 🆗❓

PR #5809, opened 2016-08-03
Tag MFCC v0.1.0

davidavdav:
Makes this package compatible with Julia v0.5 again.


tkelman:
Note that your LICENSE file doesn't indicate the GPL license of https://github.com/davidavdav/MFCC.jl/blob/master/src/feacalc.jl

davidavdav:
OK, thanks for pointing this out.  It is the result of an earlier merge of packages, but I hadn't adapted the license.

PR #5815, opened 2016-08-03
Pull request/43cc9db2

davidanthoff:
Mostly julia 0.5 compat

### DIFF
--- a/ExcelReaders/versions/0.6.0/requires
+++ b/ExcelReaders/versions/0.6.0/requires
@@ -1,0 +1,6 @@
+julia 0.4
+DataArrays
+DataFrames
+Dates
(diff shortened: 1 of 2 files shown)


ararslan:
Could you remove the Mimi commit from this PR since you have that in a separate PR?

davidanthoff:
Alright, I did.

Is there some way to prevent this when I'm tagging multiple packages in a row and then want to have PRs for each tag separately?

tkelman:
We should probably change PkgDev to make a branch before committing your tags to metadata, then go back to metadata-v2 after publish or something like that. https://github.com/JuliaLang/PkgDev.jl/issues/57

PR #5820, opened 2016-08-04
add ForwardDiff 0.2 upper bound to Celeste.jl 0.1.0 tag

tkelman:
fixes GradientNumber not defined


tkelman:
cc @jeff-regier this is an old release, but it wasn't compatible with ForwardDiff 0.2

PR #5822, opened 2016-08-04
Tag NormalizeQuantiles v0.0.11

oheil:
Added keyword argument call syntax for function sampleRanks


stevengj:
@oheil, if you're adding new functionality, shouldn't you bump the minor version instead of the patch version?

andyferris:
@stevengj Shouldn't we follow that when semver < 1 then patch level tags may include new features and minor tags are breaking changes?

tkelman:
the feature vs bugfix distinction is useful even before 1.0. this is fine for now, but keeping semver in mind going forward is always a good idea.

PR #5823, opened 2016-08-04
Pull request/9aee94e7

tawheeler:
To include updates with Discretizers.jl UniformBinWidth algorithms


tkelman:
You should probably rebase out the DC registration and revert from your local metadata so they stop showing up

tawheeler:
I was wondering why that is still floating around. I'll try and figure that out.

PR #5837, opened 2016-08-05
Register ConicIP

gabgoh:
mpf (my supervisor) and I have decided to call this package ConicIP. Here's the pull request.

ConicIP.jl: A pure Julia conic programming solver. Includes, LP, SOCP, and SDP.

PR #5839, opened 2016-08-05
ConicIP - now with a tag

gabgoh:
sorry.

https://github.com/JuliaLang/METADATA.jl/pull/5837


mlubin:
Hi @gabgoh. It's great to have this in Julia, feel free to announce on the julia-opt list.

Regarding the name, what do you think about ``ConicIPM``? (for interior-point method). [https://github.com/mlubin/Pajarito.jl](Pajarito) is a conic IP (integer programming) solver so I'm hoping we can avoid confusing potential users. Just expressing my preference here to keep the julia optimization ecosystem more coherent, but this isn't a veto.

CC @chriscoey

mlubin:
CC @mpf

timholy:
And to many people IP means Intellectual Property.

ConicIP/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4+
```
tkelman: this `+` means it won't be installable on any 0.4 release

gabgoh:
I'm thinking ConicQP - QP stands for quadratic program. It has the same ring to IP but doesn't have the baggage of all the things it already means

chriscoey:
But aren't you going to support SDP one day perhaps?

mlubin:
SDP is already there from what I saw. QP sounds restrictive when the solver is in fact much more general.

mlubin:
(Side note: What's the motivation for supporting convex quadratic objectives explicitly when they could just be handled via second-order cone constraints and an epigraph?)

gabgoh:
ok back to the drawing board. QP does seem confer the notion of "polyhedral constraints".

@chriscoey SDPs are supported. 

@mlubin we had a brief debate about this, but it turns out that the extra freedom in having the Q explicitly as an option helps. For example, in my own work I use Q's with special structure (e.g. Q can be of type `Woodbury`) and the KKT systems are way nicer with Q explicit than implicit in an epigraphical SOC. Also, Q can be set to a sparse zero matrix if unused, so there's minimal overhead.

mlubin:
@gabgoh, makes sense, thanks for the explanation!

mpf:
This thread seems to have stalled. Is this an implicit veto for `ConicIP`? Otherwise, would be good proceed with registering the package with METADATA.

tkelman:
If you feel strongly about keeping the name as-is, that's fine.

What is the intended minimum Julia version? Would at least like to get the require file correctly reflecting the supported versions.

tkelman:
Looks like you changed it in the package repo, but that won't propagate to METADATA unless you either make another tag, or modify the existing files in this branch of your metadata fork.

mlubin:
No veto here

gabgoh:
ok closing this PR, creating a new one.

PR #5841, opened 2016-08-05
Tag BDF v0.1.6

sam81:
this version supports julia v0.5, including pre-release versions of 0.5, but not julia v0.4. I've put `julia 0.5-` in `REQUIRE` to specify this, I hope that's correct.

### DIFF
--- a/BDF/versions/0.1.6/requires
+++ b/BDF/versions/0.1.6/requires
@@ -1,0 +1,4 @@
+Compat 0.7.18
+Docile
+HDF5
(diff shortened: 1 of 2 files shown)


tkelman:
yes, but when you drop support for a version of julia, the package minor (or major) version should be bumped via `Pkg.tag("BDF", :minor)`

sam81:
OK, thanks for the comment. I'll bump the minor version and submit another pull request then. I'm closing this one.

PR #5842, opened 2016-08-05
Register DocStringExtensions

MichaelHatherly:
The package provides a number of extensions to the base docsystem. The [docs](https://juliadocs.github.io/DocStringExtensions.jl/latest/) provide an overview of what the package currently includes, with more features to come in future releases.

### DIFF
--- a/DocStringExtensions/url
+++ b/DocStringExtensions/url
@@ -1,0 +1,1 @@
+git://github.com/JuliaDocs/DocStringExtensions.jl.git
--- a/DocStringExtensions/versions/0.1.0/requires
+++ b/DocStringExtensions/versions/0.1.0/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.8
--- a/DocStringExtensions/versions/0.1.0/sha1
+++ b/DocStringExtensions/versions/0.1.0/sha1
@@ -1,0 +1,1 @@
+e4bfe3537cc033c161d7e65bff776c15dcf3e9a4


tkelman:
Pretty cool that this is doable outside of Base to try things out. I don't think we actually need the code that we use to generate the docs to always be part of base, if it's possible to tie into it then I think having `make docs` depend on installing a Julia package would be acceptable.

MichaelHatherly:
> Pretty cool that this is doable outside of Base to try things out.

Yeah, the docsystem is decoupled enough from the rest of `Base` now that it's pretty simple to do this kind of thing. As a side-effect it's also really easy to move the entire thing (`Docs` and `Markdown`) back out of `Base` and into packages if we want. Only `docs/core.jl` would need to remain in `Base` with a couple minor changes to get things bootstrapped and working correctly with precompilation. I had a branch a while back that tried it out, seems to work well enough.

(Whether that's a worthwhile goal at the moment is debatable, but maybe worth keeping in mind for https://github.com/JuliaLang/julia/issues/5155.)

> I don't think we actually need the code that we use to generate the docs to always be part of base, if it's possible to tie into it then I think having make docs depend on installing a Julia package would be acceptable.

Seems reasonable to me. Once we've finished off the GSoC work I'll start figuring out the best way to get docs building automatically for `Base`, whether on travis with each merge to `master` or perhaps nightly on the buildbots?

---

I assume 24 hours is enough time for this to be open for comments/review? I'll merge in a couple of hours if that's the case.

PR #5850, opened 2016-08-05
Bump CustomUnitRanges, Grid, GtkUtilities, HDF5, and Images

timholy:


tkelman:
It's saying that GtkUtilities should have used a new minor version after 0.0.2.

GtkUtilities/versions/0.1.0/sha1
```diff
@@ -0,0 +1 @@
+99d837f559536eafafd507ffa46e9b25523eda10
```
tkelman: not pushed yet as a git tag to https://github.com/timholy/GtkUtilities.jl/releases

timholy: fixed

PR #5856, opened 2016-08-06
Tag Bootstrap v1.0.0

juliangehring:


ChrisRackauckas:
You're not supposed to tag a v1.0 before Julia v1.0. It should signify the major release with the stable version of Julia, and it's impossible to know if this will work on the first Julia release. Instead tag another minor release, Bootstrap v0.4.0.

BTW, found this because we were talking about your package in the Julia Gitter. Nice work!

Keno:
> You're not supposed to tag a v1.0 before Julia v1.0

This is not a rule that has been discussed or written down anywhere.

ChrisRackauckas:
I swear I read that it was a rule and it was discussed. Let me find it.

ChrisRackauckas:
> Adding a tagged version of your package will expedite the official registration into METADATA.jl by collaborators. It is strongly recommended that you complete this process, regardless if your package is completely ready for an official release.

I was just trying to help out, but mis-remembered (maybe someone on the Gitter explained it like that?). Thanks for correcting me.

Keno:
> I was just trying to help out, but mis-remembered (maybe someone on the Gitter explained it like that?). Thanks for correcting me.

No problem. Thanks for taking the time to review these things. However, it is important that package developers don't get frustrated by the METADATA process (sometimes people already do because of the name bikesheding that happens), so I wanted to stop by quickly and make sure that wasn't gonna happen here ;). Carry on everyone.

dpsanders:
I have certainly been told this by the authors of a well-known package when I asked them when they would consider releasing v1.0. 

On the other hand, the public facing API of a package could easily be considered stable even if the internal implementation might have to change to reflect future changes to the language.

Keno:
> I have certainly been told this by the authors of a well-known package when I asked them when they would consider releasing v1.0.

Sure, how to version your package is up to each package developer, if they want to tie it to Julia versions, no problem ;).

tkelman:
I'd hesitate to call a package 1.0 before Julia itself does so, but it's not a big deal. Use semver, if things need to break then it becomes 2.0.

juliangehring:
Thank you all for the comments and suggestions.

The tagged version introduced a new API, practically breaking existing workflows. Following [semantic versioning](http://semver.org/), incrementing the major version number to 1.y.z seems to be the most logical consequence:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> MAJOR version when you make incompatible API changes

PR #5864, opened 2016-08-07
Tag DifferentialEquations v0.2.0

ChrisRackauckas:
Large changes occurred. This signals that the major changes are complete. The internals have been re-hauled to allow for really good type inferencing on each solver call. This also enhanced precompilation, and a SnoopCompile generated precompile hinting script was added. ForwardDiff was added back as a dependency in order to do caching of Dual arrays. Bug fixes to Feagin adaptivity and displaying of Monte Carlo simulation results are also included. Next on the plate are improvements to the current FEM solvers, a benchmarking suite, improving the Monte Carlo simulations, adding back the SDE adaptivity, and as always, adding more problems and solvers.


DifferentialEquations/versions/0.2.0/sha1
```diff
@@ -0,0 +1 @@
+dbd13a2c8122d97b90d93bff7405dc7523631c75
```
tkelman: could you push the git tag to https://github.com/ChrisRackauckas/DifferentialEquations.jl/releases ?

ChrisRackauckas:
Newest tag fixes the precompile issue.

PR #5866, opened 2016-08-08
Tag Rotations v0.2.0 [https://github.com/FugroRoames/Rotations.jl]

andyferris:


Rotations/versions/0.2.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+Compat
```
tkelman: it's best to declare a minimum version dependency that implements the features you're using

andyferris: @tkelman In general, or specifically for *Compat.jl*?

tkelman: both

PR #5872, opened 2016-08-08
revise sha1 for Celeste.jl after modifying the git history for Celeste.jl

jeff-regier:


tkelman:
best to avoid force pushing over a branch that anyone has cloned from...

PR #5874, opened 2016-08-08
remove depwarns LibExpat, GZip

SimonDanisch:
Not my packages, hopefully `Pkg.tag; Pkg.publish` is doing the job alright ;)!

@amitmurthy 
@kmsquire
@davidanthoff 

In reply to:
https://github.com/amitmurthy/LibExpat.jl/issues/50
https://github.com/JuliaIO/GZip.jl/issues/57

Best,
Simon


LibExpat/versions/0.1.5/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
```
tkelman: this is going to need to be a new minor package version

PR #5876, opened 2016-08-08
Tag Distributions v0.10.1

ararslan:


Distributions/versions/0.10.1/sha1
```diff
@@ -0,0 +1 @@
+a71f7b48d2feb554459491039d4ca7da3384252e
```
tkelman: this commit is identical to 0.10.0

ararslan: ...what. Okay, I did something weird and I'm not entirely sure what.

tkelman: tagged without checkout first?

ararslan: Nope, did the usual checkout master, tag, publish...

ararslan: `Pkg.tag` should have a check for when the tag you're adding is identical to an existing one.

tkelman: yep, open an issue in PkgDev?

tkelman:
This looks like it broke a lot of dependent packages, and should probably have been at least a minor bump - http://pkg.julialang.org/pulse.html

To fix this short-term, what would you think about creating a branch from right before the major rewrite, re-applying the deprecation fixes and tagging that as 0.10.2?

PR #5878, opened 2016-08-08
Tag DataFrames v0.7.7

ararslan:


ararslan:
This has the same problem as the DataFrames PR.

DataFrames/versions/0.7.7/sha1
```diff
@@ -0,0 +1 @@
+416795217df0eb77b149a205ef9cb1f3c23828a9
```
tkelman: same commit as 0.7.6

PR #5880, opened 2016-08-09
Tag new versions of Lazy and MacroTools

ranjanan:
cc: @MikeInnes


Lazy/versions/0.10.2/requires
```diff
@@ -0,0 +1,3 @@
+MacroTools
+julia 0.3
```
tkelman: is this still tested on 0.3?

tkelman:
please push the git tags to the corresponding packages

tkelman:
Lazy.jl looks like it renamed a bunch of functions. Is that breaking for any dependent packages using older versions?

ranjanan:
Actually, a dependency for Lazy (MacroTools) works only for v0.4 and above. So I guess I should bump from v0.3 to v0.4. 

>please push the git tags to the corresponding packages

I don't have access to the repos. Will ask @MikeInnes to do it or give me access.

MikeInnes:
> Lazy.jl looks like it renamed a bunch of functions. Is that breaking for any dependent packages using older versions?

The function renames are necessary due to the fact that Base added a bunch of exports, which was breaking in itself.

I'll push the tags.

tkelman:
That doesn't answer the question. On existing versions of julia where the exports didn't change, will these renamings in Lazy break packages that were using it? If so then this should be more than a patch release, and we should add upper bounds on existing package tags' dependencies on Lazy to avoid breaking them all immediately.

MikeInnes:
> On existing versions of julia where the exports didn't change

That version is Julia 0.3, which Lazy no longer supports anyway.

tkelman:
It did until a few minutes ago. Mind your REQUIRE files.

MikeInnes:
Even then I'm pretty sure there are no packages which use those functions.

tkelman:
https://github.com/MikeInnes/MacroTools.jl/releases/tag/v0.3.1 is the tagged commit 5e5eca8791de4128e902481bdb524ce0d3b31e36, what needs updating?

MikeInnes:
Ah whoops, I meant to tag the latest commit (ffce8f322b4779ae6c7bb79cca08fcc14c354284) but was in the wrong package dir, fixed.

PR #5893, opened 2016-08-10
Tag DataFrames 0.7.8

ararslan:
Tagging a commit from before contrast coding was introduced, which seems to have caused failures in other packages.

### DIFF
--- a/DataFrames/versions/0.7.8/requires
+++ b/DataFrames/versions/0.7.8/requires
@@ -1,0 +1,7 @@
+julia 0.4
+DataArrays 0.3.4
+StatsBase 0.8.3
+GZip
+SortingAlgorithms
+Reexport
+Compat 0.8
--- a/DataFrames/versions/0.7.8/sha1
+++ b/DataFrames/versions/0.7.8/sha1
@@ -1,0 +1,1 @@
+084bfa7bff89ac5f52b9f36c41f0adea1df0f1e5


tkelman:
Awesome. Note that these aren't really personal branches if releases have been made from them, should probably call them `release-x.y` or similar and turn branch protection on for them.

ararslan:
That's good to know, I'll choose better branch naming for such things in the future. Regarding branch protection, never heard of it. Sounds nifty (and useful).

ararslan:
Oh, I'm not an owner in JuliaStats so I don't think I can protect branches

tkelman:
Ah, bummers. You can "rename" a branch by creating a new one from the same commit with a different name then deleting the old copy.

Also looks like this is missing https://github.com/JuliaStats/DataFrames.jl/pull/1028, was that only touching lines added by the contrasts PR?

ararslan:
Yes it was

ararslan:
Should I go back and rename those branches then?

tkelman:
May as well, as long as the same commit sha is in the new branch name then this one doesn't matter. It would be good to bug someone who has admin access to turn on branch protection for any branch that a release tag has been made from.

andreasnoack:
Great work both of you but are you sure that the StatsBase issue is fixed on that commit? The issue is that the values in this table https://github.com/JuliaStats/StatsBase.jl/blob/master/test/statquiz.jl#L74-L78 have changed. I've checked out the version before the merge of the contrasts pr and I still see the issue.

ararslan:
@andreasnoack Uh oh. Can you try checking out the commit before Doug's model matrix refactoring and see if that does it?

ararslan:
Oh hey @andreasnoack, when you get a chance would you be willing to protect the release-* branches on Distributions and DataFrames like @tkelman mentioned?

andreasnoack:
@tkelman Has there been any changes to the LAPACK/OpenBLAS setup on Travis? I can reproduce the test error locally on a machine with MKL but not with the normal OpenBLAS. The difference ends up being a scaling factor in the variance term that is respectively `0.0` and `eps()/2`. The test might be too sensitive to rounding.

@ararslan I'll take a look at it tomorrow.

tkelman:
No changes, but what core type is your local system using? Different openblas kernels may be getting called.

PR #5894, opened 2016-08-10
SALSA v0.1.0

jumutc:


tkelman:
Since support for Julia 0.3 is being dropped here, this should use a new minor package version via `Pkg.tag("SALSA", :minor)`

jumutc:
Or should I create a separate branch for this pull request?

tkelman:
This is okay I suppose. It's not too bad to have the 0.0.6 tag and the 0.1.0 tag be identical, since if there's ever a need to make a future bugfix tag that does support Julia 0.3 it can just be 0.0.7 and not exactly in commit order.

tkelman:
Ah, but you should push the git tags so they show up at https://github.com/jumutc/SALSA.jl/releases

jumutc:
OK, I have pushed tag v0.1.0 ... should I do it for v0.0.6 as well?

tkelman:
Unless you'd like to remove it from here, may as well?

jumutc:
OK, that's fine for me to have v0.0.6 as well

PR #5899, opened 2016-08-10
Tag WebSockets v0.2.0 [https://github.com/JuliaWeb/WebSockets.jl]

MikeInnes:


tkelman:
Could you try to be more patient with these, and wait until at least one other person has had a chance to look over these before merging them?

PR #5901, opened 2016-08-10
Tag Contour v0.1.1 and v0.2.0

tlycken:
v0.1.1 incorporates all changes needed to work without warnings on 0.5.

v0.2.0 removes an iteration API that was previously deprecated.

Users who rely on the iteration API and want to work with 0.5 should use 0.1.1, everyone else should use 0.2.0.


tkelman:
Do you know if many other packages are relying on this? Would it be worth adding upper bounds to existing tags until they've had a chance to upgrade?

rsrock:
Good call. The big one is Gadfly.  Looks like it's using the deprecated form here, at the very least.

https://github.com/dcjones/Gadfly.jl/blob/776c26c51b4235e028f78179c67f6fc82c7ce008/src/statistics.jl#L1531

tlycken:
@tkelman: It's been deprecated for a while, so I was assuming that these would be fixed. It seems that Gadfly would either need to pin its dependency to 0.1.1 or update that line too.

The only other package which has a dependency on Contour is PGFPlots (at least according to `Pkg.dependents("Contour")`), which would also have to be updated here: https://github.com/sisl/PGFPlots.jl/blob/414b168ee27096cd1c2c32dbf42b09fe470340a4/src/PGFPlots.jl#L449

tkelman:
Pinning the dependency would have to be done via upper bounds here in metadata on all past tags of Gadfly and PGFPlots.

tlycken:
Yeah, probably easier just to file PRs to upgrade them to the new API then. Would such PR's (and subsequent tags) be blocking for these changes to be merged? Would that change if I removed the second commit from this PR?

tkelman:
If the first commit here doesn't break any dependent packages, we can merge it without needing to do anything else.

Adding upper bounds would be safe to do whether or not you want to wait for dependent packages to be fixed before getting the breaking changes released.

tlycken:
I hope I got that right; I'd be grateful for a quick review.

tkelman:
That's the right idea, was that every tag in Metadata that depends on Contour? A possibly surprising side-effect of this in its current form might be that Pkg could potentially downgrade Gadfly or PGFPlots to some very old version that predates adding Contour as a dependency, in order to then get a newer version of Contour. It can be hard to predict the dependency resolution sometimes.

What we've done in the past for situations like this is to retroactively add unnecessary dependencies on `Contour 0.0.0 0.2-` to versions of Gadfly and PGFPlots that wouldn't otherwise depend on Contour, in order to avoid this downgrade scenario.

tlycken:
Yes, that were all hits for `grep -e Contour */versions/*/requires`. I've added a commit that adds dummy requirements to the older versions as well.

tlycken:
@tkelman, thanks for your help with this!

PR #5910, opened 2016-08-11
Adding Vectorize.jl (tag v0.0.3)

rprechelt:
This pull request contains `Vectorize.jl`, the cross-platform vectorization library developed for Julia's GSOC program. Contains tags for `v0.0.1` through `v0.0.3`. 

@ArchRobison @simonbyrne @vtjnash


rprechelt:
Strange git error here: `ERROR: LoadError: Invalid url scheme http for package Vectorize. Should be 'git' or 'https'`. 

I'll see if I can't look into why this is happening.

Vectorize/url
```diff
@@ -0,0 +1 @@
+http://github.com/rprechelt/Vectorize.jl
```
tkelman: this should be https

Vectorize/versions/0.0.3/requires
```diff
@@ -0,0 +1 @@
+julia 0.5
```
tkelman: this will not be installable yet

rprechelt: Good point. Maybe this ought to wait until `0.5` is released.

tkelman: or you could do `0.5.0-rc0` or similar and it'll work for RC's or newer

PR #5913, opened 2016-08-11
Tag Spectra v0.1.3

charlesll:
New release of Spectra with addition of new functionalities. Some functions use SciKit Learn so PyCall is in the REQUIRE file.


tkelman:
Are you sure you want to add 700,000 lines of text files under examples?

charlesll:
Several spectra should be present in the raw_2012 folder, but not that much: I forgot to clear the example generated files.  And despite having .DS_Store in my .gitignore, I just saw that there was .DS_Store files in this release.

All is corrected and pushed on the github repo. How do I do to make those changes appear with this particular pull request?

charlesll:
Ok,

I squashed the last 3 commits together. However my .git folder size did not decrease after the squashing? It is around 61M...

tkelman:
You would need to squash together the commit that added the files with the commit that deleted them.

charlesll:
Yes, this is what I did...

tkelman:
was that only locally? the "Updated docs" commit still on master changes about 300k lines

charlesll:
Ok so I stashed all the commits since the last merge for the version requirement change... I also reduced the size of the example dataset. My .git folder still is 62 M but I don't know what else I can do to reduce it... ?

At this point I think I am going to put a new tag and do a new pull request.

tkelman:
your local copy would maybe need a git gc, but yes probably best to do the tag over again.

charlesll:
Ok thanks, sorry for all this! I did not realise that I had generated so much stuffs in the example folder while working on this new version...

PR #5918, opened 2016-08-11
Register GitLab

mdpradeep:


tkelman:
can you push the git tag so it shows up under https://github.com/JuliaComputing/GitLab.jl/releases ?

tkelman:
I think JLD should still be listed in `test/REQUIRE`, not commented out? https://github.com/JuliaComputing/GitLab.jl/blob/60354dec877ab64ae06310c807490736952ca667/test/event_tests.jl

mdpradeep:
Updated the test dependency - https://github.com/JuliaComputing/GitLab.jl/blob/master/test/REQUIRE

tkelman:
That would need a new tag to get released. That can be done separately.

PR #5919, opened 2016-08-11
Various updates including Julia v0.5

EricForgy:
- Add PlotlyJS example
- Add support for IJulia (experimental)
- Updates for Julia v0.5


tkelman:
If you use Compat, it needs to be listed explicitly in your REQUIRE file, with a minimum version number that implements the features you need.

EricForgy:
Ok. Thanks. Do I need to tag a new version to make this minor change? Is there a way to revise and keep the same tag? If so, how? Thank you.

EricForgy:
I tried rebase, but looks like I made a mess:

image

tkelman:
Yes, you could have modified the `requires` file of the 0.0.5 tag in-place in this branch of your METADATA fork.

EricForgy:
Ok. Since I went ahead and tagged a new version, is it ok to merge as is?

PR #5922, opened 2016-08-11
Tag OpenCL v0.3.7

vchuravy:
Is it okay that the tagged commit only lives on a branch? I plan to continue supporting v0.4. but only for bugfixes from a branch.


tkelman:
> Is it okay that the tagged commit only lives on a branch?

Yes but it would be ideal if you enabled branch protection for it.

tkelman:
Since 0.3.6 supported Julia 0.3, this will need to use a new minor package version via `Pkg.tag("OpenCL", :minor)`

PR #5930, opened 2016-08-11
Next round of tags

quinnj:


tkelman:
Is work pending to fix Feather.jl and SQlite.jl for this? Both were broken on pkgeval, probably need upper bounds.

PR #5931, opened 2016-08-11
Tag OpenCL v0.5.0 [https://github.com/JuliaGPU/OpenCL.jl]

vchuravy:
Prepares OpenCL.jl for the Julia v0.5 release.


OpenCL/versions/0.5.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5
```
tkelman: this won't be installable yet

vchuravy: That should be fine for now, just so that people get the right version once it is release or should I use `v0.5-`?

tkelman: could set one of the rc's as a minimum

vchuravy: That would just be `v0.5-rc0`?

PR #5936, opened 2016-08-12
Tag Rotations v0.3.0 [https://github.com/FugroRoames/Rotations.jl]

andyferris:
New v0.5 - only version, based on *StaticArrays.jl*. Each rotation is now a 3x3 matrix (computed/sparse).


andyferris:
FYI @awbsmith

MichaelHatherly:
Since you're importing `Compat` it should probably be listed in the `REQUIRE` file as well.

andyferris:
Good catch. I removed the `using Compat` line from my code and updated the SHA. Thanks.

Rotations/versions/0.3.0/sha1
```diff
@@ -1 +1 @@
-6196140ef7377b3f2f2cf18363d1413a49f056c8
+760f1fa3100adbfea9916630ead7cfa3abfeb503
```
tkelman: can you fix up the git tag too? https://github.com/FugroRoames/Rotations.jl/tree/v0.3.0

andyferris: Done!

andyferris:
Should be good-to-go now

PR #5943, opened 2016-08-12
Tag Optim v0.6.0 [https://github.com/JuliaOpt/Optim.jl]

pkofod:


tkelman:
Will downstream packages have to adjust to this in any way, like many of them did for Optim 0.5?

pkofod:
The only way I could imagine that would be the case, is if they are using the old `NelderMead` constructor with keywords. I couldn't really see how to deprecate that. The old constructor had three arguments, and the new one has four. I guess I could make the keyword-based constructor something like 
```julia
function NelderMead(; kwargs...)
    check if old names are in kwargs && throw warning
    NelderMead(....)
end
```
right? I hadn't thought of that before. Is it possible to make a commit, and have v0.6.0 refer to that commit instead?

tkelman:
Yeah you could modify the 0.6.0 tag, call the initial one a do-over. Or make an immediate 0.6.1 that replaces it.

PR #5944, opened 2016-08-12
Atom and CodeTools

MikeInnes:


KristofferC:
Cant install these on RC because of v0.5 requirement.

PR #5946, opened 2016-08-12
Update v0.3.0 sha1

Ward9250:
Corrects incorrect sha1


Bio/versions/0.3.0/sha1
```diff
@@ -1 +1 @@
-ea70abb72143687b6d4c5a7e1e1f82e5e4e63c3f
+2870dfc06abf87702a57408be3e25116de57e57a
```
tkelman: should probably be a new tag, not modifying the existing one? https://github.com/BioJulia/Bio.jl/compare/0.3.0...v0.3.0

Ward9250: There are two tags, one is 0.3.0 (mistake) and the other is v0.3.0 (correct), all tags in Bio.jl follow the vX.X.X so the metadata should point to the v0.3.0 tag, not the 0.3.0 one. ea70... points to 0.3.0, the new on 2870df points to v0.3.0. So we want 0.3.0 in METADATA to point to v0.3.0. This is what I think this change will achieve, but I don't usually do this manually, so I'm not 100% sure.

PR #5949, opened 2016-08-12
Tag OpenCL v0.5.0 [https://github.com/JuliaGPU/OpenCL.jl]

vchuravy:
Replaces #5931


OpenCL/versions/0.5.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5-rc0
```
tkelman: I guess this should work
```
julia> v"0.5-rc0"
v"0.5.0-rc0"
```

PR #5950, opened 2016-08-12
Tag BioMedQuery v0.0.3

mirestrepo:
Fixed a bug when saving to SQLite database.
For now put a limit to version of DataStreams as v0.0.7 breaks sqlite


BioMedQuery/versions/0.0.3/requires
```diff
@@ -0,0 +1,11 @@
+julia 0.4
+ArgParse 0.3.0
+DataStreams 0.0.6 0.0.7
```
tkelman: Note that what this might mean is the package manager might choose to downgrade BioMedQuery to an older version that didn't have an upper bound in order to upgrade DataStreams. If that would also be broken, another option would be adding the DataStreams upper bound to all past tags of BioMedQuery here in Metadata.

mirestrepo: Thanks. How do I edit the REQUIRE for past tags? This seems a cumbersome solution in general --- is there a better way to take care of limiting updates to  packages that break my package? What are the best practices to take care of this?

tkelman: you modify the files for the tags here in METADATA under `BioMedQuery/versions/*/requires`

mirestrepo: @tkelman - great. did that #6006

tkelman: that PR will cover 0.0.1, but 0.0.2 within this PR will also need the same change

mirestrepo: I got it. Done!

mirestrepo: Ok - I think did that too now

PR #5953, opened 2016-08-13
Tag CoordinateTransformations v0.2.0

c42f:
@andyferris


tkelman:
can you push the git tag so it shows up under https://github.com/FugroRoames/CoordinateTransformations.jl/releases ?

c42f:
Done, thanks Tony.

tkelman:
looks like Geodesy.jl needs upper bounds and/or an update http://pkg.julialang.org/logs/Geodesy_0.4.log

c42f:
Oh darn, versioning closely evolving packages is quite tricky it seems.

The problem is that we couldn't have anticipated the upper bound at the time Geodesy-0.1.0 was released.  A version Geodesy-0.1.1 could be released, but that wouldn't fix the version breakage with the declared REQUIRE in 0.1.0.

Is the right thing to fix the Geodesy REQUIRE upper bound for CoordinateTransformations here in metadata?

tkelman:
Yes. Do all past tags of Geodesy require on CoordinateTransforms? The upper bound would need to be added to all past tags (assuming they'd all be equally broken with the newest CoordinateTransforms version) to avoid downgrading Geodesy instead of holding back CoordinateTranforms if you want both.

c42f:
No, Geodesy-0.0.1 didn't depend on it.

tkelman:
So what's worse: downgrading to Geodesy 0.0.1 if you want the latest version of CoordinateTransforms, or keeping CoordinateTransforms held back if you want to use the latest version of Geodesy? Unless we add a "fake" dependency from Geodesy 0.0.1 on CoordinateTransforms with an upper bound, either might happen depending on which solution the dependency resolution thinks is better.

c42f:
Good question. Consistent Geodesy states are:

Geodesy-0.0.1
Geodesy-0.1.0 with CoordinateTransformations-0.1.0 # API here will be truly obsolete soon
Geodesy-0.2.0 with CoordinateTransformations-0.2.0 # Should be doable in short order.

CoordinateTransformations-0.1.0 was a bit of an API aberration since it was the first version - let's get people off that ASAP.

PR #5954, opened 2016-08-13
Register and tag FunctionWrappers.jl v0.0.1

yuyichao:
As described in the [README](https://github.com/yuyichao/FunctionWrappers.jl/blob/aa1318c83d1c4c80586700f7a43e54ac49b9c1d1/README.md#limitations) this is a proof of principle implementation of JuliaLang/julia#13984 with several limitations. However, it seems that some people are interested in using it so here it is.

This requires Jameson's cfunction overhaul so it's 0.5+ only.

CI tests passes on 0.5/master Linux/OSX/Windows with (nominally) 100% coverage.

Closes https://github.com/yuyichao/FunctionWrappers.jl/issues/1


yuyichao:
= = .... apparently `PkgDev.register` picks up my git alias ....................

PR #5947, opened 2016-08-12
Tag Optim v0.6.1 [https://github.com/JuliaOpt/Optim.jl] 

pkofod:
Better deprecations, ref #5943 
changes: https://github.com/JuliaOpt/Optim.jl/blob/276c98d3a61432a14fad388debce91d056653b88/src/nelder_mead.jl#L47

might be a weird way to do it, but it works :)

PR #5957, opened 2016-08-13
Tag Optim v0.6.0

pkofod:
Okay, taking this as a learning experience, I decided to start over. replaces #5947


pkofod:
aaand, we're building https://travis-ci.org/JuliaOpt/Optim.jl/builds/152024162

This has the proper deprecation of the constructor, and also tests checking that the old syntax works.

tkelman:
Looks like this broke the tests of DSGE.jl: http://pkg.julialang.org/logs/DSGE_0.4.log

So I think we'll need to apply some upper bounds again.

pkofod:
I didn't even realize we exported that type, as we use it internally for traces.

I'm not sure if this already exists in `PkgDev`, but it would be wonderful to have a `PkgDev.test_dependents("Optim")`'ish function.

I guess it isn't sufficient if I push a `:patch` that fixes this from `Optim`'s side?

tkelman:
Fixing from Optim's side could work, if you have an idea for how to fix the issue there.

pkofod:
I have to go afk for a while, but I will fix it later. I thought something like the following would work
```julia
abstract Optimizer

immutable NelderMead <: Optimizer
end

immutable OptimizationState{T <: Optimizer}
    iteration::Int
    value::Float64
    g_norm::Float64
    metadata::Dict
end

typealias OptimizationTrace{T} Vector{OptimizationState{T}}

OptimizationTrace{NelderMead}()

OptimizationTrace(NelderMead()) # Doesn't work
OptimizationTrace{T}(method::T) = OptimizationTrace{T}() # Doesn't work:
# ERROR: cannot define function OptimizationTrace; it already has a value

Base.convert{T<:Optimizer}(::Type{OptimizationTrace{T}}, method::Type{T}) = OptimizationTrace{T}()

OptimizationTrace(NelderMead())
```

`OptimizationTrace` wasn't an abstractarray earlier, so we used a `OptimizationTrace(method)` constructor like the one I am trying to define, but when `OptimizationTrace` is a typealias, I am having some trouble finding out how to make `OptimizationTrace(method)` where `typeof(method)<:Optimizer` return `OptimizationTrace{typeof(method)}()`.

pkofod:
I see that what I originally wanted to do
```julia
abstract Optimizer

immutable NelderMead <: Optimizer
end

immutable OptimizationState{T <: Optimizer}
                  iteration::Int
                  value::Float64
                  g_norm::Float64
                  metadata::Dict
              end

typealias OptimizationTrace{T} Vector{OptimizationState{T}}

OptimizationTrace{T<:Optimizer}(method::T) = Vector{OptimizationState{T}}()
OptimizationTrace(NelderMead())
```
works on v0.5, but unfortunately not on v0.4. I guess I have to fix it in DSGE.jl, and for now put an upper limit on DSGE.jl.

PR #5961, opened 2016-08-13
Tag GitLab v0.0.2 [https://github.com/JuliaComputing/GitLab.jl]

mdpradeep:


GitLab/versions/0.0.2/sha1
```diff
@@ -0,0 +1 @@
+60354dec877ab64ae06310c807490736952ca667
```
tkelman: this is the same sha as 0.0.1

mdpradeep: I used `PkgDev.tag()`. Let me check.

tkelman: the git tag https://github.com/JuliaComputing/GitLab.jl/tree/v0.0.2 should also be updated

PR #5964, opened 2016-08-13
Register BeaData.jl

sbnicar:
Requesting publish for package BeaData.jl


kmsquire:
BureauOfEconomicAnalysisData.jl is a bit too long, but I wonder if, e.g., BEA_Data.jl would be a better name?  

AKAIK, there aren't any other packages which have underscores, but then, there also aren't many packages which begin with an acronym and end with a word...

Anyway, unless this is a clear win, don't let this comment prevent registering.

sbnicar:
The name was modeled after BlsData and FredData, both registered packages that query economic data APIs and start with acronyms.

sbnicar:
I should add that I don't really have any objection to changing the name, but iI thought an underscore or "BEAData" would be bad style, given the examples provided in the documentation.

kmsquire:
Okay, thanks for the explanation.  The name should be fine then.

milktrader:
FWIW, I don't mind BeaData.

tkelman:
If your tests are all in a try, any failures wouldn't be very visible.

PR #5966, opened 2016-08-14
Tag CoordinateTransformations v0.3.0 [https://github.com/FugroRoames/CoordinateTransformations.jl]

andyferris:
This is a Julia v0.5 version of CoordinateTransformations, meant to work in parallel with CoordinateTransformatoins v0.2.0 #5953 which is the v0.4-compatible version.

The main differentiator here is that this uses *StaticArrays* instead of *FixedSizeArrays*, and the associated Julia v0.5 version of *Rotations*.

CC @c42f - does this seem right to you?


c42f:
The require should have a minimum version for Rotations I guess.  Other than that, looks good.

andyferris:
Now Rotations is tagged in METADATA, we can do that.

andyferris:
OK this should be good now!

andyferris:
Now we are doubly-good-to-go. @tkelman :)

PR #5967, opened 2016-08-14
Tag DSP v0.0.12

simonster:


tkelman:
this is the first version to drop julia 0.3 support, should use a new minor package version

PR #5974, opened 2016-08-15
Atom and CodeTools

MikeInnes:


tkelman:
You're still using CodeTools in Atom.jl, so removing it from REQUIRE isn't going to work.

PR #5975, opened 2016-08-15
Tag Dopri v0.3.0, Tag JPLEphemeris v0.4.0

helgee:
0.5-compatibility and refactoring.


JPLEphemeris/versions/0.4.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BaseTestNext
+Compat
```
tkelman: best to include a minimum version requirement

PR #5978, opened 2016-08-15
Tag ImmutableArrays v0.0.12

sjkelly:
This fixes deprecation warnings on 0.5.


ImmutableArrays/versions/0.0.12/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3
+Compat
```
tkelman: best to have a minimum version requirement that implements the features you're using

PR #5981, opened 2016-08-15
Register ConicIP: v0.1.0, v0.1.1

gabgoh:
https://github.com/JuliaLang/METADATA.jl/pull/5839

third time's the charm, hopefully


mlubin:
@tkelman, okay to merge this?

ConicIP/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4+
+MathProgBase
```
tkelman: not having an upper bound here is likely to cause issues when MathProgBase 0.6 happens

ConicIP/versions/0.1.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4 
+MathProgBase 0.5 0.6
+WoodburyMatrices
+Compat
```
tkelman: should declare a minimum version dependency that implements `Compat.view` (or any more recent functionality you need)

ConicIP/versions/0.1.0/requires
mlubin: Probably best to just drop the 0.1.0 tag?

tkelman: @gabgoh, Miles was referring to here in metadata. It looks like you deleted the git tags from https://github.com/MPF-Optimization-Laboratory/ConicIP.jl/releases, so let's re-do this again. Add a minimum version requirement on Compat (from a quick glance it didn't look like you were using anything more recent than `Compat.view`) then we can re-tag and merge.

PR #5987, opened 2016-08-16
Update requirements to reflect changes to dependency

andyferris:
Rotations 0.3 has a different API to Rotations 0.1/0.2, and this version of CoordinateTransformations requires one of the latter.

### DIFF
--- a/CoordinateTransformations/versions/0.2.0/requires
+++ b/CoordinateTransformations/versions/0.2.0/requires
@@ -1,4 +1,4 @@
 julia 0.4
 Compat 0.8
 FixedSizeArrays 0.2.2
-Rotations
+Rotations 0.1.0 0.2+


andyferris:
CC @tkelman

CoordinateTransformations/versions/0.2.0/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.4
 Compat 0.8
 FixedSizeArrays 0.2.2
-Rotations
+Rotations 0.1.0 0.2+
```
tkelman: I don't think you want the plus

andyferris: Oh? I wanted to allow for any version 0.2 but not any version 0.3... 

For lower bounds `0.2` must mean `0.2.0`, but for upper bounds is it `0.2.x` for any `x`?

tkelman: Ah okay. Upper bounds are exclusive. So I think this will work, but might be clearer as `0.3`

PR #5988, opened 2016-08-16
Tag JSON 0.6.1, 0.7.0

kmsquire:
v0.6.1:
 * use `getfield(foo, bar)` rather than `foo.(bar)` for 0.5
 * fix off-by-one error in length check
 * various deprecation fixes
 * add set of serialization microbenchmarks
 * add v0.5 to Travis and Appveyor

v0.7.0:
    Add JSON.lower for a better serialization API

   - Introduce `JSON.lower`, designed to be overloaded, that converts an object to another “primitive” object that `JSON` can serialize (`Dict`, `Array`, etc.).
   - In the generic `JSON._print` fallback, call `JSON.lower` instead of the current dumping behaviour
   - Add a generic `JSON.lower` fallback that does the current dumping behaviour
   - Rename `JSON._print` to `JSON._writejson`, and add a deprecated fallback to a new `JSON._print` with no methods defined

Cc: @TotalVerb 
Closes https://github.com/JuliaLang/JSON.jl/issues/156


JSON/versions/0.7.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3
```
tkelman: I'd actually recommend dropping Julia 0.3 support while you're at it. I plan on freezing METADATA for Julia 0.3-supporting packages relatively soon, so I can move PackageEvaluator to testing against 0.6 and not worry much about leaving 0.3 no-longer-tested.

TotalVerb: See https://github.com/JuliaLang/JSON.jl/pull/164. Is it necessary to drop 0.3 on these versions (0.6.1 and/or 0.7.0) as well, or would it suffice to drop on future versions?

tkelman: future would be fine. near future would be advised.

tkelman: one consequence of having things this was is that you'd be introducing deprecations in a version that supports 0.3, but other packages aren't going to be updated there. so probably better not to have a bunch more warnings there.

TotalVerb: So best would be 0.6.1 with 0.3 support, but 0.7.0 with minimum 0.4?

tkelman: I think that would be a good idea given the new deprecation in 0.7

TotalVerb: @kmsquire Could you bump this minimum version to 0.4?

kmsquire: @TotalVerb, I changed the tag to point to your commit which actually drops v0.3 support.

PR #5989, opened 2016-08-16
Tag HttpParser v0.1.2

shashi:


tkelman:
This was practically rewritten to use a new version of the library, should be a new minor version at least.

tkelman:
This also does no verification of the binary it may find, so upgrading to it for users who have existing binaries in place isn't going to work at all.

shashi:
cc @ranjanan

PR #5991, opened 2016-08-16
Tag QuartzImageIO v0.1.5

kmsquire:
Cc: @rsrock @timholy


QuartzImageIO/versions/0.1.5/sha1
```diff
@@ -0,0 +1 @@
+e5dfdd154bcba00404d936af269c31541303c2bd
```
tkelman: this is the same sha as 0.1.4

kmsquire: I checked it out on v0.5, but tagged and pushed on v0.4...  fix coming.

kmsquire:
Okay, never mind.  I thought this wasn't tagged, but it's just that I hadn't run `Pkg.update()` on Julia v0.5.

PR #5992, opened 2016-08-16
Tag Coverage v0.3.1 [https://github.com/JuliaCI/Coverage.jl]

timholy:
Coverage hasn't been tagged in a long time (~9 months), so I thought I'd do it (CC @IainNZ).


tkelman:
https://github.com/JuliaCI/Coverage.jl/pull/113 might be a good idea to fix, and I believe https://github.com/JuliaCI/Coverage.jl/pull/99 will fail on windows

PR #5993, opened 2016-08-16
Tag Geodesy v0.3.0 [https://github.com/JuliaGeo/Geodesy.jl]

andyferris:
Julia 0.5 version based of Julia 0.5-only deps CoordinateTransformations 0.3 and StaticArrays


andyferris:
The second commit should mean that version 0.2.0 will still function properly on Julia 0.5 (similar to the recent CoordinateTransformations fixup, #5987)

andyferris:
CC @c42f

PR #5994, opened 2016-08-16
Put upper limit on Optim

pkofod:
xref to fix in DSGE.jl: https://github.com/FRBNY-DSGE/DSGE.jl/pull/6
xref to discussion #5957

### DIFF
--- a/DSGE/versions/0.1.2/requires
+++ b/DSGE/versions/0.1.2/requires
@@ -3,6 +3,6 @@ Calculus
 DataStructures
 Distributions
 HDF5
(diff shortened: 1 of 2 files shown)


tkelman:
Does 0.1.2 not have the same problem?

pkofod:
Thank you, yes it does. I re-pushed.

tkelman:
Good thing it was an easy fix, thanks for this!

pkofod:
> Good thing it was an easy fix, thanks for this!

Yeah, sorry for the noise.

PR #6000, opened 2016-08-16
Compatibility for v0.5

cormullion:


tkelman:

cormullion:
@tkelman  I'll read up about that... (I think your video at JuliaCon  might be a place to start...)

tkelman:
Try https://travis-ci.org/cormullion/ColorSchemes.jl, the default yml file template that was already pre-generated should be good enough to start with

cormullion:
@tkelman thanks Tony. It wasn't as painful as I thought:)

PR #6002, opened 2016-08-16
Pull request/bdebeeba

montyvesselinov:
Fixes to address Lora and Optim changes.


Mads/versions/0.2.2/sha1
```diff
@@ -0,0 +1 @@
+caae69cd519007fbbae7389cff3613a7b4c016cd
```
tkelman: you should push these commits to master so they're actually available on a branch

montyvesselinov: Pushed. Sorry. Forgot. Thank you very much!

PR #6007, opened 2016-08-16
Tag Celeste v0.1.1 [https://github.com/jeff-regier/Celeste.jl]

jeff-regier:


Celeste/versions/0.1.1/requires
```diff
@@ -0,0 +1,12 @@
+julia 0.5-
```
tkelman: since this drops support for julia 0.4, it should use a new package minor version via `Pkg.tag("Celeste", :minor)`

tkelman:
can you rebase so you aren't pulling in unrelated commits here?

PR #6015, opened 2016-08-17
Bump CSV

quinnj:


CSV/versions/0.0.10/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4.1
+DataStreams 0.0.8
+DataFrames
```
tkelman: the package require file and the metadata requires content should match. this looks right, but the package copy removed things it shouldn't have

quinnj:
What's the best way to fix this? Should I edit the CSV release tag to something like "broken release" and then do another v0.0.10 release? Is that ok to do?

tkelman:
Since this hasn't gone into metadata, you can remove the 0.0.10 tag and re-do it if that's easiest. How did this mismatch happen?

quinnj:
Ok, think I got this sorted out now.

PR #6023, opened 2016-08-17
Tag IntervalTrees v0.1.0 [https://github.com/BioJulia/IntervalTrees.jl]

bicycle1885:
* Drop support of Julia 0.3.
* Fix deprecated warnings on Julia 0.5.


IntervalTrees/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: what's the minimum version that you need?

PR #6035, opened 2016-08-18
Tag ImageProjectiveGeometry v0.1.2

peterkovesi:


tkelman:
Can you push the git tags so they show up at https://github.com/peterkovesi/ImageProjectiveGeometry.jl/releases?

It's best to declare a minimum version dependency on Compat that implements the features you're using.

peterkovesi:
I have updated my REQUIRE to request Compat 0.8.7 and I have pushed my git tags so I now have some releases listed for the package.  Do I need to submit a new push request for METADATA so that it incorporates my Compat requirement?  I am not sure how to do this through the package manager without creating (yet) another tag.

Sorry to be a bit helpless here.  The Package docs are fine for helping you create a new package but I have been struggling to work out and keep track of all the different things needed to keep a package maintained and updated (for example I did not realise I had to do an explicit pushing of tags to get releases to show up!)  Thanks for your patience

tkelman:
you can modify the contents of the files in place on this branch of your metadata fork. or it should just get picked up at your next tag. 0.8.7 is likely too tight of a bound unless you're using the features that were just implemented in the most recent version, but it's better to have bounds too tight than too loose.

peterkovesi:
I've set the Compat requirement to 0.8.7  It may be tight but I guess everyone keeps (or should keep) Compat up to date

PR #6036, opened 2016-08-18
Bump SQLite

quinnj:


tkelman:
NullableArrays should still be listed in test/REQUIRE, shouldn't it?

quinnj:
Hmmm.....probably. Though DataStreams requires it. I'll make the update, but would prefer to merge this now if that's ok.

PR #6041, opened 2016-08-18
Tag TypedTables v0.1.2 [https://github.com/FugroRoames/TypedTables.jl]

andyferris:


tkelman:
I suspect Compat 0.8.6 is more recent than you absolutely need.

PR #6042, opened 2016-08-19
Tag Documenter v0.3.0

mortenpi:
cc @MichaelHatherly


MichaelHatherly:
Thanks. Probably worth posting a short announcement on the julia-users mailing list today if you have time.

PR #6044, opened 2016-08-19
Register RData

alyst:
As discussed in JuliaStats/DataFrames.jl#955, `RData` is the package to read R data files.


tkelman:
Were you the original author of all the code imported from DataFrames? Either this should use or acknowledge DataFrames' MIT license, or anyone who contributed to the code while it was there would have to grant permission for it to be relicensed to BSD.

RData/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+DataFrames
+DataArrays
+FileIO
+GZip
+Compat
```
tkelman: best to list minimum version requirements for the features you're using

alyst:
@tkelman Oh, thanks, I've completely overlooked the autogenerated LICENSE. Updated to MIT and referenced the main authors.

Versions of the required packages are now specified as well.

nalimilan:
I would have been good to preserve history when importing source from DataFrames. I guess it's still possible to do that, rebasing all commits on top of that won't create any conflicts (though you'll have to force-push to master).

alyst:
@nalimilan Yes, that would nice. Do you know how I should call `git filter-branch` (or, perhaps, some other tool) to keep only `src/RDA.jl` and `test/RDA.jl`?

tkelman:
you wouldn't have to force push, you could do a merge commit to combine the histories

nalimilan:
@alyst I've never done that, but I could find several blog posts giving instructions. For example: http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/

PR #6050, opened 2016-08-20
Tag KNITRO v0.1.2

yeesian:


yeesian:
Oh shucks, I just remembered it's a [breaking change](https://github.com/JuliaOpt/KNITRO.jl/pull/48#issue-150511648). Should I be tagging a minor release instead?

cc @mlubin

PR #6051, opened 2016-08-20
Tag CxxWrap v0.1.6

barche:
Bugfixes and new features following reports by @jstrube


tkelman:
why were you using `long` to begin with?

barche:
@tkelman Thanks for the quick merge. I was hoping `long` would be 32 bits on 32 bit systems and 64 bits on 64 bit systems, i.e. behave as `Int`, but testing on Windows brought me back to the real world ;)

PR #6053, opened 2016-08-20
Register BARON: v0.1.0

joehuchette:


tkelman:
your require file is missing Compat. best to add it explicitly, with a minimum version

joehuchette:
Nice catch, thanks.

PR #6060, opened 2016-08-21
TopicModelsVB v0.0.1

esproff:
I've never registered/published/tagged a package before, not sure if this is what I'm supposed to do.  Do I need to wait for someone to verify something before it gets officially listed?


TopicModelsVB/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Distributions
```
tkelman: looks like this is missing OpenCL?

esproff: @tkelman Yes you're right, thanks, I just added it.

tkelman:
Would be a good idea to add some unit tests and enable Travis CI (https://travis-ci.org/esproff/TopicModelsVB.jl) going forward.

PR #6061, opened 2016-08-21
Register SugarBLAS [https://github.com/lopezm94/SugarBLAS.jl.git]

lopezm94:


lopezm94:
Is PkgDev.tag required before PkgDev.publish? Possible PR to the [docs](http://docs.julialang.org/en/latest/manual/packages/#tagging-and-publishing-your-package) would be good if thats the case.

tkelman:
Yes, we don't allow registered packages without a tagged version any more, since those can't be added with `Pkg.add`.

ViralBShah:
Nice idea!

lopezm94:
Thank you, I'm in progress to add the rest of the functions putting all the parameters inside of square  brackets. If anyone has a better idea feel free to write in the only issue of the package.

PR #6068, opened 2016-08-22
tag Restful v0.3.0

ylxdzsw:


Restful/versions/0.3.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
```
tkelman: This won't be installable until 0.5.0 final is tagged. May want to use one of the rc's as a minimum in the meantime.

ylxdzsw:
Thank you for pointing out it. Now I have changed it to `julia 0.5.0-rc2`

PR #6070, opened 2016-08-22
Pull request/33411c9d

denizyuret:
AutoGrad.jl is an automatic differentiation package for Julia. It is a Julia port of the popular Python autograd package: It can differentiate native Julia code that includes loops, conditionals, closures etc. and it can handle higher order derivatives.


tkelman:
Neat. cc @mlubin @jrevels. Does this do forward mode, reverse mode, both? How does this compare to some of the existing packages like ReverseDiffSource?

denizyuret:
AutoGrad does reverse mode differentiation.  My understanding is that forward mode is too expensive for machine learning applications I am interested in where the input is large and the output is scalar.  Unlike ReverseDiffSource it does not differentiate from source, it keeps track of primitive operations during forward pass and uses them to backprop.  This has the advantage of easily supporting conditionals, loops, helper functions, closures etc. which I believe are not very easy to do if you are working with the source.

mlubin:
Always glad to see more AD packages.

PR #6092, opened 2016-08-23
Tag ForwardDiff v0.2.3 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:


tkelman:
I think this broke JuMP on PkgEval http://pkg.julialang.org/logs/JuMP_0.4.log:
```ERROR: LoadError: LoadError: MethodError: `jacobian_dual_type` has no method matching jacobian_dual_type(::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},1}, ::ForwardDiff.Chunk{2})```

cc @mlubin @joehuchette

tkelman:
Also Rotations.jl http://pkg.julialang.org/logs/Rotations_0.5.log, cc @andyferris

andyferris:
@tkelman Is there some way (e.g. for me) to initiate the package evaluator for a specific package so I can easily check if v0.2.4 fixes this? Or other than that, how frequently will package evaluator run?

tkelman:
runs nightly at 12:20 am eastern us time. you can run it locally anywhere you have vagrant but it takes 6 cores and 9-10 hours to run everything. you can locally check things out and compare, if there's a major difference then that could be looked in to.

tkelman:
pkgeval doesn't easily support running on a single package at a time yet, but that's a rewrite that I should get to if release stuff ever calms down.

andyferris:
> runs nightly at 12:20 am eastern us time. you can run it locally anywhere you have vagrant but it takes 6 cores and 9-10 hours to run everything. you can locally check things out and compare, if there's a major difference then that could be looked in to.

Right, thanks. I suppose I was aiming for the lazy option here: click a button in my browser and see if the problem disappears.

> pkgeval doesn't easily support running on a single package at a time yet, but that's a rewrite that I should get to if release stuff ever calms down.

lol, sounds great. I suppose you could run it more frequently (e.g. every PR on METADATA?) if it only re-ran the changed package and its dependency(s).

PR #6093, opened 2016-08-23
Bump DataStreams packages

quinnj:
I've tested each of these locally on 0.4 and 0.5 with all tests passing. I'm bumping them all at once to try and avoid what happened last time with existing package versions breaking due to the staggered updates.


tkelman:
If you keep breaking the API, then upper bounds between these packages would be another way of ensuring users don't see breakage.

tkelman:
SQLite.jl is broken on 0.4: http://pkg.julialang.org/logs/SQLite_0.4.log

tkelman:
Since tags of this set of packages have yet to result in things not breaking, I'd recommend applying upper bounds to existing tags when you change API's, tagging new package releases one by one and paying close attention to CI. Your local tests are missing something.

PR #6094, opened 2016-08-23
Registering and tagging a package for parallel MCMC

KrisDM:


KrisDM:
Second attempt. Package tagged correctly this time (I hope).

PR #6096, opened 2016-08-23
Tag BioMedQuery v0.1.0

mirestrepo:
Up (minor) version with updates to support MySQL backend
UP the version requirements for DataStreams and SQLite


tkelman:
Would be best to add DataFrames and NullableArrays to your REQUIRE file if you use them, instead of assuming they'll be present as indirect dependencies.

mirestrepo:
@tkelman sounds good. I'll do that in the next update. Thanks

PR #6101, opened 2016-08-23
Register Bukdu

wookay:
Hello,
Bukdu is a web development framework.
It's influenced by Phoenix framework.

It can be running on Heroku something like
https://bukdu.herokuapp.com/

thanks!


tkelman:
Neat! Just curious, does the name mean anything? That's a fun word to say.

wookay:
It's from the name of Korean constellation "The Big Dipper".
https://ko.wikipedia.org/wiki/북두칠성

The Big Dipper (US) or Plough (UK) is an asterism consisting of the seven brightest stars of the constellation Ursa Major.
https://en.wikipedia.org/wiki/Big_Dipper

tkelman:
That's great, today I learned.

Reminds me, I wonder whether everything would work correctly if someone tried to generate and register a package with a unicode name. I guess we should be prepared for that eventuality.

wookay:
thank you!

KristofferC:
In Sweden we consider it a wagon :)

PR #6104, opened 2016-08-24
Tag WriteVTK v0.3.6

jipolanco:


WriteVTK/versions/0.3.6/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+LightXML 0.2.1
+Libz 0.1.1
+BufferedStreams 0.1.1
+Compat 0.7.15
```
tkelman: `view` needs a newer minimum than this

PR #6113, opened 2016-08-24
Register and tag Query v0.0.1

davidanthoff:


tkelman:
This is a rather generic name, which has been discouraged in the past.

You use `NDSparseData` in your `rest/runtests.jl` file so it should be included in `test/REQUIRE`, not manually cloned on CI.

ViralBShah:
Facing this problem with Graphs.jl now, which is very generic and unmaintained. However, tough to tell what package will take off and once it does, the name is difficult to change anyways.

So +1 to having a slightly less generic name, but ok to have it as Query.jl too.

davidanthoff:
> You use NDSparseData in your rest/runtests.jl file so it should be listed in test/REQUIRE, not manually cloned on CI.

NDSparseData isn't registered, so I can't add it to a REQUIRE file. As soon as JuliaComputing/NDSparseData.jl#25 is done I'll change that, but for now the existing approach seems the only way to still run the NDSparseData related tests.

The name is very generic, but then the package is also quite generic and I feel it describes the role of the package well. My sense is that the name is the right one if the package takes off. If it doesn't, we could still rename it to "free" up the generic name again? Probably easier to rename a package that died than to rename a package that is a success? I should also say that this is not an experimental balloon I'm starting, I'm committed to maintain this package.

tkelman:
Tests of a registered package shouldn't depend on an unregistered one. You could make those tests optional or wait until NDSparseData is registered.

davidanthoff:
What is the harm of the current setup? I certainly want to test the NDSparseData in CI builds and don't want to have them optional. I could add something to the README saying that if you want to run the tests locally, you need to first clone NDSparseData.

tkelman:
`Pkg.add("Query"); Pkg.test("Query")` will fail (and hence so will packageevaluator). You could make the tests give a warning if NDSparseData isn't installed.

tkelman:
> we could still rename it to "free" up the generic name again? Probably easier to rename a package that died than to rename a package that is a success?

Renaming any package after it's been registered is problematic if anyone has installed it or anything else depends on it. It would need to be deprecated under its old name, which makes freeing up the name for some other use very time consuming and complicated.

nalimilan:
LINQ.jl was a nice name given that it's "heavily inspired by LINQ". Are you afraid of trademarks?

davidanthoff:
> LINQ.jl was a nice name given that it's "heavily inspired by LINQ". Are you afraid of trademarks?

Yes, plus the whole thing isn't language integrated, so the name would be essentially wrong ;) I also expect the package to diverge from the original LINQ specification over time, and I want to make sure it has its own identity once that happens, and is not just seen as a LINQ clone.

johnmyleswhite:
FWIW, I think it's fine to have the package be called Query.jl.

davidanthoff:
> Pkg.add("Query"); Pkg.test("Query") will fail (and hence so will packageevaluator). You could make the tests give a warning if NDSparseData isn't installed.

Yep, you are right, I forgot about package evaluator. I just pushed changes that put ``using NDSparseData`` behind a check whether the package is installed, so ``Pkg.add("Query"); Pkg.test("Query")`` should work now. I left the clone of NDSparseData in the travis file so that these tests run during CI, but they won't run during the package evaluator runs.

I changed the commit of the v0.0.1 tag and force pushed this PR to point to the new right sha for v0.0.1. I hope that was ok, but given that nothing was merged yet that seemed the simplest solution.

ViralBShah:
Cc @JeffBezanson For NDSparseData. IIUC, it is changing too quickly and does not make sense to register just yet, but best if Jeff chimes in.

davidanthoff:
Thanks, @ViralBShah! I think with the current setup for tests the NDSparseData situation doesn't matter that much, now that I have implemented @tkelman's recommendation of making those tests optional we should be good.

davidanthoff:
Thanks!

PR #6116, opened 2016-08-24
Tag Mamba v0.10.0 for julia 0.5.

brian-j-smith:
This is a pull request for versions 0.9.2 and 0.10.0 of Mamba that I am maintaining in separate branches for compatibility with julia 0.4 and 0.5, respectively.  Accordingly, the requires files indicate non-overlapping julia versions (0.4 = ``julia 0.4.1 0.5-`` and 0.5 = ``julia 0.5.0-rc3``) to ensure that the correct Mamba versions are installed automatically for users.  This is the approach I took in transitioning from julia 0.3 to 0.4, and it went without out hitch.  However, feel free to let me know if there are any questions or comments.  Thanks.


Mamba/versions/0.9.2/sha1
```diff
@@ -1 +1 @@
-12a502c24b6673d6073fa79fa9fed1a292702048
+042449fec196ee27d7386ad983c8f593932a6e28
```
tkelman: you generally shouldn't modify the sha of existing tags

Mamba/versions/0.9.2/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5-
```
tkelman: Note that upper bounds are actually not very effective unless applied to all past tags. Otherwise if someone is using an 0.5 prerelease prior to rc3 so 0.10.0 won't work, this will just downgrade them to 0.9.1. The upper bound should be unnecessary if you have newer tags for 0.5.

tkelman:
Aside from the specific comments above, the strategy of maintaining separate branches for different Julia versions is a totally valid alternative to worrying about using Compat correctly. I think more package authors should actually consider doing things this way.

brian-j-smith:
Thanks for the comments.  As noted, one shouldn't change the sha in an existing tag.  Also, since julia installs the latest compatible package version (Mamba 0.10 will be installed on julia 0.5 and Mamba 0.9 on julia 0.4), the upper bound is unnecessary.  Hence, I removed my 0.9.2 tag changes from the pull request.

Having the experience of using this approach in the past, I am a fan.  In my case, some of the changes could not have been handled solely with the Compat package.  But, more generally, I feel it has kept my code cleaner, easier to maintain, and more up-to-date with julia features/syntax.  Plus, people seem to be moving to the most recent julia versions more quickly than in some other software projects where it might be more important to maintain backward compatibility longer.

PR #6117, opened 2016-08-25
Tag LibSndFile v0.3.0

ssfrr:


tkelman:
you're `using Compat` without declaring a dependency on it

ssfrr:
ah, whoops. I didn't notice because it's pulled in by a dependency. Fixed in 8fda745. Thanks for the catch.

tkelman:
> it's pulled in by a dependency

That's not guaranteed to always be the case for all versions of your dependencies, so best to re-tag that or make a new 0.3.1 tag.

PR #6118, opened 2016-08-25
Tag Interact v0.3.2

shashi:


tkelman:
Can you think of an organization that we could eventually move Interact (and Reactive and a few other similar ones) to in order to lessen JuliaLang's CI load? I'm not sure what would be best for it. Maybe JuliaInterop, or just make a new one?

PR #6119, opened 2016-08-25
Register and Tag NetworkLayout: v0.0.1

abhijithanilkumar:
[NetworkLayout.jl](https://github.com/JuliaGraphs/NetworkLayout.jl) is a package that houses graph and tree layout algorithms in Julia. Uses algorithms from GraphLayout, GraphPlot and PlotRecipes.


tbreloff:
Congrats on the package. Looking forward to incorporating this into
PlotRecipes.

tkelman:
Please push the git tag so it shows up under https://github.com/JuliaGraphs/NetworkLayout.jl/releases

tkelman:
`src/stress.jl` uses FixedSizeArrays but your REQUIRE file does not declare that dependency. You shouldn't do `Pkg.clone("LightGraphs")` on travis, let `test/REQUIRE` handle that.

abhijithanilkumar:
@tkelman The release version of LightGraphs throws errors with 0.5. That was why `Pkg.clone("LightGraphs")` was included.

tkelman:
>  I've added FixedSizeArrays as dependency

You added it to `test/REQUIRE`. That's not right for a dependency that you use in `src`, it needs to be in `REQUIRE`

abhijithanilkumar:
> You added it to test/REQUIRE. That's not right for a dependency that you use in src, it needs to be in REQUIRE

Sorry. My bad. Corrected it.

tkelman:
the requirement needs to be reflected in the requires files of the metadata tags too.

abhijithanilkumar:
Made the changes.

NetworkLayout/versions/0.0.2/sha1
```diff
@@ -0,0 +1 @@
+20a54d9a28d1aff7b471b9f99e2e73f16a7b2db3
```
tkelman: these sha's are inconsistent with your git tags https://github.com/JuliaGraphs/NetworkLayout.jl/releases

abhijithanilkumar:
@tkelman Is it okay now? I've reset the tags.

PR #6124, opened 2016-08-25
Tag Gallium and dependencies

Keno:


ASTInterpreter/versions/0.0.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
+JuliaParser 0.7.2
+AbstractTrees 0.0.2
```
tkelman: `getnode` doesn't exist until 0.0.4

tkelman:
I think last time you tagged, the checks weren't actually running on 0.5-only packages. ELF and DWARF can just be re-tagged as 0.1.0 and it should make it happy.

Keno:
That doesn't make much sense to me given that every version since v0.0.1 has required 0.5-. It's just v0.0.0 that doesn't, from way back when we still did that.

tkelman:
It checks that the minimum Julia version doesn't change during the same package minor series, so 0.0.0 is considered the same minor as 0.0.5. We could add an exception for 0.0.0, but given we don't allow any new packages to be tagged 0.0.0 it's probably easier to just re-tag these.

Keno:
Or we could delete 0.0.0, or we could override the status check. I was planning to tag all of these at 0.1.0 at once, once 0.5 is released.

tkelman:
We shouldn't delete existing tags. Overriding the status check is a bit more work, is it really that important to have the version numbers of these all the same?

Keno:
Overriding the status check is no work at all, I'll just click the button.

tkelman:
No, that would then cause every metadata PR to fail. Don't do that.

Keno:
Ugh, fine. I'll retag these.

tkelman:
Only ELF and DWARF should need it from what I can see. And ASTInterpreter's dependency on AbstractTrees is too loose.

ASTInterpreter/versions/0.0.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
+JuliaParser 0.7.2
+AbstractTrees 0.0.4
```
tkelman: should also be reflected here https://github.com/Keno/ASTInterpreter.jl/blob/master/REQUIRE

PR #6128, opened 2016-08-26
Tag DifferentialEquations v0.3.0

ChrisRackauckas:
Tagging a new version. This is the last version that will have v0.4 compatibility. Many things were added. Sundials.jl is now wrapped. Over 100 Runge-Kutta methods were added and stepsize selection was optimized. A benchmark suite was added and notebooks showing the differences between algorithms are included (summary: ODE.jl is much slower than the others with lots of error, ODEInterface is fast, DifferentialEquations.jl achieves by far the least error but is slightly slower) and the documentation was extended to recommend the most efficient algorithms for non-stiff problems (essentially: native DifferentialEquations.jl algorithms or dop853). 

Other interesting features were added. There's a set of macros to make defining systems of ODEs look more like psudocode, for example to define the Lorenz system:

```julia
f = @ode_define begin
  dy₁ = -k₁*y₁+k₃*y₂*y₃
  dy₂ =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
  dy₃ =  k₂*y₂^2
end k₁=>0.04 k₂=>3e7 k₃=>1e4
```

The API was changed to match ODE.jl/ODEInterface/Sundials, i.e. functions are defined as (t,u,du) instead of (du,u,t) (although it doesn't match Julia standards since `du` is in-place altered, (t,u,du) was chosen at least by ODEInterface and Sundials because it was the Fortran standard, so it makes sense to conform to them). 

Note that code coverage went way down because after changing the premade problems to be defined by anonymous functions (best for design in v0.5), the fact that [Julia segfaults on inlined Bigs in packages](https://groups.google.com/forum/#!topic/julia-users/cSliYJ-a_wE) made me have to avoid those tests on CI (though they can still be ran by copy pasting into the REPL, then they won't segfault).

By tagging this version, I am dropping v0.4 support so that way I can go forward with multi-threading. This is already implemented, though all of those parts are commented out so that it could build on v0.4. Given that the DifferentialEquations.jl algorithms were already shown to be vastly more efficient than the other Julia non-stiff solvers, I hope that within-method parallelism can increase the gap even more. Then while Julia fixes the segfaulting and threading boxing issues, I'll get together a large set of benchmarks for non-stiff methods and optimize some loose ends. That will be a v0.3.1 in probably 2 weeks. Then the next minor will be about adding support for parameters, i.e. functions (t,u,du,p), completing support for SIUnits (there's already a branch where it works. Plot recipes need to be defined though). I might optimize some Adams methods for fun, this will be v0.3.2. Then v0.4 will be about adding some stiff methods (BDF / NDF), optimizing the FEM methods, adding some FDM reaction-diffusion solvers, adding SDE adaptivity, and put together a solve function for DAEs (wrapping ODEInterface and Sundials). That should be done by next month if I keep working like this (a lot of this is already done, but needs more testing before being exposed in documentation).

This is a large summary but I think it's good to document it somewhere.


DifferentialEquations/versions/0.3.0/requires
```diff
@@ -0,0 +1,11 @@
+julia 0.4
+Parameters
+ForwardDiff
+GrowableArrays
+ChunkedArrays
+EllipsisNotation
+IterativeSolvers
+GenericSVD
+Compat
```
tkelman: you should declare minimum versions that implement the features you need for all dependencies

tkelman:
> By tagging this version, I am dropping v0.4 support

That's not the case until your `REQUIRE` file lists a minimum Julia version higher than 0.4.

ChrisRackauckas:
Yes, the REQUIRE file on master will be adding v0.5- shortly.

PR #6131, opened 2016-08-26
Bump Stan.jl to version 1.0.0

goedman:
Upgrade  Stan.jl to Julia v0.5


tkelman:
Have you investigated enabling Travis CI?

goedman:
Hi Tony,

Thanks for merging.

The problem with Stan.jl (and Jags.jl, which is next on my list) is that to do proper testing the CmdStan and Jags binaries need to be installed. Without these installed the Travis testing is not very telling as the more complicated code deals with running the sampling commands in parallel, processing the combined chain results and using Mamba graphics to display the results.

I did follow your suggestions regarding not using Pkg.dir() and the construct cd(...) do ... end.

Will Pkg.dir go away? I find Pkg.dir("Stan", "Examples", "Bernoulli") a bit clearer than ProjDir = joinpath(dirname(@__FILE__), "..", "Examples", "Bernoulli").

On the cd() do .. end construct, it sure looks nicer. To test the individual methods in say Jagscode.jl, typically by stepping through them using run selection  (^E in my setup of TextMate), I can temporarily go back to the cd(...) and cd(old) bracketing.

Thanks for all your help!

tkelman:
Pkg.dir won't go away, but it's not the right way for a package to refer to its own location, since the code may be loaded from elsewhere.

goedman:
Ah, I see, thank you!

PR #6134, opened 2016-08-26
SLEEF.jl and CorpusLoaders.jl

oxinabox:
I fell victim to:
https://github.com/JuliaLang/PkgDev.jl/issues/14
Still I think these are both releasable packages,
though I'ld rather have make two release PRs

 - https://github.com/oxinabox/SLEEF.jl
 - https://github.com/oxinabox/CorpusLoaders.jl


tkelman:
Please separate new package registrations into individual pull requests.

PR #6138, opened 2016-08-26
Register SLEEF [https://github.com/oxinabox/SLEEF.jl.git]: v0.1.0, v0…

oxinabox:
….2.0, v0.2.1


mlubin:
Looks pretty cool. You may want to add a link to https://github.com/JuliaMath/Yeppp.jl in the readme as well.

oxinabox:
@mlubin  good idea -- done in master. I was actual meaning to this when I created the package. But it slipped my mind by the time I got around to writing the readme.

I observe: Whereas SLEEF optimises math operations over scalars using SIMD; Yeppp! optimises math operations over vectors -- which honestly makes a lot more sense given what SIMD stands for.

KristofferC:
But you don't always have a nice fat dense vector of numbers you want to run `exp` on :)

SLEEF/versions/0.2.1/sha1
```diff
@@ -0,0 +1 @@
+30bab63bfd31780f631c00f50e212321df36e2a8
```
tkelman: this commit only exists as the tag, it's not on the master branch

tkelman:
If you enabled Travis, would your tests pass there?

oxinabox:
>If you enabled Travis, would your tests pass there?
I assume so, but I can't say for sure. I've not used Travis myself.
I'll give it a shot now.

Only thing I can think could go wrong is if the travis build/test machine doesn't have SSE2.
I'm going to work on a patch to this this morning that detects your Architecture, and changes the code to suit (Compiling a SLEEF differently, and defining a bunch of types differently)

SLEEF/versions/0.2.1/sha1
oxinabox: Oops, yeah.
Does it cause any issues? Is there a thing I should do to resolve it?
(Or something I should read)

I'm no git guru.

So it is https://github.com/oxinabox/SLEEF.jl/commit/30bab63bfd31780f631c00f50e212321df36e2a8
And now that 
https://github.com/oxinabox/SLEEF.jl/commit/1d454bd7a38fd19c0aa7c0fd11f8db7f3231a9a3
has been pushed which has it as a parent.

Thanks.

oxinabox:
It does pass on my Travis now.
I had to write different build-rules for Apple, 
and also turns out I never had v0.4 compatibility (No `VecElement`), so I tweaked the REQUIRE file.

tkelman:
If you never had 0.4 compatibility, then you should delete the 0.2.1 tag which does claim to support 0.4. Or modify the requires file of that tag in this branch of your metadata fork to prevent it from being installable on 0.4.

oxinabox:
> If you never had 0.4 compatibility, then you should delete the 0.2.1 tag which does claim to support 0.4. Or modify the requires file of that tag in this branch of your metadata fork to prevent it from being installable on 0.4.

Good point.
Thanks, I really appreciate your attention to detail, and feedback.

Is there a way to back out all tags and start again?
I guess it is not really a trivial matter of just deleting my METADATA.jl,
since to truely back out, I need to delete the tags from git.
I guess that though is probably all there is to it.

tkelman:
yeah you could start from a new branch, unmodified from upstream metadata at first. then remove the git tags from the SLEEF.jl repo, if they haven't been published yet and you want to redo them. Pkg doesn't use the git tags, but they're extremely useful in tracking the changes between versions on github

tkelman:
Can you clarify? Did you want to start over or not?

oxinabox:
>Can you clarify? Did you want to start over or not?

Ah, sorry for being unclear; I wanted to know if it were possible.
For future reference.
For now I have the fix of changing the my metadata.
Which I think is fine?

So I think this can be merged.

On the other hand, my benchmarks right now, suggest that I am going to be deprecating it soon anyway. Since when the pure julia SLEEF is merged into Libm.jl it will be faster (Probably).
See: https://github.com/JuliaMath/Libm.jl/pull/11#issuecomment-243635852

So I was thinking of backing out of registering it, since it will be to deprecate a month or so later.
And so maybe I want to avoid registering a package which could confuse people.
I can leave it unregistered for people who want to cross benchmark (like me).

What do you think?

tkelman:
Yeah if you don't think it'll be around for very long, probably don't bother registering it.

ViralBShah:
Yeah, best to keep it as a standalone repo for comparison purposes, but it may not be useful to have it in METADATA for users in general given the progress on libm.jl

oxinabox:
Yeah, I'll revisit registering it, if Libm.jl doesn't turn out so well.
Signs are really promising over there.

PR #6147, opened 2016-08-27
Register DPP.jl v0.0.1

alshedivat:
DPP package provides an efficient implementation of evaluating, fitting, and sampling from determinantal point processes.


tkelman:
Have you read http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package ? It's encouraged to make package names more explicit.

alshedivat:
Ah, I see. Thanks for pointing out. I will rename the package to `DeterminantalPointProcesses`.

What would be the best way to do that in Julia? Just close this PR, do everything locally, and create a new one with the correct name?

PR #6148, opened 2016-08-27
Tag Graphs v0.7.0

ranjanan:


goedman:
There are still a few warnings remaining regarding overwritten method definitions. I think these just indicate overlap between external constructors and the internal constructor. On my system I just removed the external constructors and Pkg.test("Graphs") still passes.

ViralBShah:
I noticed the warnings too, but was happy that the tests are passing.

PR #6150, opened 2016-08-27
Register DeterminantalPointProcesses: v0.0.1

alshedivat:
The package provides an efficient implementation of evaluating, fitting, and sampling from determinantal point processes.

Properly named the package as compared to the earlier PR #6147.


DeterminantalPointProcesses/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+6de0cc5f57d452ee6d028a6dac28e391a1a75301
```
tkelman: this commit doesn't match the git tag at https://github.com/alshedivat/DeterminantalPointProcesses.jl/tree/v0.0.1

alshedivat: fixed

PR #6151, opened 2016-08-27
Tag ImageQuilting v0.3.2

juliohm:


juliohm:
Can someone help figure out what is causing this build failure in Travis?

tkelman:
I think the error message is actually wrong, it's not Combinatorics that's the problem, it's LightGraphs. LightGraphs 0.6.1 has a Julia upper bound of `0.5-` which makes it not installable on 0.5. You don't have any upper bound here, so it's saying this package is actually not installable on 0.5. I think the upper bounds on LightGraphs should actually be removed right now, since they aren't present on all versions it just results in an older version getting used.

juliohm:
@sbromberger @jpfairbanks

sbromberger:
We could do that but we know for a fact that LightGraphs 0.6.1 doesn't work on Julia 0.5. @tkelman - what's the advised course of action here?

tkelman:
If LightGraphs 0.6.1 works better on 0.5 than LightGraphs 0.5.4 does, then it's probably best to remove the upper bounds for now and at least let people install the latest version until you have a version that does work well on 0.5.

sbromberger:
Neither will work due to the removal of `issym` in 0.5.

Also, the main Julia repo (master) is confusing:

```
By default you will be building the latest unstable version of Julia. However, most users should use the most recent stable version of Julia, which is currently the 0.5 series of releases. You can get this version by changing to the Julia directory and running...
```

But is this true? I don't see where 0.5 has been released yet.

sbromberger:
See https://github.com/JuliaGraphs/LightGraphs.jl/issues/420

PR #6160, opened 2016-08-29
Tag TikzPictures v0.3.4

mykelk:
Remove 0.5 deprecation by using Compat.


tkelman:
Note that due to missing lualatex this has been failing on PackageEvaluator: http://pkg.julialang.org/detail/TikzPictures.html

Are there still tests you could do as a fallback when lualatex isn't available? It's a test-only dependency, right?

mykelk:
Yep, it's a test-only dependency. There are some tests we could do (it can output raw tex files), but for the most part TikzPictures is pretty useless without latex installed because it uses latex to generate pictures (e.g., SVG or PDF).

tkelman:
So there are things under src that users would want to do that won't work without a latex installation present? sounds like maybe the package should check if that's the case and install one locally if it can't find an existing install?

mykelk:
I thought about doing that. The problem is that a latex distribution is pretty huge. Perhaps there should be a texlive.jl package that handles this.

tkelman:
Having a specific package for it sounds like it would be a good idea.

mykelk:
I just tagged a new release that skips some tests if lualatex is not found.

PR #6165, opened 2016-08-29
Register PowerModels: v0.1.0, v0.2.0, v0.2.1

ccoffrin:
A Julia/JuMP package for Power Network Optimization.


ccoffrin:
Additional details here, https://github.com/lanl-ansi/PowerModels.jl

tkelman:
Be aware that conditionally using packages like you're doing here https://github.com/lanl-ansi/PowerModels.jl/blob/95db77c55488bdd077acf872046ba1bb6b53bbd0/src/core/solution.jl#L109-L124 does not work well in combination with precompilation. If the set of available packages at precompile time doesn't match the set available at runtime, you'll either get missing functionality or errors. You'd be better off using a string or symbol for the solver lookup key rather than types that come from the other optional packages.

ccoffrin:
Ok, good to know.

I would be happy to move to a string or a symbol for the dict lookup, I just don't know how to accomplish that in the context of JuMP / MathProgBase.  The key challenge is that, similar to JuMP, PowerModels is totally agnostic to the solvers being used.  Given a generic solver object, is there a way to ask for a unique id, such as String or Symbol?

Thoughts @mlubin?

tkelman:
Maybe slightly hacky, but you could try `Symbol(solvertype.name.module)` ? There's a `Base.datatype_module` function for that now on master but I don't think it's in Compat yet.

mlubin:
Off topic for registration, but what's the purpose of this dictionary? Why does the solution status depend on the solver?

ccoffrin:
Ideally it should not, it's a JuMP work around for the moment.  We could discuss propagating the change in JuMP.    

There is a lot of confusion in the Power System community around what solvers  / methods provide local and global optimality.   In this tool, I wanted to make it explicit when Ipopt returns a solution it only provides a "locally optimal" or "locally infeasible" solution, so I generalized the status types to include these, and overload them specifically for Ipopt and Bonmin (in my case).

mlubin:
@ccoffrin, ``:Optimal`` means that the solver decided that it was optimal according to its convergence criteria and tolerances. It's the user's job to know what the solver's convergence criteria and tolerances are. I'd also be careful that you don't change ``:Optimal`` to ``:LocallyOptimal`` if you know the problem is convex.

tkelman:
Definitely OT, but Couenne would give you global optimality (with the caveat that it might take forever on non-convex problems), and your current check doesn't really distinguish between Couenne and Bonmin.

ccoffrin:
@tkelman I am aware that this is an issue on the current implementation, but to me that's more a problem of not having a direct interface to Bonmin or Couenne.  I actually experiment with both in my work.

@mlubin, agreed that is the semantics in JuMP.  The problem I have is that many users of PowerModels will not be aware of their solver's convergence criteria.  If they see Optimal, they will think it is globally optimal, no matter what solver they use.  All global optimal are also locally optimal, so it is still correct to call them locally optimal.  I think it would be a good thing if people had to explicitly say in their work, PowerModels with Ipopt returns "locally optimal", but I know it's also globally optimal because of x,y,z...

Thinking in the mode broad, JuMP / MathProgBase sense, I think there is a clear divide between algorithms that can provide proofs of optimality / infeasiblilty, and those that do not.  Why not distinguish these two classes of algorithms with specific status codes?

mlubin:
Would you say that a MIP solver provides a proof of optimality?

tkelman:
Happy to chat elsewhere on this, a more direct Bonmin interface via the fairly-new C interface (https://github.com/coin-or/Bonmin/blob/master/Bonmin/src/Interfaces/BonStdCInterface.h) is probably worth implementing. Couenne is messier and would need to use Cxx, or stick with what already works of going through .nl files.

ccoffrin:
If a MIP is run to completion, absolutely.  The complete search tree is the proof of optimality.  You can even use the tree as a tool for verifying the correctness of the solver.

Typically MIP solvers had not provided APIs to extract this a proof, but a number of CP solvers do this, because the complete proof is valuable.  For example, if you are building a no-good in SAT or CP, you basically do an analysis on an infeasible subset of the search tree to build a compact and valid infeasibly cut, which you then add to the root node.

ccoffrin:
Going back to the original topic, the I moved to a symbol based lookup in the master.

https://github.com/lanl-ansi/PowerModels.jl/pull/19

mlubin:
@ccoffrin we can discuss more offline or in an issue at MathProgBase

PR #6167, opened 2016-08-29
Change url for CoverageBase

timholy:
Package moved to JuliaCI, see https://github.com/JuliaCI/CoverageBase.jl/issues/17.


tkelman:
hah okay that was way overkill, sorry

timholy:
Better to be prepared, though.

2_llothcat

PR #6168, opened 2016-08-30
Tag Complementarity v0.0.6

chkwon:


tkelman:
if you `importall Ipopt` in `src`, then Ipopt should be in `REQUIRE`, not manually added on CI

chkwon:
I did some cleanup and added commits here. I am messed up with the version numbers. I hope I did other parts correctly.

tkelman:
You shouldn't be adding Ipopt in the CI yml files. List it in `test/REQUIRE` then `Pkg.test` will add it automatically.

chkwon:
I modified. 

Just one question. What does `Complementarity/versions/0.0.9/sha1` point to? I thought it would be the corresponding commit of the tag `v0.0.9`, but it is not.

tkelman:
> but it is not.

What makes you say that? Looks right to me.

chkwon:
Ooops you are right. I though the abbreviation of 6a9bb215477aad5eff80c33c52db1bf1e5f91834 would be "5f91834". I was confused. 

So technically, I just need to change `sha1` and retag the new commit, without increasing the version numbering, right?

tkelman:
> So technically, I just need to change `sha1` and retag the new commit, without increasing the version numbering, right?

What are you trying to accomplish? You shouldn't modify existing tags, but if you want to retract some of these intermediate tags that haven't been merged yet you can just delete the files for those intermediate versions here.

chkwon:
I understand. I'll delete those intermediate versions.

PR #6171, opened 2016-08-30
Tag GraphPlot v0.01

afternone:


GraphPlot/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.3-
```
tkelman: are you still testing this against julia 0.3? we're about to freeze metadata for any packages supporting earlier than 0.4

ararslan:
Looks like you have a Travis YAML file but you don't have CI testing set up for GraphPlots.jl. It's always a good idea to have CI set up for your projects. If you want to ensure your code works for all Julia versions 0.3 and higher, as currently specified in your REQUIRE file, you'll need to change the `julia` block in .travis.yml to
```yaml
julia:
  - 0.3
  - 0.4
  - 0.5
  - nightly
```
If you set it up on Travis as-is, it will only be testing on 0.5 (`release`) and 0.6 (`nightly`).

afternone:
Thanks! I will open a new PR when the problem is fixed.

PR #6174, opened 2016-08-30
Pull request/d978204e

dehann:
Related to AffineTransforms, but this package is a growing collection of Lie Group/Algebra, Quaternion, Euler, AxisAngle representations. Including convert functions between each, and overloading operators for those sets. The package already includes exponential and logarithm maps for Lie SO(3). SE(2) and SE(3) being written as we go.

Tests are still pending, but the code has been used quite a lot in previous projects etc.


dehann:
Also being used by IncrementalInference.jl and upcoming RecursiveFiltering.jl package. Oops, missing REQUIRE

PR #6177, opened 2016-08-30
Pull request/dd6ec9db

dehann:
Sorry, I only get the Travis build info relating to Packages on pull request. A few errors slipped through between my Travis and Julia requirements. It should be good now....

This package is a growing collection of Lie Group/Algebra, Quaternion, Euler, AxisAngle representations. Including convert functions between each, and overloading operators for those sets. The package already includes exponential and logarithm maps for Lie SO(3). SE(2) and SE(3) being written as we go.


dehann:
All the tags seems to have created a tag dependency issue. Sorry. Probably best to just redo my repo and start over from tag 0.0.1. I'll make sure METADATA reflects that too.

PR #6178, opened 2016-08-30
Pull request/af0696b2

dehann:
Starting over at tag v0.0.1

This package is a growing collection of Lie Group/Algebra, Quaternion, Euler, AxisAngle representations. Including convert functions between each, and overloading operators for those sets. The package already includes exponential and logarithm maps for Lie SO(3). SE(2) and SE(3) being written as we go.

Tests are still pending, but the code has been used quite a lot in previous projects etc.


tkelman:
assuming this is obsolete given #6192. it would be a lot less noisy if you just force-push over the existing branch to update a PR instead of opening so many new ones

dehann:
@tkelman yeah sorry for all the noise. Will use force-push in the future, wasn't sure about how you preferred, so I left the history intact.

PR #6184, opened 2016-08-31
Register, Tag ArbFloats v0.6.1 (extended precision)

JeffreySarnoff:
There is a first time for everything.  This is this one.  Thank you for the guidance.

### DIFF
--- a/ArbFloats/url
+++ b/ArbFloats/url
@@ -1,0 +1,1 @@
+https://github.com/JuliaArbTypes/ArbFloats.jl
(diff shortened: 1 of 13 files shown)


JeffreySarnoff:
I am on to v0.6.2, and would rather that be the one put into the ecosystem.  
What is the proper way to do that?  The METADATA I forked for v0.6.1 has become out of sync with this, the one in use.

JeffreySarnoff:
I suppose closing this and giving you the next tag cleanly makes sense.

PR #6186, opened 2016-08-31
Tag Temporal 0.0.7

dysonance:
Getting Temporal package ready for Julia 0.5 releases with updates to account for new & deprecated functionality. The PkgDev `publish` function wasn't working for me as the base `Pkg.publish` did in v0.4, so I had to do this manually in order to get the PR done, *so* if I completely botched something or missed something way off let me know and I can work on fixing ASAP.


dysonance:
So tests failed and I just fixed them and re-committed to my repo. Now, I am realizing my Temporal repo doesn't know about the 0.7 tag yet, so I'm assuming if this is to work right I have to push the tags to my remote repo and then re-open a PR?

PR #6188, opened 2016-08-31
Register package SwitchTimeOpt.jl

bstellato:
I would like to register the SwitchTimeOpt.jl package which allows the user to easily define and solve switching time optimization problems via MathProgBase.jl.


SwitchTimeOpt/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+MathProgBase
+ODE
+Ipopt
+@osx Homebrew
+@windows WinRPM
```
tkelman: looks like you aren't actually using WinRPM or Homebrew directly?

bstellato: Yes, you are definitely right. Sorry for the confusion. 
I will create another pull request with the right requires.

PR #6189, opened 2016-08-31
Tag HTTPClient v0.2.1

ranjanan:
WinRPM depends on this, so I thought I'd fix it.


tkelman:
well, only on platforms where you don't really need to use winrpm

ranjanan:
Yes, I was about to edit my comment and say just that. It still causes a bunch of annoying depwarns   while building WinRPM, so I thought it was worth fixing.

PR #6203, opened 2016-08-31
Register InteractiveFixedEffectModels

matthieugomez:


tkelman:
Since you do `using Compat`, you should depend on it explicitly in your require file instead of assuming it will be present as an indirect dependency.

matthieugomez:
Thanks!

PR #6207, opened 2016-09-01
Tag TimeSeries v0.8.6

milktrader:
Bug fixes


tkelman:
I think you should drop the upper bound. The effect of that is that on 0.5, people will end up installing the most recent version that does not have an upper bound, which is the old 0.5.9.

milktrader:
ok, it was put in place to avoid chaos from Arrayapocalypse but it appears after using this on 0.6 there are just some deprecation warnings

PR #6208, opened 2016-09-01
Register CatViews

ahwillia:
This is a lightweight package that provides a mechanism for working with data/parameters distributed across multiple arrays in a single vector.

The [README](https://github.com/ahwillia/CatViews.jl) contains self-contained documentation and examples. The motivation should be clear to the machine learning and optimization folk.

I've had requests from @ChrisRackauckas @tbreloff to register this, and have also talked a bit with @pkofod about this. Thanks to all for the input and feedback on this!


tkelman:
push the git tag so it shows up under https://github.com/ahwillia/CatViews.jl/releases

ahwillia:
Done.

Was `PkgDev.tag()` and `PkgDev.publish()` supposed to do that for me? Those commands never seem to work perfectly...

ChrisRackauckas:
It's supposed to do it. `PkgDev.publish()` is supposed to push the tags. But when it fails you have to remember to do so.

PR #6209, opened 2016-09-01
Tag BayesNets v1.0.1

tawheeler:
Now compatible with v0.5 (and backwards compatible with v0.4)


BayesNets/versions/1.0.1/requires
```diff
@@ -0,0 +1,10 @@
+julia 0.4
+Compat
```
tkelman: best to be specific about what minimum version you need that implements the features you're using

PR #6210, opened 2016-09-01
Bump DataStreams, DecFP, and WeakRefStrings

quinnj:


quinnj:
@tkelman, I think I've got the WeakRefString tags right here; mind taking a look?

WeakRefStrings/versions/0.2.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5-
```
tkelman: you shouldn't do `using LegacyStrings` even if behind an if statement that's currently false without having it in the require file, since that if statement isn't guaranteed to be false forever

quinnj: Ah, good point. I'll update WeakRefStrings

DecFP/versions/0.1.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BinDeps
+Compat
```
tkelman: missing minimum version

quinnj: I didn't bump the DecFP tag, so this commit is still consistent with the REQUIRE at the time. I can do another tag.

PR #6213, opened 2016-09-01
Register package QWTwPlot: v0.0.1

ig-or:
New 2D plotting tool for Julia


QWTwPlot/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+05b2fa629c6bf622157e512520664725ab8d5c36
```
tkelman: should replace this and the git tag with 88142b0a59361f8152efaa7038cf56bc1cd830db, or make a new 0.0.2 tag

ig-or: I made Pkg.tag(0.0.2) and Pkg.publish()  and created a new  pull request here:  https://github.com/JuliaLang/METADATA.jl/pull/6225

Is it OK? Or I supposed to modify this pull request?  I'm very new to those github-related things, so have to ask a lot of questions.

tkelman: it's less noise if you push additional commits to the same branch to update an existing PR, but opening a new one is okay

ig-or: OK!  Thanks for the info!

PR #6214, opened 2016-09-01
Tag WordNet v0.0.2

jbn:
Adds sensekey support (contribution by @oxinabox)


tkelman:
Since 0.0.1 supported Julia 0.3 but this doesn't, this should use a new minor package version via `Pkg.tag("WordNet", :minor)`

oxinabox:
Prods @jbn

jbn:
Thanks for the prod / patch, @oxinabox!

PR #6218, opened 2016-09-01
Tag MarketData v0.3.7

milktrader:


milktrader:
I fixed the travis error, how do I fix this PR?

tkelman:
Looks like it's saying you should have used a new minor version?

milktrader:
Will the test only fail if the last tag fails? (i.e., a new tag will submit two tagged versions to METADATA if I understand correctly)

If I document this tag version as a mistake and tag a new one with the fix, will METADATA stop when it sees this tag version is failing?

tkelman:
I think this minor version check should only look at the minor version series of the newest tag for any given package.

milktrader:
Ok, lets give it a try ...

The issue here btw is requiring a dependency for a package version that conflicts with its bounds.

PR #6224, opened 2016-09-02
Tag EEG v0.1.1

codles:


EEG/versions/0.1.1/requires
```diff
@@ -0,0 +1,18 @@
+julia 0.4 0.6-
+
+AWS
+Compat
```
tkelman: best to indicate a minimum version that implements the features you're using

PR #6229, opened 2016-09-02
Tag NoveltyColors v0.2.0

randyzwitch:
Update to change file reference for remote directories


tkelman:
https://github.com/randyzwitch/NoveltyColors.jl/commit/aa7fd4c0106c2254e00fa7b292b305112db49991 only fixed one instance of `Pkg.dir`

randyzwitch:
Unreal. Laziness knows no bounds.

Modified hash here to point to commit where this is fixed.

Issue #6223, opened 2016-09-02
METADATA is invalid on Julia 0.5 < rc3 due to Jags and Stan


simonster:
```
julia> PkgDev.publish()
INFO: Validating METADATA
ERROR: packages with unsatisfiable requirements found:
   Jags v1.0.0 – no valid versions exist for package Mamba
   Stan v1.0.0 – no valid versions exist for package Mamba

in check_metadata(::Set{String}) at /home/simon/.julia/v0.5/PkgDev/src/entry.jl:299
in (::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String})(::Base.LibGit2.GitRepo) at /home/simon/.julia/v0.5/PkgDev/src/entry.jl:93
in with(::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String}, ::Base.LibGit2.GitRepo) at ./libgit2/types.jl:638
in publish(::String, ::String) at /home/simon/.julia/v0.5/PkgDev/src/entry.jl:64
in publish() at /home/simon/.julia/v0.5/PkgDev/src/PkgDev.jl:70
```

According to @tkelman the problem is that Jags and Stan require Julia 0.5-, but the version of Mamba they depend on requires 0.5-rc3. cc @goedman

goedman:
Apologies, will update the REQUIRE files. Wish Pkg.publish() would work on my system :-)

simonster:
Fixed by #6231?

tkelman:
Probably. You were running with the problematic version of Julia, so what do you see with validation locally now?

simonster:
It seems so, or at least `PkgDev.Entry.check_metadata()` now passes.

PR #6231, opened 2016-09-02
Bump Stan.jl and Jags.jl to version 1.0.2

goedman:
Set Julia version lower bound to 0.5.0-rc3 in REQUIRE files. Only tests
Julia 0.5 on Travis. Updated READMEs accordingly.


tkelman:
note that fixing https://github.com/JuliaLang/METADATA.jl/issues/6223 would require modifying the requires files for the existing 1.0.0 tags here in metadata too

goedman:
Clearly I'm in the land of 'filling a hole by creating a hole'.

Don't think I can update existing tags in previously released REQUIRE files on JuliaLang/METADATA.jl 

If Stan.jl v"1.0.0"'s REQUIRE is updated, folks on Julia 0.4 would get Stan v"1.0.0", which won't work. Stan 0.3.2 needed an upper bound I guess, but then there would be no Stan.jl for Julia v0.5- < rc3.

With the majority of the REQUIRE files specifying Julia 0.5-, it seems using version tags like 0.5.0-rc3 can be problematic.

Will PkgDev.publish() still complain with the existing Stan.jl v"1.0.0" release, if if it is superseded by v"1.0.2"?

tkelman:
Yes you can edit the existing tags' require files - just modify https://github.com/goedman/METADATA.jl/blob/metadata-v2/Jags/versions/1.0.0/requires
and https://github.com/goedman/METADATA.jl/blob/metadata-v2/Stan/versions/1.0.0/requires

> Will PkgDev.publish() still complain with the existing Stan.jl v"1.0.0" release, if if it is superseded by v"1.0.2"?

Yes I believe it will, since the consistency check goes over all tags.

By the way, it's a good idea to make pull requests from specific branches rather than working from the default branch in your fork.

Jags/versions/0.2.0/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.3
+julia 0.3 0.5-
```
tkelman: Upper bounds aren't actually effective unless they're applied to all past tags. This would just result in downgrading to an older tag.

tkelman:
Just to double-check, did you intend to do 1.0.1 tags and 1.0.2 tags that point to the same sha? It's not really harmful, but worth asking.

goedman:
Well, after tagging 1.01 I noticed I had forgotten to update README. I have never found a way to untag METADATA except going back to time machine to recover all of  .julia/v0.x as it was just before I tagged the package. Can’t do that right now because for other reasons I had to install OS X Sierra which is not yet supported by Home-brew and can’t build Cairo.

Hence the 1.0.2.

I made the updates to both REQUIRE files, should I make a new PR?

tkelman:
This PR is fine (the updates show up here when you push to a branch that already has a PR open), except that your git tags for 1.0.1 don't match the sha of the metadata tags.

goedman:
Thanks a lot Tony! Couldn't have done it without your help.

tkelman:
No problem! I'll find a problem to apply Stan on one of these days so you can return the favor.

PR #5909, opened 2016-08-11
Tag DataFrames v0.8.0 [https://github.com/JuliaStats/DataFrames.jl]

ararslan:
Includes a fix for the issue caught in the StatsBase tests. Tagging a minor version because it also includes an overhaul of model matrices and the introduction of contrasts.

cc @andreasnoack @tkelman

PR #6238, opened 2016-09-02
Tag DataFrames v0.8.1

ararslan:
Accompanies #5909. It fixes some of the stuff that 0.8.0 will break.

cc @tkelman, who IIRC said we should merge these around the same time.

### DIFF
--- a/DataFrames/versions/0.8.1/requires
+++ b/DataFrames/versions/0.8.1/requires
@@ -4,5 +4,5 @@ StatsBase 0.8.3
 GZip
 SortingAlgorithms
 Reexport
-Compat 0.8
+Compat 0.8.4
 FileIO 0.1.2
--- a/DataFrames/versions/0.8.1/requires
+++ b/DataFrames/versions/0.8.1/requires
@@ -1,0 +1,8 @@
+julia 0.4
+DataArrays 0.3.4
+StatsBase 0.8.3
+GZip
+SortingAlgorithms
+Reexport
+Compat 0.8
+FileIO 0.1.2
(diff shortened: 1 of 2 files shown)


ViralBShah:
Would be nice to get this and related PRs merged to get StatsBase working again.

tkelman:
What's the status of MixedModels and FixedEffectModels here?

ViralBShah:
Perhaps those packages can catch up afterwards?

tkelman:
If DataFrames breaks them, we should add upper bounds to those packages.

ararslan:
The MixedModels breakage has been fixed and the FixedEffectsModels breakage is now obvious (i.e. the breaking case is now an explicit error). That should be easy to fix because the breaking case isn't behavior that FixedEffectsModels relied on, it's just a matter of adjusting one of the package's tests.

tkelman:
Is the change in FixedEffectModels ready? If not, should add bounds.

ararslan:
Actually it looks like there are more FixedEffectModels issues that this will cause. I'll submit another PR to add DataFrames 0.8.0 as an upper bound for FEM.

ararslan:
Done in #6244.

DataFrames/versions/0.8.1/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4
+DataArrays 0.3.4
+StatsBase 0.8.3
+GZip
+SortingAlgorithms
+Reexport
+Compat 0.8
```
tkelman: Compat.view wasn't introduced until 0.8.4

ararslan: Good catch, thanks. Fixed here and will fix on DataFrames master.

PR #6241, opened 2016-09-03
Pull request/88d328d9

kkholst:


EventHistory/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4-
```
tkelman: The trailing `-` refers to prereleases of Julia 0.4, which we no longer allow as a minimum Julia version. `0.4` would mean "a release version of Julia 0.4 or newer" which is probably what you want.

PR #6249, opened 2016-09-03
Tag Languages v0.1.0

aviks:
cc: @johnmyleswhite 

I'll tag TextAnalysis once this is done.


Languages/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat 0.8.8
```
tkelman: this is tighter than needed

PR #6253, opened 2016-09-04
Pull request/c3c43b3c

kkholst:
Trying again to fix version numbers in REQUIRE.


EventHistory/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4-
```
tkelman: you'll need to either remove the trailing `-` here or delete the files from https://github.com/kkholst/METADATA.jl/tree/pull-request/c3c43b3c/EventHistory/versions/0.0.1 so only the 0.0.2 tag will get registered

PR #6261, opened 2016-09-05
Register LibCloud

tanmaykm:
LibCloud (https://github.com/JuliaCloud/LibCloud.jl) is a Julia interface for Apache LibCloud.

Apache Libcloud is a Python library which hides differences between different cloud provider APIs and allows you to manage different cloud resources through a unified and easy to use API.


ViralBShah:
Does Julia need to be 0.4.6? Or is 0.4 sufficient?

tanmaykm:
It will likely work, but I have not tested with a version lower than `0.4.6`.
Since `0.4.6` has been out for quite some time now, I felt it is fair to use that as the starting point.

ViralBShah:
All 0.4.x releases are supposed to be fully api compatible. Unless there is a known issue with earlier 0.4 releases, this really should be 0.4.

tkelman:
New package registrations should stay open at least 24 hours.

ViralBShah:
Is there a way to make github not show a merge button for 24 hours?

tkelman:
We could set up a review timer webhook that marks the status as pending.

LibCloud/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4.6
```
tkelman: You can easily test against earlier point releases on Travis. Requiring the latest point release to run the package is not correct, unless there's a specific bug that was fixed in a backport release that you absolutely need to avoid.

tanmaykm: Yes, thanks for the tips.
Would it be alright to tag a new release with an updated `requires`?

tkelman: sure

PR #6262, opened 2016-09-05
Pull request/bfa2ac61

thofma:


Hecke/versions/0.1.3/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Nemo 0.4.1
```
tkelman: does this need an upper bound of 0.5?

tkelman:
If you're including GPL'ed C code in the repository, you should point that out in your top-level license file.

Hecke/versions/0.1.3/requires
thofma: No, I don't think so.

thofma:
I changed the license.

tkelman:
You're not the original author or copyright holder of that file are you? It says Fredrik Johansson is. IANAL, but unless you have the author's permission to relicense the entire file, I don't think the change in https://github.com/thofma/Hecke.jl/commit/e88d103573675b87a072355a250bf6b6d00b4d02 is legal.

thofma:
He actually never was the copyright holder. Someone copy pasted the wrong header in the wrong file. 

Anyway, I talked to the people involved yesterday (in person).

Hecke/versions/0.1.4/sha1
```diff
@@ -1 +1 @@
-47b6e26ee4dd2c319ca19c8e12162ff2616bcb1c
+e88d103573675b87a072355a250bf6b6d00b4d02
```
tkelman: the metadata tag should be consistent with the git tag at https://github.com/thofma/Hecke.jl/releases

thofma: Should be consistent now.

PR #6269, opened 2016-09-06
ArbFloats v0.1.2 (faster medium precision floats)

JeffreySarnoff:
There is much usage info in the README.md.
This is intended to be the substrate for a thin ArbDecimals.jl package which is to provide predefined types like ArbDec50 (50 decimal digits) and internally manage the working precision so results are more reliable than would otherwise obtain.


tkelman:
Can you push the git tags so they show up at https://github.com/JuliaArbTypes/ArbFloats.jl/releases ? They should have been auto-generated by `Pkg.tag`, but if `Pkg.publish` wasn't working then you need to manually do `git push --tags` to push the tags to github.

JeffreySarnoff:
done -- two tags are present 
(there are no earlier tags .. this release modulo two patches is the *revised* ground,
 earlier versions were quite different -- and I really don't want that as part of this)

tkelman:
No problem.

Since you have `using Compat` in your `test/runtests.jl`, you should add `Compat` (preferably with a minimum version requirement that implements the features you need) in either `REQUIRE` or `test/REQUIRE`.

tkelman:
Have you tried enabling Travis CI? Also note the pitfalls of using `Pkg.dir`, as described in https://github.com/JuliaLang/julia/issues/18329 if the package is ever installed elsewhere in `LOAD_PATH` it's better to use `dirname(@__FILE__)` to refer to the location of your own package. For referring to other packages, you might request that Nemo.jl add a constant that contains the absolute path to its own libraries that you'd like to use.

JeffreySarnoff:
Thanks for the good advice.
After I make those changes, can I stick them in the current tag, or would I add a patch -- which is ok ... I don't know how to get my copy of METADATA up-to-date and get it assimilate the new tag info.  I did enable Travis CI for this a few weeks ago -- will redo that.

tkelman:
You can make an 0.1.3 tag later (doesn't even have to be in this same PR if it's simpler to wait) that has these items addressed.

JeffreySarnoff:
Waiting on those changes makes sense -- this version serves the purpose at hand,  which is to exist as the computational strata atop which ArbDecimals.jl (exporting ArbDec25, ArbDec50, .. ArbDec300 (25, 50..300 decimal digits) does little to accomplish more than a little.   I am in the midst of that and my hope is to make that with the 0.0.3 alterations to ArbFloats.jl be the next  released thing.

PR #6270, opened 2016-09-06
Tag MXNet v0.0.9

pluskid:
* Migrated documentation system to `Documenter.jl` (@vchuravy)
* Simplified building by using Julia's OpenBlas (@staticfloat)
* Freezing parameters (@vchuravy)
* Support `DType` for `NDArray` (@vchuravy)

Also ref: dmlc/MXNet.jl#126


MXNet/versions/0.0.9/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4+
```
tkelman: The `+` most likely doesn't mean what you think it does - this would prevent the package from being installed on any version of Julia 0.4, in which case you're better declaring `0.5-` or one of the rc's as a minimum and making this MXNet 0.1.0.

pluskid: @tkelman It should work with Julia 0.4. Thanks! I will double check and fix this.

tkelman:
In https://github.com/dmlc/MXNet.jl/blob/ae1d45b4566114c1f271c07fdc36e5f3bd4824eb/deps/build.jl#L60 you should be checking out a tag or sha, not master. Otherwise all the existing tagged Julia package version that work right now could easily break if something changes on mxnet master.

PR #6278, opened 2016-09-07
Tag GR v0.15.0

jheinen:


tkelman:
note that we've frozen metadata for package versions that support julia 0.3. adjust the require to set 0.4 as minimum?

jheinen:
I changed the requirements.

tkelman:
exporting `gradient` doesn't cause a warning?

jheinen:
It does. I wanted to be MATLAB compatible, which causes this ambiguity.
```
WARNING: both GR and Base export "gradient"; uses of it in module Main must be qualified
```
In my examples, I use the qualified name:
```
using GR
x = linspace(-2, 2, 21)
y = x
X, Y = meshgrid(x,y)
z = X.*exp(-X.^2 - Y.^2)
contour(x, y, z)
u, v = GR.gradient(x, y, z)
quiver(x, y, u, v)
```
Should I remove the export?

tkelman:
If you're defining it on types that belong to the GR package, you could change it to an import and extend with new methods. If you're defining it on base types, remove the export, don't import it from base,  and have it called it the qualified `GR.gradient` form.

PR #6279, opened 2016-09-07
Tag RDatasets v0.2.0

randyzwitch:
Fixes RData deprecation errors, which cascaded into several other packages


ararslan:
Since the 0.1.4 tag has been removed from the repo, you should rebase out that commit here and just leave the one for the 0.2.0 tag.

tkelman:
if you do `using FileIO` then you should explicitly depend on FileIO in REQUIRE.

tlnagy:
`RData` depends on `FileIO`, but if `RDatasets` does an explicit `FileIO` call then it should be in `REQUIRE`

tkelman:
@randyzwitch can you please add the require changes here?

randyzwitch:
Yeah, doing it right now.

Do I have to do rebase? I feel like I’m going to mangle that if I try.

tkelman:
not needed. you can modify the files in place in this branch of your fork. you did a rename so the previous tag won't end up in the registration, and I'll squash on merge when ready.

PR #6291, opened 2016-09-08
Bump DataStreams and cap existing DataStream packages

quinnj:
CC @tkelman: I think I have the upper-bounds correct here, but I'd appreciate a look; essentially, the current tagged versions of CSV/SQLite/Feather should only ever be allowed to use v0.0.12 of DataStreams. New tags for those packages will follow this merge.

### DIFF
--- a/CSV/versions/0.0.1/requires
+++ b/CSV/versions/0.0.1/requires
@@ -2,4 +2,4 @@ julia 0.4
 Compat
 NullableArrays
 Libz
-DataStreams
+DataStreams 0.0.1 0.0.13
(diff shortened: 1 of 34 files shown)


tkelman:
upper bounds will not be effective unless you apply them to all past tags of the packages in question

tkelman:
Are there early tags of Feather or SQlite that didn't depend on DataStreams? The resolver may choose to downgrade those in order to get the latest DataStreams. There are a few different things we could do about that.

quinnj:
Just one other older version of SQLite, all versions that had DataStreams as a dependency should now be capped.

tkelman:
the issue is the old versions that didn't have DataStreams as a dependency

quinnj:
ah, what's the best solution then? not worry about capping?

tkelman:
can add an artificial datastreams dependency on the older versions. though for the versions that support older julia versions than datastreams' minimum, better to upper bound julia

quinnj:
Alright, I put julia upper bounds on all the old SQLite verions.

PR #6293, opened 2016-09-08
Register LineSearches v0.0.1

anriseth:
Moving the line search functionality of Optim to a standalone package JuliaOpt/Optim.jl#277


tkelman:
The copied code should preserve some attribution in the license - ideally you could do something with git filter-branch or similar to preserve the history of the relevant files?

anriseth:
Oh yeah, keeping the history and credit/blame is important - got a bit
carried away in the transition and stopped thinking.
I'll try to understand all this git magic and sort it out.

PR #6295, opened 2016-09-08
Tag MIToS v1.2.0

diegozea:
This solves https://github.com/diegozea/MIToS.jl/issues/32 and the `swap!` problem with `FixedSizeArrays`

Thanks!

### DIFF
--- a/MIToS/versions/1.2.0/requires
+++ b/MIToS/versions/1.2.0/requires
@@ -1,0 +1,14 @@
+julia 0.4
+LightXML
+GZip
+AutoHashEquals
+DataStructures
+Formatting
+IndexedArrays
+PairwiseListMatrices
+ArgParse
+ROCAnalysis
+Clustering
+FixedSizeArrays 0.2.3
+FastaIO 0.2
+RecipesBase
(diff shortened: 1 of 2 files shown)


tkelman:
are you sure you want multiple 20k+ line files checked in under docs?

tkelman:
doesn't need to be fixed right now, but note https://github.com/JuliaLang/julia/issues/12120, you shouldn't usually use `Pkg.dir` to refer to a package's location since it could be installed elsewhere

diegozea:
> are you sure you want multiple 20k+ line files checked in under docs?

Sorry, I didn't understand that.

> doesn't need to be fixed right now, but note JuliaLang/julia#12120, you shouldn't usually use Pkg.dir to refer to a package's location since it could be installed elsewhere

Thanks, I will do it and I will tag a `:patch` tomorrow.

tkelman:
`docs/Information.ipynb` and `docs/compiled/Information.html` are rather large files which may make your git history noticeably larger if you include them in a tagged version.

diegozea:
Thanks for the clarification. Will It help If I move the docs into the `gh_pages` branch?

tkelman:
Good question. I think Pkg currently clones all branches, but it's an outstanding todo item that we need to test whether it works properly with single-branch clones.

PR #6297, opened 2016-09-08
Update LibBSON version

pzion:
I've had to create this pull request manually since I'm getting the following from Julia:

```
julia> Pkg.publish()
INFO: Validating METADATA
    BioMedQuery v0.0.1 – no valid versions exist for package SQLite
    BioMedQuery v0.0.2 – no valid versions exist for package SQLite
    BioMedQuery v0.0.3 – no valid versions exist for package SQLite
    BioMedQuery v0.1.0 – no valid versions exist for package SQLite
    Metamath v0.0.1 – no valid versions exist for package CSV

 in error at error.jl:21
 in check_metadata at pkg/entry.jl:635
 in publish at pkg/entry.jl:372
 in anonymous at pkg/dir.jl:31
 in cd at file.jl:22
 in cd at pkg/dir.jl:31
 in publish at pkg.jl:61

julia> 
```


tkelman:
what's your `versioninfo()` ? this might be similar to https://github.com/JuliaLang/METADATA.jl/issues/6223

tkelman:
can you do `git push --tags` manually so the git tags show up under https://github.com/pzion/LibBSON.jl/releases ? `Pkg.tag` creates them, but you have to push them manually if `Pkg.publish` fails

pzion:
OK tags pushed.  Yes that certainly looks similar.  Here's my versioninfo():

```
Julia Version 0.4.0
Commit 0ff703b* (2015-10-08 06:20 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i7-3520M CPU @ 2.90GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.3

```

tkelman:
cc @quinnj is there a bug in Julia 0.4.0 that was fixed in 0.4.1 that you need to avoid? ref.  https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/SQLite/versions/0.3.3/requires - either we should loosen that to 0.4, or tighten BioMedQuery to 0.4.1 to match recent SQLite tags.

This is tracked at https://github.com/JuliaLang/PkgDev.jl/issues/71, at least for 0.5 and newer.

tkelman:
@quinnj please respond to the above question

tkelman:
@quinnj again, PLEASE RESPOND:

> is there a bug in Julia 0.4.0 that was fixed in 0.4.1 that you need to avoid? ref. https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/SQLite/versions/0.3.3/requires - either we should loosen that to 0.4, or tighten BioMedQuery to 0.4.1 to match recent SQLite tags. Same with https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/CSV/versions/0.0.5/requires and Metamath.

quinnj:
Hey folks, sorry for the slow response; I tend to mute METADATA.jl conversations pretty quickly; unfortunately gmail doesn't let mentions unmute a conversation.

Yes, CSV.jl specifically needs 0.4.1 because of the usage of libjulia's exported `jl_strtod`, which was only officially exported in 0.4.1. Anything that depends on CSV.jl then should also have 0.4.1.

tkelman:
Thanks. Should be easy enough to fix then.

PR #6299, opened 2016-09-09
Tag Mongo v0.2.1

pzion:
As before, Pkg.publish() not working here so pushing manually.


tkelman:
Do you need to be running `'Pkg.add("FactCheck"); Pkg.add("BinDeps"); Pkg.add("LibBSON"); Pkg.add("DataStructures")` on Travis? Shouldn't `REQUIRE` and `test/REQUIRE` have those all covered?

pzion:
Ah, I wasn't aware of that -- that code is quite old.  I'll close the request and re-open it with these removed in a new tag.

tkelman:
Travis-only changes don't really need a new tag, that won't effect the code users get.

pzion:
In that case I'll reopen it.  Tests are still passing on Travis after the Pkg.add commands were removed, thanks!

PR #6302, opened 2016-09-09
Tag NaiveBayes 0.3.0

OmriRoames:


NaiveBayes/versions/0.3.0/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+Distributions 0.6+
```
tkelman: what are you intending to set as the minimum version here? the `+` means a version greater than any other 0.6 release, possibly not what you intend here

OmriRoames: @JoshChristie is there a reason you set this to 0.6+?

r-chris: `Distributions 0.6+` has been in the `REQUIRE` for `NaiveBayes` since version `0.1.0` - this should be fine to stay that way

PR #6303, opened 2016-09-09
ArbFloats v0.1.3+c (plays better with Nemo, more examples)

JeffreySarnoff:
I am using version strings suffixed "+a","+b".. to manage this stage of many small adjustments and layered package interactions without climbing versions too quickly.


tkelman:
It would be preferable to stick to semantic versioning. There's absolutely no harm in having too many version number tags, but deviating in version number formats from the standard means less predictability in writing tools to programmatically deal with metadata.

carlobaldassi:
To add to what @tkelman wrote, [semantic versioning](http://semver.org) prescribes the use of `+...` suffixes for build metadata only.

JeffreySarnoff:
ok .. I will redo the numbering and submit that

PR #6318, opened 2016-09-11
Register JuliaArrays packages

timholy:
This registers the unregistered packages in https://github.com/JuliaArrays:

- `IndirectArrays` implements an `AbstractArray` type in which elements are looked up in an index. Used to implement "colormap" images or "indexed" arrays.
- `TiledIteration` implements a number of constructs useful for cache-efficient programming and exploiting Julia's native multithreading.
- `CatIndices` implements `indices`-preserving concatenation and `push!`/`pop!`/`unshift!`/`shift!` operations.
- `FFTViews` implements a wrapper that endows an array with periodic indices and accesses the zero-frequency bin at 0.


nalimilan:
OT, but it looks like `IndirectArrays` is similar in some aspects to `CategoricalArrays`. I think there are good reasons to keep them separate, but still interesting to compare.

timholy:
Not OT at all, thanks for the pointer. I agree there is considerable similarity esp. with `CategoricalArrays.OrdinalArray`. For the purposes of image processing (which is why I wrote this), however, it's important for `x[3]` to return a "plain value" rather than an `OrdinalValue`. So I agree there's room for both.

tkelman:
(re) enable Travis on FFTViews

tkelman:
FFTViews should have OffsetArrays in `test/REQUIRE`

timholy:
Fixed. Thanks for catching that.

FFTViews/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+7fd168b6f1690df986f7fff180fc82d9769ab8c1
```
tkelman: re-push the git tag? https://github.com/JuliaArrays/FFTViews.jl/releases

timholy: Again, thanks.

PR #6323, opened 2016-09-11
Register QuickHull [https://github.com/tlycken/QuickHull.jl.git]: v0.1.0

tlycken:
QuickHull has a whitepaper, pure-Julia implementation of the [quick hull algorithm](https://en.wikipedia.org/wiki/Quickhull).

Documentation <strike>will be</strike> is up on https://tlycken.github.io/QuickHull.jl/latest <strike>once I manage to get past https://github.com/JuliaDocs/Documenter.jl/issues/257</strike>.


tkelman:
push the git tag so it shows up at https://github.com/tlycken/QuickHull.jl/releases

tlycken:
I did push the tag, but it doesn't seem to have affected the releases shown on the docs page. Not sure why. The latest docs are up though (and nothing has really changed in between).

How does `mkdocs` pick the release states? Do the tags have to be parents of the `gh-pages` branch (in this repo they're not)?

PR #6324, opened 2016-09-11
Tag DocOpt v0.2.0 [https://github.com/docopt/DocOpt.jl]

bicycle1885:
* Fix docstring style
* Drop 0.3 support
* Remove Compat.jl dependency


tkelman:
still using Compat in `test/check_compat.jl` ?

bicycle1885:
It is not included in `test/runtests.jl`, so it never runs in unit tests. It should be included in the tests but one test doesn't pass actually...

PR #6333, opened 2016-09-12
Adding SimpleTraits

mauro3:


timholy:
This tried to register two versions, which are not compatible. Probably the easiest is to `git reset` your METADATA locally, [delete both tags locally and on the remote](https://nathanhoad.net/how-to-delete-a-remote-git-tag), and start from scratch.

SimpleTraits/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: what minimum version implements the features you need?

mauro3:
Ok, let's see whether this work.  I wanted to keep v0.0.1 as that still works with Julia 0.3.  But I deleted that entry in METADATA.

timholy:
Ah, OK then. Your original plan would have worked if, upon tagging the newer one with a higher Julia requirement, you had said `PkgDev.tag("SimpleTraits", :minor)`.

tkelman:
possibly not, since we've frozen metadata for julia 0.3

SimpleTraits/versions/0.1.0/sha1
```diff
@@ -0,0 +1 @@
+f00aac7d2202ad888828b45fc23eab236885a8f9
```
tkelman: push the git tag for this so it shows up under https://github.com/mauro3/SimpleTraits.jl/releases

mauro3: Ok, done.

PR #6340, opened 2016-09-12
Bump Feather

quinnj:


tkelman:
This will likely cause the package to fail on pkgeval unless the tests handle installation of the necessary python packages.

quinnj:
Yikes.....I guess that would kill my travis + docker idea as well then. Would there be a way to selectively run the tests? I from the PkgEval side or the Travis side? I think from Travis I could detect if it was a Travis env then only run the python portion then.

tkelman:
I generally prefer packages to handle their own dependencies at install time rather than add more PkgEval-specific setup code, but if this is a test-only dependency then maybe it's not such a big deal. Most users who didn't have the python packages set up properly would have `Pkg.test("Feather")` fail, if that matters.

Maybe you could have PyCall as a test-only dependency and do the `pyimport_conda` thing?

quinnj:
Ok, this has been updated based on the new testing infra using docker. This should be valid to test anywhere, regardless of whether docker is installed or not (there's a check for it) and has resulted in a much simplified development and testing process.

tkelman:
so this version will fail if someone happens to have docker installed, since it requires sudo in most configurations.

tkelman:
it'll also error if `which` is not installed

quinnj:
Potentially yes; I don't have a good sense for how common that would be or what a work-around would be. At least how docker is typically configured on windows/osx, it doesn't cause any issues.

tkelman:
it will on linux.

tkelman:
if you're intending for things to only run on travis, just do them in the yml.

quinnj:
I'm not. I run the test suite locally as well. It sounds like the best solution for linux is to create a docker user group first and allow it the right permissions. Happy to put this on the package readme if people need/want to run the tests locally on linux.

tkelman:
that's a lot of setup work to get Pkg.test to pass if you happen to have docker on your path. for running locally, make it a separate script. unless it's going to be reliable, it shouldn't be part of Pkg.test.

quinnj:
3 commands is a lot of setup? It seems pretty trivial to me in the (I think rare) case that someone already has docker, on linux, not setup to run docker without sudo. Happy to even print a useful message when we detect this case in Pkg.test. I really feel like the setup is "reliable" as is since we're covering the majority of use-cases.

tkelman:
those commands aren't possible to run as a non root user, if the system has docker globally installed by a sysadmin or someone else. the default docker installation on linux does not do this for you, so it will be common.

quinnj:
Fair enough. I've changed the script to detect if *non*-sudo docker is installed and useable before running the python roundtrips.

quinnj:
I've also tested this personally on osx, linux, with docker installed and not, and on linux with sudo linux and non-sudo.

tkelman:
better, but you should be testing against a fixed tag of the docker inage, otherwise this is the same "latest master" reproducibility issue.

quinnj:
Luckily there's only one tag for the docker image, so no worries for now. :) I'll tag another version in the future though and update the runtests.jl

PR #6347, opened 2016-09-13
Register SCIP

mlubin:
In-memory interface to SCIP (incl. some callbacks and nonlinear expressions) primarily written by @leethargo and @fserra with some prodding from me. Currently linux, from-source only, the SCIP license prevents us from distributing binaries.

csipteam

CC @chriscoey @IainNZ @joehuchette @madeleineudell


tkelman:
Nice.

Uses Compat, should list it explicitly in REQUIRE, preferably with a minimum version that implements the used features.

mlubin:
Removed the dependency on Compat

IainNZ:
Great work!

PR #6350, opened 2016-09-13
Tag GLM v0.5.5 [https://github.com/JuliaStats/GLM.jl]

andreasnoack:


GLM/versions/0.5.5/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+Distributions 0.4.6-
+StatsBase 0.8.3
+StatsFuns 0.3.0
+Reexport
+Compat 0.2.12-
```
tkelman: this is really old, the package really doesn't use anything newer from Compat?

dmbates:
The uses are to define `Void` as a type, `Union` declarations and the `.` syntax for vectorization of a scalar function.  There is no real problem with bumping the version of `Compat`.  Getting rid of the dependence on `Compat` is one of the objectives fixing a 0.4 branch and moving master to julia 0.5

tkelman:
Dot vectorization is the newest of those. I don't see Compat as a very problematic dependency if you mark your version requirement accurately, but up to you.

dmbates:
It turns out that dot vectorization is only used in the `test` directory and there is a `test/REQUIRE` that specifies `Compat 0.9.2`

PR #6351, opened 2016-09-13
Register & tag the first two pieces of the new Images

timholy:
These are components of what will eventually become a rewritten Images package. A couple of other components are basically ready except for needing some of their dependencies being tagged and registered.


tkelman:
FixedPointNumbers should be explicitly listed in ImageFiltering's REQUIRE instead of assuming it will be present as an indirect dependency

timholy:
Good point. I've fixed that problem and incorporated quite a few new commits (mostly for documentation) with these retagged v0.0.1 releases.

PR #6356, opened 2016-09-14
Tag Match v0.3.0 [https://github.com/kmsquire/Match.jl]

kmsquire:
v0.5 updates.

Also updated to work only on v0.5- or greater.  
Match v0.2.2 still supports v0.4 (with better support than v0.5)


tkelman:
push the git tag so it shows up under https://github.com/kmsquire/Match.jl/releases

kmsquire:
Tags pushed.

PR #6358, opened 2016-09-14
Tighten FixedEffectModels bound on DataFrames further

tkelman:
0.7.8 didn't fix the issue with reducing over an empty collection
and FixedEffectModels 0.3.0 is Julia-0.5-only
ref http://pkg.julialang.org/detail/FixedEffectModels.html

#6244 wasn't tight enough, cc @matthieugomez @ararslan and ref https://github.com/JuliaStats/DataFrames.jl/pull/1017#issuecomment-243803777

If one of the more recent DataFrames tags has fixed the issue, then we can instead loosen or remove the upper bounds, let me know.

### DIFF
--- a/FixedEffectModels/versions/0.0.1/requires
+++ b/FixedEffectModels/versions/0.0.1/requires
@@ -3,5 +3,5 @@ Distributions 0.4.6
 Distances 0.1.1
 StatsBase 0.3
 DataArrays
-DataFrames 0.6 0.8.0
+DataFrames 0.6 0.7.7
 GLM
--- a/FixedEffectModels/versions/0.0.2/requires
+++ b/FixedEffectModels/versions/0.0.2/requires
@@ -3,5 +3,5 @@ Distributions 0.4.6
 Distances 0.1.1
 StatsBase 0.3
 DataArrays
-DataFrames 0.6 0.8.0
+DataFrames 0.6 0.7.7
 GLM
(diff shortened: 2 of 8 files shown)


ararslan:
Shound be fixed as of DataFrames 0.8.1. The fix was introduced in https://github.com/JuliaStats/DataFrames.jl/pull/1038. However, it also includes an explicit error for a case that FixedEffectModels may be relying on working, not sure on that.

matthieugomez:
There is a test in fixedeffectmodels checking that in a particular case, an
empty vector is returned . Since Dataframes throws an error in this case
after #1038, the function now throws an error, instead of returning an
empty vector. Therefore only a dataframes version before #1038 can pass
this test.

ararslan:
@matthieugomez If having an empty vector is important for a use case in FixedEffectModels, you could simply wrap it in a `try`/`catch` block and handle the empty case in `catch`.

matthieugomez:
The current version of fixedeffectmodels works around this. I think were
talking about the previously tagged version (right?)

ararslan:
Oh sorry, my mistake.

tkelman:
This doesn't touch FixedEffectModels 0.3.0, but since that is Julia-0.5-only, the package is still failing on Julia 0.4.

ararslan:
Does tightening to DataFrames 0.7.7 fix it on Julia 0.4?

tkelman:
I can't remember if I tested this earlier, but just did on this METADATA branch and it passed with DataFrames 0.7.6.

ararslan:
Nice. Looks good to me, then.

tkelman:
This may downgrade DataFrames further for people who have FixedEffectModels installed, so @matthieugomez if it's at all possible to backport some of the fixes and release an 0.4-compatible 0.2.3 version from a branch that could work with the latest DataFrames versions, that would be useful.

PR #6362, opened 2016-09-14
Register QuantumInfo [https://github.com/BBN-Q/QuantumInfo.jl.git]

marcusps:


tkelman:
Will need a tagged version.

marcusps:
Now tagged.

tkelman:

QuantumInfo/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Cliffords
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

tkelman:
It's discouraged to extend Base methods like normalize, trace, etc on Base types in packages. That means importing your package can change the behavior of all other code. If you want to use the same function name, it's better not to import and change Base, but for users of the package version to qualify with `QuantumInfo.trace` etc.

tkelman:
https://github.com/BBN-Q/QuantumInfo.jl/commit/1be50ce3b19132a41baa8a841a749435b38dae1e looks great. We can merge this now and incorporate that in the next tag if that's easier?

marcusps:
Sure, that sounds good.

tkelman:
Oh, and do enable Travis CI when possible. https://travis-ci.org/BBN-Q/QuantumInfo.jl

PR #6364, opened 2016-09-14
Register QuantumTomography [https://github.com/BBN-Q/QuantumTomography.jl.git] 

marcusps:


tkelman:
Needs a minimum Julia version requirement (only 0.4 or higher is being accepted in METADATA now). It would also be preferred if you could separate the commits to one PR per package, if possible (there's a bug we haven't fixed yet to make Pkg.publish do the right thing automatically)

marcusps:
I separated the packages, but `QuantumInfo` is a dependency, so the the build will fail until  `QuantumInfo` is merged (#6362).

tkelman:
Yep, we'll restart CI here when that happens. Thanks!

tkelman:
You're doing `using SchattenNorms` in here, is that a separate package not yet registered?

marcusps:
Ah, I missed that. Yeah, I'll be registering that right away as well, and
will make sure it is in the dependencies.

blakejohnson:
LGTM

tkelman:
Could you restart Travis in the package to see if it passes now? Same with the other recently-merged one.

blakejohnson:
@tkelman both packages are now passing their tests.

PR #6368, opened 2016-09-15
ArbFloats v.0.1.6

JeffreySarnoff:
all known bugs vanquished


JeffreySarnoff:
updating to v0.1.7 (minor bug appeared, shucks)
I know there is a way to glom onto this pull request -- 
   uneasy about trying it for the first time now that the pkg is in good shape
so I am going to redo the Publish by hand now

PR #6369, opened 2016-09-15
Tag new version of Pages. Generalizes from localhost.

EricForgy:


Pages/versions/0.0.7/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5
```
tkelman: this won't be installable until 0.5 final is out, but that should just be a few days

tkelman:
if you're dropping 0.4 support, should use a new minor package version

EricForgy:
Hi @tkelman 

I'm new to this stuff. My package is updated to work with new v0.5 syntax, but I'm using Compat to continue to support v0.4. What is the best thing to do?

tkelman:
then your minimum julia version is still 0.4. just be sure to continue to test against that version of julia.

tkelman:
you'll also need to push the git tag so it shows up at https://github.com/EricForgy/Pages.jl/releases

EricForgy:
Thanks. I changed REQUIRE back to `julia 0.4` and pushed the new tag. Hope it's alright now.

PR #6375, opened 2016-09-15
Register GraphPlot: v0.0.1

afternone:


tkelman:
re: https://github.com/JuliaGraphs/GraphPlot.jl/issues/3 you have several functions copied from https://github.com/IainNZ/GraphLayout.jl, but please read https://github.com/IainNZ/GraphLayout.jl/blob/master/LICENSE - specifically the part that says "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

You also are `using Compat` without declaring the dependency in `REQUIRE`.

You should enable Travis CI. Your tests likely won't pass because you're `using FactCheck` in `test/runtests.jl` but don't declare the test dependency in `test/REQUIRE`.

afternone:
Thanks for your remarks. Maybe we can close the PR first , and reopen after I solve the problem.

PR #6376, opened 2016-09-15
Tag Elemental v0.1.2 [https://github.com/JuliaParallel/Elemental.jl]

andreasnoack:


Elemental/versions/0.1.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat
```
tkelman: needs a minimum version

PR #6378, opened 2016-09-16
Tag Elly v0.0.9

tanmaykm:


tkelman:
It looks like this new version fails with a connection refused on pkgeval

tanmaykm:
I don't think the PkgEval environment has hadoop installed. Test scripts for older releases were just loading the package. Can hadoop be included in PkgEval?

tkelman:
Ah. I suppose it could, though that might get pretty heavyweight, would need to test. You have some setup incantations that are working on travis?

tanmaykm:
Yes: https://github.com/JuliaParallel/Elly.jl/blob/master/test/test_setup/setup_hdfs.sh
It assumes the JVM to be already installed.

tkelman:
There is a JVM installed during the PkgEval VM setup: https://github.com/JuliaCI/PackageEvaluator.jl/blob/ef79bd7e90cb6c43ec5430cbad610a69bac51403/scripts/setup.sh#L87-L93

tanmaykm:
Oracle JDK should also be fine. Yes, it may turn out a bit heavy, though not much. And it will be used by just Elly.jl and Hive.jl. I can prepare a PR for PkgEval for this. Or we can just exclude these from PkgEval if that's more appropriate for now.

tkelman:
Let's try it and see how it goes. Testing more packages seems like a good thing to me.

tanmaykm:
Ok.

tanmaykm:
@tkelman the PR is here: https://github.com/JuliaCI/PackageEvaluator.jl/pull/145

PR #6379, opened 2016-09-16
Tag TikzPictures v0.3.5

mykelk:
Fixed issue with error messages for texlive 2016 and address issue with testing when lualatex is not present.


tkelman:
note that `success` actually throws an ENOENT if the program isn't found

PR #6355, opened 2016-09-14
Tag NFFT v0.1.2

robertdj:

PR #6380, opened 2016-09-16
add an upper bound on GeneralizedSampling's dependency on NFFT

tkelman:
ref https://github.com/JuliaLang/METADATA.jl/pull/6355

cc @robertdj - if you later tag a version of GeneralizedSampling that works with the latest NFFT, then this won't be noticeable.


robertdj:
I'm working on updating GeneralizedSampling to fit the new types from NFFT. It should be fixed soon.

tkelman:
I'll merge this for now in case your new fixed version isn't ready by the next PackageEvaluator run.

robertdj:
Thanks. I wont have an update until next week.

PR #6381, opened 2016-09-16
Misc upades

mauro3:
See https://github.com/mauro3/Parameters.jl/blob/master/NEWS.md


mauro3:
There was a performance regression.  I'll re-open once fixed.

PR #6382, opened 2016-09-16
Tag UnicodePlots v0.2.0 [https://github.com/Evizero/UnicodePlots.jl]

Evizero:
Update to 0.5 strings and hist replacement


UnicodePlots/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
```
tkelman: this won't be installable until 0.5.0 final is out, but that should be soon

tkelman: you should also update https://github.com/Evizero/UnicodePlots.jl/blob/master/.travis.yml if you're dropping 0.4 support

Evizero:
Thanks. Will fix

PR #6386, opened 2016-09-16
Tag AutoGrad v0.0.3

denizyuret:
Cleaned up the Pkg.dir.


AutoGrad/versions/0.0.3/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+# Compat 0.7.9
```
KristofferC: This is commented out?

tkelman: only used in an example https://github.com/denizyuret/AutoGrad.jl/search?utf8=%E2%9C%93&q=Compat

denizyuret:
Yes, I am not using Compat in this version, will add it back when needed.

PR #6387, opened 2016-09-16
Register Knet: v0.7.3, v0.8.0

denizyuret:
Second try: I removed the tags for the older versions, just left the ones actively used.


tkelman:
Be aware that the way in which you're doing optional dependencies here https://github.com/denizyuret/Knet.jl/blob/v0.8.0/src/cuda44.jl and here https://github.com/denizyuret/Knet.jl/blob/v0.8.0/src/karray.jl#L452-L457 interacts poorly with precompilation. If the set of packages available at precompile time does not match the set of packages available at load time, you will get missing functionality or errors.

denizyuret:
Thanks for noticing this.  I've already had users report this problem.  My
temporary workaround is to put a precompilation command in the build.jl
script so a Pkg.build("Knet") will fix the problem.  Do you have a better
suggestion I can use before Julia has better support for optional
dependencies?  I really don't want to put nonessential packages in REQUIRE.

tkelman:
You can move the optional features that have additional dependencies to a separate package that can then depend unconditionally on both. This is the only method that works reliably right now.

denizyuret:
Will remove the optional dependencies and resubmit...

PR #6390, opened 2016-09-17
Tag CRC v1.1.2

andrewcooke:
Merging a pull request that changes a test dependency.


tkelman:
METADATA is frozen for packages that support Julia 0.3. Best to drop support for Julia 0.3 and re-tag as 1.2.0 of the package.

tkelman:
Particularly because your Travis isn't testing against 0.3, and Libz (https://github.com/andrewcooke/CRC.jl/pull/8) doesn't support 0.3 - https://github.com/BioJulia/Libz.jl/blob/v0.0.1/REQUIRE

tkelman:
redundant with #6578

PR #6398, opened 2016-09-18
Tag Conda v0.2.5

tkelman:
Heads-up @Luthaf @stevengj - this backports https://github.com/JuliaPy/Conda.jl/commit/4e25afd0c1e9cc8979d6f398169c907386332e06 to a new `release-0.2` branch in that package that still supports Julia 0.3, since otherwise the package is broken by the latest conda release that asserts because the `condarc-julia` file doesn't end in `condarc` or `.yml`.


stevengj:
Thanks!

PR #6401, opened 2016-09-19
Tag OpenGene v0.1.10

sfchen:
Small fix and work around for SHA.sha1() incompatibility

https://github.com/JuliaLang/METADATA.jl/pull/5525

@tkelman


OpenGene/versions/0.1.10/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+GZip
+DataStructures
+DataFrames
+SHA
+Compat
```
tkelman: should really declare a minimum version dependency that implements the features you need

PR #6407, opened 2016-09-19
Tag BlossomV v0.1.0

mlewe:
* added some convenience functions
* lowered requirements for C++ compiler version

fixes mlewe/BlossomV.jl#7


tkelman:
BaseTestNext is a test-only dependency so could probably be moved to `test/REQUIRE`, right?

mlewe:
You're right. Thanks for the hint, I'll change it in the next release.

PR #6415, opened 2016-09-20
Tag LightGraphs v0.6.4

sbromberger:
Ok, the final final :)


tkelman:
Travis-only changes don't really need a new tag, that's not user-visible at all.

sbromberger:
Ah, ok. I'll close this out and remove it.

PR #6419, opened 2016-09-20
Tag NetCDF v0.4.0 [https://github.com/JuliaGeo/NetCDF.jl]

meggart:


NetCDF/versions/0.4.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+BinDeps
+Conda
+Compat
```
tkelman: best to include a minimum version requirement that implements the features you need

tkelman:
push the git tag so it shows up under https://github.com/JuliaGeo/NetCDF.jl/releases

PR #6424, opened 2016-09-20
Tag Juno.jl 0.2.0 and Atom.jl 0.5.2

pfitzseb:
Some bugfixes.


Juno/versions/0.2.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5-
```
tkelman: Should Juno 0.1.0 be 0.5-only? It fails tests on 0.4.

pfitzseb: I think so, yes (cc @MikeInnes though). There should be no way Juno could get automatically installed on a 0.4 setup, so I'm not sure if it's worth changing.

tkelman:
does it serve any purpose to allow it to be installed on 0.4 at all?

tkelman:
does it serve any purpose to allow it to be installed on 0.4 at all?

pfitzseb:
Don't think so -- I'd say it was wrong to have the 0.4 requirement in 0.1.0.

tkelman:
which isn't this 0.1.0, but could be an 0.1.1.

PR #6425, opened 2016-09-20
Fix Julia 0.5 deprecation warnings in LibBSON.jl

pzion:
Not sure why I'm getting authentication errors using the PkgDev.publish() interface, so doing a manual pull request.


LibBSON/versions/0.2.5/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+BinDeps
+Compat
```
tkelman: should really declare a minimum version requirement that implements the features you need (like `unsafe_string`, `Compat.String`, etc)

tkelman:
Note that `release` on Travis CI will soon be changed to mean 0.5, so to continue testing against Julia 0.4 if you're still supporting it you should add an `0.4` line in your yml file. You also probably don't need to do `Pkg.add("FactCheck"); Pkg.add("BinDeps"); Pkg.checkout("Compat")` on Travis, if these packages are in `REQUIRE` or `test/REQUIRE`

pzion:
OK thanks, it's fairly unclear to me how exactly to support both Julia 0.4 and 0.5.   Are there any guides or threads that suggest best practices here?  I wanted to make things work cleanly under Julia 0.5 as it was recently released.

pzion:
And apologies if I have missed something obvious in the docs!

tkelman:
If you mean on Travis, see https://docs.travis-ci.com/user/languages/julia#Choosing-Julia-versions-to-test-against, just do
```
julia:
  - 0.4
  - 0.5
  - nightly
```

pzion:
No, sorry, I was referring to the previous comment you made about unsafe_string, Compat.String, and so on.  I had only tested to make sure that the module passes all tests on Julia 0.4 and 0.5 on my local machine, which I guess happened to have a "new enough" version of Julia 0.4 and/or Compat and/or wherever unsafe_string comes from exactly.. I'm not sure sure what the best practice is for giving the right version numbers here, should I should specify the versions I have locally as the minimum versions for Julia and the module dependencies?  Or is that overkill?

tkelman:
There's a note on this here - https://github.com/JuliaLang/Compat.jl#tagging-the-correct-minimum-version-of-compat

I'd say look at what functionality you're needing to use, and check what tags they were first released in. Set your minimum version as the highest of those tags.

pzion:
Great, thank you!

pzion:
OK I've made the suggested changes.  It seems that in particular unsafe_string requires Compat 0.9- so we need a pretty recent version anyway.  I've verified that the minimum versions specified actually work by pinning the package versions.

tkelman:
The trailing dash means a prerelease, are you sure that's what you intended?

pzion:
OK, I've removed the trailing dash.  Hopefully that's it!

PR #6352, opened 2016-09-13
Tag CategoricalArrays v0.0.5 [https://github.com/nalimilan/CategoricalArrays.jl]

nalimilan:
@quinnj Do you think this will break your packages? Do we need to set upper version bounds?

PR #6426, opened 2016-09-20
Bump DataStreams

quinnj:
Note this depends on https://github.com/JuliaLang/METADATA.jl/pull/6352 being merged first. Tested locally on OSX for julia 0.5-rc4/0.4.6.

I also have tags for CSV, SQLite, and Feather ready to push as soon as this is merged to prevent weird transition states.


quinnj:
Alrighty, proper CategoricalArrays upper bounds in place and travis seems happy.

tkelman:
Since not all past versions of these depend on CategoricalArrays, this won't do what you expect. It will likely downgrade to a past version that doesn't depend on CategoricalArrays to try to upgrade that package.

quinnj:
I don't think I understand. What would be the solution here then? it seems to work fine for me in a few local testing runs I've done (0.4, 0.5, etc.)

tkelman:
upper bounds are only effective when applied to all past tags. otherwise the resolver can downgrade your package to something very old instead of holding back your dependency

quinnj:
So I should add fake dependencies on CatArrays to all past versions of DataStreams/Feather? That's not what I did last time (with CSV/SQLite and upper bounds on DataStreams), and that seemed to work out fine.

I have all the other tags ready to go, so I'd rather just make the bump and then re-tag the rest and then everyone can safely upgrade to the current releases that all support both 0.4/0.5.

tkelman:
Fake dependencies would get the resolver to do the right thing. Which was the last time, did you merge that before I looked at it or did I check it first?

Simple tests won't show potential strange resolution solutions.

tkelman:
https://github.com/JuliaLang/METADATA.jl/pull/6291#issuecomment-245765717 had a different solution because the julia minimums were different. here they're all 0.4, right?

quinnj:
You checked and merged. https://github.com/JuliaLang/METADATA.jl/pull/6291

quinnj:
Why does that make a difference?

tkelman:
julia upper bounds remove those versions as feasible solutions the resolver can return. trying to add fake dependencies when the new one doesn't support the same set of julia versions would make that version infeasible on the old julia version it formerly supported.

quinnj:
Ok, so I'll add fake dependencies to all past versions.

quinnj:
Ok, should be good to go here.

PR #6429, opened 2016-09-21
Register TaylorIntegration: v0.0.1

PerezHz:
First release of [TaylorIntegration.jl](https://github.com/PerezHz/TaylorIntegration.jl), a package for ODE integration in Julia, using Taylor's method


tkelman:
When you branch depending on VERSION it's best to be specific, in case this package triggers a bug in base Julia that would require bisecting.

PerezHz:
Thanks for merging, @tkelman ! Do you have any ideas on how we could fix this?

tkelman:
You could either do something involving reshape to avoid making the version-dependent code necessary, or check if there are similar conditions in other packages (Compat has a lot, though maybe not this specific indexing one). Or if you know which NEWS entry the behavior change corresponds to and that has a PR link to where it was implemented, you can check out the merge commit for that PR in a clone of Julia and run the `contrib/commit-name.sh` script.

tkelman:
Looks like this fails on the tagged version of TaylorSeries http://pkg.julialang.org/logs/TaylorIntegration_0.5.log - I should have noticed that in your .travis.yml, should try to get TaylorSeries tagged, then declare a minimum version in your REQUIRE file and remove the checkout from .travis.yml.

PerezHz:
Thanks for noticing this @tkelman , this was meant as a temporary solution while TaylorSeries v0.2.1 is released, since TaylorIntegration currently requires TaylorSeries' latest master. I hope this is ready in a couple of days. As soon as v0.2.1 is ready, I will fix the REQUIRE file and remove the checkout in .travis.yml. Or do you know if there is a way to select either the latest master or a particular commit for a given package in the REQUIRE file, even if it is an untagged commit? Cc: @lbenet

tkelman:
REQUIRE only supports tagged releases. A revision to the package manager for future Julia versions should support using sha's as dependencies, but that will only be acceptable in personal/organization package registries, not the official default one.

lbenet:
I'll tag a new version of TaylorSeries today. Thanks for noticing this!

lbenet:
I've tagged a new version in TaylorSeries. I guess this should solve the problems. Sorry, I 'm too slow in tagging.

PerezHz:
Great! Thanks, @lbenet ! I'll fix the issue at hand as soon as #6483 is merged

PerezHz:
@tkelman TaylorSeries v0.3.0 has already been released, and in TaylorIntegration's latest master I have fixed accordingly the issues currently present at [http://pkg.julialang.org/logs/TaylorIntegration_0.5.log](http://pkg.julialang.org/logs/TaylorIntegration_0.5.log) that you noticed. In this case, would it be best to tag a new version (v0.0.2), or should re-tag v0.0.1?

tkelman:
you can modify the require file of the existing tag in metadata if that's the only change

tkelman:
if there are code changes, best to make a new tag though

PerezHz:
Ok, well I did make a few changes, so I'll make a new tag, thanks @tkelman !

PR #6430, opened 2016-09-21
Tag GSL v0.3.5 [https://github.com/jiahao/GSL.jl]

ranjanan:


tkelman:
looks like tests are failing on mac?

ranjanan:
Ah, you're right. It seems that the function `sf_legendre_Plm_array` is deprecated from GSL 2.0 onwards and possibly removed GSL 2.4 which is what is being downloaded on Mac. It would pass on Windows and Linux because they're probably downloading GSL 1.16 which the package has been tested on.

ranjanan:
I have fixed this in the original package (https://github.com/jiahao/GSL.jl/pull/58) and have update this PR

tkelman:
x-ref https://github.com/jiahao/GSL.jl/issues/53 for tracking the issue with gsl 2

PR #6431, opened 2016-09-21
Tag DifferentialEquations v0.4.0

ChrisRackauckas:
Added dense output via call-overloading, so that way the interpolated solutions are given by `sol(t)` using the same solution object as the timeseries of the output. The new version is also compatible with unit-checked arithmetic via SIUnits.jl (and should be automatically compatible with Unitful.jl, haven't checked it yet). Added a benchmarking suite, and it shows really good results for native DifferentialEquations algorithms over Sundials, ODE, and ODEInterface on nonstiff equations. Extensions to handle explicit parameters for additional features was handled by the creation of ParameterizedFunctions.jl (still experimental). 

Included this time are minimum versions for each dependency.

### DIFF
--- a/DifferentialEquations/versions/0.3.1/requires
+++ b/DifferentialEquations/versions/0.3.1/requires
@@ -1,0 +1,10 @@
+julia 0.5
+Parameters 0.3.1
+ForwardDiff 0.2.4
+ChunkedArrays 0.0.2
+IterativeSolvers 0.2.2
+GenericSVD 0.0.2
+Compat 0.8.8
+Plots 0.9.2
+InplaceOps 0.0.5
+SIUnits 0.0.6
--- a/DifferentialEquations/versions/0.3.1/sha1
+++ b/DifferentialEquations/versions/0.3.1/sha1
@@ -1,0 +1,1 @@
+d4f35822bbc9882fbfc6951330acfbd26d99a7ed


DifferentialEquations/versions/0.4.0/requires
```diff
@@ -0,0 +1,10 @@
+julia 0.5-
```
tkelman: can wait a while, but when you're ready to drop support for prereleases, take off the trailing dash

tkelman:
you still mention GrowableArrays in `src/precompile.jl` but it's not in REQUIRE any more?

tkelman:
similarly some of the tests use EllipsisNotation, so that should be explicitly in either REQUIRE or test/REQUIRE

ChrisRackauckas:
Those two packages longer used for this (at least for now). The precompilation file is old so it's not being used right now (it's commented out). I plan on updating it after this release, but I'd like to give it a good amount of time on master before I'd put a new one into release. (I'll remove the tests which use EllipsisNotation, I wonder why that wasn't causing test failures?), and fix the tests (had release tests disabled since I dropped v0.4 compatibility, forgot to re-add release for tests yesterday!). I don't want to require v0.5 instead of v0.5- yet because some installations, like the Julia COPR, still have v0.5-rc4 on nightly. I'll post when these changes are made.

tkelman:
One of your other dependencies must indirectly be pulling in EllipsisNotation still. Also if you're going to have `TEST_CONDITIONAL_DEPS` set to true, then `Sundials` should stay in `test/runtests.jl` - people are going to want to run tests locally, and best if `Pkg.test` could work by default.

ChrisRackauckas:
Thanks. I'll work that out.

PR #6436, opened 2016-09-21
Register displaz

c42f:
Julia bindings for a hackable 3D plotting tool for large point clouds


Displaz/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

tkelman:
Turn Travis on, add some tests? Is the underlying C program all that hard to build for testing?

c42f:
Good questions.  My thinking was that as version 0.0.1 (with likely a lot of breaking API change still to come) and people already using the package, it might be better to just make a tagged version than to keep people on their toes with master changing under them.

In the short term I can do some tests of the serialization code and mock something out fairly easily to test command line parameter handling.

Testing with the actual C++ companion program in a virtualized headless environment will be "interesting" as it normally depends on a working X windows and working OpenGL 3.3 implementation to actually do anything useful.  Steps for an end-to-end integration test could be something like:

1. Ensure displaz-gui can even start in a headless environment (looks possible via `xvfb` https://docs.travis-ci.com/user/gui-and-headless-browsers/ ?)
2. Ensure I can get an OpenGL 3.3 context on travis (dubious? https://github.com/travis-ci/travis-ci/issues/5645 )
3. Implement a proper scriptable screenshot command for the C++ program (semi-working one here https://github.com/c42f/displaz/pull/98, needs more work.)
4. Send various data series to displaz-gui via the Displaz.jl interface, point the camera somewhere.
5. Take a screenshot
6. Compare screenshot against reference with an image diff algorithm of some kind.

Certainly seems useful and doable, just a lot of work and getting the time to do it all may be months away.  The story becomes more messy on windows where I've got some existing qt packaging woes and would need to build a working software-based OpenGL (mesa softpipe or llvmpipe may be promising, https://mesamatrix.net/ though building it could be a saga).

tkelman:
Sounds a bit complicated then. x-ref https://github.com/tbreloff/VisualRegressionTests.jl which may have some useful bits.

c42f:
Yes, testing this fully is excessively complex.  However visual testing should be mostly out of scope for the bindings here, since the C++ program itself is scripting language agnostic - visual tests largely belong there rather than with the bindings.

Having thought about it a bit more, I think tests for the bindings here should simply verify that the data to be plotted is serialized as expected and passed with the correct flags to the external program.

c42f:
I'll attempt to get some simple tests in place for the next version, as per the last comment.

PR #6442, opened 2016-09-21
Bump CSV

quinnj:


tkelman:
Failing CI on 0.4?

CSV/versions/0.0.15/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4.1
+DataStreams 0.0.14
```
tkelman: It's best if this matches https://github.com/JuliaData/CSV.jl/blob/master/REQUIRE - why aren't you at least using `PkgDev.tag` ? Even if publish doesn't work, tag should.

quinnj:
Yeah, 0.4 failure is CategoricalArrays precompilation.

tkelman:
Solution is what? Disable precompilation for the whole stack on 0.4, or make whatever versions are triggering the problem 0.5-only?

quinnj:
@nalimilan, do you have any thoughts on the precompilation woes? I was originally thinking of taking the same approach as last time and try to see if we can just temporarily disable something in CAs.jl to avoid the segfault, but maybe it's time to just move on to 0.5 with DataStreams.

quinnj:
Closing for now. My plan right now is help https://github.com/JuliaStats/DataFrames.jl/pull/1008 get merged and ironed out in the next week or two. At that point, I'll actually move the NullableArrays/CategoricalArrays code over to DataFrames (it's the DataFrames implementation code for the DataStreams interface anyway). That will remove all the dependencies from DataStreams (as it should be), and allow the rest of the packages to breathe a sigh of relief. Until then, master on any DataStreams packages will be in working condition.

tkelman:
We need to upper bound and make that latest DataStreams tag not installable since the latest tags of CSV and SQLite are now broken everywhere

quinnj:
Can we just revert the DataStreams tag?

tkelman:
might be better to re-tag 0.0.13 as 0.0.15

nalimilan:
We could easily disable precompilation for CategoricalArrays on 0.4 for now. Slow packages are better than nothing.

tkelman:
Every package that depends on it will have to do the same. Having the entire DataFrames stack not precompilable isn't acceptable IMO. Would rather leave things alone as they are before this refactoring on 0.4, and make things 0.5-only if that fixes the issue.

tkelman:
(not exactly as they are right now, there's still work to do to get things into a working state by either making the 0.0.14 tag not installable, or obsoleting it with a higher-versioned tag (doesn't necessarily have to be a newer commit though) so it will not get used.

nalimilan:
Yeah, I was only talking for CategoricalArrays, for which working slowly on 0.4 is better than nothing. That doesn't mean we should use for DataFrames in 0.4.

tkelman:
Should probably make it `__precompile__(VERSION >= v"0.5-pre")` or similar then, maybe bisect to be specific if you have time to do so.

PR #6444, opened 2016-09-21
Tag TikzGraphs v0.1.2

mykelk:
Address issue with repeated labels.


tkelman:
METADATA is frozen for package versions supporting Julia 0.3 now. If you change the require file to set the minimum julia version to 0.4, we can tag this with a new minor version number of the package.

mykelk:
Sorry, I missed that up. I should have specified minor. I'll fix that.

PR #6446, opened 2016-09-22
Move TikzGraphs require to Julia 0.4 or later

mykelk:
Earlier PR did not increment the minor version. This one should have it right now.


tkelman:
probably best to remove the files related to the 0.1.2 and 0.1.3 tags here

TikzGraphs/versions/0.2.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+TikzPictures
+LightGraphs
+Compat
```
tkelman: also a good idea to declare a minimum version dependency that implements the features you need to use

mykelk: I don't think there is a specific version dependency (i.e., this update doesn't depend on any new features of these packages). Would it be bad practice to just list the most recent version numbers for TikzPictures, LightGraphs, and Compat?

mykelk: Getting rid of the v0.1.2 should resolve the Travis issue.

tkelman: using the most recent version would be too tight, but maybe better that way than too loose? whatever you've used Compat for to fix deprecations has probably not been implemented in that package forever.

PR #6452, opened 2016-09-22
Register / tag GraphPlot v0.1.0

sbromberger:
Adds graphplot v0.1.0 (manual push).


tkelman:
push the git tag so it shows up at https://github.com/JuliaGraphs/GraphPlot.jl/releases

tkelman:
Why are LightGraphs and Graphs being imported in a try-catch?

tkelman:
gadfly.js is copied straight from Gadfly without license attribution?

tkelman:
You should use `dirname(@__FILE__)`, or `@__DIR__` from Compat 0.9.1, instead of `Pkg.dir`, otherwise the package won't work if installed elsewhere - ref https://github.com/JuliaLang/julia/issues/12120

sbromberger:
@tkelman the code isn't mine, but we moved it into JuliaGraphs since loads of people are using it and it needs to be registered. I'm not aware of the licensing issues - @afternone will have to weigh in here.

The reason we're importing LightGraphs and Graphs the way we are is because there's STILL no good way to do conditional dependencies.

I'll make the change to `Pkg.dir` once we get this thing registered, or when I have to make the next set of commits.

Bottom line: it seems there are multiple issues here that are preventing you from registering this package. Could I ask you for a checklist instead of what we're doing right now, which is more of a "bring me a rock" exercise? Thanks.

tkelman:
The dependencies aren't conditional, they're in REQUIRE. The Pkg.dir change can wait, but the license attribution needs to actually adhere to the licenses of the various projects this has copied code from. A one liner comment of "copied from ..." is better than violating copyright with no attribution at all, but essentially the only condition of the MIT license is that you preserve the copyright and license in full when you copy code. Claiming otherwise that this code is copyrighted by someone who did not write it is not something we can allow.

sbromberger:
Understood. Not to sound too defensive, but this isn't my code. :) I'm just trying to solve a dependencies problem within JuliaGraphs. For what it's worth, this IS in the README:

>Graph layout and visualization algorithms based on Compose.jl and inspired by GraphLayout.jl.

> The spring_layout and stressmajorize_layout function are copy from IainNZ's GraphLayout.jl.

> Other layout algorithms are wrapped from NetworkX.

> gadfly.js is copied from Gadfly.jl

... but I realize this is not sufficient. I'll go through it and see what we need to do license-wise.

sbromberger:
Before I try again, comments on https://github.com/JuliaGraphs/GraphPlot.jl/blob/master/LICENSE.md ?

tkelman:
Yeah I think that's fine.

sbromberger:
Thanks. I made the other changes (dependencies) as well. I'll wait on the `Pkg.dir()` until I have some time to test it. Closing out and will resubmit.

I appreciate your help on this.

PR #6454, opened 2016-09-22
Register GeneralizedSchurAlgorithm

neveritt:
URL: https://github.com/KTH-AC/GeneralizedSchurAlgorithm.jl
Version: `v0.0.1`


ararslan:
Looks like you have Travis and coverage set up but no actual tests. It would be good to set up tests for your repo to ensure everything is working as expected before it's registered.

dpsanders:
And some information about what the package actually does. Is this related to https://github.com/JuliaMath/IterativeSolvers.jl  for example?

Also, the most widely used license in the Julia community is the MIT license. Is there any particular reason why you chose GPL? This is likely to make the package less widely used by the community,(Anybody who wants to use your code will be forced to change their own license to GPL, due to the viral nature of GPL; instead of doing so, they will likely prefer not to use your package.)

aytekinar:
@tkelman and @dpsanders , I know this might not be the correct venue for this type of conversation, but since in the other repo this thread is mentioned, I wanted to ask a couple of questions.

For us, there is no serious reason not to use MIT or MPL --- it was more of a taste for us to choose GPL. However, I could not understand the "*(Anybody who wants to use your code will be forced to change their own license to GPL, due to the viral nature of GPL; instead of doing so, they will likely prefer not to use your package.)*" part.

If you are to require the package in `Julia`, for instance, will the license prevent you from adding `ControlCore` or `GeneralizedSchurAlgorithm` whatsoever, unless your package is GPL-compatible? If so, I believe the `PkgDev` should not provide any licenses which are viral in this sense... LGPL, MIT or MPL should be the ones to go for, then?

Sorry for the long message... I would appreciate if you gave some more intuition behind it, tho.

tkelman:
If you want weak copyleft on your own code, LGPL or MPL are preferable - GPL's strong copyleft means there's legal ambiguity (IANAL) about what makes for a derived work when using a library (see the GPL FAQ entries at https://www.gnu.org/licenses/gpl-faq.html#NFUseGPLPlugins and the following #LinkingWithGPL entry) about when other code also needs to be released under a GPL-compatible license, preventing it from being used within a proprietary product for example. There's some language in the FAQ about "dynamically link[ing] plug-ins, and mak[ing] function calls to each other and shar[ing] data structures" that may be relevant to how Julia packages work. Many organizations are risk-averse and would rather avoid using any code in which there's any legal ambiguity.

tkelman:
> Anybody who wants to use your code will be forced to change their own license to GPL

Not exactly, but derivative works would have to be distributed under GPL.

aytekinar:
@tkelman we thought any derivative work forced to be distributed under GPL would be a good thing, at start, but then I can see your concerns regarding the vague/ambiguous descriptions which might prevent that package usage in `Julia`.

Thanks for taking the time and clarifying the issue. We will change the license to LGPL.

GeneralizedSchurAlgorithm/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+ToeplitzMatrices
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

neveritt: Do you mean a minimum version of both ToeplitzMatrices and Compat?

tkelman: unless every tagged version ever released of either of those packages works, then yes it would be good to declare a minimum on one or both

neveritt:
@dpsanders, this package might indeed be related to [iterativesolvers](https://github.com/JuliaMath/IterativeSolvers.jl). This package is playing around with solvers for matrices with displacement structure. My main interest being BlockToeplitzmatrices. The main reference is [Fast Reliable Algorithms for Matrices with Structure](http://epubs.siam.org/doi/book/10.1137/1.9781611971354).

PR #6463, opened 2016-09-22
Tag TikzGraphs v0.2.0, BayesNets v1.0.3, and Discretizers v0.3.1

tawheeler:
Fix warnings in 0.5


BayesNets/versions/1.0.3/requires
```diff
@@ -0,0 +1,10 @@
+julia 0.4
+Compat
```
ararslan: It would be good to add a minimum version for the dependencies, if not for all then at least for Compat. That ensures that the package will always be using a version that has the features it needs.

tawheeler: right - I removed that while trying to get ConjugatePriors to work in 0.5....
can add back.

PR #6469, opened 2016-09-23
Tag SampledSignals v0.2.1 [https://github.com/JuliaAudio/SampledSignals.jl]

ssfrr:
Switches to using resampling from DSP.jl instead of my own crappy linear-interpolating resampler. Big thanks to @JayKickliter for making that work with super low overhead in a streaming context.


tkelman:
Modifying the behavior of another package as in https://github.com/JuliaAudio/SampledSignals.jl/commit/81d03692e71039525651d71cc1b6c612f3263030 is a bit questionable, but I trust you plan on getting rid of that.

ssfrr:
Yeah, I agree it's pretty dirty, and I'm keeping an eye on DSP.jl so that I can remove it and bump the version number in my REQUIRE.

PR #6471, opened 2016-09-23
Tag IntervalWavelets v0.0.4

robertdj:
The new version *is* compatible with v0.5


IntervalWavelets/versions/0.0.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Wavelets
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

robertdj: It should be fixed now.

tkelman: yep, looks good, though be sure to make the same change at https://github.com/robertdj/IntervalWavelets.jl/blob/master/REQUIRE so it'll get picked up for the next tag too

robertdj: I've just pushed an update.

PR #6472, opened 2016-09-23
Bump DataStreams (back to 0.0.13 release)

quinnj:


tkelman:
should push a new numbered tag for this to https://github.com/JuliaData/DataStreams.jl/releases too

quinnj:
I did, it's [here](https://github.com/JuliaData/DataStreams.jl/releases/tag/v0.0.15). I think it doesn't show up at the top because it must order by commit?

tkelman:
my bad, guess it orders by commit date

DataStreams/versions/0.0.15/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+DataFrames
+NullableArrays 0.0.8
+CategoricalArrays 0.0.4 0.0.5
```
tkelman: thanks for remembering this bit

tkelman:
Hm, the failing jobs at https://travis-ci.org/JuliaData/DataStreams.jl did something funny, they saw they were on a tagged commit for 0.0.13 then upgraded to 0.0.14! Could you restart those failing builds now that this has been merged?

tkelman:
Cool, thanks. Lots of ambiguity warnings on 0.4 but at least thing pass. I think it would be a really good idea to make the next tag (both here and on CategoricalArrays, cc @nalimilan) 0.1.0 and 0.5-only, so if someone wants to deal with these warnings they can backport a new bugfix tag from a branch.

quinnj:
yeah, that's been my line of thinking as well. Hopefully we can get the new DataFrames into shape pretty soon, tag a 0.5-only and then do new releases for a bunch of downstream stuff, including a 0.1 for DataStreams

nalimilan:
> I think it would be a really good idea to make the next tag (both here and on CategoricalArrays, cc @nalimilan) 0.1.0 and 0.5-only, so if someone wants to deal with these warnings they can backport a new bugfix tag from a branch.

I'm not sure I follow. Do you suggest that not supporting 0.4 at all would be better than printing all those warnings (which I guess I could fix at some point)? But then we should also apply that change to already existing releases, which have the same problem.

tkelman:
Just don't try to make new risky features work on 0.4 is all. Fix the warnings if you can, but that can be done off a branch of the package that's just bugfixes.

PR #6475, opened 2016-09-23
Tag Mads v0.2.6

montyvesselinov:
Fixed to work with Julia 0.5


tkelman:
Do you plan to add testing on 0.5 to https://github.com/madsjulia/Mads.jl/blob/master/.travis.yml?

tkelman:
`String` will throw a deprecation warning on 0.4 if not imported from `Compat` (which would need to be explicitly in your require file)

montyvesselinov:
done!

-monty

montyvesselinov:
On Fri, Sep 23, 2016 at 11:22 AM, Tony Kelman <notifications@github.com>
wrote:

> Compat


I know this. But I am not sure how to address this.

I am also not sure what to do with AbstractString in 0.5:

function loadmadsfile(filename::AbstractString)

Should I drop it?

Thank you very much for your help!

-monty

tkelman:
You have a bunch of dependencies already, so adding Compat to your REQUIRE file (with a minimum version of 0.7.15 I believe) will be essentially free. Just do the first few lines in https://github.com/JuliaLang/Compat.jl#usage inside any modules where you would like to use `String` without any deprecation warnings across both 0.4 and 0.5. Be aware that `Compat.String` on 0.4 is not a concrete type, so it's better for function arguments than for fields of types. Non-concretely typed fields might hurt performance, but if the fields are already abstract it shouldn't make much difference.

tkelman:
There are still a few alternative string types other than `String` on Julia 0.5, so `AbstractString` still serves a purpose. But some of the functions you call on that string might not support other string types like SubString or other unicode encodings from LegacyStrings.jl, so you'd need to test to say for sure.

montyvesselinov:
Thank you very much! added.

-monty

PR #6477, opened 2016-09-23
Register / tag GraphPlot

sbromberger:
License issues fixed. Dependencies fixed. Todo: `Pkg.dir()` fix.


sbromberger:
Another dependency issue. Closing out - will hope that submission number 5 is the magic one.

PR #6481, opened 2016-09-23
Tag Unitful v0.0.2 [https://github.com/ajkeller34/Unitful.jl]

ajkeller34:
I have fixed a nasty promotion bug in v0.0.1 where `[1.0m, 2.0m] ./ 3` would throw a `Unitful.DimensionError()`. I'm inclined to work on improving promotion—a few things still don't work quite like unitless numbers, for instance `[1m, 2m]./3` will throw an `InexactError()`—but I'll probably wait to fix that until https://github.com/JuliaLang/julia/issues/18465 is patched so that I know which `@inferred` errors are my fault when I play around with `promote_op`. Unitful v0.0.2 also adds a new unit conversion syntax.

Is there a summary of the state of `promote_op` as of v0.5, as well as how to write code that is amenable to type inference? I'm sorry if that's a bit vague. I feel my `promote_op` functions could be simpler and better but I have tried writing them other ways and I get `@inferred` failures. If I remove the `promote_op` definitions entirely then I get a lot of `@inferred` failures, although I have some vague notion that `promote_op` is being phased out... it's not even in the 0.5 manual anymore.


tkelman:
@pabloferz why did you remove promote_op from the docs?

pabloferz:
`promote_op` was documented as a possible way of fixing some limitations with the promotion system but the whole thing changed before 0.5 was released and it wasn't supposed to be a method to be preferably override outside Base. I deleted for that reason, I guess I could have changed the section by one explaining the changes instead.

pabloferz:
@ajkeller34 If the problem was related to using `promote_op` with generated functions as you pointed out in https://groups.google.com/forum/#!topic/julia-users/EAoanGDSc6g, then the problem might be on the use of generated functions and not in `promote_op` itself. I just found out recently that are some restrictions when using generated functions, some of these are stated in the manual.

PR #6485, opened 2016-09-23
Test the sanity of package tags when running Travis on a PR

ararslan:
When someone submits a pull request on METADATA, this addition to the CI tests will ensure that, for each modified package,

1. All local tags are present on the remote,
2. If so, that the commit SHAs match between the local and remote tags, and
3. ~~If more than one tag is available for a package, that multiple tags don't share a commit SHA.~~

Items 1 and 2 produce errors, ~~3 produces warnings.~~

I've also made the (unrelated) change of adding 0.6 to the testing list. This is a separate commit so it can be removed if necessary.

cc @tkelman

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -234,6 +234,12 @@ for pkg in readdir("METADATA")
     end
 end
 
+if haskey(ENV, "TRAVIS_PULL_REQUEST") && ENV["TRAVIS_PULL_REQUEST"] != "false"
+    info("Checking repository tags...")
+    Pkg.add("JSON")
+    include("check-pr.jl")
+end
+
 info("Verifying METADATA...")
 if isdefined(Pkg.Entry, :check_metadata)
     Pkg.Entry.check_metadata()
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -1,0 +1,66 @@
+using JSON
+
+const PR_NUM = ENV["TRAVIS_PULL_REQUEST"]
+const BUILD_DIR = ENV["TRAVIS_BUILD_DIR"]
+const API_URL = "https://api.github.com/repos/JuliaLang/METADATA.jl"
+const CURL_URL = "$API_URL/pulls/$PR_NUM/files?per_page=100"
+
+# Get files modified in the PR from the GitHub API
+changes = JSON.parse(readchomp(`curl "$CURL_URL"`))
+
+# If all went well, `changes` will be an array, or a dict without that key
+if typeof(changes) <: Dict && haskey(changes, "message")
+    msg = changes["message"]
+    error("GitHub API gave result \"$msg\" from `$CURL_URL` GET request")
+end
+
+# Get the associated package and version for each affected file
+modified = Dict() # package => [versions...]
+for diff in changes
+    pkg = split(diff["filename"], "/")[1]
+    if isdir(joinpath(BUILD_DIR, pkg)) && pkg != ".test"
+        v = VersionNumber(match(r"^[^/]+/versions/([\d.]+)", diff["filename"]).captures[1])
+        if haskey(modified, pkg)
+            in(v, modified[pkg]) || push!(modified[pkg], v)
+        else
+            push!(modified, pkg => VersionNumber[v])
+        end
+    end
+end
+
+# Check that tagging makes sense
+for pkg in keys(modified)
+    dir = joinpath(BUILD_DIR, pkg)
+
+    remote_url = readchomp(joinpath(dir, "url"))
+    lines = split(readchomp(`git ls-remote --tags --refs -q $remote_url`), "\n")
+
+    remotetags = Dict(map(lines) do s
+        t = split(s, "\t")
+        VersionNumber(t[2][11:end]) => t[1]
+    end)
+
+    localtags = Dict(map(readdir(joinpath(dir, "versions"))) do v
+        sha = readchomp(joinpath(dir, "versions", v, "sha1"))
+        VersionNumber(v) => sha
+    end)
+
+    notpushed = setdiff(localtags, remotetags)
+    getmad = intersect(modified[pkg], notpushed)
+
+    if isempty(notpushed)
+        # Tags match, so we can compare SHAs 1-1
+        for tag in keys(localtags)
+            if localtags[tag] != remotetags[tag]
+                msg = string("The commit SHA for $pkg $tag does not match between METADATA ",
+                             "and the upstream package repository.")
+                error(msg)
+            end
+        end
+    elseif !isempty(getmad) # just those added in this PR
+        msg = string("The following tags for $pkg have not been pushed to upstream ",
+                     "repository: ", join(getmad, ", "), ".\nTo fix this, navigate to ",
+                     "the package directory and run `git push --follow-tags`.")
+        error(msg)
+    end
+end
--- a/.test/travis.sh
+++ b/.test/travis.sh
@@ -4,7 +4,7 @@ cd $(dirname $0)/..
 git checkout -b localbranch
 cd ..
 ln -s $PWD/METADATA.jl METADATA
-for ver in 0.4 0.5; do # add 0.6 once nightly is 0.6-dev
+for ver in 0.4 0.5 0.6; do
   mkdir -p ~/.julia/v$ver julia-$ver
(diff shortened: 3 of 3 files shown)


tkelman:
Warnings that only appear in the Travis log won't be very useful. I think this will need to result in at least an autogenerated comment or a custom PR status message for people to notice them.

ararslan:
Is checking whether the most recent commit SHA is the same as for a previous tag worth figuring out how to do that? That seems like something that should really be handled in PkgDev anyway, I just figured it wouldn't hurt checking for it here.

timholy:
Seems like the easy thing would be to throw an error for all three.

ararslan:
@timholy In general I agree, but there was some discussion in https://github.com/JuliaLang/PkgDev.jl/issues/62 about situations where it makes sense to use the same commit for multiple tags. That's why I made it a warning here rather than an error.

tkelman:
I don't think it should be an error here unless there's a way to add a manual exception list for it. Same sha is better as a local PkgDev warning, since it doesn't actually hurt anything it just may not be what the user intended.

ararslan:
Should I take that part out of here then?

.test/METADATA.jl
```diff
@@ -242,3 +242,8 @@ else
     import PkgDev
     PkgDev.Entry.check_metadata()
 end
+
+if haskey(ENV, "TRAVIS_PULL_REQUEST") && ENV["TRAVIS_PULL_REQUEST"] != "false"
+    Pkg.add("Requests")
+    include("check-pr.jl")
```
tkelman: better turnaround time if we do this before check_metadata, since that takes some time and doesn't fail very often

.test/check-pr.jl
```diff
@@ -0,0 +1,69 @@
...
+
+    if isempty(notpushed)
+        # Tags match, so we can compare SHAs 1-1
+        for tag in keys(localtags)
+            if localtags[tag] != remotetags[tag]
+                error("Commit SHA does not match between local and remote for $pkg $tag")
+            end
+        end
+    else
+        error("$pkg local tags not pushed to remote: " * join(notpushed, ", "))
```
tkelman: since this may fail for people pretty often when they forget, let's be super verbose so the travis log explains exactly what they forgot to do and how to fix it

ararslan: It's `git push --follow-tags`, right?

.test/travis.sh
```diff
@@ -4,7 +4,7 @@ cd $(dirname $0)/..
 git checkout -b localbranch
 cd ..
 ln -s $PWD/METADATA.jl METADATA
-for ver in 0.4 0.5; do # add 0.6 once nightly is 0.6-dev
+for ver in 0.4 0.5 0.6; do
```
tkelman: a downside of this is if a change on master happens to break pkgdev or requests, all metadata tagging would come to a halt until they're fixed from either a new tag of the package or a new nightly. this shouldn't be necessary until someone tags an 0.6-only package, I think. If it doesn't add much to the travis time, we could run it but not require it to pass, so you'd get errors in the log but only people looking closely would notice them

.test/check-pr.jl
tkelman: I usually tell people `git push --tags`, what's the difference?

.test/travis.sh
ararslan: Might be worth holding off on this then, since I very seriously doubt there are any packages even close to having a 0.6-dev-only release.

.test/check-pr.jl
```diff
@@ -0,0 +1,69 @@
...
+        VersionNumber(v) => sha
+    end)
+
+    notpushed = setdiff(localtags, remotetags)
+
+    if isempty(notpushed)
+        # Tags match, so we can compare SHAs 1-1
+        for tag in keys(localtags)
+            if localtags[tag] != remotetags[tag]
+                error("Commit SHA does not match between local and remote for $pkg $tag")
```
ararslan: "Commit SHA does not match between METADATA and remote" doesn't seem much clearer, IMO.

.test/check-pr.jl
ararslan: http://stackoverflow.com/a/26438076

.test/check-pr.jl
tkelman: don't say remote either. package upstream git tags or something. without context, local and remote aren't meaningful

.test/travis.sh
tkelman: does adding the additional run make a noticeable difference to travis time? I don't know exactly how many cores they have

.test/METADATA.jl
tkelman: or do both asynchronously if they're both read only operations and that makes things any faster

.test/travis.sh
ararslan: I think they have 2 vCPUs. I didn't notice much of a difference but I haven't done any rigorous timing.

.test/check-pr.jl
ararslan: Okay, made some updates

ararslan:
Issues I just realize that I need to deal with before this can be merged, aside from any comments from others:

- [ ] Check whether the repository is hosted on GitHub before trying to look for it using the GitHub API
- [ ] Deal with paginated results from the API

ararslan:
Changed to shell out to Curl rather than using the Requests package, which has to build all kinds of stuff before it can work.

.test/travis.sh
```diff
@@ -18,6 +18,6 @@ for ver in 0.4 0.5; do # add 0.6 once nightly is 0.6-dev
     touch success-$ver &
 done
 wait
-if ! [ -e success-0.4 -a -e success-0.5 ]; then # add success-0.6 once nightly is 0.6-dev
+if ! [ -e success-0.4 -a -e success-0.5 ]; then
```
tkelman: leave the comment but as "once there are any 0.6-only package versions"

.test/check-pr.jl
```diff
@@ -0,0 +1,63 @@
...
+                error(msg)
+            end
+        end
+    else
+        msg = string("The following tags for $pkg have not been pushed to the remote ",
+                     "repository: ", join(notpushed, ", "), ".\nTo fix this, navigate ",
+                     "to the package directory, ensure the tags exist locally by running ",
+                     "`git tag --list`, and if they do, run `git push --follow-tags`.\nIf ",
+                     "the tag does not exist, run `", VERSION < v"0.5-" ? "Pkg" : "PkgDev",
+                     ".tag(\"$pkg\")` and try again.")
```
tkelman: is Pkg.tag the right thing to recommend? the user submitted a PR somehow, so wouldn't Pkg.tag create new metadata files?

tkelman:
we should simulate this on existing packages. many might be missing git tags for a lot of older metadata tags, and I think in the current form this would fail on those. maybe we need a PkgDev function first to create any git tags that don't exist yet based on metadata tag info.

.test/check-pr.jl
ararslan: Good call. Should I just leave out the recommendation entirely?

ararslan:
Why would there already be tags in METADATA that don't correspond to upstream tags?

tkelman:
there were a lot of metadata tags that never had the git tags pushed.

ararslan:
Okay, changed it to only check whether the most recent tag has been pushed. Does that seem reasonable?

tkelman:
Probably all new tags added in a PR would be better

ararslan:
Okay, implemented that.

.test/check-pr.jl
```diff
@@ -0,0 +1,66 @@
...
+# Get the associated package and version for each affected file
+modified = Dict() # package => [versions...]
+for diff in changes
+    pkg = split(diff["filename"], "/")[1]
+    if isdir(joinpath(BUILD_DIR, pkg)) && pkg != ".test"
+        v = VersionNumber(match(r"^[^/]+/versions/([\d.]+)", diff["filename"]).captures[1])
+        if haskey(modified, pkg)
+            in(v, modified[pkg]) || push!(modified[pkg], v)
+        else
+            push!(modified, pkg => VersionNumber[])
```
tkelman: should this be initialized with `v` ?

ararslan: Yes

PR #6487, opened 2016-09-24
Register StructuredQueries [https://github.com/davidagold/StructuredQueries.jl.git]: v0.0.1

davidagold:


tkelman:
There's only one package that starts with a lower-case letter, and it is very old and has no tags. Julia naming conventions suggest capitalized module names, and packages by extension.

davidagold:
@tkelman Okay. I figured I may as well change the whole name, then.

I tagged v0.0.2 just because I read somewhere that trying to force revise tags is a bad idea. Or something like that. Is that okay?

tkelman:
Revising tags that haven't been pulled by anyone or merged to metadata is fine. Since the commit at which 0.0.1 was tagged was using a different module name, it won't work and the files for that tag should just be removed from this branch of your metadata fork

davidagold:
How's this?

tkelman:
Best to make the git tags and metadata tags consistent - https://github.com/davidagold/StructuredQueries.jl/releases. Neither technically has to start at 0.0.1, gaps aren't a big deal.

davidagold:
What's the best way to delete the tags I've pushed to `origin`?

quinnj:
`git tag -d v0.0.1`
`git push origin :v0.0.1`

davidagold:
Woo! Thank you both. Okay, https://github.com/davidagold/StructuredQueries.jl/tags should be good now.

PR #6490, opened 2016-09-24
Tag FixedPointNumbers v0.1.7

timholy:
Changes (@timholy unless stated otherwise):
- Reliably detect impossible conversions
- Produce informative errors for failed conversions
- Add unchecked conversions (@timholy and @vchuravy) and improve efficiency of some conversions
- Support promotion among UFixed types
- Fix bugs in <=, iteration, operations with high-precision UFixed types
- Deprecate lowercase conversion functions
- Turn on AppVeyor
- Turn on coverage analysis
- Improve the README


tkelman:
The change in exception types here broke several downstream packages' tests.

timholy:
Argh. I didn't think of that. Do you have a list?

Or should I revert? It really is a huge usability step forward to provide useful error messages, so I'm much more inclined to fix problems elsewhere than revert this.

tkelman:
see http://pkg.julialang.org/pulse.html at least for packages that were previously passing

tkelman:
as an immediate mitigation, we can add upper bounds on the dependencies of all past tags of the packages that broke

timholy:
Easier to just fix the problems. (If we want to be serious about upper bounds, we need a bot.) #6513.

tkelman:
pkgeval is a bot. not exactly what you mean, but the bounds would still be correct to add

timholy:
Sure, but if we really wanted correctness we'd probably have to add upper bounds on >90% of the tags in METADATA. It's just that we don't know what those bounds should be, because the vast majority of bugs get fixed the way I fixed this one. It's a tedious and thankless job, so if we want to be serious about doing that we're better off with a recommender bot + implementer function.

PR #6491, opened 2016-09-24
Register and tag RestrictProlong

timholy:
Grid.jl has one component that is not covered by Interpolations.jl: implementations of key operators for multigrid methods, namely `restrict` and `prolong`. This new standalone package provides those operators (and is considerably more efficient than Grid).

The other motivation for making it standalone is that it's useful in other operations, too; for example, IJulia is pretty slow at displaying large images, so it will be used to reduce the resolution before sending the data to Jupyter.


tkelman:
Yay, multigrid!

PR #6493, opened 2016-09-24
Tag TaylorSeries v0.3.0

lbenet:
This tags a new minor version, with correct minimum version of Compat.


tkelman:
Do you want both the 0.2.1 tag and the 0.3.0 tag in here?

lbenet:
I don't mind; thanks for asking. 

Clearly it was a mistake to tag as 0.2.1, in the sense that it was not a patch, but a minor version...

tkelman:
So do you want to remove the 0.2.1 tag then? Or if the need arose to make an 0.2.2 tag would you create a branch and do things out of order?

lbenet:
I didn't know you could erase such tags. I think it's ok to leave it as it is.

lbenet:
Thanks!

PR #6496, opened 2016-09-25
Tag LCIO v0.4.2 [https://github.com/jstrube/LCIO.jl]

jstrube:


LCIO/versions/0.5.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

LCIO/versions/0.5.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps 0.4.5
+Compat 0.9.2
```
tkelman: you're using something that was only implemented in the most recent version?

jstrube: I don't have any users right now and don't want to spend time figuring out how far I can go back. If there's a user request to support an older setup, I'll look into it.

tkelman: there are specific instructions on the compat readme, it's a few clicks

jstrube:
Thanks for merging. I appreciate the pointers, too

PR #6499, opened 2016-09-25
Register ExpressionPatterns [https://github.com/fcard/ExpressionPatterns.jl.git]

fcard:
Hello there!
I spent more than a year holding back registering this since there already is MacroTools, but since both PatternMatch and Match are registered I though I'd give it a try, to make a counterpart to a counterpart*.

Also, mine supports the fabled multiple and nested slurps! (which makes the whole code 10x more complex and a lot slower [working on the latter...], so I get why we don't see it very often)

*<sub>A counterpart to a counterpart both in the functionality and act of counterparting, therefore making the duo a counterpart of counterparts to the other counterparts as well as a meta counterpart to the functional packages...!? Makes sense!?</sub>

### DIFF
--- a/ExpressionPatterns/url
+++ b/ExpressionPatterns/url
@@ -1,0 +1,1 @@
+https://github.com/fcard/ExpressionPatterns.jl.git


fcard:
Oh boy! Did I misinterpret this?
```
Once the package URL for FooBar is registered in the official METADATA repo,
people know where to clone the package from, but there still aren’t any
registered versions available. You can tag and register it with the PkgDev.tag() command...
```
Makes it sound like tagging can comes after integration in METADATA. Either way, will fix asap. Since this PR is attached to a commit I guess I will need to make another PR... Is it okay if I make one manually with a branch? (automated tools with uneditable output are the bane of my existence!)

PR #6504, opened 2016-09-25
Fixed build on Julia 0.5

aaalexandrov:
Fixed building of DevIL / OpenIL for Julia 0.5.0 and bumped up the package version to 0.4.0


aviks:
To confirm, version `0.4.0` of `DevIL` will not be installable on `Julia 0.5`?

tkelman:
Can you push the git tag so it shows up at https://github.com/JuliaGL/DevIL.jl/releases ?

tkelman:
@ararslan shouldn't your check have flagged this?

aaalexandrov:
I've pushed the v0.4.0 tag to the DevIL repo, sorry about that.
And to answer the question above - no, I'm not supporting Julia 0.4 in the new release, I've changed the requirements to Julia 0.5-
Does the Julia package manager sync to the latest published package always, or does it try to sync to an earlier package version that corresponds to the actual Julia version?

aaalexandrov:
Is there a way to re-run the checks without submitting changes again?
It should be ok, but it's still showing the result from before me pushing the tags.

tkelman:
julia 0.4 will keep using the last tagged version of the package that supports 0.4 according to its require file. this new tag will be installable on prerelease (the trailing dash) of 0.5 or newer

PR #6505, opened 2016-09-25
Tag CodeTools v0.4.1 [https://github.com/JunoLab/CodeTools.jl]

pfitzseb:
Enabled travis and fixed some bugs.


pfitzseb:
Travis just failed, log [here](https://gist.github.com/pfitzseb/1541c0eeaca6b5c42765024d3fccf83f) in case someone's interested.

CodeTools/versions/0.4.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5-
+Lazy
+LNR
+JuliaParser
+MacroTools
+Compat
```
ararslan: It'd be a good idea to specify a minimum version for Compat to ensure you get the compatibility features you need. (It's a good idea for any dependency, but especially Compat.)

CodeTools/versions/0.4.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5-
```
tkelman: doesn't have to be soon, but when you're ready to stop supporting prereleaes, drop the trailing dash

PR #6506, opened 2016-09-25
update versions for LambertW.jl and ZChop.jl

jlapeyre:
This is same as the pull request that I just closed, except
the julia version should be added to all the necessary `requires` files.


ZChop/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

tkelman:
You should enable Travis CI. Note that `release` will mean 0.5, so you should add an explicit 0.4 entry in `.travis.yml` to continue testing against that version of Julia for as long as it's supported according to your REQUIRE file.

jlapeyre:
I enabled Travis-CI in the github repositories and added explicit versions in .travis.yml

LambertW/versions/0.0.5/sha1
```diff
@@ -0,0 +1 @@
+1a3ecb7000435ba8c7f1cc8516fd553d10c93776
```
tkelman: this doesn't match the git tag at https://github.com/jlapeyre/LambertW.jl/releases - are you sure you want to tag all these versions, or just the latest one?

ZChop/versions/0.1.1/sha1
```diff
@@ -0,0 +1 @@
+008a63dff6d60f3773b72e6c09a17dc46f084f66
```
tkelman: this tag was not pushed to https://github.com/jlapeyre/ZChop.jl/releases

LambertW/versions/0.0.5/sha1
tkelman: It's not particularly harmful to have too many versions, just checking. It also looks like you deleted your metadata fork?

jlapeyre: I am still learning this process.

I really want as few versions as possible. I am trying to follow the manual, but it ends in confusion. E.g. when I try  `PkgDev.publish`, I get a message saying I should first update. Then when I push to my fork I see that other peoples' recently accepted pull requests are included in mine. etc.

I have been deleting the metadata fork and forking again, because the various julia and git tools refuse to do what I ask (probably for good reason). E.g., git push jlapeyre metadata-v2 refuses to push anything. I was afraid that the pull request would no longer refer to my new metadata fork and that seems to be the case. 

I can simply issue another PR if that will work for you.  I can delete all version tags greater than the latest in the julia repo of METADATA.jl, except for one. I can check that the tagged commits match those in my METADATA and push everything and issue a new PR.  If this does not work, I can back up further in the process. This would be easiest for me.

I will try these things and then add another comment here.

ZChop/versions/0.1.1/sha1
jlapeyre: I will check this now.

LambertW/versions/0.0.5/sha1
jlapeyre: I left all the tagged commits in the two packages because the tags are already published on github. I pushed all tags from the packages to github. But, I removed unnecessary version tags for them from my fork of METADATA. I pushed this change to github. Will this work for you ?

ZChop/versions/0.1.1/sha1
jlapeyre: Done.

LambertW/versions/0.0.5/sha1
tkelman: happy to assist you if you're having trouble with the process. a new PR with just one new tag per package sounds like the best plan

jlapeyre: Thanks. That indeed sounds like the best plan.

PR #6509, opened 2016-09-25
Tag TextAnalysis v0.1.0

aviks:


TextAnalysis/versions/0.1.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+Languages 0.0.5+
```
tkelman: what are you intending with the trailing `+` ?

PR #6511, opened 2016-09-26
Only check changes to version files

ararslan:
This should fix the failure in #6510.

cc @tkelman and @fcard

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -14,12 +14,16 @@ if typeof(changes) <: Dict && haskey(changes, "message")
     error("GitHub API gave result \"$msg\" from `$CURL_URL` GET request")
 end
 
+const RGX = r"^[^/]+/versions/([\d.]+)"
+
 # Get the associated package and version for each affected file
 modified = Dict() # package => [versions...]
 for diff in changes
-    pkg = split(diff["filename"], "/")[1]
-    if isdir(joinpath(BUILD_DIR, pkg)) && pkg != ".test"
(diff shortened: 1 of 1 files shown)


.test/check-pr.jl
```diff
@@ -14,12 +14,16 @@ if typeof(changes) <: Dict && haskey(changes, "message")
...
 # Get the associated package and version for each affected file
 modified = Dict() # package => [versions...]
 for diff in changes
-    pkg = split(diff["filename"], "/")[1]
-    if isdir(joinpath(BUILD_DIR, pkg)) && pkg != ".test"
-        v = VersionNumber(match(r"^[^/]+/versions/([\d.]+)", diff["filename"]).captures[1])
+    fname = diff["filename"]
+    pkg = split(fname, "/")[1]
+    # Only look at changes to tagged versions
+    if isdir(joinpath(BUILD_DIR, pkg)) && pkg != ".test" && diff["status"] == "added" && ismatch(RGX, fname)
```
tkelman: what about modified? people shouldn't usually do that, but occasionally they do

ararslan: It'll bug them about pushing tags if they go back and modify an old tag in METADATA. Is that desirable?

tkelman: only if they modify the sha, right?

ararslan: Only if they modify an old tag that wasn't pushed... though I guess in that case they _should_ push it. So yeah, the status check probably shouldn't be in there at all.

PR #6515, opened 2016-09-26
Tag TaylorIntegration v0.0.2

PerezHz:
This PR tags TaylorIntegration v0.0.2, which fixes testing issues present in v0.0.1

PR #6516, opened 2016-09-26
Select the appropriate SHA from git ls-remote

ararslan:
Addresses the issue noted in #6515.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -37,12 +37,29 @@ for pkg in keys(modified)
     dir = joinpath(BUILD_DIR, pkg)
 
     remote_url = readchomp(joinpath(dir, "url"))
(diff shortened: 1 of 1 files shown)


.test/check-pr.jl
```diff
@@ -37,11 +37,12 @@ for pkg in keys(modified)
     dir = joinpath(BUILD_DIR, pkg)
 
     remote_url = readchomp(joinpath(dir, "url"))
-    lines = split(readchomp(`git ls-remote --tags --refs -q $remote_url`), "\n")
+    lines = split(readchomp(`git ls-remote --tags -q $remote_url`), "\n")
+    filter!(line -> endswith(line, "^{}"), lines)
```
tkelman: this won't work for lightweight tags - come to think of it, will `--follow-tags` ?

PR #6517, opened 2016-09-26
disable the tag check for now

tkelman:
cc @ararslan this needs a lot more testing on your fork


tkelman:
It's also hitting the github api rate limit, which we'd need to find a way to get around.

ararslan:
I'm close to the daily unauthenticated API request limit just from testing code.

tkelman:
we can probably do something local like `git diff --name-only origin/metadata-v2 HEAD` ?

PR #6519, opened 2016-09-26
tag UnicodeFun

SimonDanisch:
I hope the name is okay for a package that offers different unicode translation function, e.g.
`to_latex(latex_string) --> unicode_string`, or to `subscript(string) --> unicode_subscipt_string`.
Since this is incredibly `fun`, Function is abbreviated in the name ;)


tkelman:
Really shouldn't be extending base methods on base types as in `Base.findnext(A::AbstractString, v::Char, idx::Integer)`, that changes the behavior of unrelated code

SimonDanisch:
Okay, thought this is more elegant than calling it `utf8_findnext`...
Note, that it shouldn't change behavior, and only strictly improve the case were it wasn't functioning before.
See: https://github.com/JuliaLang/julia/pull/18668
If your concerns persist, I can surely change the name, though ;) Maybe the correct way forward is to move this into Compat and wait for my PR to be merged.

tkelman:
Changing something from an error to not an error is changing behavior. If the change gets merged to base then we can figure out the best way to port it to Compat.

SimonDanisch:
Okay, removed it ;)

SimonDanisch:
thanks! 👍

PR #6521, opened 2016-09-26
Tag Parameters v0.4.0 v0.5.0

mauro3:
This used to be #6381 but I could not re-open that one.  I think this is correct but I had to edit some stuff by hand.


Parameters/versions/0.5.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+DataStructures
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

mauro3: Tnx. I did some detective work and updated https://github.com/mauro3/Parameters.jl/blob/master/REQUIRE and the `requires` for 0.5.  I can't easily update the `REQUIRE` for 0.4 but should I update `Parameters/versions/0.4.0/requires` nonetheless?

tkelman: if the min ver applies, yes best to edit the metadata requires file

mauro3: Somewhat related.  In Travis it doesn't actually install the minimal required packages, correct?
See https://travis-ci.org/mauro3/Parameters.jl/jobs/162817273 .  Is there a way to do this?

tkelman: what do you mean? it installs your dependencies, looks like it's doing the right thing

mauro3: Yes, but it did not install Compat 0.7.15 but the latest 0.9.2.

tkelman: oh i see. Pkg tries to maximize versions, you have to pin or specifically ask it otherwise

mauro3: I guess that makes sense usually. But for testing a switch to checkout the minimal requirements would be good too.

mauro3: The other `requires` is also adjusted.  Probably ready to merge now.

PR #6525, opened 2016-09-26
Bump ColorVectorSpace, SimpleTraits

timholy:


SimpleTraits/versions/0.1.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
ararslan: It would probably be a good idea to add a minimum version for Compat to ensure the requisite features are always available to SimpleTraits.

mauro3: I'll check.

mauro3: It's actually only needed for the tests.  Is there a way to specify this?

ararslan: If the package itself doesn't rely on Compat, only the tests, you can remove Compat from the main REQUIRE file entirely and instead put it in test/REQUIRE. (That file doesn't need a Julia version, nor does it need to be in METADATA.)

ararslan: Looks like you're doing `using Compat`. If the main package code really doesn't need Compat, you'll have to remove the `using` as well as the REQUIRE entry.

mauro3: Yep, that `using Compat` is a hangover from the 0.3 days.

mauro3:
@timholy would you mind re-tagging SimpleTraits to 78ad7adbc4b42f641d105036b43d7a063a266f8a ?  Although better wait for the tests to finish in case I missed something.

SimpleTraits/versions/0.1.1/requires
ararslan: The days we all would like to forget...

timholy:
Done.

PR #6529, opened 2016-09-27
Simplify and (hopefully!) fix tag checking for PRs

ararslan:
There are a number of simplifications and corrections to the tag checking code:

* No more dependency on JSON.jl
* No more use of the GitHub API (which means no more quotas and rate limits)
* Get both annotated and lightweight tags when fetching remote tags
* Correctly identify when a tag hasn't been pushed to the remote

In simulations it's shown to exhibit the desired behavior in all of the circumstances I could think of. I think it runs a little faster too now. Maybe.

~~I also made the unrelated change of having the Bash script print the Julia version it's running. It's helpful for debugging when looking at the Travis log to have that be obvious.~~

cc @tkelman

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -1,29 +1,65 @@
-using JSON
-
-const PR_NUM = ENV["TRAVIS_PULL_REQUEST"]
 const BUILD_DIR = ENV["TRAVIS_BUILD_DIR"]
-const API_URL = "https://api.github.com/repos/JuliaLang/METADATA.jl"
-const CURL_URL = "$API_URL/pulls/$PR_NUM/files?per_page=100"
 
-# Get files modified in the PR from the GitHub API
-changes = JSON.parse(readchomp(`curl "$CURL_URL"`))
+function get_remote_tags(url)
+    ls = readchomp(`git ls-remote --tags -q $url`)
+    lines = split(ls, "\n")
+    n = length(lines)
 
-# If all went well, `changes` will be an array, or a dict without that key
-if typeof(changes) <: Dict && haskey(changes, "message")
-    msg = changes["message"]
-    error("GitHub API gave result \"$msg\" from `$CURL_URL` GET request")
+    tags = Vector{VersionNumber}(n)
+    shas = Vector{AbstractString}(n)
+    peel = Vector{Bool}(n)
+
+    for (i, line) in enumerate(lines)
+        sha, tag = split(line, "\trefs/tags/")
+
+        peeled = endswith(tag, "^{}")
+        peeled && (tag = tag[1:end-3])
+
+        tags[i] = VersionNumber(tag)
+        shas[i] = sha
+        peel[i] = peeled
+    end
+
+    remotetags = Dict{VersionNumber,AbstractString}()
+
+    for tag in unique(tags)
+        ispeeled = map((t, p) -> t == tag && p, tags, peel)
+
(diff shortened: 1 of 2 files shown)


.test/travis.sh
```diff
@@ -5,6 +5,7 @@ git checkout -b localbranch
 cd ..
 ln -s $PWD/METADATA.jl METADATA
 for ver in 0.4 0.5 0.6; do
+  echo "\n============[ ${ver} ]============"
```
tkelman: The `versioninfo()` already accomplishes this. Maybe prepending to every line of stdout from the Julia command would be useful though?

ararslan:
> `error: Could not access 'origin/metadata-v2'`

Well, that was unexpected.

.test/travis.sh
ararslan: Not sure what you mean. Or maybe I do but I don't know how to do that. Looking at the log this doesn't do what I hoped it would anyway so I'll just remove this commit.

.test/check-pr.jl
```diff
@@ -1,29 +1,63 @@
...
+        peel[i] = peeled
+    end
+
+    remotetags = Dict{VersionNumber,AbstractString}()
+
+    for tag in unique(tags)
+        ispeeled = map((t, p) -> t == tag && p, tags, peel)
+
+        sha = if any(ispeeled)
+            first(shas[ispeeled])
```
tkelman: we'd expect there to only be one here, verify that is the case?

.test/check-pr.jl
```diff
@@ -1,29 +1,63 @@
...
+        else
+            first([tags shas][tags .== tag, 2])
```
tkelman: can we do this without creating arrays or logical indexing? `shas[findfirst(tags, tag)]`

tkelman: maybe with verification that there's only one tag match, if being paranoid

ararslan: Who knows why I did it that way. Thanks for the much simpler, saner suggestion.

.test/check-pr.jl
ararslan: Added a check. I didn't know what to actually call the tag in that (hopefully impossible??) situation, so I just called it "malformed" in the error message. Let me know if there's something better.

.test/check-pr.jl
```diff
@@ -1,29 +1,65 @@
...
+    for tag in unique(tags)
+        ispeeled = map((t, p) -> t == tag && p, tags, peel)
+
+        sha = if any(ispeeled)
+            shas[ispeeled]
+        else
+            shas[findfirst(tags, tag)]
+        end
+
+        length(sha) == 1 || error("Upstream tag $tag is possibly malformed")
```
tkelman: findfirst will only return one (or a boundserror if 0), so maybe verify `sum(t -> t == tag, tags) == 1` in the not-peeled case?

.test/check-pr.jl
tkelman: maybe dump the raw results of the refs that are being processed if it goes wrong? we hope for it not to happen, but remote debugging and reproducing things that fail on travis is a pain

.test/check-pr.jl
ararslan: Good idea.

.test/check-pr.jl
ararslan: That seems like a good idea. I can do that in another PR since this one has already been merged.

PR #6532, opened 2016-09-27
Tag TensorOperations v0.5.0

Jutho:
For compatibility with Julia v0.5


TensorOperations/versions/0.5.0/sha1
```diff
@@ -0,0 +1 @@
+b379dbf4b222ced96c816a67cafba08450e8452c
```
tkelman: push the git tag so it shows up at https://github.com/Jutho/TensorOperations.jl/releases

Jutho: Oops. Something went wrong during `PkgDev.tag` and I therefore switched to manually updating `METADATA.jl`, but did not check to also push the tag to `TensorOperations.jl`. Unfortunately, it is at my work computer so it will have to wait until tomorrow morning. Thanks for spotting this.

Jutho:
Ok if I close this and submit a new PR?

PR #6536, opened 2016-09-27
Tag ZChop v0.1.1

jlapeyre:
Fixed bitrot.


ararslan:
The Travis failure is my fault.

This looks good to me, though I would recommend adding a minimum version for the Compat dependency.

jlapeyre:
If I add the Compat version to my local branch and push to github, will this be included in this PR ? Or should I cancel the PR, make the changes and then a new PR ?, or ...

ararslan:
> If I add the Compat version to my local branch and push to github, will this be included in this PR ?

Yep, that will work, though you'll also want to add the version to the REQUIRE file in the repository. (That part doesn't affect the PR though.)

jlapeyre:
In order to push the edited REQUIRE in ZChop to github I will have to increment the patch number and tag the ZChop commit, and also change the version of ZChop in METADATA to 0.1.1, as well. Does this sound correct ?

ararslan:
Based on my understanding, that shouldn't be necessary. I think you can just go to your local clone of METADATA, apply the change to `ZChop/versions/0.1.1/requires`, and push the change to your fork. For example, it might look like
```bash
cd ~/.julia/METADATA
vim ZChop/versions/0.1.1/requires   # or whatever your editor of choice may be
# apply change
git add ZChop/versions/0.1.1/requires
git commit -m "Add minimum version for Compat"
git push jlapeyre metadata-v2
```
You can make the change to the package repository's REQUIRE file separately without affecting the commit used for the v0.1.1 tag. What matters is having the version present in METADATA, since that will dictate the installed version for users.

(@tkelman, please correct me if anything I said was nonsense or otherwise blatantly false)

ararslan:
Wait a sec--ZChop doesn't actually depend on Compat at all? `using Compat` doesn't appear anywhere in the code. If that's the case, this discussion is moot and you can just remove Compat from REQUIRE entirely. :)

jlapeyre:
So, the tagged version of the package repo will not have the Compat version number; this change will go into a future version.

But, I just discovered that the dependency on Compat is bogus. The package does not in fact depend on Compat. In this case the version does not matter. How about leaving Compat unversioned in `require`, and fixing both of these in the next release (i.e. removing all mention of Compat) ?

tkelman:
that sounds fine. restarting travis here

jlapeyre:
`using Compat` does not occur in the current master branch on github, but it does in v0.1.1, which has already been pushed to github. So, if the PR is for v0.1.1, it seems I have to leave `Compat` in `require`.

Also, if I understand correctly, I can remove Compat everywhere, tag v0.1.2, push both the package and METADATA, and rename the PR to "Tag ZChop v0.1.2"

tkelman:
renaming the PR is optional, but if you'd like to make the additional tag now you can do so

PR #6548, opened 2016-09-28
Tag Spectra v0.1.8

charlesll:
New version with new functions (ctxremoval, mlregressor) and correction of different errors in the rameau function. The baseline function has been improved. The examples have been updated, as well as the documentation.


tkelman:
Looks like your tests are failing.

charlesll:
Yes, thanks for pointing that. As I changed the code for the splines, I need to update the smoothing coefficients for the splines in the test. I will have a look at that in a few hours.

charlesll:
I also am running into a wall with the array assignment in Julia 0.5. Because of that, Spectra.jl probably does not work with Julia 0.5.

For instance, saying I had a x axis=

`x = collect(250:0.5:1300)`

Saying that I have a data array with the same size of x in rows, and n columns. To select specific elements in this data array, I used to do:

```
x_wanted = x[x.>500.]
y_wanted = data[x .> 500.,:]
```

The first x_wanted is working, but the y_wanted is giving me a BoundsError... Do you have any idea about this? I know something changed in 0.5 regarding the behaviour of indexing, but I don't really see to which specific change this issue is related...

charlesll:
Ok I found the origin of the indexing behaviour. The previous code works with doing

`y_wanted = data[x[:] .> 500.,:]`

Is that normal? I don't really understand why for this particular line of code doing x .>500 or x[:] .> 500 is different?

PR #6552, opened 2016-09-28
Register TablesDemo [https://github.com/davidagold/TablesDemo.jl.git]: v0.0.1

davidagold:


aviks:
We'll rerun the tests once `AbstractTables` get registered.

nalimilan:
Obligatory bike shedding session: maybe better call this `TablesDemo`? No need to make people believe this is *the* Julia table type until they read the description...

davidagold:
@nalimilan That's a fair suggestion. I've begun using this package for my own research, and would prefer to use the present name. However, if you think that there's a serious risk of confusion, I'll change it to TablesDemo.

nalimilan:
I wouldn't call it "serious", but I guess clarity would be better for everybody. Anyway, your call.

davidagold:
@nalimilan How would you feel if I registered Tables.jl for my team at UW (University of Washington) and changed all the `Tables` references in my blog post to `DataFrames`? That may be even less confusing than introducing people to a `TablesDemo` package, since they may then think that the latter is a demo of a `Tables` package that’s being developed for public use.

davidagold:
Actually, the above can be solved with a private METADATA. I've changed the name to TablesDemo.jl. Still requires AbstractTables.

TablesDemo/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5
+Reexport 0.0.3
+Compat 0.9.2
+AbstractTables 0.0.1
+NullableArrays 0.0.9
+DataStreams 0.0.15
+CSV 0.0.15
```
tkelman: no such thing yet?

davidagold: You are right. This should be better.

TablesDemo/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+942c6131a5c9d19808546a55e778b4be9bec585d
```
tkelman: this commit is present as the tag but not on master? https://github.com/davidagold/TablesDemo.jl/commit/942c6131a5c9d19808546a55e778b4be9bec585d

davidagold:
Forgot to push. That is [corrected](https://github.com/davidagold/TablesDemo.jl/commit/942c6131a5c9d19808546a55e778b4be9bec585d) now, thank you!

tkelman:
You have a pretty distinctive way of structuring your packages. I think I like it.

davidagold:
Thanks :) I'm glad it makes sense to somebody else, too. And thank you for bearing with me while I straightened everything out.

PR #6553, opened 2016-09-28
Register JSONify [https://github.com/joshbode/JSONify.jl.git] 

joshbode:
New package: JSONify.jl

cc: @JockLawrie


tkelman:
It's pretty strongly discouraged to extend methods that aren't from your package on types that aren't from your package. That can change the behavior of completely unrelated code depending on whether or not your package has been imported.

joshbode:
Thanks Tony - where would you recommend something like this living (if
anywhere?) - JSON.jl? DataFrames.jl?

tkelman:
Having DataFrames depend on JSON would probably be an easier sell than the other way around. Or it's fine to have as its own package if it either uses its own types (trivial wrappers perhaps) or its own methods.

Requires is also problematic in combination with precompilation (which you haven't yet turned on, but anyone who wanted to use this package within a package that does have precompilation enabled would have to do), there aren't any great reliable ways of doing both precompilation and conditional dependencies right now - there are lots of hacky ways but most of them cause problems.

joshbode:
Thanks Tony - work is going on to integrate some of this into JSON.jl, after which I'll make another PR to DataFrames.jl (using the new `lower` methods JSON.jl uses).
I'll close this.

PR #6555, opened 2016-09-28
tag GeometryTypes, FixedSizeArrays and FreeType

SimonDanisch:
**GeometryTypes:**
+ polygon triangulation
+ removed depwarns
+ FlexibleSimplex, FlexibleConvexHull
+ bug fixes

**FixedSizeArrays:**
+ scalar isnan
+ fixes

**FreeType**
+ removed pointer argument restriction to let ccall do the proper conversion


FreeType/versions/1.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+@windows WinRPM
+Compat
```
tkelman: should add a minimum version that implements the features being used

tkelman:
what is the license status of https://github.com/JuliaGeometry/GeometryTypes.jl/pull/58, exactly?

also when you add features, you should really be incrementing minor version

SimonDanisch:
I've asked the author and never got a reply :(

tkelman:
then you don't have permission to license it under mit

SimonDanisch:
Uh... so what are my options?

Keno:
Send another email or write a clean-room re-implementation.

SimonDanisch:
Okay, thanks!

tkelman:
Where did it come from exactly? The site may have copyright info somewhere

SimonDanisch:
It seems that the author has the same code here: http://wiki.ros.org/convex_decomposition
Which does use the MIT license, as far as I can tell.
(http://docs.ros.org/jade/api/convex_decomposition/html/triangulate_8cpp_source.html)

tkelman:
good, hopefully that contribution was done by the author instead of copy-pasted by someone else. If you attribute that MIT license then you should be fine.

PR #6557, opened 2016-09-28
Tag Reactive v0.3.5 [https://github.com/JuliaLang/Reactive.jl]

SimonDanisch:
Fixed the `stop` event loop function and added `post_empty` function.
@shashi


Reactive/versions/0.3.5/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: add a minimum version dependency that implements the features you need

tkelman:
`release` in `.travis.yml` is about to mean 0.5, add an explicit 0.4 entry to continue testing there

PR #6460, opened 2016-09-22
Tag Distributions v0.10.3

ararslan:
Got a SecureTransport error on OS X using `PkgDev.publish()`, so I'm doing it manually. Bummers.

PR #6559, opened 2016-09-28
Tag StatsBase v0.9.1

simonbyrne:


ararslan:
Looks like the 0.9.1 tag hasn't been pushed. See https://travis-ci.org/JuliaLang/METADATA.jl/builds/163420020#L182.

tkelman:
Is the removal of functions a breaking change here?

ararslan:
The removal of `xmultinom_sample!` shouldn't be breaking, since it wasn't exported nor intended to be used directly.

The removal of the StatsFuns dependency, including the reexporting of the StatsFuns functions might be marginally more problematic, though I would still be surprised if anyone was calling then directly from StatsBase rather than going through StatsFuns.

Renyi entropy and generalized mean functions were added.

Now, the change to the API of `mad` is a user-facing, breaking change. I don't know how widely used that function is, but any code that calls `mad` with more than one argument will fail under StatsBase 0.9.1.

So maybe this should be a minor version instead? (I don't know how kosher this is, but I would just delete the 0.9.1 tag, retag as 0.10.0, then force push to this PR's branch...)

simonbyrne:
You're right, a minor tag is probably better. I'll update it.

ararslan:
Awesome, thanks!

tkelman:
Looks like this broke Distributions.jl's tests, is that known? http://pkg.julialang.org/detail/Distributions.html

ararslan:
> is that known?

Nope. Looks like it was the change to the `mad` API. I'll fix and tag.

ararslan:
Actually it looks like that was fixed on master a while back, I just need to cherry pick out a 0.10.3 tag that includes it.

tkelman:
Oh yeah, the https://github.com/JuliaLang/METADATA.jl/pull/6460 issue

ararslan:
I think I may have done the cherry picking correctly this time... (fingers crossed). We'll see what happens.

PR #6562, opened 2016-09-28
Tag SDWBA v0.0.2

ElOceanografo:
Adding some new zooplankton models


SDWBA/versions/0.0.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Docile
```
ararslan: As of Julia 0.4, both Docile and Lexicon are deprecated in favor of [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl).

PR #6564, opened 2016-09-28
Add Julia bindings for SOASTA's mPulse Query API

bluesmoon:
SOASTA uses julia to analyse performance data, and to fetch some types of aggregate data, a user needs to communicate with SOASTA's REST APIs.  This module provides julia bindings to those APIs allowing anyone with a SOASTA account to fetch their data as julia friendly data structures.


tkelman:
I'll note that there's only one currently registered package that starts with a lowercase letter, and it's old and deprecated. Module naming conventions generally start with a capital letter.

bluesmoon:
Understood, but this relates to a product whose official name (trademarked)
is spelled mPulse.

bluesmoon:
thank you

PR #6567, opened 2016-09-28
Tag DifferentialEquations v0.4.1

ChrisRackauckas:
[Parameters.jl](https://github.com/mauro3/Parameters.jl) had a breaking change to its syntax, which caused JuliaDiffEq/DifferentialEquations.jl#57. This change makes DifferentialEquations.jl compatible with Parameters 0.5.0 and updates the minimum version requirement.


tkelman:
You need to push the git tag

ChrisRackauckas:
Should be good now.

DifferentialEquations/versions/0.4.1/sha1
```diff
@@ -0,0 +1 @@
+1c4001be9f6f5ce4e1c0b0a3da98f15121f00b91
```
tkelman: this commit is 2 days old and only has a doc change relative to 0.4.0 https://github.com/JuliaDiffEq/DifferentialEquations.jl/compare/v0.4.0...1c4001be9f6f5ce4e1c0b0a3da98f15121f00b91

ChrisRackauckas:
Thanks for catching that. I wonder why it tagged that commit...

Anyhow, I tagged a new version. I checked that this tag is on the most recent commit, and it's pushed to the online repo.

PR #6569, opened 2016-09-28
Tag SFML v0.2.0 [https://github.com/zyedidia/SFML.jl]

zyedidia:


SFML/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

zyedidia:
If I add a version dependency in this PR I should also change it in the actual SFML.jl repo right? But if I make a commit there then the commit hash changes and I have to retag and possibly make a new PR here? I'm not sure how to fix this without creating a mess.

nalimilan:
Yes, but you should be able to modify the first commit, and force-push the new one to this branch. If that doesn't work, you can always open a new pull request.

tkelman:
Or you can just do it separately in the requires file for the tag in this branch of your metadata fork, and not worry about touching the tags or sha's. Then as long as you make the same change in your package's REQUIRE file before you make the next tag, it'll all work.

tkelman:
yep, like that. don't forget to make the corresponding change to https://github.com/zyedidia/SFML.jl/blob/master/REQUIRE too

PR #6570, opened 2016-09-29
Tag TensorDecompositions v0.2.0

yunjhongwu:
- Fix issues for Julia 0.5 and some minor bugs
 - Drop Julia 0.4 support 
 - Allow to provide an initial guess of factors in *candecomp*


tkelman:
FactCheck is a test-only dependency, so it should be in `test/REQUIRE` instead of `REQUIRE`

yunjhongwu:
Thanks! I moved FastCheck to `test/REQUIRE`.

PR #6571, opened 2016-09-29
ArbFloats v0.1.8

JeffreySarnoff:
cleaned up


tkelman:
There are quite a few tags being added here, you probably only want the latest?

JeffreySarnoff:
YES -- and that should make the checks pass, too

tkelman:
It looks like the merge commit might be doing you in. I bet if you checked out the unmodified upstream metadata-v2 branch, did `git checkout -b arbfloats-newbranch` then `git cherry-pick c140513de6a29d074677513235dd2ccb929d0eb3` and make a PR from that new branch it would be cleaner.

JeffreySarnoff:
OK

PR #6573, opened 2016-09-29
Register GoogleCloud [https://github.com/joshbode/GoogleCloud.jl.git]

joshbode:
New Package - GoogleCloud.jl: Google Cloud JSON APIs for Julia


tkelman:
`MbedTLS` should be explicitly listed in `REQUIRE` if you're going to do `import MbedTLS`

tkelman:
You should consider adding some tests if you can. It's possible to set secure environment variables on Travis, if that helps.

joshbode:
Thanks, I missed that! I've added MbedTLS to the REQUIRE file, and we're working on some tests.

Not sure yet whether to try to mock the Google APIs (which seems like a _lot_ of work) or try and get a perpetual basic GCP project set up to run tests against (using secure environment variables in Travis to store credentials, which I wasn't aware of, thanks!). This seems easier but might end up costing us at some point.

cc: @JockLawrie

PR #6578, opened 2016-09-29
Tag CRC v1.2.0 [https://github.com/andrewcooke/CRC.jl]

ranjanan:


CRC/versions/1.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+ArgParse
+Compat
```
tkelman: needs a minimum version

CRC/versions/1.2.0/requires
```diff
@@ -1,3 +1,2 @@
 julia 0.4
 ArgParse
-Compat
```
tkelman: it's still in `test/runtests.jl`

PR #6579, opened 2016-09-29
register and tag PkgBenchmark.jl

shashi:


ssfrr:
Thanks @shashi, this looks really cool! I've just been familiarizing myself with BenchmarkTools and thinking about setting up a reproducible benchmarking rig for some of my packages, and this looks like it'll make the process more streamlined.

KristofferC:
I guess people will have to make sure to update their packages .gitignore so users won't dirty their repos just by running the benchmarks?

shashi:
@ssfrr I'm glad you think this will be useful. :) Do open issues/feature requests if you try it.

@KristofferC by default, yes, this will write to a place inside the package. Now that you mention it, `~/.julia/v0.X/.benchmark_data/<pkg>/...` might be a better place for these?

tkelman:
Wouldn't that make `Pkg.status` show a dirty non-repo at `Pkg.dir(".benchmark_data")` ? Why not make a package-scaffold auto-generation function that creates a template `benchmarks` folder including a stock gitignore?

PR #6581, opened 2016-09-29
Register and tag ClobberingReload.jl

cstjean:
[Provides](https://github.com/cstjean/ClobberingReload.jl):

- A `reload` alternative that overwrites the existing module instead of replacing it
- `Autoreload` functionality that works on 0.5
- Redefinition warning suppression

Incidentally, the travis build works on `julia: nightly` but not on `julia: release`, because the latter still refers to 0.4.7. Is that something I need to change somewhere?


tkelman:
use `julia: 0.5`

PR #6584, opened 2016-09-29
Only check SHA matches for modified package versions

ararslan:
Currently when all tags have been pushed to the upstream package repository, we check that the SHA for _every_ tag matches between what's in METADATA and what's in the upstream repo. With this change, we'll only be checking whether the SHA matches for versions that were modified in the PR.

This also includes a fix that will ignore tags in the upstream repo with names that aren't valid version numbers. It's fine to have those upstream, just not in METADATA. Attempting to create a version in METADATA with an invalid version number will still fail.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -93,10 +93,11 @@ for pkg in keys(modified)
 
     if isempty(notpushed)
         # Tags match, so we can compare SHAs 1-1
-        for tag in keys(localtags)
+        for tag in modified[pkg]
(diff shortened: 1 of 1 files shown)


tkelman:
how did this hit a `KeyError: key v"0.0.3" not found` ? whose 0.0.3?

ararslan:
I have absolutely 100% no clue.

ararslan:
That appears to have been a fluke? ¯\\\_(ツ)\_/¯

PR #6587, opened 2016-09-30
Tag LightGraphs v0.7.2 [https://github.com/JuliaGraphs/LightGraphs.jl]

sbromberger:


tkelman:
It's strongly discouraged to extend Base methods on signatures consisting only of Base types as in https://github.com/JuliaGraphs/LightGraphs.jl/blob/11b6a653e9b6690ada4fa53513ec9452f215cfd7/src/linalg/graphmatrices.jl#L136-L150 and https://github.com/JuliaGraphs/LightGraphs.jl/blob/11b6a653e9b6690ada4fa53513ec9452f215cfd7/src/linalg/graphmatrices.jl#L258, as that changes the behavior of completely unrelated code depending on whether or not LightGraphs has been imported anywhere else.

sbromberger:
cc @jpfairbanks

PR #6588, opened 2016-09-30
Tag Gumbo v0.2.2

porterjamesj:


Gumbo/versions/0.2.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4 0.6-
+BinDeps
+Compat
```
tkelman: what's the minimum version that implements features you need?

Gumbo/versions/0.2.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4 0.6-
```
tkelman: Upper bounds are not effective unless applied on all past tags. all this will do is prevent this version from being installed on 0.6-dev, so there users will be held back to version 0.2.1.

porterjamesj: oh good point. bleh, ok i will remove this. sad that past me did not have enough foresight to have been using upper bounds all along but i am too lazy to go back and update everything now.

PR #6592, opened 2016-09-30
Tag JSON v0.7.1 [https://github.com/JuliaLang/JSON.jl]

TotalVerb:
Only user-visible change is printing `Float32` in accordance with JSON standard on v0.5; see https://github.com/JuliaIO/JSON.jl/pull/172.


TotalVerb:
Second commit changes repository URL to JuliaIO.

PR #6595, opened 2016-09-30
TypedDelegation (new package)

JeffreySarnoff:
Delegation macros, easily used.
Apply functions over the values given by fields of constructed types.
Operate with field-valued operands, yielding a result of the same type.


JeffreySarnoff:
the difficulty was with the metadata (bad urls)
@tkelman this should be good .. there is very little source text!

TypedDelegation/versions/0.1.1/requires
```diff
@@ -0,0 +1 @@
+julia v0.5
```
tkelman: the `v` doesn't hurt (I think) but it's a bit nonstandard here

tkelman:
Enable Travis CI?

PR #6598, opened 2016-09-30
Tag PhyloNetworks v0.3.0

cecileane:
we wish to tag a version before making updates to work with Julia v0.5.


PhyloNetworks/versions/0.3.0/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4
+Compat
```
tkelman: it's best to declare a minimum version dependency that implements the features you need

cecileane: Many thanks! Added minimum version dependency for packages, and restricted to Julia 0.4.x . The same changes were made to the package [commit 6c5642d](https://github.com/crsl4/PhyloNetworks.jl/commit/6c5642dabbe6754b59ab943162d8bf4cf6dc59d8). The hash in METADATA/PhyloNetworks/versions/0.3.0/ is then one commit behind. I assume it's okay (after reading [this](http://docs.julialang.org/en/release-0.5/manual/packages/#fixing-package-requirements)), but I'm happy to be told otherwise and to change whatever is needed.

PhyloNetworks/versions/0.3.0/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4 0.5-
```
tkelman: applying uppe bounds isn't effective unless it's done on all past tags. the result of this will be users on julia 0.5 or 0.6-dev will be held back to the previous tag

cecileane: Do you recommend getting rid of the 0.5- upper bound? I am inclined to add the upper bound to all previous past tags.

Our plan is to continue development on the master branch with a requirement of `julia 0.5 0.6-`, and to maintain a version for julia 0.4 on a branch "julia-v0.4", this one requiring `julia 0.4 0.5-`. 

Does that sound reasonable, or do you recommend getting rid of all upper bounds on all tags?

tkelman: Until you have a tag available that supports 0.5, I would avoid adding upper bounds. It would make the package not installable on 0.5 or 0.6-dev, and would be kind of messy to deal with if anyone has installed it there yet. I think it's preferable to allow installation of the latest tag, even if it's not completely working yet.

cecileane: Many thanks Tony. I removed the upper bound in METADATA and in [PhyloNetworks](https://github.com/crsl4/PhyloNetworks.jl/commit/1a7da765f13cc69334d2b94f6870d294aa0dfdc0). I really appreciate your feedback here. You are giving so much to the community.

tkelman: Aww, you're too kind! Always impressed with what people are building using Julia.

PR #6601, opened 2016-10-01
Tag Wavelets v0.5.2 [https://github.com/JuliaDSP/Wavelets.jl]

gummif:


Wavelets/versions/0.5.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Reexport
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

gummif: Fixed

PR #6602, opened 2016-10-02
Register SJulia [https://github.com/jlapeyre/SJulia.git]: v0.0.13

jlapeyre:


tkelman:
It's recommended not to include Julia in package names - http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package

ViralBShah:
As a package, I agree that having Julia in the name doesn't help given the `.jl` suffix. Also, even if `Julia` shouldbe in the name, I feel that it should at least be `SymJulia`. The project is really interesting in my mind that it could even have its own name without the words `Sym` or `Julia` in the name.

Just some thoughts. Happy to also go with SJulia if that is what the authors want, but just requesting a short discussion once.

jlapeyre:
I would like a better name, without "Julia", if possible. I'll think about it. If we change the name, now is the time to do it.

*SymJulia* is the natural improvement, but its not great. I'm open to suggestions. Maybe bikeshedding on the user list is a good idea. But, picking [a good name]( https://en.wikipedia.org/wiki/List_of_computer_algebra_systems) for  a project is harder than picking a color, plus, it's easy to repaint.

We could name it after a person, *chyrch* for Alonzo Church. Or use *sym* somehow to connote the purpose. Google turns up nothing for *symjay*.

ViralBShah:
`Alonzo.jl`?

jlapeyre:
I like `Alonzo.jl`. It is simple, short, easy to remember.

Also `symanip` is not used. I translated *mathematics* and *symbol* in all of google's languages and found some interesting words and ideas. `symata` is interesting. So is `symmata`, which is less common on google.

nalimilan:
Or the boring-but-explicit SymbolicMath.jl?

jlapeyre:
@nalimilan The problem with this is that the name will be referred to repeatedly in documentation, and I think this could get cumbersome. `SJulia` functions as a standalone language (although it can also be used from within `julia`.)  So, it would have to be  "`SymbolicMath` does this, `SymbolicMath` does that."

MikeInnes:
How about Scythe.jl – vaguely sounds like Sym

nalimilan:
SymMath.jl then? Repeating SJulia in the docs isn't great either, as only one letter in that name actually gives information about what package you're talking about. :-)

ViralBShah:
I still like Alonzo.

jlapeyre:
I am leaning towards `Symata.jl` and `> symata`. I also still like  `Alonzo` , it fits well with tradition and is not used in software, or much in trademarks in general. Googling `symata` shows it to be much less common, of course. `6500` hits. `symmata` has no hits, but I like the shorter version. There is a design company that sells through etsy called `symata`. One reason I thought of naming it after a mathematician/computer scientist is that names like `symata` can easily get lost in a forest of similar names, like all the cars a few years ago with names like `aventa`.  I could also take the python route and name it after my favorite musician, ie `Ribot.jl`, which also happens to mean "carpenter's plane" in Catalan.

I originally had `SymJulia`, but I switched to `SJulia` because it is less typing in code and docs. Both of these were just default names, nothing considered. There are some packages called `Xja` for java. We could do `Symju`, but I don't see any reason to reference the host language.

@nalimilan I absolutely agree that repeating `SJulia` in the docs is not good. All the suggested names are an improvement on this.

`SymMath` and `SymMath` are already used in a few mathematics software projects, as well as other things.

@MikeInnes  `Scythe.jl` is nice.  Of couse, `symata` is more google-unique... I suppose online scythe sales are not growing though.

MikeInnes:
This is one of the nice things about the .jl convention – you can call it X for short (where X is some pseudo-random 1/2 syllable noun), but people can search for X.jl and it'll be the first hit.

Ribot.jl is a nice name too.

jlapeyre:
@MikeInnes  Wow, I didn't realize how well this works.  `R.jl` does not even exist, but googling it turns up R-related Julia packages. This makes `Alonzo.jl` and `Scythe.jl` more attractive, at least to the medium term. But, since the package is also a language, I can imagine at some point distributing it standalone, rather than as a Julia package.  Still, "x math" or something should always work for these, even though Church was a mathematician/logician.

I am strongly influenced by imagining typing a name and seeing it on the command line, and seeing it appear in code. And also, by having the package name the same as the one that used to invoke a REPL. I am trying to resist this, thought, because this is irrelevant to 99 percent of users, especially since the language is specifically meant to target non-programmers.

tkelman:
Symata.jl sounds neat to me. Logo possibilities of a Julia-colors fermata over some mathematical expression?

jlapeyre:
The logo idea sounds great. I'd have to learn a bit more about graphic design.

musm:
@jlapeyre my suggestion goto http://www.thinkbabynames.com/ 😄   use some sort of random scheme to select an arbitrary name that sounds interesting to you,  don't ask for opinions on said chosen name (there will be objections) rename package, move on with more important things... (case in point the Julia name)

jlapeyre:
@musm. I was hoping I would hear one name and say, "that's it!" Consensus is out of the question, of course.

But, that's good advice: the search is done. `Symata.jl` is the new name. Thanks to all.

ararslan:
Great work on this package, and excellent choice of name!

jlapeyre:
I think everything is done on my end.

jlapeyre:
Till now methods in `Base` were overwritten, eg. `show` for `Rational` and `Symbol`. This is fixed [here](https://github.com/jlapeyre/Symata.jl/commit/936c9f2140dfbcb8aeef6a4d5ee5dc3412d2795c#diff-02c799ef097a422d24361aa323ebf8f8R96).  I will wait to see if there are more problems before tagging Symata again and modifying `METADATA/Symata`.

tkelman:
You still have a few mentions of SJulia in your tests (and some `Pkg.dir`s that should be replaced with `dirname(@__FILE__)`, or `@__DIR__` from the latest Compat version), but those can be fixed up in future tags. Not sure why that didn't cause your tests to fail?

jlapeyre:
I will fix those, now. It should be easy. I thought I grepped the entire source with `-i sjulia`...

jlapeyre:
I found two `Pkg.dir`s and changed them to `dirname(@__FILE__)`.  There were mentions of SJulia in the old, unused tests in `./test`. I removed all of these files. There is one file remaining in `./test` that runs the tests in `./sjtest`.

PR #6603, opened 2016-10-02
Register AMD [https://github.com/JuliaSmoothOptimizers/AMD.jl.git]: v…

dpo:
…0.0.1


dpo:
```
METADATA/AMD/versions/0.0.1/requires: oldest allowed julia version 0.3.0 too old (>= 0.4.0 needed)
```
Is it bad to also support 0.3?

tkelman:
Metadata is frozen for 0.3 supporting packages. Please use a more descriptive name for this.

tkelman:
Or given that this is a part of SuiteSparse, add it to SuiteSparse.jl.

dpo:
It's useful on its own and has a very small footprint. AMD predates SuiteSparse. AMD's the name. I don't think it should be changed.

tkelman:
AMD is a name that violates the Julia package naming guidelines, especially as it has nothing to do with the chip maker.

When was AMD last distributed independently of SuiteSparse? You're using SuiteSparse's copy, and when SuiteSparse gets removed from base this will break and you will have to depend on SuiteSparse.jl to get the library anyway.

dpo:
AMD has been around since the mid 1990s. I only use SuiteSparse's copy because it's currently part of Julia and doesn't require compiling anything else. When SuiteSparse gets removed, I might depend on SuiteSparse.jl (that project hasn't been updated in months), or I might go with http://www.netlib.org/linalg/amd, which has far smaller footprint.

It's called AMD everywhere in software and literature, not ApproximateMinimumDegree or anything else. Anyone knowing a minimum of numerical linear algebra will look for it under that name. That's what Matlab also calls it. A glance at the README makes it clear that the chip maker has nothing to do whatsoever in the story (what would such a package do, really?). I didn't come up with "AMD" as an abbreviation. The [package naming guidelines](http://docs.julialang.org/en/release-0.5/manual/packages/#guidelines-for-naming-a-package), item 6, says that

> Packages that wrap external libraries or programs should be named after those libraries or programs.
CPLEX.jl wraps the CPLEX library, which can be identified easily in a web search.
MATLAB.jl provides an interface to call the MATLAB engine from within Julia.

Thus AMD.jl wraps the AMD library. It's not jargon. Thus I don't think the name violates the guidelines.

ViralBShah:
I completely agree about the AMD algorithm and SuiteSparse only happens to have a high quality relatively recent implementation, but by no means is the only one. However, for someone looking at the package repository, it is absolutely not clear if this package is doing stuff related to the AMD CPUs or the AMD algorithm (if they know of it). This is the challenge with short package names in general, even though this is a completely legitimate name.

For the dependency on SuiteSparse, I agree that the strategy is right. Depend on SuiteSparse.jl when it is no longer in Base, but no point worrying about it until then.

Can we use the word `sparse` somewhere in the package name or find some small way to make the name clearer to a casual reader?

ViralBShah:
`LibAMD` is another possible name that clearly refers to a library called AMD.

tkelman:
The Netlib copy has a pretty terrible license:

> NOTICE:  "The AMD routines (AMDEXA, AMDBAR, AMDHAF, AMDHAT, AMDTRU,
> and AMDATR) may be used SOLELY for educational, research, and
> benchmarking purposes by non-profit organizations and the U.S.
> government.  Commercial and other organizations may make use of the
> AMD routines SOLELY for benchmarking purposes only.

@Sacha0 has also implemented a pure-Julia version at https://github.com/Sacha0/ApproxMinimumDegree.jl, which is a much clearer name. The TOMS paper starts by calling it AMD, but most other places it's clearly used as an acronym, the other papers all spell it out.

ViralBShah:
Many of the older papers had these cryptic names for FORTRAN reasons, which have luckily gone away.

dpo:
Forget it.

StefanKarpinski:
@dpo, please don't get discouraged by a constructive discussion of library naming. Computing is a crowded space and calling a library `AMD.jl` is likely to cause a fair amount of confusion. Is the suggestion of calling this package `LibAMD.jl` not acceptable to you?

dpo:
@StefanKarpinski I appreciate your and @ViralBShah's comments. To be frank, I'm having a difficult time finding the discussion constructive or useful. As far as I can tell, the package name follows the guidelines as it wraps a very well-known linear algebra library by the same name. That it could be confused with something having to do with the chip maker is staggering to me. At that rate, `Convex.jl` could be mistaken for some revival of the Convex Computer Corporation and `JuMP.jl` should be rejected for not having a particularly clear name (not to pick on those great packages, but only for the purpose of illustration).

Over at JuliaSparse, I wrote `HarwellRutherfordBoeing.jl`. Should that also be changed to `HarwellRutherfordBoeingTheMatrixStorageFormatsNotTheResearchLabsOrTheAerospaceCompany.jl`?

I'm not specifically opposed to `LibAMD.jl` but I don't see how that couldn't be a library related to the chip maker. I believe `AMD.jl` is the right name since that's the name of the underlying library and the name under which people know it. I am troubled by the kind of criticism that flies around here. It may not have been intended that way, but it sure appears gratuitous from the receiving end. Erroneous sweeping claims about naming convention violations and arguments about the licensing of some imaginary potential future version of this package are not just frustrating, they're arguments for the heck of it.

That said, I very much appreciate the hard work put in by everyone who develops Julia, and I'd like to offer (hopefully) constructive comments from a user's perspective.

1. It seems to me that it would be useful to consider both the package name and the one-line description on the Github page when assessing a package. They could both appear in the commit and `PkgDev` might offer facilities for obtaining the one-line description of a package before `add`ing it (perhaps it already offers such facilities; I'm not sure). Packages could also fit in categories, e.g., linalg/AMD.jl or be tagged.
2. I've registered packages for Atom in the recent past and over there, nobody cares what you register (provided it's not offensive, I imagine). Registration is done from the command line and is automatic. It's simple, fast and efficient. The good packages survive, the rest sinks to the bottom. I wrote a package to debug Python code. There already existed one named `atom-python-debugger`. I named mine `python-debugger`. Soon enough, my package had many more downloads than the previously existing one despite the potential naming confusion. Users are not stupid and don't blindly install packages without looking into them. I'm not saying the Atom system is ideal but there's something to be said for the simplicity and open-mindedness of it.

ViralBShah:
`HarwellRutherfordBoeing.jl` is a perfectly fine name, and would not be as descriptive IMO as `HRB.jl`. Also, I am sure you agree that `AMD` is a better recognizable name than `Convex Computer Corporation`. 

Package namespaces are definitely planned for and we will no longer need to have naming discussions when we have that.

ViralBShah:
Also, my request was only to reconsider the name `AMD.jl` for two reasons - one reason for a library by the chip maker and the second reason being multiple AMD implementations as you yourself pointed out.

Given the multiple AMD implementations, it seems that at least some indication that this was based on SuiteSparse is what my request is, given that you pointed out that we may have a wrapper for a Netlib amd implementation in the future. `SuiteSparseAMD` would be better than `LibAMD` for that reason. 

However, I am ok with `AMD.jl` if that is what you really want.

ViralBShah:
@tkelman Even if an upstream library has a terrible license, we have package wrappers for all kinds of libraries - open source, strange licenses, proprietary, etc. I am not sure why that is relevant.

ViralBShah:
@dpo I feel that this discussion is frustrating for you - and I personally would like to say that I love the stuff you are doing, and I personally wanted these wrappers for a long time. I may even have filed an issue for it sometime back or said something on the mailing list.

Hence, I am one of those who would be most pained to see this not getting merged.

dpo:
@ViralBShah It's good to hear that namespaces are in the cards. The AMD that's on Netlib is Fortran instead of C and implements a few variants. It's otherwise the same method as the one in SuiteSparse. I'm not claiming I'll interface that in the future; I was just responding to the first far-fetched criticism. Adding to SuiteSparse.jl makes sense but I can also see that AMD by itself, having such small footprint, is useful. Personally, I use AMD but rarely use the rest of SuiteSparse. If licensing is such an issue, I'm happy to ask Tim Davis nicely if he would relicense it.

I don't mean to dwell on the chip maker argument, and I appreciate your effort to find a middle ground. But I honestly don't see that the same criticism can't be used against `LibAMD.jl`.

I'm not trying to be inflexible. If there's a legitimate objection, I'll take it. If not, I'm happy to get rid of 0.3 support. I'll let you decide and reopen the PR if you feel like it.

ViralBShah:
Ok - let's go with it.

tkelman:
The company named AMD isn't the main point, it's how understandable will the name be for a non expert user. I think you're overestimating how well known AMD is, particularly among users who don't have a graduate level background in linear algebra.

The name is the hardest thing to change about a package after it's been registered, so there's always some discussion whether the name makes sense. Once namespaces are implemented, personal registries can be anything goes, but for now the namespace is shared and global.

ViralBShah:
I think we have discussed this a lot. All the points are made - and we'll go with @dpo 's final decision and of course whatever is necessary to make this green.

dpo:
Doctoring up the PR didn't work. I'll open a new one.

PR #6604, opened 2016-10-02
Register Profiles [https://github.com/JuliaSmoothOptimizers/Profiles.…

dpo:
…jl.git]: v0.0.1


tkelman:
Another not-very-descriptive name here. PerformanceProfiles? SolverProfiles? ComparisonProfiles? MoreProfiles? It's a pretty specific A-vs-B comparison that these refer to.

And what's the license on the original code, if these are direct translations?

ViralBShah:
I had the same question when I saw the this package. @dpo I understand this nitpicking can be frustrating, but it helps make the packages descriptive and visible and gain more users.

PR #6607, opened 2016-10-02
Row echelon

blegat:
Create a package for `rref` (see https://github.com/JuliaLang/julia/pull/9804).
I also added support for Rational types, tests and and a help message.


ararslan:
It would be nice to preserve copywrite information regarding the authors of the origin code in the license, and even better, preserve the Git authorship information for the code, which is possible (albeit not easy).

As an aside, it looks like you're branching on the type of the input--better to use dispatch, i.e. separate methods for different types.

Other than those things, looks okay to me.

tkelman:
Looks like the original version is almost entirely written by Jeff - https://github.com/JuliaLang/julia/blame/d47926a3a02157f4c3a7efdd99d0d00b91f78628/base/linalg.jl#L442-L474 with only minor adjustments / moves of the function to different files by other people later. What we've done in other cases is used the same license as Julia but say the code was formerly part of Julia, and add yourself as an ongoing maintainer.

blegat:
@tkelman Do you have a package in mind that does that (just be sure of what I should write) ?

ararslan:
@blegat [Git.jl](https://github.com/JuliaPackaging/Git.jl)

PR #6612, opened 2016-10-02
Tag PairwiseListMatrices v0.4.0

diegozea:
It supports Julia 0.5.


tkelman:
Can you push the git tag so it's visible under https://github.com/diegozea/PairwiseListMatrices.jl/releases ?

diegozea:
Thanks, I did `git push --tags` and now the tag is visible under releases.

PR #6616, opened 2016-10-03
Tag OffsetArrays v0.2.12

alsam:
Expanded on use case examples:
1. `scalar_law/PROGRAM1`
2. `acoustics_1d_example1`
updated timings for `scalar_law/PROGRAM1` vs. Fortran version.


tkelman:
Do you need to check in a copy of clawpack or code translated from it? That has its own license, "research and instructional use only" is quite a bit stricter than your MIT license.

alsam:
Thanks @tkelman for pointing to the license issue. Some time ago I contacted with Professor Randall J. LeVeque about he opportunity to translate clawpack Fortran codes to Julia. He kindly got me written permission in email message on translation of the codes and clarified license issues, his answer on my question may I translate and copy the codes:

> Certainly!  The BSD license allows unlimited copying and modifying of
> the code (with some attribution to the original) and our hope is that
> people will extend and reuse it in multiple ways.

The developers of the clawpack were very excited on the opportunity of their code be translated to Julia.

The clawpack codes are distributed under BSD license:
[The Clawpack 5 license](http://www.clawpack.org/license.html)

I should have added a reference to the license. I'll add the reference. 
Anyway if you have doubts I might move the original `acoustics1d` Fortran clawpack code and `Julia` translated one from `OffsetArrays` to another package. 
What is your opinion?

tkelman:
Oh if it's all BSD then yeah no problem, as long as you include proper attribution. The "research and instructional use only" wording struck me as unusual.

tkelman:
Looks like the scalar_law examples are by a different author and similarly have a research-only clause on them. If these are just examples and you're not using them as part of the test suite for your code, then I think it would be better from a license perspective to have them in a separate repo.

alsam:
You are quite right: `scalar_law` examples are distributed under another license. The examples are given from the book [Numerical Solution of Hyperbolic Partial Differential Equations by John A. Trangenstein](https://www.amazon.com/Numerical-Solution-Hyperbolic-Differential-Equations/dp/052187727X). To be on a safe side I also contacted with author and got an email permission for translation and publishing.
Below are excerpts from Prof. Trangenstein answers:
>Thank you for your email. It is nice that you asked for permission.
...
> Generally, I support what you have done, but I would like to know a bit more about which of my routines you plan to post on github. Could you send me a brief description of the codes from the book that you would like to post?
...
> There may be licensing problems with the publisher, so I need to check with Cambridge University Press. They are generally pretty quick to respond.
...
> Cambridge U Press approved your posting to github. See below. Sorry it took so long.
...

Despite the fact that I have a written permission to publish the codes, you are quite right, that the license is different, so I'm ready to remove `scalar_law` from `OffsetArrays` and put it into another repository. On the other hand I'm using `scalar_law/PROGRAM0` and `scalar_law/PROGRAM1` for comparison purposes: the performance of the original Fortran code vs. translated Julia one. Ok, I will move both `scalar_law` and `acoustics1d` to different package, say `OffsetArrays_examples.jl`. Is it ok?
Thanks.

tkelman:
Great that you asked for and got permission, you could reference that you received that permission next to the license statement, and add a note in your top-level license file. Or moving the code to a different repo sounds fine too. Unless they are part of your unit / correctness tests, I don't think you necessarily need examples to be within the same repo.

alsam:
Ok, thank you. Certainly I will move the examples to another repo (or two repos, as examples are from the different sources) with all references about licenses and received permissions.

alsam:
Oops, didn't move the tag in METADATA, just updated `sha1`, sorry.

alsam:
Hello @tkelman, I moved all the examples out of this repository into the two different repositories. Sorry it took too much time - was extremely busy at work.
Is it ok? Thanks!

PR #6618, opened 2016-10-03
Tag all the things

MikeInnes:


ararslan:
Looks like the Atom tests are failing because it can't find `@init`. I think that's something that came from Requires, which was removed as a dependency of Atom after the previous release.

MikeInnes:
Which is also fixed by this PR :) The latest Lazy.jl has `@init` for now so once this is merged the tests will pass.

ararslan:
Awesome 👍

tkelman:
Enable Travis on Media.jl. Given https://github.com/JunoLab/Media.jl/issues/4 it won't be testing much, but will at least verify that the package loads.

tkelman:
If functions have been moved around and these won't work with earlier versions, then the dependencies in REQUIRE need lower bounds

CodeTools/versions/0.4.2/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.5-
-Lazy
+Lazy 0.11.3
```
tkelman: codetools and mux don't use `@init` do they? the others should have this change made in the package REQUIRE too before the next tag

PR #6619, opened 2016-10-03
Tag FileIO v0.1.3 [https://github.com/JuliaIO/FileIO.jl]

timholy:
Changes:
- add missing BigTIFF magic bytes (@romainFr)
- prevent StackOverflow if `load`/`save` is not defined by a module (@timholy)

### DIFF
--- a/FileIO/versions/0.1.3/requires
+++ b/FileIO/versions/0.1.3/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.7.19
--- a/FileIO/versions/0.1.3/sha1
+++ b/FileIO/versions/0.1.3/sha1
@@ -1,0 +1,1 @@
+c38ef5ab4176f6f5c642caccd22d69f76e9d08fd


tkelman:
@alyst looks like RData is having trouble with this? e.g. http://pkg.julialang.org/logs/FreqTables_0.4.log

tkelman:
ImagineFormat too http://pkg.julialang.org/logs/ImagineFormat_0.4.log

alyst:
It seems like the problem is that `RData` exports `load`, and then the new 0.1.3 `FileIO` check fails, because `RData.load == FileIO.load`.
I'm not sure how to properly fix that. I would like to preserve the current behaviour: `using RData` makes `FileIO.load(filename::String)` available for the user. Maybe `FileIO` check could be a little bit more elaborated and look for the existence of the specific signature?

timholy:
This might be fixed by https://github.com/JuliaIO/FileIO.jl/pull/84. There are some downsides to this approach, unfortunately---this is getting to be quite a hack. What are the odds of forcing users to say `using FileIO`? I recognize that may not be ideal, either.

Cleanest would be to get https://github.com/JuliaLang/julia/issues/14472.

PR #6622, opened 2016-10-03
Register Tensors [https://github.com/lanaperisa/Tensors.jl.git]: v0.0.1

lanaperisa:


tkelman:
This looks neat, did you change your mind about wanting to register it? How does it compare to @Jutho's https://github.com/Jutho/TensorOperations.jl ?

Jutho:
Maybe it would be good to add a little more description of what this package can do in the README. If the focus is on the tucker format, maybe a less generic package name would be favoured? I am a bit skeptical about reserving a generic name for a package with a specific functionality.

lanaperisa:
I will not register it for now, but after I write more detailed README and I will definitely consider choosing a bit more specified name. Thanks for your comments!

PR #6624, opened 2016-10-03
Tag PGFPlots v1.4.0 and BayesNets v1.0.5 

tawheeler:
THe I-don't-know-how-to-submit-separate-PRs-for-this edition.

PGFPlots now supports error bars.
BayesNets now works on Distributions v0.11.0.

I will delete https://github.com/JuliaLang/METADATA.jl/pull/6620


ararslan:
Currently making multiple disjoint METADATA pull requests for multiple packages at once is kind of a pain. For reference, you can remove the other commit of the other package as follows:
```bash
cd ~/.julia/METADATA
git checkout -b no-pgfplots
git rebase -i HEAD~2
# Your default editor ($EDITOR) will be opened, then you can remove the line
# corresponding to PGFPlots. This removes the commit from the current branch.
git push tawheeler no-pgfplots:pull-request/a1fbf64c --force
```
That's how I like to do it, anyway.

To avoid grouping them in the first place, this is what I do: I start with my `PkgDev` commands, e.g. `tag` and `publish`, then open the PR for package A. Then I do
```bash
cd ~/.julia/METADATA
git checkout -b tag-package-A
git checkout metadata-v2
git reset --hard origin/metadata-v2
```
This resets METADATA to a state that exactly matches what's upstream. This avoids non-fast-forward merges when doing `Pkg.update()`, plus it allows you to make changes to packages in separate PRs.

All of this could be easily remedied if PkgDev checked out a separate branch before committing changes...

PR #6626, opened 2016-10-03
Initial DataStreams 0.1 release with CSV, SQLite, Feather, and ODBC

quinnj:


quinnj:
A little synopsis here:

* Bumping DataStreams to 0.1, with julia 0.5 requirement
* Bumping all other packages to rely on DataStreams 0.1.0 and julia 0.5 as well

tkelman:
Is DataStreams 0.1 breaking for past versions of everything else? It's not ideal to do these all together (especially by hand - at least use PkgDev.tag even if publish doesn't work?) as Travis fails for everything downstream right now.

quinnj:
Yes, this is all breaking. I've done several revisions/iterations on the DataStreams API over the last few weeks to really get it ot a place where it can be stable for a good while. In that process, there have been several name changes and shifts in functionality to get as much into DataStreams as possible and allow packages to need to implement as little as possible.

I'd really rather tag everything at once and then be able to do new tags for individual packages as needed.

tkelman:
Sure you want to add a 70,000 line csv file to ODBC.jl?

quinnj:
Yes, it's important to test the "multi-fetch" behavior of ODBC when querying. There is a cutoff where it will retrieve a maximum of 65535 rows at a time, so it's important to test the code branches where multiple fetches are assumed.

PR #6630, opened 2016-10-04
Pull request/917cb59f

MikeInnes:


Requires/versions/0.2.3/sha1
```diff
@@ -0,0 +1 @@
+f89031757bc0ebac8d052f3fd8b401006d1b17fb
```
tkelman: update the git tag https://github.com/MikeInnes/Requires.jl/releases

PR #6635, opened 2016-10-04
Register Suppressor [https://github.com/Ismael-VC/Suppressor.jl.git]:…

Ismael-VC:
… v0.0.1, v0.0.2


Suppressor/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should really have a minimum version requirement here that implements the features you need

tkelman:
Dunno if this was intended or an accident, but the 0.0.1 and 0.0.2 tags here point to the same commit.

tkelman:
Should also write some tests and enable CI. Some methods of redirecting output are very susceptible to freezing on Windows for example. Not sure whether that includes what you're doing here, but you should test it.

Ismael-VC:
> Dunno if this was intended or an accident, but the 0.0.1 and 0.0.2 tags here point to the same commit.

I'm having trouble finding a way to test this, but I'll do it.

The double tagging the same commit was an unintended mistake, do you know how could I undo the last tag, I've not found any reference on that.

Ismael-VC:
superseded by https://github.com/JuliaLang/METADATA.jl/pull/6659

PR #6644, opened 2016-10-05
register PooledArrays 0.0.1

JeffBezanson:
This implements a compressed representation of arrays with few unique values. It's designed to have the exact same behavior as `Array` except for performance / memory footprint.


ararslan:
Can you push the tag to the upstream repo so it shows up under Releases? Ref https://travis-ci.org/JuliaLang/METADATA.jl/builds/165389876#L194.

JeffBezanson:
Oops. Done.

nalimilan:
Just FYI, sounds very similar to https://github.com/JuliaArrays/IndirectArrays.jl.

JeffBezanson:
Yes, true. The main extra feature I want here is that the pool is sorted, giving fast `sort` and `sortperm`. Mutation is allowed, and sortedness is maintained when new items are added. Maybe it should be called `SortedPoolArrays`?

nalimilan:
No idea. I guess it depends whether you plan to make it more general at some point.

timholy:
I would definitely call this different enough to merit a different package. I'll add a link to this from IndirectArrays. (Might be good to do the converse, too.)

tkelman:
Turn Travis on here.

PR #6645, opened 2016-10-05
Register and tag Ranges

timholy:
This package is essentially a way of backporting https://github.com/JuliaLang/julia/pull/18777 to julia 0.5, and supporting the creation of ranges of objects like unitful-quantities, vectors, etc, that are not currently supported by Base.


timholy:
OK to merge?

PR #6647, opened 2016-10-06
Tag Fontconfig v0.0.3

tlnagy:


tlnagy:
dcjones/Fontconfig.jl#5 is the issue tracking part of the problems here.

Fontconfig/versions/0.0.3/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.3
```
tlnagy: Thanks, I also ran into some issues with tests on OS X so I'm going to patch all of that first.

PR #6636, opened 2016-10-05
Tag Bio v0.4.2 [https://github.com/BioJulia/Bio.jl]

bicycle1885:
* Bug fix
* Require Julia 0.5 (not a development version)

PR #6649, opened 2016-10-06
Ignore upstream tags that don't start with v

ararslan:
This should address the issue noted in #6636.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -5,7 +5,7 @@ function get_remote_tags(url)
     lines = split(ls, "\n")
 
     filter!(lines) do line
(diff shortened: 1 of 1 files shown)


tkelman:
could go in the regex, couldn't it?

ararslan:
I'd have to reproduce `Base.VERSION_REGEX` but change `v?` to `v`... This is the approach that PkgDev uses, which is why I chose it. I can do whatever's best though.

tkelman:
I meant the `r"^\w+\trefs/tags/([^\^]+)(\^{})?$"` regex

ararslan:
Oh, that's a good idea. I'll do that.

PR #6650, opened 2016-10-06
register FixedSizeStrings 0.0.1

JeffBezanson:
I thought a package for this might already exist, but I couldn't find one. This is very handy e.g. for table columns of 2-character ASCII strings.


tkelman:
Turn Travis on. Does this pass tests on 0.4?

JeffBezanson:
Done. And yes it does; I tried it.

PR #6653, opened 2016-10-06
Register KSVD.jl

IshitaTakeshi:


JeffBezanson:
Looks like you need to push the version tag.

PR #6654, opened 2016-10-06
Register Losses [https://github.com/JuliaML/Losses.jl.git]

Evizero:
The package is a result of a collaboration to create a comprehensive collection of loss functions used in machine learning

The core difference to similar (registered) packages are:
- a large number of queryable properties (e.g. `isconvex(myloss)`)
- dispatchable on being margin- or distance- based (classification vs regression)
- very flexible concerning the types of the parameters

cc: @tbreloff @ahwillia @joshday


tkelman:
> collection of loss functions

Should this be called LossFunctions.jl then?

ahwillia:
👍  I actually like LossFunctions better.

tbreloff:
Hmm. It clashes with the convention of Penalties/Transformations but agrees
with ObjectiveFunctions. I'd prefer to keep the name mostly out of
laziness, but I won't put up a huge fight.

joshday:
This package provides subtypes of `Loss`, and not a collection of functions (i.e. this is not like `StatsFuns`).  I would say `Losses` is the more appropriate name.  

From the Julia docs:
> Packages that provide most of their functionality in association with a new type should have pluralized names.

tkelman:
Without a little bit of application specific context, "Loss" isn't all that meaningful.

ahwillia:
"LossFunction" is very verbose though...

colintbowers:
I had no imminent plans to register LossFunctions.jl, so if you prefer that name then please go with it. Sounds like you've done a fair bit more work than in mine anyway...

Evizero:
Well the most consistent name would be `MLLosses`, but the double L is evil and must not gain a foothold in this world.

I do see that maybe `Losses` is uninformative for domain outsiders, but to be fair I can't quite think of a term collision that would make the name ambiguous. Plus it follows the convention since the main type is `Loss`

The problem with `LossFunctions` is that it is not really true. `LossFunctors` would be more appropriate in that regard, but even uglier.

Any other name ideas?

ahwillia:
I still like LossFunctions.jl even if it breaks the convention mentioned by @joshday - it is more explicit. But I think we should keep `Loss` as the main type exported by the package.

> The problem with LossFunctions is that it is not really true. LossFunctors would be more appropriate in that regard, but even uglier.
 
Not sure that I follow. You mean that our `Loss` isn't a function? Or that everyone calling them "loss functions" is technically wrong? Either way, I don't think this is fatal to naming the package LossFunctions.jl

All this being said - I'm still fine with Losses.jl unless @tkelman seconds me with strong feelings.

Evizero:
> Not sure that I follow. You mean that our Loss isn't a function? 

That our losses are types. Personally I am ok with pretty much all proposed solution. Thinking about it a little I have to say `MLLosses` is growing on me for consistency reasons

ahwillia:
Well then Penalties would have to become MLPenalties? I'm not a big fan of
the double L in MLLosses. Plus these loss functions are entirely usable
outside of machine learning -- and this should be encouraged!

kmsquire:
I think the term `Function` as used above, is confusing two different meanings of the word. Losses generally are functions in the mathematical sense, even if they're not implemented as Julia functions, and "loss function" is a common way to refer to losses in machine learning literature, so users shouldn't be too confused by it.

Evizero:
Fair points I suppose. Well, I'd like to move this along one way or the other.

@joshday @tbreloff Can you accept LossFunctions.jl ? If you both agree I shall rename.

tbreloff:
If it won't be merged as Losses, then fine. But it seems silly do go to the
trouble if renaming ***and** fixing all the other packages that reference
it, on a semi-arbitrary whim.

joshday:
By accepting LossFunctions, are we also dedicating ourselves to TransformationFunctions and PenaltyFunctions?  I'm still not convinced this is a problem that needs solving.  That being said, I'm just happy I'll be able to use this in my registered packages soon.  Losses/MLLosses/LossFunctions all work for me.

tbreloff:
> are we also dedicating ourselves to TransformationFunctions and PenaltyFunctions?

:-1: Yuck.

In fact, this conversation is making me go the other way and want to rename ObjectiveFunctions to Objectives.

Evizero:
@tkelman can we hear a final ruling. Can we accept `Losses.jl` or do we require a renaming?

ahwillia:
I would vote strongly against PenaltyFunctions and TransformationFunctions.
LossFunctions just seems slightly more explicit and is still concise enough
for me.

This should be @evizero's decision since he has put in the most work to
this package. I don't have super strong feelings.

tkelman:
Objectives is a much less useful or descriptive package name than ObjectiveFunctions, and IMO Losses is similarly less clear than LossFunctions would be. If you feel strongly about it we can register the shorter name, but to me it would fail the "what is this package for" test. Try to think a bit about visibility and clarity to potential users instead of optimizing for author keystrokes.

tbreloff:
I defer the decision to Christof.

Evizero:
I though long about this on my walk home from the office. I shall rename to `LossFunctions.jl` for the sake of clarity and community. At the end of the day I am thankful that the Julia namespace is not that flooded with uninformative package names from domains I know nothing of.

That said I am strongly against renaming the types to reflect package name conventions (i.e. renaming `Loss` to `LossFunction`), because `LossFunctions.Loss` is clear enough by any means.

Concerning other packages we are working on: I don't see a reason this decision should constraint their names in any way.

I will get to it soon. I shall open a new PR and link here. Thanks everyone for taking part in the discussion.

PR #6655, opened 2016-10-06
Tag Gadfly v0.5.0 [https://github.com/dcjones/Gadfly.jl]

shashi:
Minor release because "added functionality in a backwards compatible manner" (around Themes and Juno integration)

@tlnagy @MikeInnes


tlnagy:
This is probably one of the biggest Gadfly releases in awhile. Hopefully nothing breaks. 😁

PR #6659, opened 2016-10-06
Register Suppressor [https://github.com/Ismael-VC/Suppressor.jl.git]

Ismael-VC:
Continuation of: https://github.com/JuliaLang/METADATA.jl/pull/6635

Sorry @tkelman


Ismael-VC:
@tkelman I think I've got the hang of this, sadly PkgDev is unusable on Windows at the moment but I learned how to do everything manually.

Please let me know if there is something else that I need to do in order for this to be merged.

Thanks for everything!

tkelman:
probably best to remove all but the most recent tag here

Ismael-VC:
Thanks for you input I will certainly do that.

Ismael-VC:
@tkelman I've just updated the PR, finally everything seems to be in order, I'll be paying attention to any further notice, thanks again.

PR #6660, opened 2016-10-06
Tag StatsBase v0.11.0

ararslan:
Minor version since it deprecates a few things.

### DIFF
--- a/StatsBase/versions/0.11.0/requires
+++ b/StatsBase/versions/0.11.0/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.8.4
--- a/StatsBase/versions/0.11.0/sha1
+++ b/StatsBase/versions/0.11.0/sha1
@@ -1,0 +1,1 @@
+aa5851d83a5b1c5cb3969db0d3b1a6b0bc28b5c7


tkelman:
Looks like DataArrays, MixedModels, and GLM have failures due to this http://pkg.julialang.org/pulse.html

dmbates:
Ah, I see that `GLM` has already been fixed by @ararslan.

ararslan:
@nalimilan Looks like the failure in DataArrays is due to difference between `findat` and `indexin`. Would you mind taking a look at that since you're the one who made that change? It looks like `indexin` returns things in a different order than `findat`.

nalimilan:
Woops. Actually, `findat` returns the first matching index, while `indexin` returns the last one. So I'm not sure what to do. Just print a deprecation warning pointing to `indexin`, but keep the old code?

ararslan:
How could one mimic the behavior of `findat` using `indexin`?

nalimilan:
That's not trivial, since you need to handle zeros. It could be useful to add a boolean argument to `indexin`, but that won't work anyway for Julia 0.4 and 0.5. But so far the uses I could find in DataArrays could be either removed or replaced with `indexin`, so it doesn't look like a critical distinction.

ararslan:
Okay. Since it's not generally trivial to convert code that uses `findat` to use `indexin`, perhaps we want to rethink the deprecation for Julia 0.4 and 0.5, and submit a base PR to get that into 0.6? Not sure what the best course of action is.

nalimilan:
OTOH DataArrays appears to be the only user (until my PR at least):
https://github.com/search?l=&q=language%3AJulia+findat&ref=advsearch&type=Code&utf8=%E2%9C%93

So I'm inclined to deprecating it anyway. StatsBase is already large enough, we don't need to provide generic utilities like that.

ararslan:
That's excellent news! Then yes, let's just leave the deprecation as-is.

nalimilan:
See https://github.com/JuliaStats/StatsBase.jl/pull/218.

PR #6664, opened 2016-10-07
Tag Spectra v0.2.1

charlesll:
Improving compatibility with Julia v0.5


Spectra/versions/0.2.1/requires
```diff
@@ -0,0 +1,12 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

tkelman: you also need to import Compat.String to avoid warnings on 0.4

charlesll: On Compat? 0.9.2 I guess is the right one. Ok, thanks for Compat.String !

charlesll: Also, I just saw that I still have a Warning when loading Spectra because Base.WORD_SIZE is deprecated. However, I don't find what I need to use instead of Base.WORD_SIZE... Do you have any idea?

tkelman: What features are you using from 0.9.2? There's a section in the Compat readme that explains how to identify when features were introduced

`Sys.WORD_SIZE`

charlesll: I am using the @static macro to test the system OS before Fortran compilation.

So from this page: https://github.com/JuliaLang/Compat.jl/commit/f59536084236dfa0437601111a20aab4c82cfba8

I guess the minimum version supporting this macro is 0.7.19?

tkelman: Right, though `@static` is rarely actually needed.

charlesll: Yes I only need it to test the system OS with @static is_unix(), for the compilation of GCVSPL.f (which still does not work for Windows...). I might remove the Compat things later, but some of the users still use 0.4.6 at this time.

I will create another version number for all those changes. Thanks for the help!

tkelman: a normal `if` should be fine, the `@static` is rarely necessary

charlesll: I tried the following:
if VERSION < v"0.5"
...
else 
...
end

But this was not working at all. I was advised to use Compat on Julia Gitter...

PR #6666, opened 2016-10-07
Register and tag AxisArrays

timholy:
AxisArrays provides a rich API for indexing via named axes, "dimensional" or categorical axes, and intervals. As such it has some overlap with DataFrames, NamedAxes, and functionality implemented in Images. Aside from its support for arbitrary dimensionality, what distinguishes it is that its internal design makes such operations inferrable and consequently fast.


AxisArrays/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5-
+IntervalSets
+Iterators
+Compat
```
tkelman: should have a minimum version requirement (if needed for an 0.5-only package?)

timholy: Good catch, thanks.

AxisArrays/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+b334d9055283f00b59ca116ac8e0a783b2db4b31
```
tkelman: travis spoiler alert, git tags will need to match https://github.com/JuliaArrays/AxisArrays.jl/releases

PR #6668, opened 2016-10-07
Bump ColorTypes, ImageCore, ImagineFormat, Immerse

timholy:


tkelman:
ImagineFormat needs to have Compat in its REQUIRE if it's used there

tkelman:
Really shouldn't deprecate things in a patch release. Images and Winston are going to need upper bounds here.

timholy:
OK. Because of the deprecations I didn't expect this to be at all breaking, but it looks like I needed another method.

But I guess I also see your point about even the deprecation warning, given that's what julia does. Many more minor bumps in the future, it seems. (Prob. a good thing.)

Fixed by:
https://github.com/JuliaGraphics/ColorTypes.jl/pull/58
https://github.com/nolta/Winston.jl/pull/273
https://github.com/timholy/Images.jl/pull/562

Will tag after they get through CI (and likely, dinner).

PR #6673, opened 2016-10-07
Tag LCMGL v0.0.2 [https://github.com/rdeits/LCMGL.jl]

rdeits:
Removes a call to the (deprecated) `@linux` macro, and adds a Compat dependency to keep things working on julia 0.4.


LCMGL/versions/0.0.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+BinDeps
+@osx Homebrew
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

rdeits: Good point. Done, and I've updated the SHA to match.

PR #6674, opened 2016-10-07
Tag Fontconfig v0.0.3

tlnagy:
This adds:

- basic testing support
- installation and linking of the libfontconfig binary across platforms
- various bug fixes for v0.5
- drops 0.3 support

Thanks @tkelman and @ararslan for the help!


ararslan:
Since you're bumping the required Julia version, you'll have to tag this as a minor release rather than a patch. That's why Travis is failing.

tlnagy:
Ah. So should I delete the tag and re-tag it?

ararslan:
Yep

tlnagy:
I'm running into this now: https://github.com/JuliaLang/PkgDev.jl/issues/81

PR #6680, opened 2016-10-08
Tag DecisionTree v0.5.0

bensadeghi:


DecisionTree/versions/0.5.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+ScikitLearnBase 0.0.4
+Compat
```
tkelman: should declare a minimum version that implements the features you need

PR #6681, opened 2016-10-08
Tag GeometricalPredicates v0.0.6

robertdj:
Re: https://github.com/JuliaGeometry/GeometricalPredicates.jl/issues/20


tkelman:
Looks like you'll need to re-tag https://github.com/JuliaGeometry/GeometricalPredicates.jl/commit/daf55ad776aa5331eec77fd5e0a4a5db6e913852 for this to work on 0.4. Since you have "allow maintainers to make edits" checked, I added a little test exception. It was being picky about a very old tag which wasn't worth worrying about.

robertdj:
Argh... I forgot that I tagged the second last commit. Is it as easy as deleting the current v0.0.6 tag locally and on GitHub and then tag the last commit? Or is that too late now with this PR?

tkelman:
that should work, you can change the sha and the git tag since this hasn't been merged yet

robertdj:
I'm running into something that may be a problem, but I'd prefer to be careful: Deleting the old v0.0.6 is successful, both locally and on GH. Manually (in the shell) adding a new v0.0.6 tag is successful, but `git push origin :v0.0.6` returns "error: unable to delete 'v0.0.6': remote ref does not exist".

It's also a bit odd that the [latest tag in the repo](https://github.com/JuliaGeometry/GeometricalPredicates.jl/tags) is v0.0.4...

tkelman:
Looks like you may have tried to delete it twice or something? Go ahead and re-create the missing tags based on the contents of METADATA.

robertdj:
Still getting the same error. It may be possible to fix with a `--push` option. Is that too risky?

tkelman:
You don't need to delete it again if it's already been deleted.

robertdj:
I haven't deleted it. I've just added a v0.0.5 and v0.0.6 tag locally based on the two sha's in METADATA and now I want to push them to the repo on GH.

robertdj:
I'm not deleting anything now. I just try to push a local v0.0.5 tag to GH.

tkelman:
does fetching first help at all?

robertdj:
Unfortunately not. Is it possible to add a tag directly on GH?

tkelman:
I think if you "Draft a new release" under https://github.com/JuliaGeometry/GeometricalPredicates.jl/releases/new it should be possible to make a tag from a specific commit through the web interface. It'll show up slightly differently in the list than other tags, but oh well.

Another option might be trying from a completely new clone and seeing if adding tags from that works any better.

robertdj:
I first deleted GeometricalPredicates from `~/.julia/v0.5/` and `~/.julia/.cache`, added it from Julia with `Pkg.add("GeometricalPredicates")` and then started messing with the tags. Are you suggesting that I should clone the repo manually?

How about this: I "draft a new release" for the tag v0.0.5 that matches the sha in METADATA and add a v0.0.6 tag from Julia. I don't think the last tag should be problem.

tkelman:
If you deleted it from both those places then that should be a clean clone, so equivalent to doing it manually I would think. I'm not sure why git/github are getting mad at you here. If you can get something to work and don't mind the release list looking a little funny with a mix of web-releases and normal tags, then hopefully that'll be good enough. Sorry this has been such an ordeal.

robertdj:
Sorry about my long response time; I've been travelling. 
I've added the v0.0.5 tag using GH. Unfortunately, trying to push a v0.0.6 tag (from the shell) is also failing with the same error:

> $ git push origin :v0.0.6
> error: unable to delete 'v0.0.6': remote ref does not exist
> error: failed to push some refs to 'git@github.com:JuliaGeometry/GeometricalPredicates.jl.git'

Should I use GH again?

tkelman:
if there's no remote v0.0.6, then no need to try deleting it

robertdj:
It's still the same situation: I'm not trying to delete anything. In Julia:

```
Pkg.add("GeometricalPredicates")
Pkg.checkout("GeometricalPredicates")
PkgDev.tag("GeometricalPredicates")
```

Then I run `git push origin :v0.0.6` in the shell and get the above error message.

tkelman:
the colon means you are trying to delete the remote tag. do it without the colon.

robertdj:
Ahh...! That probably explains all the difficulties. Let me try it without the colon when I get close to the right computer later. Thanks for your patience!

robertdj:
Now the v0.0.6 tag is on GH as well. I'm terribly sorry for being such an idiot with the Git tags!
Do I need to do more here? Otherwise I'll move on to the packages depending on GeometricalPredicates :-)

PR #6683, opened 2016-10-08
Bump ColorTypes, Colors

timholy:


tkelman:
Should this fix the failures in Images and Winston?  http://pkg.julialang.org/pulse.html

PR #6687, opened 2016-10-09
Tag Symata v0.0.15 [https://github.com/jlapeyre/Symata.jl]

jlapeyre:
Symata v0.0.15 tags the last commit made to Symata when v0.0.14 was registered. v0.0.15 resolves all the issues discussed when v0.0.14 was registered.


jlapeyre:
This request is not quite right. I will make a new request

PR #6689, opened 2016-10-09
Tag FixedPointNumbers v0.2.0

timholy:
Main changes:
- implement `isapprox`
- be more consistent about choice of `Float32` or `Float64` when converting to `AbstractFloat`

The latter has a chance of being breaking for operations that really care about the specific element type. Not sure whether this should be a minor version bump as a consequence (it's a pretty tiny change, really).


tkelman:
Better to change minor version number too often than not often enough.

PR #6690, opened 2016-10-09
Register ImageAxes and ImageMetadata

timholy:
These two packages leverage AxisArrays to endow images with extra behaviors. In the upcoming rewrite of Images, they will replace most of the old dictionary-based functionality in Images.jl, making many more operations inferrable.


tkelman:
ImageMetadata should depend on FixedPointNumbers explicitly if it's being used directly

tkelman:
SimpleTraits should also be in `test/REQUIRE`

timholy:
Good catch, fixed.

PR #6694, opened 2016-10-09
Bump ColorTypes, Images, Winston

timholy:
Fixes issues noticed in #6668.


tkelman:
ColorTypes' tests are failing here

timholy:
Fixed by adding a new tag for FixedPointNumbers.

tkelman:
There are also some new inference failures in ImageFiltering, not sure if related http://pkg.julialang.org/logs/ImageFiltering_0.5.log

PR #6701, opened 2016-10-10
Tag ROCAnalysis v0.0.2 [https://github.com/davidavdav/ROCAnalysis.jl]

davidavdav:
Update to julia-0.5 (and nightly)

 - removed dependence on NumericFuns
 - small bugfixes


tkelman:
METADATA is frozen for packages that support Julia 0.3. If you could drop 0.3 support and re-tag as 0.1.0 we can merge. Also best to declare a minimum version requirement that implements the features you need, with Compat and other dependencies.

tkelman:
closing in favor of #6710

PR #6702, opened 2016-10-10
Tag GeoStatsImages v0.0.5 [https://github.com/juliohm/GeoStatsImages.jl]

juliohm:


tkelman:
What is the copyright status on these images, if you're reproducing some of them from other sources?

juliohm:
They are all publicly available online, some of them from a textbook
written by my adviser, and some of them from a lab in Minnesota that also
shares the data publicly.

I point to the sources in the README.

-Júlio

tkelman:
Is the image data itself copyrighted? Do you have permission to redistribute them under your own copyright and the MIT license?

juliohm:
How can I confirm for sure these images aren't copyrighted? You can see all the download links have no copyright statement:

http://trainingimages.org/training-images-library.html
https://www.esci.umn.edu/orgs/seds/Sedi_Research.htm

I can email the authors if necessary, I don't think they would complain though.

tkelman:
You probably should anyway. If they're associated with a book, the publisher probably owns the copyright. Downloadable off the internet without an explicit license is not the same thing as granting permission for others to redistribute under other terms.

juliohm:
Thank you Tony, I'll double check with them. Would downloading the data
from their link on demand from the Julia package be different? Like
explicitly pointing to an URL and downloading the data whenever the user
asks for it in Julia.

-Júlio

tkelman:
Then you wouldn't be redistributing the data, just the code to download it, so I'd say that would be clearer from a copyright standpoint.

juliohm:
Good to know, I'll consider refactoring the code. However, in general the
cleaning of the data varies dramatically depending on the source. This
package puts everything in the same format.

-Júlio

PR #6708, opened 2016-10-11
Bugfix release for Feather

quinnj:


tkelman:
This is going to interact poorly with precompilation. Can you make the dependency unconditional until the NullableArrays version of DataFrames is released and you can depend on it?

quinnj:
Huh? What dependency are you referring to exactly?

tkelman:
DataArrays

quinnj:
Feather actually isn't precompiled, so it should be ok, right?

tkelman:
If you mark it with `__precompile__(false)` so no one tries to import it in their own packages that are precompiled, then sure. DataArrays is pure Julia right? So just adding it to REQUIRE until the rewritten DataFrames is ready doesn't seem like it would be that problematic?

quinnj:
Ok, updated with an explicit requirement for DataArrays.

tkelman:
Not in https://github.com/quinnj/METADATA.jl/blob/16dea3f44b6e28ea4cbd49e59658db1d93b7f860/Feather/versions/0.2.1/requires - you should really be using PkgDev if it works at all, have you at least tried using `PkgDev.tag` then manually pushing, if publish isn't working?

quinnj:
Just re-pushed. Does PkgDev have functionality to delete tags and retag? It seems like a situation where I'd need to do things manually anyway.

tkelman:
No one has implemented PkgDev.untag yet (the issue has been opened at least 3 times by now), it's the manual creation of tags I'd like to discourage.

PR #6709, opened 2016-10-11
Register NLPModels [https://github.com/JuliaSmoothOptimizers/NLPModel…

dpo:
…s.jl.git]: v0.0.1

### DIFF
--- a/NLPModels/versions/0.0.1/requires
+++ b/NLPModels/versions/0.0.1/requires
@@ -1,0 +1,3 @@
+julia 0.4
+Compat 0.7.15
+LinearOperators 0.2.0
--- a/NLPModels/url
+++ b/NLPModels/url
@@ -1,0 +1,1 @@
+https://github.com/JuliaSmoothOptimizers/NLPModels.jl.git
--- a/NLPModels/versions/0.0.1/sha1
+++ b/NLPModels/versions/0.0.1/sha1
@@ -1,0 +1,1 @@
+2afe2ef7fce406571ab8196ea33ed9a54400a4ea


NLPModels/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

tkelman:
Conditional dependencies interact poorly with precompilation. If you're going to vary functionality based on which other packages are installed, you should explicitly mark the package with `__precompile__(false)`. Or separate the conditional pieces into their own packages.

NLPModels/versions/0.0.1/requires
dpo: Minimum version of what?

dpo:
Aha.  Let me think about that.

NLPModels/versions/0.0.1/requires
tkelman: Compat specifically, but more generally all dependencies if you're using features that don't exist in every single tagged release

dpo:
Done. I'll consider separating the optional bits for the next release, but I'm not too big on creating multiple tiny packages.

NLPModels/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat 0.9.2
```
tkelman: `Compat.String` has been there since 0.7.15, are you relying on anything newer than that?

tkelman: a too tight bound is better than no bound at all, not a big deal. there's a little note on the Compat readme with a blame-based method. we should probably annotate all the readme entries with "since vx.y.z" notes.

dpo: Thanks. Should be good now.

dpo: Does `test/REQUIRE` also need versions?

tkelman: Yes, if you're using any features that aren't present in every single feasible tagged version of your test dependencies.

dpo: Finding the earliest version of everything that works is a combinatorial problem that could be automated, but that will be for later. `test/REQUIRE` now also has versions.

tkelman: blame helps narrow it down, but yeah being too tight with bounds is better than being too loose

dpo:
Any chance of getting this merged then?

tkelman:
There's an unfortunate acronym collision between nonlinear programming and natural language processing, and I don't see what this does that MathProgBase doesn't, but not much will change on either of those fronts.

dpo:
Some of NLPModels was part of MathProgBase at some point but was pulled out. NLPModels provides a model API for writing solvers in Julia, regardless of the provenance of the underlying problem. AmplNLReader.jl and CUTEst.jl build upon it. The approach is quite different from MPB, where models contain an explicit reference to the solver they are associated with.

PR #6710, opened 2016-10-11
Tag ROCAnalysis v0.1.0 [https://github.com/davidavdav/ROCAnalysis.jl]

davidavdav:
Dropping support for julia-0.3

### DIFF
--- a/ROCAnalysis/versions/0.1.0/requires
+++ b/ROCAnalysis/versions/0.1.0/requires
@@ -1,0 +1,4 @@
+julia 0.4
+Compat 0.8.8
+Requires 0.2.2
+DataFrames 0.7.8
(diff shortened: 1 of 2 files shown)


tkelman:
cc @diegozea looks like this caused a test failure in MiToS.jl  http://pkg.julialang.org/logs/MIToS_0.4.log

davidavdav:
Two packages declaring `AUC()`.  `ROCAnalysis.AUC()` is a courtesy for people thinking in terms of positive metrics, higher is better.  Personally I believe in error metrics, lower is better.

diegozea:
Thanks!

diegozea:
I import `AUC` from `ROCAnalysis` now.

PR #6711, opened 2016-10-11
add some more debugging output to the travis script

tkelman:
I think the `git fetch origin +refs/pull/####/merge:` is sometimes getting stale data,
causing some builds to pass when they shouldn't


tkelman:
closing and reopening from my fork, maybe there's a difference... pardon the noise

PR #6713, opened 2016-10-11
Bump MappedArrays, TestImages, ImageCore

timholy:


TestImages/versions/0.1.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Images 0.5
```
tkelman: should this have been changed to FileIO ?

timholy: Thanks.

tkelman: https://github.com/JuliaImages/TestImages.jl/releases too or travis will (at least should) fail

tkelman: oh nevermind not if you didn't change the sha, this should be fine

PR #6717, opened 2016-10-11
Bump some stuff

tbreloff:


Plots/versions/0.9.4/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.5-
+RecipesBase
+PlotUtils
+Reexport
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

tbreloff:
Hmm. I don't know if I'm even using Compat anymore.  Is this merge-blocking?

Plots/versions/0.9.4/requires
tkelman: if you delete it here it should also be deleted in the package require

tbreloff:
Any chance you can merge PlotThemes? I deleted compat but I also have
commits that add that as a dependency, so it would be great to include both
changes.

tkelman:
new package registrations stay open 24 hrs for initial feedback. you don't need to change the sha here to remove the dependency as long as it didn't require any code changes

Plots/versions/0.9.4/requires
tkelman: Should be deleted from https://github.com/tbreloff/Plots.jl/blob/master/REQUIRE before the next tag.

tbreloff:
Thanks Tony. I can tag again after PlotThemes is merged.

PR #6718, opened 2016-10-11
New package // BioEnergeticFoodWebs

tpoisot:
Version `0.1.0` of `BioEnergeticFoodWebs`, a package to simulate biomass dynamics on food webs. The name comes from the "bio-energetic" model of biomass, which will be recognisable to ecologists (our target audience).

Tested on `release` and `nightly`.


BioEnergeticFoodWebs/versions/0.1.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
```
ararslan: You said that the package is being tested on `release` and `nightly` on Travis? `release` refers to 0.5 and `nightly` is 0.6, so you're allowing 0.4 but not testing on it.

tpoisot: We are testing on 0.4 as well (on our own CI machine). I added this to travis (and it passes), but I don't think it warrants tagging a new version.

ararslan: Oh I don't mean tagging a different version, I just wanted to make sure that if you claim to support 0.4 that you're also testing on it. :) Sounds like all's good then! 👍

PR #6720, opened 2016-10-11
Tag Gillespie v0.0.2 [https://github.com/sdwfrost/Gillespie.jl]

sdwfrost:


ararslan:
Looks like you don't have a Julia version specified in the requires file, which is why Travis is failing.

Gillespie/versions/0.0.2/sha1
```diff
@@ -0,0 +1 @@
+dc8e829603a62c953033d796b2b74709a25ade0d
```
tkelman: are you sure this is the commit you want to tag? it's from July https://github.com/sdwfrost/Gillespie.jl/commit/dc8e829603a62c953033d796b2b74709a25ade0d

Gillespie/versions/0.0.2/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

Gillespie/versions/0.0.2/sha1
sdwfrost: My bad; I didn't update the package from the Git repo before publishing, as I started with a fresh install. I actually want to tag this commit: [https://github.com/sdwfrost/Gillespie.jl/commit/39fb53516cc1d736c18ba7ac62d51e81287f9246](https://github.com/sdwfrost/Gillespie.jl/commit/39fb53516cc1d736c18ba7ac62d51e81287f9246)

PR #6723, opened 2016-10-11
Tag CUSPARSE v0.5.0 [https://github.com/JuliaGPU/CUSPARSE.jl]

kshyatt:
- Now supports Windows
- Uses `Base.Test` `@testset`s
- Tests pass on Windows and Linux


CUSPARSE/versions/0.5.0/sha1
```diff
@@ -0,0 +1 @@
+e7f2be5b071c09026079ccb41545695e542ad3c2
```
kshyatt: Thanks again for the catch ❤️

PR #6724, opened 2016-10-12
Tag Fontconfig v0.1.1

tlnagy:
- Fixes build on Windows, thanks to @jmcastro2109 for helping me figure this out


tkelman:
Should probably enable AppVeyor so it stays that way

tlnagy:
Good point, now that it's in JuliaGraphics, I can add it without bugging Daniel.

PR #6731, opened 2016-10-12
Tag YT v0.4.0 [https://github.com/jzuhone/YT.jl]

jzuhone:


tkelman:
PyArray comes from PyCall, right? You shouldn't be extending getindex on a type from a different package, since that will change the behavior of unrelated code depending on whether or not your package has been imported.

jzuhone:
@tkelman decided to avoid using PyArray altogether now--I'll re-tag and re-submit.

PR #6734, opened 2016-10-12
Tag CovarianceMatrices v0.1.2 and v0.2.0

gragusa:
- v0.1.2 is the last release to support Julia v0.4
- v0.2.0 minor release Julia v0.5 only


CovarianceMatrices/versions/0.1.2/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4 0.5-
+StatsBase 
+GLM 0.5.5
+DataFrames
+Compat
```
tkelman: should declare a minimum version that implements the feature you need

gragusa:
Hopefully this will fix things.

CovarianceMatrices/versions/0.1.2/requires
tkelman: still need a minimum version on the 0.1.2 tag

CovarianceMatrices/versions/0.2.0/requires
```diff
@@ -2,4 +2,3 @@ julia 0.5
 StatsBase 
 GLM 0.6
 DataFrames
-Compat
```
tkelman: you're still doing `import Compat.view` on master which you'll need to remove for this to work

gragusa: Removed `import Compat.view` from master, retagged and updated sha of the tag. 

For the v0.1.2 tag, the tag REQUIRES 0.8.4 of compat, but I forget to update the REQUIRE of the METADATA PR. Now I did it.

gragusa:
Thanks!

PR #6740, opened 2016-10-13
Tag Gadfly v0.5.2 & Reactive v0.3.6

shashi:
Test fixes


Reactive/versions/0.3.6/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat 0.9.2
```
tkelman: what are you using that's only in 0.9.2?

shashi: I guess @SimonDanisch just went with the latest tag of Compat. And it works. Is this a problem?

tkelman: It's too tight of a bound. What features do you actually need?

ararslan:
While you're at it, want to update the URL for Gadfly to reflect its new location?

tkelman:
@araslan I actually put that in https://github.com/JuliaLang/METADATA.jl/pull/6712 but it was hitting a bug in the test script that I haven't figured out yet

ararslan:
Ah, okay. Sorry for the noise, Shashi, and thank you Tony for looking into that!

tkelman:
"looking into" = running https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/update-redirects.jl

Reactive/versions/0.3.6/requires
shashi: a NEWS.md would be nice to have. Anyway, I could fix this by removing the dependency :-p

tkelman:
Yeah looks like that should work too, though check that it still works on 0.4 and doesn't have depwarns on 0.5.

Since Travis is so backed up right now, that reminds me - time to create a JuliaUX org for Reactive, Interact, maybe a few more?

PR #6745, opened 2016-10-13
Tag Plots v0.9.5 [https://github.com/tbreloff/Plots.jl]

tbreloff:


tbreloff:
Since this wasn't merged yet, I'm adding a fix for an issue with PlotThemes gradient names.

tkelman:
These are both failing tests on Travis

tbreloff:
PlotThemes tests pass on master (one tiny change to a test... no change to the package).  I don't think the Plots tests will pass until PlotThemes is tagged, but they pass locally.

tkelman:
Best to re-tag or make a new PlotThemes tag that includes the test fix then

tbreloff:
Or... you could just merge this and I'll restart travis.

tkelman:
restarting travis on plotthemes won't make a difference, the version you tagged here fails its tests - it only has one dependency

tbreloff:
I'm not tagging again anytime soon, so send people to the gitter if they have trouble

tkelman:
I don't understand what the problem was here. If you already fixed the issue, just adjust the sha and the git tag, or add a new one.

PR #6747, opened 2016-10-13
Tag Gillespie v0.0.2 [https://github.com/sdwfrost/Gillespie.jl]

sdwfrost:
Trying again with up-to-date repo


Gillespie/versions/0.0.2/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+Compat
```
tkelman: https://github.com/JuliaLang/METADATA.jl/pull/6720#discussion_r82973731

sdwfrost: Do you mean a minimum version for Compat as well?

tkelman: yes, whatever first implements all the features you need from compat. there's a note on the readme there about how to identify that

tkelman:
it's a good idea to do so if you're relying on any features that haven't been present in every tagged release of those packages, but can be a bit trickier to determine for other packages.

Gillespie/versions/0.0.2/requires
tkelman: be sure to also make the same change here https://github.com/sdwfrost/Gillespie.jl/blob/master/REQUIRE

PR #6749, opened 2016-10-13
Tag MixedModels v0.5.8

dmbates:


tkelman:
Should use a new package minor version when dropping support for Julia 0.4

tkelman:
And similar to https://github.com/JuliaStats/GLM.jl/issues/159, the package will remain broken on 0.4 if this tag is Julia 0.5-only. If there have been features or new code that relies on 0.5-only features, would it be possible to backport a 0.5.8 release that does support Julia 0.4 and passes tests?

tkelman:
Another option would be to upper-bound the past tags to StatsBase < 0.11 on both packages. Then StatsBase would get held back on 0.4 as long as either of those packages are installed.

dmbates:
@tkelman To make sure I understand, do I need to change the `requires` file in all previous releases of `GLM` and `MixedModels` to upper-bound the version of `StatsBase` and that this would be done according to the instructions [here](http://docs.julialang.org/en/release-0.5/manual/packages/#fixing-package-requirements).   In other words I change the `requires` file in the METADATA repository only.

Currently the requires file for 0.5.7 has the line
```
StatsBase 0.8.2
```
which I should change to
```
StatsBase 0.8.2 < 0.11
```

Finally, in Julia-0.5.0 I
```
PkgDev.tag("MixedModels", :minor)
PkgDev.publish()
```

tkelman:
No `<` in the require file. I can do the bounding for you if you'd prefer. Though as in https://github.com/JuliaStats/GLM.jl/issues/159#issuecomment-253874062, if releasing an 0.4-compatible version with just the bugfix is possible without too much trouble, it would be preferred to avoid holding people back to old releases of StatsBase.

dmbates:
I'm sorry but I am hopelessly confused here and convinced that I would foul things up big time if I were to attempt to reconcile the situation.  Because the version of this package for julia-0.5.0 depends on version 0.6.0 of `GLM`, I guess I should try to straighten out https://github.com/JuliaStats/GLM.jl/issues/159#issuecomment-253874062 first.

tkelman:
Great, GLM is now fixed. To get MixedModels up and running, can you start by creating a `release-0.5` branch in that package starting from the commit https://github.com/dmbates/MixedModels.jl/commit/d1e90928225bc5d8e531b6a5366563a061fa4527 ? I think with one change of `wrkwt` back to `wrkwts` for GLM 0.5.x we can make that pass its tests on Julia 0.4, tag it as 0.5.8, then move on and tag the Julia-0.5-only GLM 0.6.0.

dmbates:
@tkelman You suggested creating a `release-0.5` or `v0.5` branch.  Did you mean to write `release-0.4` or `v0.4` branch.  I already have a `v0.4` branch.  I made a couple of changes in it and it passes tests under julia-0.4.7.  What I plan to do is to remove checking under 0.5 and nightlies for travis and appveyor on this branch and check in this version to the branch only (i.e. keep master as it is).  Then I will do `Pkg.tag()` and `Pkg.publish()` under v0.4.  Should I change the `REQUIRE` file to have
```
julia 0.4 0.5
```
or just go ahead with tagging a 0.6.0 release with minimum julia requirement of 0.5

Thanks for all you patience and help on this and sorry for being such a PITA.

tkelman:
I noticed that you had a `v0.4` branch, but I wasn't sure what your naming convention was, if that was referring to Julia version 0.4 or MixedModels version 0.4. Your MixedModels v0.5.7 tag was at commit ab841caffba6f307adde8f52fb931c522ce031e4, which it looks like you do have present on that branch. I may have not checked that closely enough. I think what you're doing so far sounds good. I wouldn't bother with the upper bound if you plan on tagging a newer Julia-0.5-only version soon. Happy to help.

tkelman:
we can revisit this julia-0.5-only version as MixedModels 0.6.0 whenever you're ready

dmbates:
@tkelman I would like to get a julia-0.5-only version out but the master branch is now failing tests on [travis](https://travis-ci.org/dmbates/MixedModels.jl/builds/168930879) and on [appveyor](https://ci.appveyor.com/project/dmbates/mixedmodels-jl), apparently when trying to load `Feather`.  The immediate cause seems to be `Feather` using `__precompile__(true)` and `CategoricalArrays` not.

I'm not actually using `Feather` or `CategoricalArrays` in the tests because I set the version of `DataFrames` to be 0.8.4, the latest version, which, I think, does not have the `PooledDataArray -> CategoricalArray` and `DataArray -> NullableArray` changes in it.

tkelman:
You have `using CategoricalArrays` first thing in both `src/MixedModels.jl` and `test/runtests.jl` https://github.com/dmbates/MixedModels.jl/search?utf8=%E2%9C%93&q=CategoricalArrays&type=Code

dmbates:
@tkelman I have removed those references to `CategoricalArrays`.  I think it was actually `Feather` that was causing the problems but I can't be sure.

I tied myself up in knots trying to get my local version of METADATA so that I could pull from origin because the files in MixedModels/versions/0.5.8 were inconsistent.  I spent a fun time trying to do a merge and, after a couple of hours, reverted to removing Julia and all my packages, rebooting the computer and starting again from scratch.  Right now it is taking forever to download the DLL for NLopt.  I hope I am able to commit something for 0.6.0 later tonight.

tkelman:
That was a lot more drastic than needed (though consistent with https://xkcd.com/1597/ ...) - would have been happy to offer guidance at any point that git started doing confusing things.

tkelman:
closing, #6810 looks good

PR #6750, opened 2016-10-13
Tag Iterators v0.2.0 [https://github.com/JuliaLang/Iterators.jl]

iamed2:
Performance improvements, support 0.5, drop 0.3.


Iterators/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
iamed2: It turns out we didn't need Compat anymore at all, which made my job easier.

PR #6753, opened 2016-10-14
Register and tag ShowItLikeYouBuildIt

timholy:
This package allows one to simplify the `summary` of `AbstractArray` types, with the design of presenting type information via the sequence of commands one might use to construct such an object. Discussed in https://github.com/JuliaLang/julia/issues/18909.


timholy:
Seems like another bug in `check-pr`?

tkelman:
Looks that way. And not only affecting this PR. Worth throwing https://github.com/timholy/DebuggingUtilities.jl#test_showline at it to see where it's going wrong?

tkelman:
Hm, okay restarted and it went away. My current theory for what might be causing these is Travis getting outdated information when it fetches the merge commit. Will experiment with some debugging output in #6712 to try to substantiate that theory.

tkelman:
Should Compat be listed in `test/REQUIRE`, or is it not needed?

timholy:
Tagged a newer commit (see https://github.com/JuliaArrays/ShowItLikeYouBuildIt.jl/pull/1).

PR #6764, opened 2016-10-14
Tag RCall v0.6.0 [https://github.com/JuliaInterop/RCall.jl]

randy3k:
Update RCall to v0.5.3.


RCall/versions/0.5.3/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5
```
tkelman: since this is the first 0.5-only tag, it should use a new package minor version

randy3k:
Tag RCall v0.6.0 instead.

PR #6766, opened 2016-10-15
Tag ImageQuilting v0.4.0 [https://github.com/juliohm/ImageQuilting.jl]

juliohm:


tkelman:
Your optional dependencies in a try-catch interact poorly with precompilation. You should really be marking your package `__precompile__(false)` if you're going to put try-catches around optional dependencies.

juliohm:
Thank you @tkelman, I will fix the precompilation statement in a future patch release. Could you please merge the tag and elaborate on what mechanism would be better to solve optional dependencies?

tkelman:
Separating them into independent packages. Or disabling precompilation. Or ensuring they don't get loaded until run time, instead of module load time.

juliohm:
Ok, I will disable precompilation or try to load at runtime in a future patch. My colleagues at work are waiting for a working version of my package in Julia v0.5. I am involved with other responsabilities at the moment and cannot work on this fix now.

juliohm:
@tkelman could you please comment on the harm of having precompilation with optional dependencies like I am doing? What is the scenario in which precompilation fails?

tkelman:
when the set of available dependencies at precompilation time does not match the set available at runtime, you get either errors or missing functionality. that change in state does not propagate to the staleness check correctly.

juliohm:
@tkelman is there any plan to support optional dependencies in a robust way? The OpenCL packages are definitely a good use case, not everyone has a GPU to perform calculations and it is a pain to configure drivers for these devices. I definitively want to have this feature in my package, but I also like the idea of having everything precompiled once. It improves the user experience.

tkelman:
There's a ppan but it hasn't been inplementes yet. You could move the optional GPU components to a separate package and that would work now.

juliohm:
It is not easy to move the GPU components to a different package, at least I don't see how I would do it without major refactoring of the code. Could you please elaborate on how one would Pkg.add("separate") and get the functionality in package "original" affected? Is it through redefinition of methods with the import statement?

tkelman:
You could arrange things to pass a type representing the computation backend, with a default unspecified setting that gets determined by a runtime priority list. importing the gpu package could insert itself into that priority list.

juliohm:
Interesting, I will think about it, for now I will just disable precompilation.

PR #6770, opened 2016-10-15
DevIL / OpenIL version 0.4.2 with precompilation

aaalexandrov:


tkelman:
Should probably add an appveyor.yml file. And some tests.

PR #6773, opened 2016-10-16
Tag ImageAxes, move sha on ImageCore to merge commit

timholy:
Chose the wrong commit in #6772


ImageCore/versions/0.0.5/sha1
```diff
@@ -1 +1 @@
-52f75085d31183b13a318566a24dba40029e06a5
+c94238a82d185c7c713b206a3c40ad4e420d314e
```
timholy: Didn't know that. What's so bad about changing the sha? FYI I also changed the tag, of course. AFAICT it's only if someone tried to update in that window between the two that it could have caused a problem; to me that doesn't seem any worse than a transient DNS outage or random Travis failure.

tkelman: there was a window, and this repository gets cloned thousands of times per day. don't think that's even counting updates to existing clones.

timholy: OK. Given that you routinely encourage folks to change bounds (which can also change the Pkg state), it didn't seem so different to me. Might be worth spelling out somewhere.

timholy: Also, keep in mind that we're discussing the difference between the merge commit and the one that immediately preceded it. So they're indistinguishable in terms of what code gets installed on the user's machine.

tkelman: bounds change the pkg resolution solution, which will result in changing the versions that get installed. the shas should stay fixed for each given version.

tkelman: if it was indistinguishable, why did it need to be changed at all?

timholy: Because we usually recommend tagging the merge to master, whereas this tag was on a branch.

PR #6779, opened 2016-10-17
Add guideline stating not to modify existing `sha1` files

tkelman:


tkelman:
for precedent, there are 2 references in https://github.com/JuliaLang/julia/pull/4391 that "the hash associated with a version is permanent," or at least that was the intent that things shouldn't be modified once published and potentially pulled by any users

PR #6780, opened 2016-10-17
Tag Sundials v0.3.0

ChrisRackauckas:
This version includes @acroy's large PR, a large breaking change. The API in this version is still the same as in the previous version, but the improved API, coupled with a bump in the Sundials version change, is still a ways off. So because this version has large bug fixes but has a compatible API with code from before, we are tagging this for the time being so that we can take our time with the API PRs.


Sundials/versions/0.3.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BinDeps 0.3
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

ararslan:
Looks like you need to push the tag to the repository. See https://travis-ci.org/JuliaLang/METADATA.jl/builds/168380522#L209

ChrisRackauckas:
No it's there: https://github.com/JuliaDiffEq/Sundials.jl/releases

It took me awhile to find out that I had to do it manually though.

ararslan:
Oh, I guess the Travis build just needs to be restarted then.

If you use PkgDev, the tags are pushed upstream for you. Highly recommended 👍

ChrisRackauckas:
> If you use PkgDev, the tags are pushed upstream for you. Highly recommended 👍

I used PkgDev on Windows, so the tags aren't pushed upstream for you... highly not recommended 👎

tkelman:
Report an issue.

ChrisRackauckas:
Yeah I always forget to. Sensible minimum version requirements were added. I'll go report the issue right now.

Sundials/versions/0.3.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BinDeps 0.4.3
```
tkelman: which is it, 0.4.3 or 0.4.5 https://github.com/JuliaDiffEq/Sundials.jl/commit/0e13eddd9095525cfc87185170ee346d7e233678? those are both very new, is there actually a brand new feature or bugfix in BinDeps you're relying on?

tkelman: Ok. Not much has changed in BinDeps for some time, it's primarily been deprecation fixes.

PR #6784, opened 2016-10-17
Register ParameterizedFunctions v0.0.1

ChrisRackauckas:
ParameterizedFunctions is a setup for callable types which allow for parameters inside of functions. This allows differential equations solvers to generalize the functionality in ways such as taking parameter derivatives for sensitivity analysis, parameter estimation, bifurcation plots, etc. This setup also allows the same function to be used old-school ODE solvers which are do not implement such functionality (ODE.jl, ODEInterface.jl), allowing for seamless interaction.

While the type on its own doesn't warrent a package, the package implements some macros for defining ParameterizedFunctions. These will parse a LaTeX-like definition and build the correct function with the parameters built-in, and will even symbolically build the Jacobian, inverse Jacobian, and (in the near future) the parameter derivative functions automatically, which will enhance the speed of solvers that are aware of the functionality. This means it will both help DifferentialEquations.jl and the other JuliaDiffEq packages interop more seamlessly with the other packages while giving even more performance gains.


ChrisRackauckas:
Oh wow, something was messy here. I'll fix it.

PR #6789, opened 2016-10-18
Tag LightGraphsExtras v0.0.4 [https://github.com/JuliaGraphs/LightGra…

sbromberger:
…phsExtras.jl]


tkelman:
Since support for Julia 0.4 is being dropped, should use a new package minor version 0.1.0

PR #6791, opened 2016-10-18
Tag Compat v0.9.3 [https://github.com/JuliaLang/Compat.jl]

tkelman:
Some Nullable-related additions by @davidagold https://github.com/JuliaLang/Compat.jl/pull/287

You'll want to migrate StructuredQueries over to using these from Compat as soon as you can, since otherwise having the same definitions in two different places may cause warnings or more serious problems.


davidagold:
Duly noted. Thank you.

PR #6792, opened 2016-10-18
Tag CategoricalArrays v0.1.0 [https://github.com/JuliaData/CategoricalArrays.jl], move to JuliaData

nalimilan:
Supporting Julia 0.4 turns out to be harder than I thought, especially with last release, so I'm giving up that goal. Raise requirement everywhere to ensure users don't use old versions without noticing.


tkelman:
Did any of these work? Several other packages have tags that depend on these versions and do support Julia 0.4, which this would break. Would rather leave it in place as-is and make the new 0.5-only version 0.1.0.

nalimilan:
They did work, but with occasional crashes, and the new version crashes systematically. I'd rather not have people use the old versions as the new one fixes important issues, and the package is still likely to evolve significantly.

That said, we can wait for a while  before raising the requirement so that we can adapt dependencies. I guess most of them are @quinnj's packages?

(It's funny, I think we had reverse positions in a previous issue, in which I wanted to keep 0.4 support... :-)

tkelman:
Pkg.update() will error with this for anyone who has the package installed on 0.4. can we just leave the existing versions alone as they are?

nalimilan:
Done.

tkelman:
Great, thanks. Should use a new package minor version though.

nalimilan:
Here you go. What a milestone!

PR #6794, opened 2016-10-18
adding test GreatCircle.jl v0.1.0

acrosby:
Let me know if there are any issues, I have to do this manually usually.


tkelman:
missing the `requires` file

acrosby:
Thanks, is this just supposed to look like "REQUIRE" in the root of the package repo?

tkelman:
same content, but named `requires` for some reason that isn't written down anywhere I can find

acrosby:
yeah, I tried to find an explanation for the METADATA.jl layout, but couldn't.

PR #6795, opened 2016-10-18
Tag MixedModels v0.5.8

dmbates:
Version installable on julia-0.4.x


MixedModels/versions/0.5.8/requires
```diff
@@ -0,0 +1,10 @@
+julia 0.4
+DataFrames 0.8
+Distributions 0.10
+GLM 0.5.2
+NLopt 0.3
+Showoff
+StatsBase 0.8.2
```
tkelman: the `dof` renaming was done in StatsBase 0.11, right @ararslan?

ararslan: Yep.

PR #6798, opened 2016-10-19
Register and tag QuickTypes.jl

cstjean:
Two simple macros for [more concise type definitions](https://github.com/cstjean/QuickTypes.jl). 

```julia
@qtype Window{T}(price::Int, color=:blue; opacity::T=1.0)
```

Is this already defined elsewhere?


kmsquire:
👍 I've wanted this for a while, but never took the time to write the macro. Thank you!

PR #6808, opened 2016-10-19
Tag Celeste v0.2.0 [https://github.com/jeff-regier/Celeste.jl]

andreasnoack:


tkelman:
Compat is still used at https://github.com/jeff-regier/Celeste.jl/blob/ae3ae4f083157204bfe68efe912441b5540f6267/test/test_transforms.jl#L5, so should be at least in `test/REQUIRE` if not the main `REQUIRE` - or that line removed if no longer needed

andreasnoack:
Ok. I've updated `Celeste.jl` and moved the tag.

PR #6812, opened 2016-10-19
Tag POMDPs v0.3.0 [https://github.com/JuliaPOMDP/POMDPs.jl]

etotheipluspi:
Update to 0.5


POMDPs/versions/0.3.0/requires
```diff
@@ -1 +1 @@
-julia 0.4
+julia 0.5
```
tkelman: should also make this change here https://github.com/JuliaPOMDP/POMDPs.jl/blob/master/REQUIRE

tkelman: https://github.com/JuliaPOMDP/POMDPs.jl/pull/108

PR #6817, opened 2016-10-20
Tag GR v0.16.0 [https://github.com/jheinen/GR.jl]

jheinen:


tkelman:
Conditional dependencies like you're doing here https://github.com/jheinen/GR.jl/blob/v0.16.0/src/jlgr.jl#L7-L11 are not going to interact well with precompilation.

PR #6824, opened 2016-10-21
Tag PhyloTrees v0.5.0 [https://github.com/jangevaare/PhyloTrees.jl]

jangevaare:


tkelman:
`throw`ing strings is technically allowed, but probably not what you want

PR #6826, opened 2016-10-22
Register DifferentialEquationsBase

ChrisRackauckas:
DifferentialEquationsBase is the lightweight Base library which contains the shared types of the differential equations solvers. This will become key during the breakup of DifferentialEquations.jl into component packages as specified in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59 , but is also already necessary for the new MultiScaleModels.jl to be able to properly extend DifferentialEquations' functionality without the strict requirement (since the MultiScaleModel is "an array type" which has uses beyond differential equations, as already discussed in Gitter chat).


ChrisRackauckas:
Changing the name to DiffEqBase to match the split.

PR #6830, opened 2016-10-22
Tag FredData v0.1.3 [https://github.com/micahjsmith/FredData.jl]

micahjsmith:
I know that my build shows as failing, but this is only for nightly on Linux (v0.5 on Linux and macOS and nightly on macOS are all okay). For some reason, Travis fails to download the DataFrames package (https://travis-ci.org/micahjsmith/FredData.jl/jobs/169761289#L168), so I'm not worrying about it.


tkelman:
known issue that we'll hopefully fix before too long. you can mark nightly as an allowed failure so it'll run but won't mark your status red

PR #6831, opened 2016-10-22
Register RigidBodyDynamics 0.0.1

tkoolen:
`RigidBodyDynamics` is a basic rigid body dynamics package implemented completely in Julia. See https://github.com/tkoolen/RigidBodyDynamics.jl/blob/master/examples/Mechanism.ipynb for some usage examples.

### DIFF
--- a/RigidBodyDynamics/url
+++ b/RigidBodyDynamics/url
@@ -1,0 +1,1 @@
+https://github.com/tkoolen/RigidBodyDynamics.jl.git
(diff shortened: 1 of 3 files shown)


tkelman:
You're using Compat at https://github.com/tkoolen/RigidBodyDynamics.jl/blob/v0.0.1/test/runtests.jl#L7, so should declare the dependency (with a minimum version requirement that implements the features you need) in either `REQUIRE` or `test/REQUIRE`

tkoolen:
Good call. I don't actually need anything from Compat anymore, so I'll just get rid of the using statement.

tkelman:
can tag that as an 0.0.2 then

tkoolen:
Sounds good, thank you!

PR #6833, opened 2016-10-22
DifferentialEquations.jl Modularization

ChrisRackauckas:
Implements the changes previously discussed at https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59

- The solvers were split out of DifferentialEquations.jl as `____DiffEq`.
- Auxiliary components, such as DiffEqBase and DiffEqDevTools. follow similar naming but are distinguished from the actual solversr
- VectorizedRoutines and ChunkedArrays get a tag as well for compatibility
- ParameterizedFunctions has a compatibility tag
- Every package has sensible minimum version requirements
- Tests pass locally but do not pass on Travis because they are waiting for each other to be tagged. I hope that an exception can be made for this case to accept without the tests passing (the code is directly from DifferentialEquations.jl for the most part) so that tests can pass in the normal way without obstruction, and a patch bump can be issued if necessary (again, tests pass locally on a Windows computer and a Linux computer so most of them should be okay!)


ChrisRackauckas:
Tags for DifferentialEquations.jl, making it a metapackage which holds all of this together, will come later, after the pieces of the ecosystem all have tests passing.

ararslan:
The repository names for all of the packages have to end in `.jl`; currently StokesDiffEq doesn't. The URLs in this PR will similarly need to be updated so that they point to the right place.

ChrisRackauckas:
Thanks for the check. I updated StochasticDiffEq.jl and StokesDiffEq.jl's URLs. I realized I also missed AlgebraicDiffEq.jl.

OrdinaryDiffEq/versions/0.0.1/requires
```diff
@@ -0,0 +1,10 @@
+julia 0.5
+Parameters 0.5.0
+ForwardDiff 0.2.4
+GenericSVD 0.0.2
+Compat 0.8.8
+InplaceOps 0.0.5
+NLsolve 0.7.3
+Sundials 0.3.0
+Ranges 0.0.1
+RecipesBase 0.1.0
```
tkelman: DiffEqBase is used without being `REQUIRE`d

tkelman: likewise with `IterativeSolvers` and `ParameterizedFunctions`

and `Plots` is used in tests but not listed in `test/REQUIRE`

ChrisRackauckas: Fixed

StochasticDiffEq/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+ParameterizedFunctions 0.1.0
+Parameters 0.5.0
+ChunkedArrays 0.1.0
```
tkelman: DiffEqBase missing here

PR #6837, opened 2016-10-23
Tag BloomFilters v0.2.0 [https://github.com/johnmyleswhite/BloomFilte…

boydgreenfield:
…rs.jl]


BloomFilters/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need - there's a note about this on the Compat readme

kmsquire: Bump!

ViralBShah:
I am not convinced that finding the minimum Compat.jl version is the most straightforward thing to do. I would rather just people use the current Compat version instead, which may be a bit of overkill but certainly something that is much simpler.

tkelman:
If you're using Compat it's almost always to fix some specific well-defined deprecations.

Bounds that are too loose allow version combinations that don't work together to be potentially selected by Pkg. Bounds that are too tight prevent version combinations that do work from being installed. Both are incorrect, but with different consequences.

KristofferC:
It seems that Compat is pretty much the poster child for when this is never an issue in practice.

boydgreenfield:
I've fixed this, though now I think more Compat changes and a v0.3.0 are needed for Julia v0.6.0... 

Can this get merged in now at least? Thanks!

BloomFilters/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4.0
```
tkelman: should make this change on the require file of master as well

tkelman:
you seem to have force pushed over something and deleted https://github.com/johnmyleswhite/BloomFilters.jl/commit/07d7fe1ba27cb039deb52339149015f3051974a1

boydgreenfield:
Ugh. I borked and reset this. Unfortunately, now when I try to tag v0.2.0 with `Pkg` locally I get the following (unrelated?) error:

```
julia> Pkg.publish()
INFO: Validating METADATA
    Lora v0.5.0 – no valid versions exist for package ForwardDiff

 in error at error.jl:19
 in check_metadata at pkg/entry.jl:602
 in publish at pkg/entry.jl:339
 in anonymous at pkg/dir.jl:28
 in cd at file.jl:20
 in cd at pkg/dir.jl:28
 in publish at pkg.jl:59
```

Preference on debugging the above vs. force pushing onto this branch the proper SHA1 and tagging the main repo manually?

tkelman:
what version of julia are you using?

boydgreenfield:
@tkelman I don't use Julia regularly (at all), so until trying to update this for folks I was on 0.3.x. I just got 0.5.0 via Homebrew and get the above error even after nuking my `~/.julia` directory and trying to start from scratch.

tkelman:
most likely, that error has to do with a bug that was fixed in 0.5.1. don't think the staticfloat tap is updated yet, but the cask version might be

boydgreenfield:
Got it... what's the easiest/fastest fix here? I have to take off momentarily and this was my "don't be an open source slacker" task for the day I'd like to get fixed but have obviously screwed up...

tkelman:
should be fine

boydgreenfield:
@tkelman Thanks for the flexibility (and attentiveness)!

PR #6838, opened 2016-10-23
Tag DataCubes v0.0.7 [https://github.com/c-s/DataCubes.jl]

c-s:
This version is compatible with Julia v0.5.


tkelman:
should use a new minor package version if dropping support for julia 0.4

PR #6843, opened 2016-10-23
Pull request/e51d9f75

c-s:
Previously I submitted a pull request for DataCubes v0.0.7 to support Julia v0.5, but was told that I need to update the minor version number. I don't know how to update the pull request, so created another pull request. Thanks.


ararslan:
> I don't know how to update the pull request

It's actually pretty simple, though like most things in Git, it isn't exactly obvious. For future reference, what you do is push the Git commit to the branch that's being used for the pull request. So in the case of this PR, the remote name is `c-s` and the remote branch associated with the PR is `pull-request/e51d9f75`. To update the PR, you'd commit some changes to your local clone of METADATA, say for the sake of argument you're on a branch called `update-pr`, and then do
```
git push c-s update-pr:pull-request/e51d9f75
```
More information about this can be found in [this handy GitHub Help article](https://help.github.com/articles/pushing-to-a-remote).

I hope that helps!

ararslan:
Also since you've opened this new PR that supersedes the old one, you should close the old one, #6838.

c-s:
Thanks for helpful information! I closed the old pull request.

tkelman:
You should not be defining these methods: https://github.com/c-s/DataCubes.jl/blob/v0.1.0/src/common.jl#L19-L23, instead do what the warning suggests and use `permutedims`. Otherwise you're extending base methods on base types, which changes the behavior of all other unrelated code depending on whether or not your package has been loaded.

c-s:
Thanks for the suggestion. I deleted unnecessary transpose definitions.

PR #6844, opened 2016-10-23
Tag DifferentialEquations v0.5.0

ChrisRackauckas:
This is the pre-modularization tag for [DifferentialEquations.jl](https://github.com/JuliaDiffEq/DifferentialEquations.jl). However, it is a minor tag because there are many feature which were added, including (but likely not limited to):

- A benchmarking suite (which shows [order of magnitude speedups for the ODE algorithms over ODE.jl](https://github.com/JuliaDiffEq/DifferentialEquations.jl/tree/master/benchmarks))
- [High order interpolations for every optimized Runge-Kutta method (ranging from 3 to 9 on 10? algorithms, with a default of 3 on all other 100+ ODE algorithms, including the ODE.jl wrapped algorithms)](http://juliadiffeq.github.io/DifferentialEquations.jl/stable/solvers/ode_solve.html)
- Overloaded calls so the [solution object acts both like an array and the interpolation function](http://juliadiffeq.github.io/DifferentialEquations.jl/latest/tutorials/ode_example.html)
- Sophisticated [event handling and callback functions](http://juliadiffeq.github.io/DifferentialEquations.jl/latest/man/callback_functions.html)
- Wrapper for the Sundials [Differential-Algebraic Equation (DAE) solver](http://juliadiffeq.github.io/DifferentialEquations.jl/latest/tutorials/dae_example.html)
- A DSL via function definition macros. These have already been expanded upon as ParameterizedFunctions.jl, but will release with these still present for compatibility

```julia
g = @ode_define begin
  dx = σ*(y-x)
  dy = x*(ρ-z) - y
  dz = x*y - β*z
end σ=>10. ρ=>28. β=>2.66
```

- Analysis methods for tableaus, such as plotting of stability regions
- Sophisticated defaults handling, having different stabilization parameter defaults for different algorithms
- The documentation now uses the native HTML rendering
- Improved default plots

Next, 

- DifferentialEquations.jl will be split into component packages to make it easier for other contributors to follow the fast development and for other developers to use this package without requiring the whole thing. 
- Native algorithms for DDEs will be added to the ecosystem (using the interpolations)
- "Free" symbolic (inverse) Jacobians from ParameterizedFunctions.jl will increase the performance of the stiff solvers
- "Models" add-on packages which pre-define standard models used in disciplines for easy use by financial quants, biologists, etc.
- NDF/BDF algorithms for stiff ODEs
- A logo. It's already made!

I hope to do quicker tags by having the pieces more modular, but plan to continue this pace.

### DIFF
--- a/DifferentialEquations/versions/0.5.0/requires
+++ b/DifferentialEquations/versions/0.5.0/requires
@@ -1,0 +1,12 @@
+julia 0.5
+Parameters 0.5.0
+ForwardDiff 0.2.4
+ChunkedArrays 0.0.2
+IterativeSolvers 0.2.2
+GenericSVD 0.0.2
+Compat 0.8.8
+Plots 0.9.2
+InplaceOps 0.0.5
+NLsolve 0.7.3
+Sundials 0.3.0
+Ranges 0.0.1
--- a/DifferentialEquations/versions/0.5.0/sha1
+++ b/DifferentialEquations/versions/0.5.0/sha1
@@ -1,0 +1,1 @@
+eab10c60d3e81a0328c5fbbd4272d41df9ba57bb


ararslan:
Looks like you still have to push the v0.5.0 tag to the package repo.

ChrisRackauckas:
Thanks for notifying me. The tag has been pushed.

ChrisRackauckas:
Does Travis need to be re-run?

ararslan:
Yep. You can close and reopen the PR to trigger a new build.

PR #6845, opened 2016-10-23
Tag SGP4 v0.1.1 [https://github.com/crbinz/SGP4.jl]

crbinz:
New build process


SGP4/versions/0.1.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+PyCall 1.6
+Dates
+Compat
```
ararslan: It's usually a good idea to specify a minimum version for `Compat` that includes the features you need.

ararslan:
You can ignore the current Travis failure about the commit SHA not matching; the SHAs actually do match, so the check is wrong.

crbinz:
I've added the Compat version as requested.

tkelman:
`Compat.ASCIIString` was introduced a bit earlier than 0.9.0, there's a note on the Compat readme with instructions on how to find more specific versions.

crbinz:
Thanks, that's useful.

PR #6846, opened 2016-10-23
Tag Reel v0.2.0

tawheeler:
A new tag for Reel!


ararslan:
The `requires` file here needs a Julia entry that specifies a minimum version necessary for the package to work. That's what's causing the Travis failure.

PR #6848, opened 2016-10-24
Nonlinear squares fitting + PrettyNumber

RedPointyJackson:
General fitting package + number prettificator.


nalimilan:
The last commit shouldn't be there: you should rebase on `origin/metadata-v2` (instead of merging).

Naming is up to you, but I think you'll get more users if you find a more explicit name than Fitotron. :-)

Also, regarding PrettyNumber, the maintainer of [Humanize.jl](https://github.com/IainNZ/Humanize.jl) might be interested in taking similar functionality. Better avoid creating packages with only one function when possible.

thehellion:
I would say it's a fitting name.

RedPointyJackson:
Thank you @nalimilan, I will try to make a new pull request correcting that. Regarding `PrettyNumber`, I think it will be more correct to improve `Humanize.jl`; like you said it's better to concentrate funcionality in less packages.

nalimilan:
Note that you can also update a PR in place using `git push --force`.

RedPointyJackson:
Thanks, you are very helpful! Nevertheless I prefer to make a new one so I can share `Fitotron` only and keep aside `PrettyNumber` for the moment.

PR #6850, opened 2016-10-24
Register ProximalOperators v0.0.1 [https://github.com/kul-forbes/ProximalOperators.jl.git]

lostella:


dpsanders:
It's hard to guess what Prox.jl might be about. Maybe ProximalAlgorithms.jl or similar would be a clearer name?

lostella:
@dpsanders: "ProximalAlgorithms.jl" would not be appropriate in my opinion, because "proximal algorithms" usually refer to optimization algorithms which use the proximal operator that the package implements. On the other end, "prox" is the commonly used notation for such proximal operator, and is widely known in the optimization community, hence "Prox.jl".

But here are some viable options I see so far:
- Proximal.jl
- ProximalOperators.jl

Prox/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: see where it says "lostella wants to merge 1 commit into `JuliaLang:metadata-v2` from `lostella:pull-request/6beb70ae`" right under the title? That means this PR is associated with the branch named `pull-request/6beb70ae` in your fork of this repository at https://github.com/lostella/METADATA.jl. You can make modifications to these files and push additional commits to that same branch, and they will appear here in this PR. If you rename the repository, change the URL and move the files accordingly. You'd also likely need to change the sha1 to point to a commit where the file names and module name have all been changed.

dpsanders:
`ProximalOperators.jl` seems to be a good option.

lostella:
Changed name and added minimal version of Compat in REQUIRE.

ProximalOperators/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat 0.7.13
```
tkelman: that'll give you deprecation warnings on 0.4, won't it?

PR #6853, opened 2016-10-24
Tag Reel v0.2.0

tawheeler:
Travis added, tests pass on Julia 0.4, 0.5, and release.
Added minimum requirements.


Reel/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat 0.9.2
```
tkelman: This package isn't using cholfact is it? https://github.com/JuliaLang/Compat.jl/commit/972b1c78b08e6f1dca9d790d2cbbe13850161890 looks more like the newest thing being used from what I can tell.

PR #6854, opened 2016-10-24
Tag CRlibm v0.3.0 [https://github.com/dpsanders/CRlibm.jl]

dpsanders:
0.5 deprecations, and include the source code of the `CRlibm` library in the repository rather than downloading it.

### DIFF
--- a/CRlibm/versions/0.3.0/requires
+++ b/CRlibm/versions/0.3.0/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.8.8
--- a/CRlibm/versions/0.3.0/sha1
+++ b/CRlibm/versions/0.3.0/sha1
@@ -1,0 +1,1 @@
+00e8f51aa20e3ab35b59ebebdfb6d074018d1ef3


tkelman:
> include the source code of the CRlibm library in the repository rather than downloading it

Why? From a licensing perspective, downloading at install time keeps things more clearly separate than committing them both as part of the same project.

dpsanders:
The download site has been down several times. Also, the upstream project is now deprecated and won't be updated (see the home page).

ViralBShah:
I don't see how downloading vs. having in the repo has anything to do with licensing. The licenses have to be clearly specified irrespective of how the sources are obtained. One could make other arguments for downloading - but in this case it doesn't make sense given the upstream issues.

tkelman:
If you're checking in a copy of the sources in your repo, then you're clearly redistributing them and need to obey the terms of the original project's license as far as redistribution is concerned.

dpsanders:
According to http://softwareengineering.stackexchange.com/questions/86142/what-exactly-do-i-need-to-do-if-i-use-a-lgpl-licenced-library : 

"If you are dynamically linking the LGPL library then you can either distribute you application alone, without the library and tell people where to download it from and how to include it, to use it. Or you can include a copy of the library binaries and its source with your application."

Which aspect of "need to obey the terms of the original project's license" are you referring to, @tkelman ?  To be honest it is not clear to me what the correct solution is, but it certainly seems to be allowed, or even encouraged, to redistribute the source.

dpsanders:
See section (4.d.0) of the license text: https://www.gnu.org/copyleft/lesser.html

tkelman:
4.b, primarily

dpsanders:
I have added the text of the `CRlibm` license and the LGPL to `LICENSE.md`. Is this sufficient?

PR #6856, opened 2016-10-24
NaiveBayes 0.3.1 and Boltzmann 0.5.0

dfdx:


NaiveBayes/versions/0.3.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+Distributions 0.6+
```
tkelman: best to use actual version numbers here, the + isn't all that useful as a lower bound

dfdx: Ah, leftover from the times I didn't really understand version specs. Should I update and re-submit committed tags or is it ok to fix it in master only?

tkelman: can fix it in-place just in this metadata branch, as long as the same change gets made on master prior to the next tag

dfdx: fixed, thanks for the tip

PR #6858, opened 2016-10-25
ParallelAccelerator and CompilerTools v0.2.0

ehsantn:


ParallelAccelerator/versions/0.2.0/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.4
+CompilerTools
+DataStructures
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need - there's a note about this on the Compat readme

ehsantn: We only use Compat for OS check so I'm not sure if we need it. Maybe it would be useful for CompilerTools.

PR #6867, opened 2016-10-26
Register+tag BasisMatrices.jl, bump CompEcon and QuantEcon

spencerlyon2:
This is a set of changes that breaks out a chunk of CompEcon.jl into a new package BasisMatrices.jl

Some other tools also went into QuantEcon.jl

I should also note that I transferred CompEcon.jl from my personal github account to the QuantEcon organization. Should I adjust the url for CompEcon to reflect this?


tkelman:
What licenses are the sources used for BasisMatrices under? If they're derived from other code, either need permission to re-license the translation, or should fully attribute the license of the original code. Same with anything translated from scipy.

spencerlyon2:
No code from scipy here.

The original code is not licensed, but we obtained written permission from the original authors to create the derivate work and release under a BSD-3 license.

spencerlyon2:
The one file using code based on scipy routines now includes the scipy license at the top.

I _think_ we are good to go on that front, but if there is anything else I should be doing please let me know.

tkelman:
Sounds good. I'd make a note that it was relicensed with permission.

spencerlyon2:
I added a note at the top of the readme for both CompEcon and BasisMatrices.

Thanks for checking on that stuff -- open source licensing is always a bit mysterious so it's helpful to have someone else checking in on it.

PR #6874, opened 2016-10-26
Tag NamedArrays v0.5.3 [https://github.com/davidavdav/NamedArrays.jl]

davidavdav:
Fixing several issues and including some PRs.


NamedArrays/versions/0.5.3/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

davidavdav: You are right.  I'll close and re-submit.

PR #6879, opened 2016-10-27
Tag ZipFile v0.3.0 [https://github.com/fhs/ZipFile.jl]

fhs:
update https://github.com/fhs/ZipFile.jl/issues/29

### DIFF
--- a/ZipFile/versions/0.3.0/requires
+++ b/ZipFile/versions/0.3.0/requires
@@ -1,0 +1,2 @@
+julia 0.4
+Compat 0.8.4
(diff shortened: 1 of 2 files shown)


ZipFile/versions/0.3.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat 0.7.18
```
tkelman: `Compat.view` needs 0.8.4

fhs:
updated with new requires

PR #6881, opened 2016-10-27
Register PlyIO [https://github.com/FugroRoames/PlyIO.jl.git]: v0.0.1

c42f:


c42f:
By the way @SimonDanisch, you might be interested in this one - a generic ply IO package, which should be able to handle any elements or properties you throw at it.

PR #6883, opened 2016-10-27
Tag BioMedQuery v0.2.0 [https://github.com/bcbi/BioMedQuery.jl]

mirestrepo:
Drop support for Julia 0.4


tkelman:
Missing `DataStructures` in REQUIRE?

BioMedQuery/versions/0.2.0/requires
```diff
@@ -12,3 +12,4 @@ XMLconvert 0.0.1
 MySQL 0.0.2
 DataFrames 0.8.3
 NullableArrays 0.0.8
+DataStructures 0.4.6
```
tkelman: good, be sure to also do this here https://github.com/bcbi/BioMedQuery.jl/blob/master/REQUIRE before the next tag (don't need to worry about changing the sha or git tag though)

mirestrepo: Done. Thanks!

PR #6886, opened 2016-10-27
Pull request/1e8311fa

phaverty:
Requiring julia 0.5
Added tapply functionality with docs and tests


tkelman:
Need to remove https://github.com/phaverty/RLEVectors.jl/blob/e9f73cd54b264d55a5512ffc5353279b2f96e7aa/src/RLEVectors.jl#L5-L6 if `Compat` isn't in `REQUIRE`

PR #6890, opened 2016-10-28
Register NamedColors [https://github.com/JuliaGraphics/NamedColors.jl…

oxinabox:
….git]: v0.1.0
### Http URL for the repository of your proposed package:

```
https://github.com/JuliaGraphics/NamedColors.jl
```
### Proposed name of the package:

```
NamedColors.jl
```
### One-line description:

```
More named colors than you ever thought you needed
```
### Alternative package names:

> Give alternative package names (applicable if your package is not a wrapper of another package).

ColorNames
DescribedColors
### Documentation:

>  You are required to provide adequate documentation.
>  Confirm whether documentation is provided in README.md or give a link to the separate documentation

In README.md and 2 example Jupyter notebooks. Fairly extensive use of docstrings, for what this is.
### Tests and continuous integration:

> Your package is required to have tests in the test/runtests.jl file,
> and to be linked to a continuous-integration service. Provide the relevant information.

Travis is enabled on the repository. Tests pass on 0.5 and nightly
AppVeyor is enabled on the repository. Tests pass on 0.5 and nightly

```
96% coverage
```
### Related packages in the Julia ecosystem:

> Describe the similarities and differences with related packages.

Color.jl provides a colorant macro (see below), that accepts named colors. But it only does about 660 named colors. NamedColors  does about 3,500 (I have aspirations towards 8,000). Color.jl is mostly focused around color conversion/manipulation. 
### Macro stealing

 Arguably this engages in "macro stealing".
It replaces all uses of Color.jl's `colorant` string macro, with it's own.
(with a more specific type signature requirement)
But to be fair, that is nearly the entire point of this package.
It defers to the original functionality of the `colorant` macro,
unless that results in an error due to the color not being found.
In the case the color is not found, then it swings in, and search's it's own much larger list of color names.

I could put in my own string macro: `namedcolor`, thus making them separate.
But I figure that if someone is setting colors on a plot, and they are mixing named colors, with colors in hex etc, they don't really want to be switching macros.
The point is to be able to avoid thinking about the colors, and just use the first name the pops into your head.
### Doesn't loading 3500 colors into an global variable increase startup time?

Surprisingly not.
Loading those colors only takes about 0.04 seconds on my PC.
### Some of the licenses on the data files are scary

Yes, yes they are (another reason that i don't want to merge this into Colors.jl).
One should be aware that you can not copyright a color.
You can trademark it, but that is a different thing, and not a concern here.
Similarly you can trademark its name, but that is still not a concern.

What is a concern is that you can copyright the lists as a whole.
And most of the lists (except XKCD) have indeed been copyrighted as such, but licensed for use.
The lists I am distributing are being distributed according to those licenses.
The scariest license is actually the CC-SSA from wikipedia for the Crayola color names.

But in it is all fine, the licenses are on the data, not on the code.
The code is not derived from the data -- the package as a whole is a "Collection" incorperating several works.
roughly speaking, CC-SSA imposes it's viral restrictions on Derivative Works, not on Collections (the license actually goes into some detail about this).


tkelman:
@timholy opinions on this using the same macro name?

Seems to me like you'd be asking for code that gives runtime errors without it being obvious that it depends on NamedColors to work.

timholy:
As long as the ones that have the same names also return the same values, I have no objections.

oxinabox:
>As long as the ones that have the same names also return the same values, I have no objections.

They do.
Because this only searches the extended list of names if Color.jl's `colorant` string macro gives an error.

PR #6892, opened 2016-10-28
Tag GMT v0.0.2 [https://github.com/joa-quim/GMT.jl]

joa-quim:


GMT/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

joa-quim: Sorry, don't understand.
My local copy as well as https://github.com/joa-quim/GMT.jl/blob/master/REQUIRE both have those two lines, so why is it missing and where?

tkelman: this line has no version dependency on Compat, which says even the very first released version of Compat is acceptable

joa-quim: Ah, right. Hadn't think on that.
So I updated [REQUIRE ](https://github.com/joa-quim/GMT.jl/blob/master/REQUIRE)Is this enough?

joa-quim: Tony, I appreciate your help but this is damn complicated and requires a lot of work when I only want to make update my package. 
I have no idea on how to update my fork of metadata to a particular commit when I'm not even able to see the GMT package when scolding my fork METADATA's repo.
And I have no idea the minimum compat version required to run GMT.jl from Julia 0.4. I think I have no big special needs but there were tons of deprecation messages so I added Compat and it happens to currently be 0.9.3

tkelman: You're using `unsafe_wrap` which was introduced in 0.8.0: https://github.com/JuliaLang/Compat.jl/commit/d22506770d4b52bd2409fc70fa879e707b573949

From looking at your diff, it doesn't look like you're using anything newer than that.

Edit the file here - https://github.com/joa-quim/METADATA.jl/blob/pull-request/09bebd95/GMT/versions/0.0.2/requires

PR #6893, opened 2016-10-28
Register COBRA.jl - [PR/f3f0464f]

laurentheirendt:


tkelman:
You should be listing test requirements in `test/REQUIRE`, instead of adding them only on CI.

tkelman:
CC BY-NC-ND is also a fairly restrictive license, and not OSI approved: https://opensource.org/licenses/alphabetical

rmtfleming:
Is Attribution-NonCommercial-ShareAlike 4.0 International acceptable? It is not on the OSI list, but the OSI list is not compatible with valorisation of research investment, which is a key criteria for our research funding.

laurentheirendt:
Thanks @tkelman. Test requirements are now listed in `test/REQUIRE`.

tkelman:
NC-SA is better than NC-ND, but restricting commercial use is unfortunate.

johnmyleswhite:
To add to Tony's concerns, I think there's a reasonable argument that the Julia package manager should not make it easy to install software that they cannot use without violating laws. If we allow the package manager to distribute software that generally prohibits commercial use, we may want to default to prohibiting installation of that category of software until users explicitly opt in.

laurentheirendt:
Thank you for your comments, @tkelman and @johnmyleswhite.

The license has been changed to GPL v3.

PR #6900, opened 2016-10-29
Tag Polyhedra v0.1.1 [https://github.com/blegat/Polyhedra.jl]

blegat:


tkelman:
You can't have an unregistered package LRSLib in your `test/REQUIRE`. Register that first?

blegat:
It is kind of a circular dependency. I had to change a few things in `Polyhedra` to make `LRSLib` work so currently (but that is temporary), `LRSLib` needs the latest version of `Polyhedra` to work.
I was thinking to first tag `Polyhedra` then update `.travis.yml` and `appveyor.yml` of `LRSLib` then register `LRSLib`. Then I will also be able to update `.travis.yml` and `appveyor.yml` of Polyhedra to use the registered version of LRSLib.

I can also register them at the same time but that would require either having failing test when they are registered or having `Pkg.clone`s in the `*.yml`s.

What is better ?

blegat:
Let's do something simple. I'll just remove `LRSLib` from `test/REQUIRE` and add it back when `LRSLib` is registered.

blegat:
Done :)

PR #6901, opened 2016-10-29
Tag PlotlyJS v0.5.1 [https://github.com/spencerlyon2/PlotlyJS.jl]

spencerlyon2:
Update version of plotly.js

Added ohlc, candlestick, scatterternary, and pointcloud trace types.

Added `purge!`, `to_image`, and `download_image` routines to match plotly.js function API


PlotlyJS/versions/0.5.1/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.5-
```
tkelman: when you're ready to no longer support prereleases, drop the trailing dash

PR #6907, opened 2016-10-29
Tag v0.0.3 OrdinaryDiffEq.jl and v0.0.2 StochasticDiffEq

ChrisRackauckas:
This patch should make the DifferentialEquations.jl metapackage not throw any errors or warnings anymore. It updates the premade_problems to use new macros from ParameterizedFunctions which use less derivatives.


StochasticDiffEq/versions/0.0.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+ParameterizedFunctions 0.1.0
+Parameters 0.5.0
+ChunkedArrays 0.1.0
```
tkelman: still missing DiffEqBase

PR #6909, opened 2016-10-29
Tag ReverseDiffSource v0.3.0 [https://github.com/JuliaDiff/ReverseDiffSource.jl] 

fredo-dedup:
This new release adds : 
- compatibility to the released Julia 0.5 version (but compatibility with 0.4 is abandoned)
- enables precompilation
- changes the API of `rdiff` : values for variables are no longer necessary, only types need to be provided. The package makes use of Julia's type inference to choose the differentiation rules instead of recalculating actual values of the whole graph. This should bring better performance, simpler code and open up new simplifications possibilities for the generated code.


ReverseDiffSource/versions/v0.3.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Compat
```
tkelman: should declare a minimum version that implements the features you need

fredo-dedup:
I added a version to Compat in requires, but also had update accordingly the REQUIRE in the package, ....which lead to a new commit, ... for which I had make a new tag .... and finally forced me change the sha1 in this PR. I hope I haven't forgotten anything  ;-) 

Noticed too that I had named the directory version "v0.3.0" instead of "0.3.0",  but this is corrected now.

scidom:
Thanks @fredo-dedup, will look into this in relation to the required updates in Klara in the next few days.

PR #6910, opened 2016-10-29
Tag BioEnergeticFoodWebs v0.1.1

tpoisot:
- require `0.5` 
- updated tests (and fixed a bug)
- updated documentation


tkelman:
Since this is dropping support for Julia 0.4, it should use a new package minor version

tpoisot:
If possible, no. We have a paper un review with this tag and this version, so we'd rather not change the version number now.

tkelman:
You can simply make a new copy of the same sha and tag it as 0.2.0 and that should work, leaving 0.1.1 in place unmodified is not the greatest from a versioning perspective but should be able to pass the checks here if there's any newer minor version tag.

tpoisot:
Can't I just re-add support for 0.4 in this tag? We have a doi associated with this version, which is linked to a paper, and it would be a lot of trouble for me to change things at this point.

tkelman:
Did this sha work on 0.4? If so you can adjust the requires file here. If not, you could make a new 0.1.2 tag that does support 0.4.

tpoisot:
Ok, I'v added 0.4 to this sha, and will do a 0.2.0 later that drops 0.4 support. Thanks.

PR #6911, opened 2016-10-30
Tag Weave v0.2.0 [https://github.com/mpastell/Weave.jl]

mpastell:


mpastell:
I just noticed a rather nasty bug in this version. What's the right thing to do here? Merge this one and release a fix right after it? It seems I can't tag v0.2.1 because Julia seems to think there is no 0.2.0.

tkelman:
How nasty of a bug? Since this hasn't been published, could also remove the 0.2.0 tag and replace the sha here.

mpastell:
Too nasty to publish the package... See my comment on https://github.com/mpastell/Weave.jl/pull/49 . I will remove this tag and retag the corrected version.

PR #6915, opened 2016-10-30
Register ParameterEstimation.jl

ChrisRackauckas:
ParameterEstimation.jl is an addon to the JuliaDiffEq suite parameter estimation from differential equation models. The documentation will be contained in the DiffEq docs


ChrisRackauckas:
This requires RecursiveArrayTools.jl to go through.

tkelman:
ParameterizedFunctions is missing from `test/REQUIRE`

tkelman:
This is also a fairly generic name for a pretty specialized, small piece of code.

ChrisRackauckas:
>This is also a fairly generic name for a pretty specialized, small piece of code.

Small code, yes. I'm just tagging early and often, for now to make it easier for others to play with and test https://github.com/JuliaOpt/Optim.jl/issues/297. I've already shown that the code this is part of won't be small and will be quite feature-filled in a few months.

However, it is specialized: it's parameter estimation for differential/difference equations, right now only for ODEs but code for SDEs and PDEs are in the pipeline. When I Google parameter estimation, everything that pops up is for some kind of differential or difference-based model, so I thought it would be redundant to add a qualifier for that. However, I am open to suggestions.

ChrisRackauckas:
DiffEqParamEstim sounds like it's reasonable given some chats. What do you think of this?

pkofod:
> When I Google parameter estimation, everything that pops up is for some kind of differential or difference-based model, so I thought it would be redundant to add a qualifier for that.

I think this is an artifact of your of search history being used by google (to prevent you from broading your horizon I guess). I don't get any hits about differential or difference-based models.

ChrisRackauckas:
>I think this is an artifact of your of search history being used by google (to prevent you from broading your horizon I guess)

Damn Google. When I go incognito it shows parameter estimation stuff for regressions and neural networks mixed in there as well (though neural nets and ODEs actually have a correspondence relation....). But yeah, I am fine with a more DiffEq naming, just let me know what's acceptable please.

tkelman:
DiffEqParamEstim sounds fine to me, and hard to confuse with anything

tkelman:
Though your other packages have DiffEq at the end, right? Either way.

ChrisRackauckas:
The solver components end in DiffEq (OrdinaryDiffEq, StochasticDiffEq, etc.), the extra tools prefix it (DiffEqDevTools, DiffEqBase, etc.).

I'll change the name and let know.

PR #6916, opened 2016-10-30
Register PyDSTool.jl

ChrisRackauckas:
PyDSTool.jl is a wrapper for the Python package PyDSTool. It's a low-level wrapper, providing some helper functionalities to make it easier to interface with PyDSTool be independent of the DifferentialEquations.jl ecosystem. A package which is JuliaDiffEq specific, Bifurcate.jl, will come soon and will make the ecosystem interact with PyDSTool easier, but this allows those who are interested in directly working with PyDSTool to do so.

Note that the Travis tests fail due to a known segfaulting problems with PyCall: https://github.com/JuliaPy/PyCall.jl/issues/327 and https://github.com/JuliaPy/PyCall.jl/issues/328. When these nondeterministic segfaults don't occur, the tests pass locally, but since they are nondeterministic there's no guerentee of them passing. Thus the reliable passing of the tests is waiting on an upstream fix.


tkelman:
Since you do `using Conda` in `deps/build.jl`, you should list Conda explicitly in `REQUIRE`.

tkelman:
Your tests should probably be checking the results of what they run?

ChrisRackauckas:
I can't think of a good test other than the plotting (which is why there's a bunch of plot commands commented out). I guess I can add spot checking of some values given that the plots show it's alright.

ChrisRackauckas:
Another thing to note with the tests is that more extensive tests will come from [Bifurcate.jl](https://github.com/JuliaDiffEq/Bifurcate.jl). That makes it easy to have it plug into the whole DiffEq machinery and pull a bunch of test problems from there. PyDSTool.jl is a little too close to the syntax of PyDSTool which doesn't play well with others (all functions and such are defined as strings...) which is why there's a "middle-man" to make it play well with the rest of JuliaDiffEq (as one backend to Bifurcate, the others looking to be AUTO and Julia-based methods, likely from NLsolve if they end up going through with their homotopy methods). Although Bifurcate.jl does already work, the PyCall segfaults make it tough to develop/test the whole stack together right now, which is why I'm holding off on registering that.

tkelman:
This was still missing Conda in REQUIRE

ChrisRackauckas:
I'll put a patch in.

PR #6919, opened 2016-10-31
Tag Weave v0.2.0 [https://github.com/mpastell/Weave.jl]

mpastell:


Weave/versions/0.2.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

PR #6922, opened 2016-10-31
Tag TimeZones v0.2.4 [https://github.com/JuliaTime/TimeZones.jl]

omus:


tkelman:
looks like you forgot to push the git tag?

omus:
I did forget to push the tag. Should be pushed now. Can you restart the CI or do I need to make a new PR?

tkelman:
there was another one http://pkg.julialang.org/logs/TimeZones_0.5.log

I think this package should strongly consider downloading a fixed named version of the db, and make a new Julia package tag for each update. Otherwise this isn't really reproducible over time.

omus:
I'll take that under consideration. As of v0.2.5 I remove the problematic test with testing for the number of abbreviations which has been the reason for these failures.

PR #6923, opened 2016-10-31
Register and tag RigidBodyTreeInspector

rdeits:
RigidBodyTreeInspector.jl is a simple tool for visualizing and manipulating the structure of a kinematic tree (for example, a humanoid robot). From a RigidBodyDynamics.jl description of a robot, it generates visual geometry to show the basic skeleton of the robot, along with (optionally) the robot's inertial properties or visual meshes. 

When used inside an IJulia notebook, it creates a set of sliders to interactively manipulate the configuration of the robot. 

It relies on functionality from:

* [tkoolen/RigidBodyDynamics.jl](https://github.com/tkoolen/RigidBodyDynamics.jl) to parse and represent kinematic structures
* [RobotLocomotion/director](https://github.com/RobotLocomotion/director) for 3D visualization
* [rdeits/DrakeVisualizer.jl](https://github.com/rdeits/DrakeVisualizer.jl) to communicate with `director`
* [rdeits/PyLCM.jl](https://github.com/rdeits/PyLCM.jl) for message-passing and communication

(note that PyLCM is not listed as an explicit dependency because it is *only* used through the higher-level API provided by DrakeVisualizer.jl, which does depend directly on PyLCM).


tkelman:
It's usually discouraged to extend base methods on base types as in https://github.com/rdeits/RigidBodyTreeInspector.jl/blob/dfbce52dc2e418f0dd6084481f7e47c935310e50/src/animate.jl#L1, as that'll change the behavior of unrelated code

tkelman:
Assuming this file https://github.com/rdeits/RigidBodyTreeInspector.jl/blob/dfbce52dc2e418f0dd6084481f7e47c935310e50/test/runtests.jl#L27 will be installed at a particular path under `ENV["HOME"]` also seems fragile and unlikely to pass for any users (or pkgeval) who attempt to run `Pkg.test`. Your tests should probably handle obtaining any example resources they're going to load, and put them in a package-local path.

tkelman:
`CoordinateTransformations` and `ColorTypes` should be listed explicitly in `REQUIRE` if they're being imported

rdeits:
@tkelman thanks, and good points all around. I've eliminated the extension `Base.one(::Array)`, eliminated the test dependency on the file you mentioned by including a much smaller file of the same type in the repo, and updated the REQUIRE in the package and in METADATA

PR #6925, opened 2016-10-31
Register InformationMeasures [https://github.com/Tchanders/Informatio…

Tchanders:
…nMeasures.jl.git]

InformationMeasures.jl implements a range of multivariate information measures, from joint entropy and mutual information to the less well-known total correlation and partial information decomposition, among others. InformationMeasures.jl has a flexible API that can accept raw (continuous) data, discrete data, or probabilities. A number of different methods are offered for discretizing data and for estimating probabilities.


tkelman:
Would be good to enable Travis CI here.

PR #6927, opened 2016-10-31
Register DiffBase [https://github.com/JuliaDiff/DiffBase.jl.git]

jrevels:


DiffBase/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version that implements the features you need

tkelman:
Would AutoDiffBase be any clearer to distinguish it from the recently registered, similarly named DiffEqBase?

jrevels:
> Would AutoDiffBase be any clearer to distinguish it from the recently registered, similarly named DiffEqBase?

Ah, I missed that. DiffBase is intended to extend to non-automatic differentiation packages as well, such as FiniteDiff. DifferentiationBase, maybe? Though that to me just seems needlessly verbose, especially since I was purposely conservative on `export`ing package bindings. Typing out `DifferentiationBase._____` everywhere could be annoying.

jrevels:
Added a minimum version to Compat, does everything look okay now?

I'm mainly against changing the name..."differentiation" and "differential equations" are so linguistically/conceptually coupled that I feel like most reasonable names could be construed as ambiguous. `DiffBase` is the obvious name for the shared library backing the `FiniteDiff`, `SymbolicDiff`, `ForwardDiff`, and `ReverseDiff` packages. Anyway, this naming ambiguity problem should (hopefully) go away once we have namespaces.

PR #6933, opened 2016-11-02
Tag DiffEqBase, OrdinaryDiffEq, ParameterizedFunctions, and DiffEqDevTools

ChrisRackauckas:
This propogates a refactor of RecursiveArrayTools out of DiffEqBase, the abstract definition of ParameterizedFunction and sensitivity related types into DiffEqBase, and the changes to reliant packages. Tests in this group are all dependent on the changes of each other, so they all pass locally on each other's master but not on Travis.


ParameterizedFunctions/versions/0.2.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+SymEngine 0.1.1
+DataStructures 0.4.6
```
tkelman: missing `DiffEqBase`

ChrisRackauckas: Fixed

tkelman: be sure to also add it here https://github.com/JuliaDiffEq/ParameterizedFunctions.jl/blob/master/REQUIRE before the next tag

ChrisRackauckas: Yes, I pushed that probably at the same time you were writing that 👍

PR #6934, opened 2016-11-02
Tag AlgebraicDiffEq, FiniteElementDiffEq, StochasticDiffEq

ChrisRackauckas:
This set of patches finishes the refactor of DiffEqProblemLibrary out of DiffEqBase. They are pretty much untouched except now the addition of DiffEqTestLibrary as a test requirement. (Should I, for tags like this in the future, mix this tag with the registration tag of DiffEqProblemLibrary? I split the this to two parts to show that tests pass, but having it as one patch may make it easier for you?)


tkelman:
the fewer packages per pr, the better

ChrisRackauckas:
Good to know. Thanks for clarifying.

PR #6939, opened 2016-11-02
Register ContMechTensors [https://github.com/KristofferC/ContMechTensors.jl.git]: v0.1.0

KristofferC:
Trying again :)


ContMechTensors/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: what minimum version?

KristofferC: Doesn't matter. I will not go back and retag and update minimum version for an obsolete Julia version.

tkelman: And if you do want to tag a version that supports 0.4, it's not necessary to re-tag or change the sha to modify version requirements, this file can be changed in place just in metadata.

KristofferC: I deleted the old tag and removed the version from the METADATA pr.

PR #6941, opened 2016-11-02
Tag SugarBLAS v0.0.2 [https://github.com/lopezm94/SugarBLAS.jl]

lopezm94:


tkelman:
Packages shouldn't extend base methods on base types as in https://github.com/lopezm94/SugarBLAS.jl/blob/bd8a8c51eb35d25726e5fead28d6d90f55ffc872/src/SugarBLAS.jl#L15 - that is changing the behavior of unrelated code depending on whether or not your package is loaded, hiding a deprecation instead of fixing it properly.

lopezm94:
Thank you for reviewing the code! Now it should be good to go.

PR #6942, opened 2016-11-02
Tag Weave v0.2.1 [https://github.com/mpastell/Weave.jl]

mpastell:


Weave/versions/0.2.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Compat
```
tkelman: should declare a minimum version that implements the features you need

mpastell:
Sorry, I forgot. I actually managed to remove Compat from depencies completely. I will close this and reopen with modified version.

PR #6949, opened 2016-11-03
Pull request/129a2c1b

spencerlyon2:
Renames a type in BasisMatrices.jl


CompEcon/versions/0.2.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+QuantEcon
+BasisMatrices
```
tkelman: minimum 0.2 then?

spencerlyon2: Yep, that's right. Will fix and push an update

PR #6952, opened 2016-11-04
new version for DASSL

pwl:


tkelman:
METADATA is frozen for package versions that support Julia 0.3. Unless you have a really pressing need to still support that old of a Julia version, can you change the minimum Julia version to 0.4 or newer and re-tag with a new package minor version?

tkelman:
The fact that the requires file content doesn't match the package's version suggests you're creating this manually. It's best if you didn't do that, and used `Pkg.tag` (on julia 0.4 or earlier) or `PkgDev.tag` (on julia 0.5 or later). Even if you're having authentication problems with publish (in which case you can push to your fork manually), you should use tag to handle creating these files.

PR #6955, opened 2016-11-04
Register AutoDiffSource [https://github.com/gaika/AutoDiffSource.jl.git] v0.0.1

gaika:


tkelman:
Looks good. How does this compare to the various other existing autodiff packages?

gaika:
https://github.com/JuliaDiff/ReverseDiffSource.jl is pretty close and has more features. This one is faster,  supports multiple return values, and since forward and backward passes are separate it can differentiate nested functions or be plugged in to your own gradient learning code with ease.

PR #6956, opened 2016-11-04
Pull request/cac954f6

danspielman:


tkelman:
The Travis failure is saying that the 0.0.2 sha doesn't match between the metadata tag here and the git tag in the repo.

PR #6959, opened 2016-11-05
Pull request/5a3ec5fe

montyvesselinov:
Various fixes and improvements


tkelman:
If Mads is going to import PkgDev (https://github.com/madsjulia/Mads.jl/blob/e14a05de4b7b5a659cbef50dd78328ff69a47656/src/MadsPublish.jl#L2), then it needs to depend on it in REQUIRE. Though since there are no versions of PkgDev that support Julia 0.4, that would require dropping 0.4 support (or making two branches, one for 0.4 and one for 0.5). Though the contents of the MadsPublish.jl file really shouldn't be a part of the package's computational API at all, rather it's just a convenience script that developers should include separately on their own.

RobustPmap/versions/0.1.10/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+JLD
```
tkelman: need to add `Compat 0.7.15` https://github.com/madsjulia/RobustPmap.jl/blob/2d350f47f7b622fe54b5d829abbf932e79576fd5/src/RobustPmap.jl#L35

montyvesselinov:
On Sat, Nov 5, 2016 at 8:38 PM, Tony Kelman <notifications@github.com>
wrote:

> Compat 0.7.15


Thank you very much! I added it.

-monty

tkelman:
closing this in favor of #6964

PR #6963, opened 2016-11-06
Fix SHA1s for DifferentialEquations.jl

ChrisRackauckas:
Let me start by saying that I realize this was a mistake and I don't know of a good way to backtrack.

DifferentialEquations.jl had balooned in size to 750MB due to animated gifs and Jupyter notebook files which were in the repository as documentation. So I looked up how to clear these files (since they are not part of the code, and are now available over at DiffEqTutorials, DiffEqBenchmarks, and DiffEqDocs) and successfully did so using BFG repo cleaner. However, soon after pushing this change, I realized that this changed the SHA1s on every tag commit, thus breaking compatibility with METADATA. I tried force pushing a previous clone but that did not fix the issue.

This PR fixes the SHA1s, fixing the issue.

I am sorry for the error. Since I was not touching any .jl file, I thought the cleaning would not change anything related to Julia (just the documentation). However, it's clear I was wrong and should've asked first.


tkelman:
> I tried force pushing a previous clone but that did not fix the issue.

It should have. What was wrong?

DifferentialEquations/versions/0.5.0/sha1
```diff
@@ -1 +1 @@
-eab10c60d3e81a0328c5fbbd4272d41df9ba57bb
+0f6830a7339e2e2634516ce7bac3f5f070cfb538
```
tkelman: considering https://github.com/JuliaDiffEq/DifferentialEquations.jl/commit/eab10c60d3e81a0328c5fbbd4272d41df9ba57bb looks fine on master, it appears the force pushing was fine, just your tags are wrong now since they're all from a different version of the history

ChrisRackauckas: Okay, I guess I tested too early. Now it all works: I just tried a brand new Julia installation and it installs/precompiles/works fine. I guess my last check was before the force push was complete and I freaked out because the clean install didn't work. So it looks like it's restored, but it's back up to 270MB. I will open up an issue to discuss this.

PR #6966, opened 2016-11-06
Viridis

tpoisot:
Viridis, Inferno, Plasma, and Magma colorscales for Gadfly


KristofferC:
What advantage does having Viridis as its own package as opposed to just a PR to Gadfly?

tpoisot:
@KristofferC modularity? Less burden of the Gadfly maintainer to maintain additional code?

KristofferC:
Having it in Gadfly would mean it would be tested continuously together with Gadfly instead of being liable to breaking / not working with the latest Gadfly version. For a small piece of code that is as tightly coupled to a package like this, my personal view is that it would be much better to live with the package that it actually enhances.

tpoisot:
I disagree -- the endgoal is to support other packages than gadfly.

lobingera:
I agree withTimothée that we need a modular solution for importing named colormaps (and the data).
(btw: that was the reason i started https://github.com/lobingera/ColorMaps.jl, which includes this data)

I disagree with Kristoffer that this is Gadfly/Compose centric and needs to be included there.

tkelman:
Should acknowledge the source and license of the data in your top-level license file in addition to in the code itself. Would also be good to write some actual tests and enable Travis CI.

tpoisot:
@tkelman is there a way to cleanly untag, or should I make a new tag with the tests?

tkelman:
A new tag would be fine

PR #6968, opened 2016-11-07
Register EzXML [https://github.com/bicycle1885/EzXML.jl.git]: v0.0.1

bicycle1885:
Another libxml2-based package with consistent APIs, automatic memory management, XPath support, etc.

https://github.com/bicycle1885/EzXML.jl


nalimilan:
Just curious, what's the difference with LightXML.jl? Could these packages be merged at some point, or should one replace the other?

bicycle1885:
There are many differences:
* Documents and XML trees will be garbage collected.
* Simpler types: `Document` and `Node` only.
* Attribute values can be accessed like a dictionary.
* Namespace support.
* XPath support.

Not sure whether people love EzXML.jl but I believe EzXML.jl will supersede LightXML.jl.

nalimilan:
OK, thanks. Then why not call it LibXML2.jl?

bicycle1885:
Because there is no reason to do so; EzXML.jl is not a legitimate successor of LightXML.jl and there is no shared code between them.

nalimilan:
Note I said LibXML2.jl, not LightXML2.jl. Using the name of the wrapped library is classic for Julia packages.

bicycle1885:
Ah, sorry. My eyes are slippy because it's already 7 p.m. here.
LibXML2 is okay, but I'm not willing to wrap all features of libxml2.

bicycle1885:
Ping. Is there any reason not to merge?

tkelman:
There is a pre-existing C library also called ezxml http://ezxml.sourceforge.net, might be a bit confusing that this is based on a different library?

bicycle1885:
Oh, no. I didn't realize it. Maybe I need to rename my package HardXML.jl or something. But the C library is not so common and looks almost abandoned (no updates for 5 years).

tkelman:
nevermind, should have checked the dates there

bicycle1885:
Thank you!

PR #6969, opened 2016-11-07

dysonance:
Still having to do this from the command line (PkgDev.publish still not working for me), so if any issues just let me know and I can jump on it.


tkelman:
It's discouraged to extend Base methods like `diff` on signatures where all the types are also from Base, as that can change the behavior of unrelated code depending on whether or not your package has been imported.

dysonance:
@tkelman Yeah I can see that. I will change it back. Should I go through the whole process of retagging and resubmitting the PR? Or is there a more efficient way of making that change?

tkelman:
You could either make it an 0.1.2 tag and add it to this branch with a new commit, or modify the existing 0.1.1 tag since it hasn't been published yet

dysonance:
@tkelman Did that do the trick?

tkelman:
yep, looks good

Temporal/versions/0.1.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
```
tkelman: ah but as soon as you don't need to support prereleases any more, would be good to drop the trailing dash here from REQUIRE - no rush at all though

nalimilan:
I've discovered these very interesting packages -- I was intrigued by the name Indicators.jl in the PR's title. I'd just like to point out that it would likely be easier to find if it's name was more explicit, like TimeSeriesIndicators.jl or FinancialIndicators.jl.

PR #6977, opened 2016-11-08
Tag HDF5 v0.6.7 [https://github.com/JuliaIO/HDF5.jl]

simonster:


simonster:
*sigh* this is why i try to avoid tagging packages

PR #6984, opened 2016-11-08
Tag MotoServer v0.2.0 [https://github.com/invenia/MotoServer.jl]

iamed2:
Adds do block syntax to get around https://github.com/JuliaLang/julia/issues/19258


tkelman:
Looks like Travis isn't liking your newly added tests?

iamed2:
Ah, I thought this would happen at some point. They switched from including the server package in `moto` to having an "extra" in `moto[server]` (which I didn't even know before was a feature of PyPI/pip). I'll fix up and make a new PR tomorrow.

tkelman:
If it ends up being travis-only changes needed then this tag would be alright, just keep us posted.

PR #6985, opened 2016-11-08
Tag Mimi v0.2.1 [https://github.com/anthofflab/Mimi.jl]

davidanthoff:
This should fix the current package test failures.

### DIFF
--- a/Mimi/versions/0.2.1/requires
+++ b/Mimi/versions/0.2.1/requires
@@ -1,0 +1,6 @@
+julia 0.4
+DataStructures
+Distributions
+DataFrames
+Documenter
+Compat 0.8.6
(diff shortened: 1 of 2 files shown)


tkelman:
Looks like this fails on travis with `setbestguess` not defined?

davidanthoff:
That is just a problem on master. [Here](https://travis-ci.org/anthofflab/Mimi.jl/builds/174560966) is the travis build that corresponds to the version tagged here.

The master travis build is also passing [now](https://travis-ci.org/anthofflab/Mimi.jl/builds/174561582) but we won't tag a new version from that for a while.

tkelman:
Ah my mistake, sorry for misreading that.

PR #6990, opened 2016-11-09
blegat/SemidefiniteModel -> JuliaOpt/SemidefiniteModels

blegat:
blegat/SemidefiniteModel -> JuliaOpt/SemidefiniteModels


SemidefiniteModel/url
```diff
@@ -1 +0,0 @@
-git://github.com/blegat/SemidefiniteModel.jl.git
```
tkelman: Deleting this is going to cause nasty errors if anyone has installed the package in the day it's been registered. Best to leave this in place, even if it won't be the name going forward.

tkelman: Basically SemidefiniteModel can be left registered with the existing tag and url file, the github redirect will still work but new tags will be under the new name. That's what happened when React.jl was renamed to Reactive.

blegat: Done :)

PR #6995, opened 2016-11-10
Tag StaticArrays v0.1.0 [https://github.com/JuliaArrays/StaticArrays.jl]

andyferris:


StaticArrays/versions/0.1.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5-
```
tkelman: as soon as you don't need to support prereleases any more, drop the trailing dash

PR #6998, opened 2016-11-10
Tag RecurrenceAnalysis v0.1.0 [https://github.com/heliosdrm/Recurrenc…

heliosdrm:
…eAnalysis.jl]


RecurrenceAnalysis/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Distances
+Compat
```
tkelman: best to declare a minimum version dependency that implements the features you need

heliosdrm: Ok, done in a separate pull request

PR #7000, opened 2016-11-11
Tag PowerModels v0.2.2

ccoffrin:


PowerModels/versions/0.2.2/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4.6
+JSON
+MathProgBase 0.5.4
+JuMP 0.14.0
+Compat
```
tkelman: it's best to declare a minimum version dependency that implements the features you need

ccoffrin: Thanks for the tip.  Has been added to master, https://github.com/lanl-ansi/PowerModels.jl/pull/39, will be in the next release.

tkelman: should be added in metadata as well

PR #7014, opened 2016-11-12
Tag DifferentialEquations v1.0.0 [https://github.com/JuliaDiffEq/DifferentialEquations.jl]

ChrisRackauckas:
This tag finalizes the modularization change https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59 . It should be merged after https://github.com/JuliaLang/METADATA.jl/pull/7006 . It changes DifferentialEquations.jl to a metapackge which re-exports the functionality of its components. 

While that itself sounds very simple, it still has a major purpose within the JuliaDiffEq ecosystem as it is the "batteries included - everything just works" MATLAB-like toolbox. Some of the features require fitting together multiple packages and would lead to an awful user experience, but by using the metapackage everything in the documentation (in [DiffEqDocs.jl](https://juliadiffeq.github.io/DiffEqDocs.jl/latest/)) works seamlessly by `using DifferentialEquations`. However, in this new setup, users and developers are also given the option to use only the functionality they need, which was a highly requested feature since the package grew so large!

In addition, this distributed but still unified interface will make it easier for contributors to join the ecosystem. A 30,000 line of code package presents itself as a massive wall to new developers. However, in this new setup, developers will be able to "join the ecosystem" and have their algorithms available from the common interface via a standard plug-in interface (and moving their package to JuliaDiffEq for maitanence).  This is all explained in [DiffEqDevDocs.jl](https://juliadiffeq.github.io/DiffEqDevDocs.jl/latest/) (still in development))

Not only will this drastically lower the entrance-barrier, but it will also distribute the interface components to the individual solver packages. This moves the "common interface" code out, with a PR to Sundials already in motion, and code already being distributed to ODEInterface and ODE.jl. This will effectively end the conditional dependency problem DifferentialEquations.jl once had.

The end result is more features, more scalable development, easier for contributors, a method to get rid of the conditional dependencies, and as always, faster/more performant solving of differential equations.


aviks:
HI Chris, 

I noticed that compared to earlier tags for this package, you have removed minimum versions for the dependencies. Is that on purpose? Just checking. 

Also, this package no longer needs Compat?

ChrisRackauckas:
The minimum versions were gotten rid of because the package completely changed from implementing 30,000 lines of code to a metapackage at the center of a modular ecosystem. To understand it, you can look at the entire source in 3 seconds: https://github.com/JuliaDiffEq/DifferentialEquations.jl/blob/master/src/DifferentialEquations.jl .

The full explanation of why this is the case and what is gained is in JuliaDiffEq/DifferentialEquations.jl#59 that I linked. I have not put minimum versions on any of these packages yet (though they have minimum versions on their parts) because likely a few patches will need to be done, and so I will put them in when everything stabilizes to force them to the newer stabler versions. Until then, keeping it open is a sign to me that that dependency is "not fully patched up" yet.

This is one part of the large change (nearing the end though!), and will likely have a few more parts.

tkelman:
Not having lower bounds means "works with any tagged version." If that is not the case, then you should have bounds.

ChrisRackauckas:
That is the case.

PR #7017, opened 2016-11-12
Tag ERFA v0.2.0 [https://github.com/JuliaAstro/ERFA.jl]

nirinA:


aviks:
METADATA no longer accepts packages with a requirement of `julia 0.3`. This is because most developers do not have a 0.3 version to test on. If you really want to tag this with `julia 0.3`, please assert that you have tested the package on julia versions 0.3, 0.4 and 0.5. Otherwise, please change the minimum julia version requirement to 0.4.

ERFA/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.3
+BinDeps
+Compat
```
tkelman: it's also best to declare a minimum version dependency that implements the features you need

nirinA:
@aviks : i've changed the required julia version to 0.4
@tkelman : added versions for BinDeps and Compat
hope they are correct now.

tkelman:
those aren't correct minimum versions unless you're relying on a feature that was first released in that version. very latest is too conservative and not accurate.

nirinA:
i'll change to BinDeps 0.3.21 and Compat 0.7.10. these versions are that currently running on my julia 0.4.2 installation and with ERFA.jl tagged 0.1.2, so i think they should be right.

ararslan:
Now that the tags have been pushed upstream, can someone restart Travis here?

nirinA:
thanks!

PR #7022, opened 2016-11-13
Update Juno stack

pfitzseb:
Please do **not** merge this once approved, but rather ping me or someone else from the Juno org so we can time the release of the Atom packages with merging this (Juno will break again if we don't keep Julia and Atom packages in sync). Thanks!


pfitzseb:
Alright, if you're around this can be merged, @tkelman. :) I'll tag the Atom packages right afterwards.

pfitzseb:
Thanks, @aviks!

PR #7023, opened 2016-11-13
Tag Klara v0.6.2 [https://github.com/JuliaStats/Klara.jl]

scidom:
Updates to make Klara work with Julia v0.5

### DIFF
--- a/Klara/versions/0.6.2/requires
+++ b/Klara/versions/0.6.2/requires
@@ -1,0 +1,6 @@
+julia 0.5
+Distributions 0.4.7
+Formatting 0.1.4
+ForwardDiff 0.2.0
+ReverseDiffSource 0.2.1
+StatsBase 0.7.4
--- a/Klara/versions/0.6.2/sha1
+++ b/Klara/versions/0.6.2/sha1
@@ -1,0 +1,1 @@
+ccb5bea54e1bf5c34290835323ebafafe4a51710


scidom:
The checks fail, according the error messages, because I tagged v0.6.2 of Klara instead of v0.7.0 as proposed by the system (it seems this proposal is based on the fact that I now support only Julia v0.5). The problem is that I have my own system of version tagging, and I have my issue and milestone management in Klara settled accordingly. Can we proceed with Klara v0.6.2 please? This will protect my time from having to readjust my issue and milestone tracking, plus it will give me the space of developing Klara as it feels fit. Thanks

tkelman:
The previous tagged versions support Julia 0.4, this doesn't. That's a breaking change with respect to Julia 0.4 users. How is re-tagging this as 0.7.0 possibly a burden?

tkelman:
Alternately, if this still supports Julia 0.4, could just change the lower julia version bound back to 0.4 then this would be fine as 0.6.2.

scidom:
Hi @tkelman, what I was trying to say is that I had labelled the issues and milestones in Klara, and would take a little more extra time/work to get change these. A second matter is that I would like to be given the space as a developer to sort out the tagging of versions the way it suits me, as it used to be.

If my kind request can't be accommodated, let me know and I will change to 0.7.0 as you requested (for a number of reasons the newest Klara won't run in a straightforward way in Julia 0.4). Thanks for your time.

scidom:
@tkelman, a clarification is pending as to whether we can merge this or you won't allow it. If not, may I ask you to let me know please so that I can take appropriate action and the tagging is not stalled for no reason? Thanks.

Keno:
This can't be merged as-is due to the failing status check, which would make status checks fail on all future PRs. This either needs to be retagged as 0.7.0, or we need to add a special exception for this package to the CI system

scidom:
Thanks @Keno, in this case I will close this PR. I will open a new one in the next couple of days, as it might take an hour or so to sort out the labelling of issues/milestones in Klara, dropping the 0.6.2 tag from the Klara repo etc.

PR #7024, opened 2016-11-13
Tag OrdinaryDiffEq v0.1.1, DiffEqBase 0.1.3

ChrisRackauckas:
This is a small tag that adds minimum version requirements and patches the isinplace parameter usage in the common interface. Will be required for tests to pass on a Sundials PR which is why it's important.


ChrisRackauckas:
This requires the Juno tags: https://github.com/JuliaLang/METADATA.jl/pull/7022

ChrisRackauckas:
This other patch I through in has something that will allow some other tests to pass (which I will tag in soon) and so I put it in for good measure.

tkelman:
Your tests are failing with a stack overflow

ChrisRackauckas:
Yes, it's expected that the DiffEqBase tests fail because it needs the tag for OrdinaryDiffEq which updates it to the interface change, and the OrdinaryDiffEq tests will fail because it's looking for the DiffEqBase interface change tag. Everything passes locally so I have to go with that, and then maybe patch afterwards.

If you want to know specifics, this causes the problem on Travis:

```julia
  function solve(prob::DEProblem,args...;kwargs...)
    alg,extra_kwargs = default_algorithm(prob;kwargs...)
    solve(prob,alg,args...;kwargs...,extra_kwargs...)
  end
``` 

But it's because `prob` had a slight change in its parameters, and the dispatches in OrdinaryDiffEq.jl's tag make it so that way this will go to the right `solve`, but without OrdinaryDiffEq.jl's tag this will recurse.

This is why I ask for a way to test on master. Everything passes locally. The interface should be much more stable now, making DiffEqBase change less and causing less of these tags (at least with OrdinaryDiffEq).

tkelman:
The interface of DiffEqBase changes between 0.1.1 and 0.1.2?

ChrisRackauckas:
There's a patch which changes a parameter in the `ODEProblem` type from `Val{isinplace}` to `isinplace` (since it's a boolean). 

(Should that have been a minor version?)

ChrisRackauckas:
Is this waiting on anything?

tkelman:
if you break your api, that's not a bugfix patch release

PR #7028, opened 2016-11-13
Tag ParameterizedFunctions v0.3.0 [https://github.com/JuliaDiffEq/ParameterizedFunctions.jl]

ChrisRackauckas:
This tag releases a bunch of new symbolically-computed dispatches for ParameterizedFunctions. Specifically, the ones for Rosenbrock methods are cool since they allow for pre-computed inverses of matrices which are specific to the internals of numerical methods, speeding things up tremendously.

This updates the minimum version of SymEngine to the recent 0.1.2 since that fixes a common build failure. This also moves out the abstract ParameterizedFunction type to DiffEqBase so that it fits in the with the rest of the ecosystem better (no required dependency here). 

This is separate from the other DiffEq tags and should pass on its own.


ChrisRackauckas:
The patch that's also included adds the change from here: https://github.com/symengine/SymEngine.jl/issues/49 . That way if SymEngine doesn't build on a user's machine this is all still usable (since the usage is all in a try/catch, it will just disable all the extra symbolic calculations). It's kept in a separate patch tag for safety.

ChrisRackauckas:
Also included is a patch which makes it all safer when SymEngine can't build a user's computer.

PR #7032, opened 2016-11-14
Tag ApproxFun 0.4

dlfivefifty:
This is a major update to ApproxFun, in particular, revamps how PDE solving is done so that it is more efficient for variable coefficients.

It also adds upper bounds to the requires for SingularIntegralEquations (with a new tag forthcoming there as well).

### DIFF
--- a/ApproxFun/versions/0.4.0/requires
+++ b/ApproxFun/versions/0.4.0/requires
@@ -1,0 +1,9 @@
+julia 0.4
+FastGaussQuadrature 0.1
+FastTransforms 0.0.6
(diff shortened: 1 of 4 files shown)


aviks:
I'm not 100% sure if existing users of SingularIntegralEquations will have any issues with this change. I think this should be fine, but I'll wait for @tkelman  to chime in before merging.

dlfivefifty:
I also manage SingularIntegralEquations.jl, and can assure you that users will not have issues.  I am going to tag an updated version as soon as this pull request is merged.

PR #7035, opened 2016-11-14
Tag Nemo 0.5.2

thofma:


aviks:
Copying the error from the test failure: 

ERROR: LoadError: New tag 0.5.2 of package Nemo requires julia 0.5.0, but version 0.5.0 of Nemo requires julia 0.4.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.6.0 using `Pkg.tag("Nemo", :minor)`.

cc: @thofma

thofma:
@wbhart Do we want to bump Nemo to version `0.6.0`?

wbhart:
I don't think so. We haven't added anything new, that I can recall.

Bill.

aviks:
You have dropped support for `julia 0.4` in this version, and that needs a new `minor` version per the METADATA tests.

wbhart:
I guess we'll have to wait until Nemo 0.6.0 is ready then.

wbhart:
Alternatively we can just add 0.4 back in to the list of supported
versions. I believe Nemo 0.5.1 didn't support Julia 0.4.x either, though we
still had it listed as supported.

thofma:
@aviks Are these rules for versioning written down somewhere? When do I have to bump the revision number and when the major number?

wbhart:
I'm confused. Are we talking about the Nemo version number here or the
Julia version number?

We set the Nemo version number. Since the convention is that packages
should not increment the major version number because Julia hasn't reached
1.0 yet, that means our major releases have to have the minor version
number incremented.

That means that very small changes, such as the one from 0.5.1 to 0.5.2
need the final digit changed, not the minor version number.

Apologies if I'm misunderstanding what the issue is. I've not been
following the discussion.

Bill.

aviks:
Hi @thofma : the current policies are documented in the METADATA.jl README : https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md . See point 6 for the relevant paragraph (Note that they were written a while ago, and talk abut 0.3 vs 0.4, but the same rules apply for 0.4 vs 0.5)

@wbhart We are talking about Nemo version. While it is true that small changes can and should have the final number changed, the argument here is that dropping support for a version of Julia is a reasonably big change that warrants the increment of the second digit (the `minor` in `major.minor.patch`. The underlying reason is that it leaves the option of creating a bugfix patch release for users of the older julia version.  Hence, given `Nemo 0.5.1` supported `julia 0.4`, the tests fail when `Nemo 0.5.2` declares `julia 0.5` to be its minimum supported julia version. 

This set of rules is baked into the automated tests for METADATA, and as a result we cannot merge this PR as is, since then tests will start failing for all future PR. So the  options are: 
  * tag this revision as `Nemo 0.6.0`  (which you've said you disagree)  
  * change `Nemo 0.5.2` to support `julia 0.4`
  * Tag `Nemo 0.5.3` with support of `julia 0.4` and close this PR
  * Do not tag a new version of Nemo till 0.6 is ready.

wbhart:
Thanks for the explanation @avik

I appreciate there is nothing that can be done on the Julia/METADATA side.
We'll discuss and decide on which is the better option for us.

I don't recall which issue it was that forced us to drop support for 0.4
mid-way between feature releases.

Bill.

ViralBShah:
Bump. Should we close this and await a new PR, or will this one be updated?

PR #7036, opened 2016-11-14
Tag QuantEcon v0.8.0 [https://github.com/QuantEcon/QuantEcon.jl]

spencerlyon2:
Added iterators to simulate over markov chain values and indices


tkelman:
You need to test on 0.4 if your REQUIRE says you support it.

@aviks are you looking over these at all before you merge them?

QuantEcon/versions/0.8.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.4
```
tkelman: this needs to be 0.5, or the package will break on Pkg.update for users on 0.4. @spencerlyon2 do you intend to fix this?

spencerlyon2: We are ok -- the only failure is with the tests. The library still works 100% ok on 0.4

PR #7046, opened 2016-11-15
Tag EventHistory v0.0.3 [https://github.com/kkholst/EventHistory.jl]

kkholst:


EventHistory/versions/0.0.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
```
tkelman: this fails tests on julia 0.4 https://github.com/kkholst/EventHistory.jl/pull/1

PR #7061, opened 2016-11-16
Pull request/af139728

AndreyKolev:


GARCH/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4 0.5-
```
tkelman: upper bounds are not effective unless added to all past versions of the package. all this will do is result in older versions getting used on julia 0.5.

PR #7070, opened 2016-11-17
Tag the DiffEq Group

ChrisRackauckas:
This PR is for minor tags on:

- DifferentialEquations
- OrdinaryDiffEq
- StochasticDiffEq
- Sundials
- DiffEqDevTools
- DiffEqProblemLibrary
- DiffEqBase

The biggest change which spurs most of this is due to:

https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/5

It moved a bunch of types out of DiffEqBase into the respective packages, and thus is a major breaking change which will cause all tests to fail until all of these tags are through because there will otherwise be cases of conflict in exporting. Note that the majority of Sundials is untouched here, just the common DiffEq interface has a change (and the only change is moving the algorithm types from DiffEqBase to Sundials, which is breaking in code but not in API, but it still gets a minor tag for safety).

However, the only breaking API change from this occurs on StochasticDiffEq which upgrades its API to the common interface.

Because of this large change requiring each other in tandum, minimum version requirements were set so that way these tags all require each other, meaning previous versions will still be untouched.


ChrisRackauckas:
We're good now 👍

PR #7087, opened 2016-11-18
Tag BayesianDataFusion v1.2.0

jaak-s:


tkelman:
your tests are failing on travis. release is now 0.5, so if this package still supports 0.4 then you should be sure you continue testing against it

PR #7088, opened 2016-11-18
Register LanguageServer

davidanthoff:


LanguageServer/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+Compat 0.9.4
```
tkelman: unless you're relying on a feature that was first released in this version, this is too conservative

davidanthoff: Actually, we don't need Compat at all, I removed that.

PR #7102, opened 2016-11-20
Support for Julia v0.5

ma-laforge:


ma-laforge:
Cancelling... Apparently I must increment tag with :minor.

PR #7132, opened 2016-11-23
New Temporal version

dysonance:
Big thing with this update is that TS objects can now take on non-numeric value types.

Also made some changes to the `show` method to better divide up the job and hopefully make it easier to understand.

Also, I _think_ I _might_ have finally figured out how to get `PkgDev.publish` working on Windows. I think my issue had to do with some confusion over using ssh vs. https credentials/authentification. However, if there is still something fundamentally wrong with this PR let me know and I can jump on it.


aviks:
I don't think you need a dependency on Dates since your package works on julia 0.5

dysonance:
@aviks Okay, I just removed it from the REQUIRE file and pushed it to my repo, how do I add that change to this PR?

dysonance:
Okay I just made the changes directly on here so that the `requires` file no longer includes `Dates`, and changed the `sha1` file so that the commit on the 0.2.0 tag matches the change to the `REQUIRE` file on my repo. Does this work?

aviks:
Should be fine.. I'll double check in a bit.

PR #7138, opened 2016-11-23
Pull request/d28e80d6

charlesll:
Drop support for Julia 0.4 with a few improvements. Versions 2.4 and 0.3.0 are equal, I followed Travis instructions to tag 0.3 because the support for Julia 0.4 was dropped.


tkelman:
What is your goal with also tagging 0.1.9? If 0.2.3 support Julia 0.4, then 0.1.9 is unlikely to ever be used.

charlesll:
0.1.9 should not be there, it already has been published. I do not
understand why this is there...

tkelman:
0.1.9 was not previously published - https://github.com/JuliaLang/METADATA.jl/commits/metadata-v2/Spectra/versions

But it looks like it's the same commit as 0.2.0, so won't matter.

charlesll:
Ok... So I do not really know what happened there. Anyway, version 0.1.9
should not matter after all. So I guess it is fine.

PR #7145, opened 2016-11-25
Tag ArbFloats v0.1.9 [https://github.com/JuliaArbTypes/ArbFloats.jl]

JeffreySarnoff:
some speed improvements
(this update cleans up more than the two files indicated -- there was a tag mismatch problem earlier)


JeffreySarnoff:
darn -- something is wrong

JeffreySarnoff:
I need help -- somehow, earlier a tag directory for 0.1.9 was entered into v0.5/ArbFloats/.git/refs/tags instead of v0.5/METADATA/ArbFloats/.git/refs/tags.  That made it impossible to update.  I saved the directories ArbFloats and METADATA/ArbFloats (tar.gz) then deleted the 0.1.9 tag that was in the wrong place and PkgDev.tag()ed a new 0.1.9 -- it went in the right place, with the wrong code.  I restored the saved version and it works properly -- what should I do?

ArbFloats/versions/0.1.9/sha1
```diff
@@ -0,0 +1 @@
+b3ce3393993c5ccb0a1a1a92d4e3d2c859106c3e
```
tkelman: it looks like the tag consistency check should be failing for this but isn't https://github.com/JuliaArbTypes/ArbFloats.jl/releases

JeffreySarnoff: It is odd -- most of those tags do not appear on my local version's METADATA, which shows 0.1.0 0.1.1 0.1.2 0.1.8. 0.1.9

JeffreySarnoff: What safe steps move this repository to where it should be?

JeffreySarnoff: Also -- do not trust the 0.1.9 version at [releases](https://github.com/JuliaArbTypes/ArbFloats.jl/releases), that change is one that I made and is valid, but the v"0.1.9" tag under /v0.5/ArbFloats/.git/refs/tags/v0.1.9 has this content:
`190801bdaf5370e75fcb17e3abefa23196e337ed`

tkelman: Your local copy is irrelevant if you haven't pushed the info to github. You need to replace the tags on github if they don't match what github says.

tkelman: It's okay for there to be tags on github but not metadata. The issue here is tags that are present on both but don't point to the same commit.

JeffreySarnoff:
closing and reopening to redo the auto checking

JeffreySarnoff:
reopen to check

tkelman:
cc @ararslan I'm confused why the Travis check isn't flagging a problem here. if dbce0537f2386c8c1cf319de05306fd83353c768 is the desired tag commit, then the metadata sha1 file should also say thay

JeffreySarnoff:
I understand that Travis did not alert to the mismatch, and that that is problematic.  Meanwhile, I want this to get through -- there is another package I cannot release without it.  I stuffed the correct SHA value into METADATA (and it is rechecking).

ararslan:
Problems with my terrible code (I wrote the thing that checks for matching SHAs here) shouldn't have to hold this up; I can investigate separately.

ArbFloats/versions/0.1.9/requires
```diff
@@ -0,0 +1 @@
+julia 0.5.0
```
tkelman: since you still have a `deps/build.jl` script that uses both BinDeps and Compat, you shouldn't remove them from REQUIRE

JeffreySarnoff: How do I fix that -- undo the change and tag that 0.1.10?

tkelman: Either that, or add them back just to this ArbFloats/versions/0.1.9/requires file in your metadata fork for now and make sure to change REQUIRE before your next tag.

JeffreySarnoff:
Changed it in metadata fork, as soon as this clears -- I will put the changes in the repository REQUIRES (there are few other doc improvements that I will add at the same time and give the bundle the next tag).  Thank you for the help.

JeffreySarnoff:
Is there some reason this has not been merged?

JeffreySarnoff:
Thanks Avik!

PR #7149, opened 2016-11-26
Pull request/5568f12d

ContaTP:
A package to calculate power and sample size for clinical trials
Fixed some bugs
Add tests
Build documentation


ClinicalTrialSampleSize/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5.0+
```
ihnorton: Why the `+` here?

ContaTP: I have already found this typo and fix that in the newest version

PR #7156, opened 2016-11-27
Tag Ipopt v0.2.5 [https://github.com/JuliaOpt/Ipopt.jl]

mlubin:
Fixes builds on travis (https://github.com/JuliaOpt/Ipopt.jl/issues/72)


tkelman:
does this only happen on travis? probably shouldn't disable secure downloads unconditionally

PR #7158, opened 2016-11-28
Tag Showoff v0.1.0 

tlnagy:
- Drops support for julia 0.3
- Fix deprecations on julia 0.6


Showoff/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tlnagy: Drat. I didn't notice that Compat was missing a version, will fix.

PR #7167, opened 2016-11-29
Tag TerminalExtensions v0.2.0 [https://github.com/Keno/TerminalExtensions.jl]

Keno:
Tagging for improvements in iTerm detection


TerminalExtensions/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5-
+Compat
```
tkelman: please declare a minimum version dependency that implements the features you need

aviks:
Can I ask, why have we started to make a minimum Compat version a hard requirement? Practically, if your package does not work with the latest version of Compat, you are going to have a problem. And nobody actually cares enough to figure out the actual limit, they'll just put a version to satisfy this requirement. I can't see any way in which this actually makes the ecosytem better.

tkelman:
It's not about the latest version of Compat. It's about what happens with older versions. And this isn't unique to Compat, but I'd like to encourage people to think more carefully about dependency versions in all cases. Compat is an easy place to start because its features are very well defined.

If you don't declare a minimum version dependency, then you're stating that your package works with every single version of that package that has ever been tagged. That is almost always not correct.

aviks:
Again, I think this does not make a difference in practice.  Compat is very likely to be upgraded to the latest version, and so you will never have to worry about working with older versions. Of the millions of pkg updates that have happened, what percentage of them do you suppose had Compat kept behind? 

Every package asserts that it works with all future versions of Julia. But if everyone starts putting max julia versions on all packages, development will pretty much grind to a halt. 

Let me be clear, I do not think this should be a free for all. And we all are grateful for the processes you have put in... it has made the ecosystem more robust. However, every rule has a cost in the friction it adds. And thus each rule should be measured against the actual harm it prevents, versus the friction it adds. We should not be using METADATA PR's as a teaching opportunity.

tkelman:
We've had these discussions with Pkg3 about bounding. It's about what information is known and what risk you're willing to assume. Applying bounds is about avoiding known breakage. When someone applies an upper bound, packages will be held back to older versions. Then "my dependencies will probably be at their latest versions" is a bad assumption to make. Not having any lower bounds at all is usually incorrect, but it's not obvious why if it's never explained.

Encouraging more careful dependency practices doesn't cost much, if people actually respond to comments. It will make things more robust however, especially in deployments that are applying unconventional package bounds trying to enforce testable stability.

tkelman:
> But if everyone starts putting max julia versions on all packages, development will pretty much grind to a halt.

I disagree here. It would just mean packages aren't installable by default on the latest version of Julia until someone says they should be.

This isn't a hard rule that I apply, but I do point it out when I see it, and I generally don't merge PR's if I've made a comment that gets no response from the submitter. "Why?" or "I don't want to bother with this" is a good enough response to generic feedback that isn't about a specific actual problem, but more best-practices. It's not really about Compat as much as encouraging package authors to think about what they're using from their dependencies, and expressing more of that intent so the package manager can make use of it in version resolution.

tkelman:
And in this particular case, it looks like Compat might not actually be accomplishing anything at all for Julia 0.5- and newer, in which case the uses of it could probably just be removed.

ViralBShah:
My point has always been that these checks should be automated and built into tooling. Suggesting  and not merging because someone did not implement a suggestion appears to me to create friction in contributing packages.

I think it should be the package author's job to create a high quality package that works for their users. I don't think METADATA is the point to insert these checks. If there are legitimate worries about a package that is widely used, shouldn't that package's issues be the place to discuss them?

tkelman:
Yes, any checks that can be automated should be. Not all can.

Didn't say I don't merge if someone doesn't implement a suggestion - I don't merge if someone doesn't respond at all to comments. A response of "okay we can fix that in a future tag" is usually fine, unless the issue is with the requirements of a tag in which case adding a new tag with different requirements can't always actually fix the issue. Due to the way dependency resolution works, old tags are still potentially usable so their requirements often do matter permanently.

tkelman:
> don't think METADATA is the point to insert these checks

Where then? METADATA is where checkpoints are defined between arbitrary development on master and what users will install by default. I do think we should have separate registries as soon as that's supported, with one (or several) a free for all with very minimal quality control, and a smaller core of curated packages that see better stability testing and checks for best practices before users install them. "Pkg.update broke my code" or "these packages didn't work at all" gives a bad impression for the entire ecosystem. Well defined standards and testing will help mitigate that for now, while there isn't any formal separation between packages that are expected to be unstable and questionably usable and those that people can use and rely on.

PR #7172, opened 2016-11-30
Tag WoodburyMatrices v0.2.1 [https://github.com/timholy/WoodburyMatri…

timholy:
…ces.jl]

CC @mlubin


mlubin:
Thanks!

WoodburyMatrices/versions/0.2.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: should declare a minimum version dependency that implements the features you need

timholy: It turns out that dependency was added by you in https://github.com/timholy/WoodburyMatrices.jl/commit/0a7f414f47e85dacafeac0aac0e74637f93653ba :stuck_out_tongue:. You made another commit around the same time that looks like it was intended to support `Tuple{Int,...}` syntax on 0.3 at the same time. Now that we require 0.4 I think we can drop it. Should I fix that in WoodburyMatrices and retag before we merge this, or is it OK to merge this as-is and then do another release that drops Compat?

timholy: I'm gonna merge this and then disentangle Compat separately.

PR #7177, opened 2016-11-30
Tag ForwardDiff v0.3.0 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:
This change breaks a couple of things in the v0.2.0 API in order to get rid of global state, reduce compile times, provide a totally non-allocating API, and support more array types. I've written an [upgrade guide here](http://www.juliadiff.org/ForwardDiff.jl/upgrade.html), and have tried to ensure that [using the old API throws depwarns when possible instead of straight up breaking code](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/deprecated.jl).


ChrisRackauckas:
When you say "totally non-allocating API", you mean there's a simpler solution to this issue?

https://github.com/JuliaDiff/ForwardDiff.jl/issues/136

jrevels:
No - if your objective function mutates values, those values still need to be typed generically enough to allow for `Dual` numbers to work. Unfortunately, that's a fundamental limitation of operator-overloading AD.

What I meant is that it is now possible to call ForwardDiff's API functions in such a way that they do not allocate any additional memory on top of whatever the objective function allocates. In other words, if your objective function is non-allocating, calls to ForwardDiff function shouldn't need to allocate, as long you provide work buffers up-front via the `AbstractConfig` types.

KristofferC:
Well.. beaten by a second.

ChrisRackauckas:
Got it. Thanks.

PR #7178, opened 2016-11-30
Register QML [https://github.com/barche/QML.jl.git]: v0.1.0, v0.2.0

barche:
This adds the QML.jl package for using the Qt5 QML system to build Julia GUIs


MichaelHatherly:
Since `Compat` is [being imported](https://github.com/barche/QML.jl/blob/1530a79e1723a348cf225c882605f2bccab8104e/src/QML.jl#L3) it should be included in the `REQUIRE` file. Should also set a minimum version for `Compat` that implements the features you're using.

barche:
Thanks, I will remove Compat from the next release, I forgot to do so after bumping the minimal Julia version to 0.5.

PR #7180, opened 2016-12-01
New MbedTLS.jl releases

quinnj:


tkelman:
The 0.4-supporting tag needs to be tested on 0.4. Major changes to security-critical packages like this one should probably stay open for review a little longer than you did here.

quinnj:
I didn't commit anything though? Or do you mean to the package? Yeah, I can understand that.

Is there actually something wrong with the travis build? The error looked weird: https://travis-ci.org/JuliaLang/METADATA.jl/builds/180310273

I just saw your comments on the 0.4 testing and thanks for catching that: I didn't realize that 0.4 testing had been removed. Let me turn that back on.

tkelman:
I meant the package, especially given how much discussion had happened on the previous attempt to change the bindeps handling there.

There shouldn't usually be a push build on PR's submitted from forks. Did you accidentally push the branch to JuliaLang but then delete it while opening the PR or something?

andreasnoack:
bump. `MbedTLS.jl` macro errors are breaking many packages on master.

quinnj:
Here are some of the thoughts I've had over the last day or two, given the main "blocking" issue that we're not sure we can rely on always having the mbedtls shared libraries that we ship with julia proper:

* Are there ways we can ensure that packaged distros include the shared libraries? @nalimilan, you do some linux package distro, is it possible to include an mbedtls dependency for julia?
* I still think it's a significant improvement for the majority of package users to rely on the julia-shipped shared libraries; indeed, 90% of the 36 currently closed issues were build-related.
* That said, I understand we should have proper fallbacks to build the library; I can try and go back thru the old build.jl code and have it be used as a fallback if the julia-shipped libraries aren't available. I'd appreciate some help as I'm not super familiar or well-versed in the usage of BinDeps.jl and how to make the build.jl as robust as possible (as I think few users are, given the amount of build-related problems we see).

tkelman:
Adding a package dependency to Julia in every distribution is the wrong solution to this issue. Distributions aren't packaging Julia packages, just Julia itself, and mbedtls is not a direct dependency of Julia itself. It's an implementation detail of how we configure libgit2. This is like packages that want to use Julia's gmp or mpfr - unless they're doing so by the high level Julia API's (of which there are none in base for mbedtls), the package needs to be responsible for obtaining the library.

The macro error can be solved by backporting the macro fix to an old branch, or reverting the build change for now.

quinnj:
Makes sense; I'll definitely defer on what's appropriate for linux pkg distros.

Let me try putting the build.jl logic back in before reverting. I'll try to get it sorted out in hte next day or so.

yuyichao:
So IIUC there are two independent issues related to this package happens unfortunately at almost the same time.

1. https://github.com/JuliaWeb/MbedTLS.jl/pull/81 was merged, which does not support 0.4 anymore and may cause issue on 0.5.
2. We need to fix the macro

Personally I think not having (1) shouldn't affect too many (any?) people but (2) is causing PkgEval to break on master. Given the response in the past few days and the related issue(s) in the past months, it doesn't seem that there's going to be a new solution for 0.5/master any time soon and I think it's better to not letting it to blocking fixing of issue (2).

IMHO we should tag a release that includes the fix for (2) without (1). This can in principle be done by creating a new branch/cherry-picking the fix but I think the package master should generally be ready for tagging and leaving an untaggable change on master for a long time will also block future fixed to the package which can be especially problematic since we are planning to merge a few other major breaking changes to base and want this package to be working or fixed as soon as possible.

Therefore I'm planing to revert https://github.com/JuliaWeb/MbedTLS.jl/pull/81 tomorrow and point 0.3.1 tag to the master after that. Please comment if there's any other changes that should be included/excluded in the tag or if there's a better solution/reason to not revert the change and do cherry pick instead.

tkelman:
Make sure all supported versions of Julia are being tested on CI before you tag. Presumably if you revert the PR 81 you'll re-enable 0.4 support?

yuyichao:
I assumed that the 0.4 CI config was dropped in PR 81 but yes I'll check that.

quinnj:
I'm going to take a stab at putting build.jl back in on master tonight, will report my results.

quinnj:
That's my plan, put the last committed version of build.jl back in as a fallback to execute when the julia-shipped libraries aren't detected.

The initial travis results look promising: https://travis-ci.org/JuliaWeb/MbedTLS.jl

tkelman:
No, I mean put BinDeps back exactly as it was. You need to run the validation checks on anything you find. Julia or linux distros may not always ship with the right version that the package was written against.

PR #7188, opened 2016-12-02
More Julia v0.4->v0.5 fixes

ma-laforge:
Also try to provide @shashi with a good notebook example, as requested in:

https://github.com/JuliaLang/Interact.jl/issues/36


ma-laforge:
https://github.com/ma-laforge/InspectDR.jl/blob/master/notebook/2_interact.ipynb

nalimilan:
FWIW, the changes you made at https://github.com/ma-laforge/LibPSF.jl/commit/7308f14cf471fd5f1ed47e7fdd9133d9bebf83e2 are not really correct, as they mean you won't accept non-standard string types. Just use Compat.jl without changing the code at all.

Other changes look OK.

ma-laforge:
Thanks for the comment.  I might be wrong here (correct me if so), but from what I understood:

1)
File paths (in Julia? / on modern computing systems in general?) are supposed to be stored/manipulated as UTF8 strings (Not UTF16, nor UTF32, nor...).

Now, from my understanding: since v0.5 of Julia, I should be using type `String` - because `UTF8String` is now deprecated and merged with `ASCIIString`.

I used to generate function signatures with `::AbstractString` on file paths simply out of laziness.  If I used the suggested `UTF8String` instead, someone making a function call using an `ASCIIString` would either have to explicitly upconvert to `UTF8String `- or I (as the module developer) would have to make duplicate function signatures to support `ASCIIString` file paths (thus the laziness).

2)
As for using `String` for signal names:  I am not convinced the PSF file format even supports UTF8 strings at all.  I was debating  throwing an exception if `!(isascii(signame))`... but opted to maintain support for UTF8 strings... just in case it worked.

kmsquire:
Bump.  I think that the discussion is ancillary, and that this can be merged--is this correct?

ma-laforge:
@kmsquire: If the question was directed at me, than yes: This PR is ready/waiting for a merge.

PR #7190, opened 2016-12-03
Register RationalFunctions

aytekinar:
[RationalFunctions.jl](https://github.com/aytekinar/RationalFunctions.jl.git): v0.0.1


dpsanders:
This looks nice, but is unusable without at least some usage instructions in the readme?

aytekinar:
Hey @dpsanders ! Thank you for your constructive feedback!

I will be providing some instructions this weekend. Should I tag again (increment the version number), or shall I just push to master and wait for some feature change for the next tag update? Any suggestions or good practices for this?

dpsanders:
Personally I don't think it will need a new tag if you just update the readme, but it doesn't hurt either.

aytekinar:
OK! Upon updating the readme, I will mention here again to get attention and possibly have the package go through.

RationalFunctions/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

aytekinar: I will modify it as `Compat 0.8.0`.

Although I can understand the need for declaring a minimum version dependency on packages, I have doubts related to the practical aspects. In my case I only depend on `Compat` and `RecipesBase`. Still, it is hard to dig and fetch the minimum version number that supports the functionality I would need.

Without enforcing a `CHANGELOG` type of requirement from packages,  any contributor needs to (1) either check all the packages' README.md files at all tagged versions, (2) or do `bisect` in some source files to find the version the feature is introduced.

Do you have any better suggestions? Of course, one obvious solution could be to mark the most recent versions of dependencies to date when the new module is to be registered  here at METADATA. However, this is not the same as declaring "minimum version dependency that implements the features you need."

Sorry, but I am confused.

aytekinar:
Ping: @dpsanders, @tkelman.

I hope this time I could address your concerns.

Thanks again for constructive feedback!

ararslan:
Looks good to me!

ararslan:
It would be good to have tests. But apart from that this looks okay to me.

PR #7196, opened 2016-12-05
Tag Rotations

andyferris:
The recent StaticArrays tag seems to be here too... (here's hoping Pkg3 makes this a bit easier to avoid).


ararslan:
> here's hoping Pkg3 makes this a bit easier to avoid

Actually that could simply be solved by making PkgDev check out a separate local branch rather than making all of its changes on the default branch.

In the meantime you can just `git checkout -b rotations-0.3.2`, `git rebase -i HEAD~2`, delete the entry for StaticArrays, `git push anyferris rotations-0.3.2:pull-request/9511da62 --force`.

simonbyrne:
Or try [AttoBot](https://github.com/attobot/attobot)

KristofferC:
I merged StaticArrays METADATA PR. Maybe you can rebase this so the extra commit goes away?

andyferris:
I'm not sure what you mean, @nalimilan, by "combined" license? 

I was going to contact the guy who wrote what inspired the Boost-licensed section to see if he's OK with me just putting it all under MIT, but I didn't see the use in delaying bugflixes just for that.

andyferris:
> Keeping two commits sound fine to me. 

RIght, I figured someone would hit "squash and merge" in github and that would be that...

tkelman:
I usually prefer doing squash-merge on all metadata PR's unless they're intentionally doing multiple tags of multiple packages in the same PR - fewer commits in the history per update that way.

nalimilan:
> I'm not sure what you mean, @nalimilan, by "combined" license?

Since the package contains files with two licenses, it needs to be under a license that comply with the terms of both. Usually this is done by choosing the most specific/restrictive license of the two (e.g. GPL when combining MIT and GPL). I'm not sure how it would work here, but it would be good to make this explicit.

PR #7204, opened 2016-12-05
Tag ForwardDiff v0.3.2 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:
This should be the final dependency to tag before [ReverseDiff](https://github.com/JuliaDiff/ReverseDiff.jl) can be registered.


tkelman:
Something is broken here - compare https://github.com/JuliaCI/pkg.julialang.org/blob/08915ae594000b161ee4f25563165a266186aad1/logs/Turing_0.5.log (passing, 0.3.1) to https://github.com/JuliaCI/pkg.julialang.org/blob/de25fb3eb9ae2384bbf9fb9163ba971839359a4d/logs/Turing_0.5.log (failing, 0.3.2)

jrevels:
Thanks for letting me know, seems like there was a typo in the code I added to prevent ambiguity errors on v0.4. Ref https://github.com/JuliaDiff/ForwardDiff.jl/pull/168

PR #7208, opened 2016-12-06
Pull request/62ed20db

dehann:
Refactoring of code into these few packages. Changes go together.


nalimilan:
Could you add badges pointing to Travis tests?

In KernelDensityEstimate and IncrementalInference, better use `AbstractString` and Compat.jl (except when typing actual objects and fields like `Dict{String, T}` and `String[]`). Complex types like `VoidUnion{Vector{String}}` will need a `T<:AbstractString}` type parameter instead.

I'm not sure what's our policy on calling `BLAS.set_num_threads(2)` automatically in IncrementalInference, but I wouldn't encourage it. It would be better to leave the user do that, as this has side effects on unrelated packages.

Finally, as regards coding style, are you aware that the convention is to use lowercase for functions?

Other than that, looks good to me.

dehann:
Hi @nalimilan, I will add Travis badges to RoME in the next tag if that is okay please. I'm busy reworking a quite a bit of code at the moment.

All packages aside of RoME worked fine on 0.4. I just need to restrict Julia 0.4 testing to use the previous tag of each package. I'm not quite sure how to do that.

I'd like to avoid deleting and redoing tags on these Pkgs if possible please, but happy to incorporate comments on the next tags.

Thanks

dehann:
Actually, @nalimilan, I think the Travis tags are there for RoME -- but this is the first register and tag pull request. They'll show up when Package listings update. All other packages are set.

Thanks

nalimilan:
> Actually, @nalimilan, I think the Travis tags are there for RoME -- but this is the first register and tag pull request. They'll show up when Package listings update. All other packages are set.

No, you're speaking about tests for tagged packages, but you also need links to Travis to check that tests pass *before* tagging.

I'd like to hear the opinion of others regarding the use of `BLAS.set_num_threads` before merging.

As regards my other comments, they are less serious, but it doesn't make a lot of sense to tag a release whose only change is going to be reverted in the next one... That's your call, anyway.

dehann:
Okay, I'll do a new pull request, but I don't how to do:
> No, you're speaking about tests for tagged packages, but you also need links to Travis to check that tests > pass before tagging.
Could you tell me how to do this please?

PR #7214, opened 2016-12-06
Updated SwitchTimeOpt

bstellato:
- Removed Ipopt Dependency
 - Updated compatibility with Julia 0.5


ihnorton:
Both versions have the same commit hash.

PR #7217, opened 2016-12-07
Register Gasp.jl v0.0.1

kpamnany:
Global Address SPace toolbox.


kpamnany:
Cc: @tkelman, as we've discussed this. Subsumes `Dtree.jl` which should be removed -- should that be done via a PR as well?

tkelman:
Given that past tagged versions of Celeste depend on Dtree, no please do not remove anything. We can deprecate Dtree by adding a Julia upper bound on all of its tagged versions, which will make it no longer installable once the (future) version of Julia in the upper bound is released.

kpamnany:
Okay, thanks.

tkelman:
You really shouldn't be cloning master of an external resource. A version of the Julia package should use a fixed version of the external repository.

kpamnany:
I can ``git clone https://github.com/kpamnany/gasp.git --branch="v0.2"``, but I get the following error when calling ``LibGit2.clone()``. Any idea why?
```
LoadError: GitError(Code:ENOTFOUND, Class:Reference, Reference 'refs/remotes/origin/v0.2' not found)
```

kpamnany:
Ah. Turns out ``LibGit2.clone()``'s branch option does not apply to tags, unlike the command line git. I have to use ``LibGit2.branch!()``.

aviks:
Is this ready to go?

kpamnany:
Yes it is, thanks.

PR #7220, opened 2016-12-07
Add julia version bounds to Iterators.jl requirements

martinholters:
For versions up to 0.1.7, `using Iterators` fails on Julia 0.4 and later.
For version 0.1.8, `using Iterators` fails on Julia 0.5 and later.
For version 0.1.9, `using Iterators` works on Julia 0.5 (and current master), but `Pkg.test` fails.

For versions up to 0.1.7, I had to add lower bounds which I did very liberally, i.e. `using Iterators` works.


martinholters:
Travis failure is due to missing tags (for ancient versions) in Iterators.jl repo. While it surely would be nice to add them, the kind of changes I'm doing in this PR should arguably be possible even without the tags (think of an unresponsive upstream).

tkelman:
Why the 0.1 and 0.2 minimum versions for the tags that had no minimum before?

martinholters:
After adding the missing tags to the Iterators repo, Travis now passes.

> Why the 0.1 and 0.2 minimum versions for the tags that had no minimum before?

Version 0.0.0 was the only one that didn't error when `using` it in Julia 0.1, for the others, up to and including 0.1.7, `using` them in Julia 0.2 worked. I did not test them any further though.

Iterators/versions/0.0.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.1 0.4
```
tkelman: I wouldn't put a lower bound at all, use 0.0

martinholters: Makes sense.
What bout the others where I put `0.2`? That would exclude in-development versions between 0.1 and 0.2. Better `0.2-`? (How was the versioning scheme back then?) Or also `0.0`?

tkelman: If you got ahold of julia 0.1 somewhere and tested it and saw failures, then `0.2-` is probably most appropriate

martinholters: > If you got ahold of julia 0.1 somewhere

`git checkout v0.1; make` - well, almost :smile: 

> `0.2-` is probably most appropriate

Will update...

martinholters:
What about the upper bound on 0.1.9? As said, `using Iterators` works on 0.5, but `Pkg.test("Iterators")` fails. Opinions?

tkelman:
what's the failure?

martinholters:
```
Error During Test
  Test threw an exception of type MethodError
  Expression: collect(chain(1:2:5,0.2:0.1:1.6)) == [1:2:5;0.2:0.1:1.6]
  MethodError: no method matching length(::Iterators.Chain)
  Closest candidates are:
    length(::SimpleVector) at essentials.jl:168
    length(::Base.MethodList) at reflection.jl:256
    length(::MethodTable) at reflection.jl:322
    ...
```

OTOH
```
julia> collect(subsets(collect(1:4),2))
6-element Array{Array{Int64,1},1}:
 [1,2]
 [1,3]
 [1,4]
 [2,3]
 [2,4]
 [3,4]
```
so it is not completely unusable. Also, the error that occurs during testing does not lead to silent corruption. So I'm tempted to change the PR to remove the upper bound from 0.1.9 again. Objections?

martinholters:
Ok, removed upper bound from 0.1.9. Good to go?

martinholters:
I think I have addressed all of @tkelman's remarks and without the upper bound on 0.1.9, I don't see how this can hurt anyone, so lacking further concerns, I'll merge later today.

PR #7222, opened 2016-12-07
Register CSDP [https://github.com/JuliaOpt/CSDP.jl]

blegat:


tkelman:
You should not be downloading master of an external resource in your build.jl file. That can work one day, then change and stop working in the future. Use a tag or sha.

Don't use Pkg.dir to refer to a package's location, use `dirname(@__FILE__)`.

blegat:
@tkelman Thanks for the suggestions, that is fixed now ! :)

CSDP/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+BinDeps
+Glob
+MathProgBase
```
mlubin: You should have tight lower and upper bounds on the MathProgBase dependency to avoid breakage if we make changes to the API (see other solver packages).

blegat: Thanks, this is fixed :)

PR #7224, opened 2016-12-07
Tag ForwardDiff v0.3.3 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:
This should hopefully fix the `UndefVarError` Tony mentioned [here](https://github.com/JuliaLang/METADATA.jl/pull/7204#issuecomment-265333768). I haven't yet tried Turing.jl's tests locally, though.


jrevels:
I've confirmed that this fixes the specific test failure reported in Turing.jl, but I can't run all their tests because I'm getting errors when building the package (I think my source builds of Julia don't have [the requisite header files Turing.jl is looking for](https://github.com/yebai/Turing.jl/tree/v0.1.0#1-install-julia)).

PR #7230, opened 2016-12-07
Tag Optim v0.7.0 [https://github.com/JuliaOpt/Optim.jl]

pkofod:
Optim is currently keeping ForwardDiff back.


tkelman:
This broke several downstream packages - see 0.4 and 0.5 changes at http://pkg.julialang.org/pulse.html

pkofod:
Thanks for always keeping an eye on the quality. I'm on mobile, but the only error I could see was in dsge, and I warned the community half a year ago that they should use the provided results getter functions and that direct field access was at own risk. This time around I pulled the lever for good and changed the field names to match the common usage of the word minimum. It was a minor tag as well, so breaking changes should be expected. The rest I optim related stuff I saw was deprecation warnings except the lsqfit error that I've filed a pr to fix. I'll have a closer look when I have time .

tkelman:
I haven't checked every single regression there either, but some may also be interactions with recent ForwardDiff tags. I can file issues letting these downstream authors know, but may be more effective if you or @jrevels are able to take a look and say what might need changing.

pkofod:
> I haven't checked every single regression there either, but some may also be interactions with recent ForwardDiff tags. I can file issues letting these downstream authors know, but may be more effective if you or @jrevels are able to take a look and say what might need changing.

Sure, and thanks again for mentioning it. I don't really have the total picture of how Optim is used out in the wild! I'll go through the logs to check.

pkofod:
https://github.com/FRBNY-DSGE/DSGE.jl/pull/9
https://github.com/STOR-i/GaussianProcesses.jl/commit/135a1bf6d83dc28afd25c9a186df945fcc3f5057
https://github.com/JuliaStats/KernelDensity.jl/pull/40

and then there's some lsqfit-problem floating around which could be related to the code migration from optim to lsqfit. Will have a look later.

PR #7231, opened 2016-12-08
Pull request/a14099ef

dehann:


tkelman:
Are you sure you want to register a version of RoME that supports 0.4 but is immediately followed by a version that doesn't? It may be better to start with only registering the 0.5-only version.

dehann:
Hi @tkelman, thanks let me redo these tags then. I have been trying to avoid redoing tags, but alas!

tkelman:
Could just revert that commit and not register the 0.0.1 tag

dehann:
Thanks, I have already deleted the tags and decided to do more code refactoring. This drops 0.4 support in favor of Julia 1.0 readability. Once the #7241 is merged I will do a new pull request to bring in changes relating to these packages.

PR #7235, opened 2016-12-08
Pull request/bbca88b7

sylvaticus:
New package OdsIO with functions for reading OpenDocument Spreadsheet (.ods) files (already working and documented) and for writing (planned for February) using the python ezodf module.

PR #7236, opened 2016-12-08
Pull request/6aea31e4

sylvaticus:
Now I am pretty lost :-/
Following documentation http://docs.julialang.org/en/release-0.5/manual/packages/ I did tagged my repo.


sylvaticus:
For what I understood you should change the documentation about writing packages (http://docs.julialang.org/en/release-0.5/manual/packages/) to specify first PkgDev.tag("FooBar") and then PkgDev.publish().

ararslan:
You'll want to remove the FooTest commit as I outlined here: https://github.com/JuliaLang/METADATA.jl/pull/7235#issuecomment-265807615

sylvaticus:
Hello, thank you very much. I think I done what you suggested in https://github.com/JuliaLang/METADATA.jl/pull/7235#issuecomment-265807615

tkelman:
Be aware that conditional dependencies as you've done in a try catch are generally unreliable and a bad idea, and they do not interact well with precompilation.

sylvaticus:
Thank you tkelman, do you refer to the `using DataFrames` only or also to the `@pyimport ezodf` one?

I was thinking that it (they) could be a problem, so I commented out the precompile statement.. after all the module is very small.
Do you suggest instead I should get rid of it (them) ?

sylvaticus:
I removed the conditional dependency on DataFrames and rather added it to REQUIRE.

aviks:
Is this ready to go?

tkelman:
it'll need a new tag to pick up the change in thr dataframes dependency, but that can be a new tag pr after registration

PR #7243, opened 2016-12-09
Pull request/ad57a173

montyvesselinov:
Switching to Optim 0.7


tkelman:
contained in #7248. it would be preferable if you could check out a new branch from upstream metadata-v2 before tagging separate packages

andreasnoack:
@tkelman Sorry. Your comment only showed up when I merged.

tkelman:
you should be squash merging here.

PR #7249, opened 2016-12-10
Tag SingularIntegralEquations v0.2.1

dlfivefifty:
Support for ApproxFun v0.4.1


tkelman:
Patch releases really shouldn't include breaking changes.

dlfivefifty:
Sorry, I just wanted to update the syntax before dropping 0.4 support (so that the usage is consistence). The syntax of ApproxFun should be stable now and so it shouldn't happen again.

dlfivefifty:
In any case, this package uses internal features of ApproxFun, so will likely continue to need to be limited to specific patch releases.

aviks:
@dlfivefifty did you want to get this merged in it's current form? Sorry, did not understand the result of the above conversation.

dlfivefifty:
@aviks Yes please.  @tkelman's complaint applies to ApproxFun.jl, not SingularIntegralEquations.jl, which has already been tagged and merged.

aviks:
Gottit, thanks. @tkelman thoughts?

PR #7250, opened 2016-12-10
Bump solvers

mlubin:
Fixes some fallout from https://github.com/JuliaLang/julia/pull/15850


Cbc/versions/0.3.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+BinDeps
+MathProgBase 0.5.5 0.6
+@osx Homebrew
+@windows WinRPM
```
tkelman: still using Compat in deps/build.jl so it still needs to be listed here

ECOS/versions/0.7.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps
+MathProgBase 0.5 0.6
+@osx Homebrew
```
tkelman: same here

tkelman:
Ipopt needs an 0.4-supporting tag that passes its tests, the current release does not.

PR #7263, opened 2016-12-12
Register ODEInterfaceDiffEq.jl

ChrisRackauckas:
ODEInterfaceDiffEq.jl is a glue package for ODEInterface with gives it JuliaDiffEq common interface bindings and all of the advantages that come with it. This code was previously included in OrdinaryDiffEq.jl and was the code which required conditional dependencies, but using this common API, conditional dependencies are no longer required. This does need to be registered as it is at least a crucial part of the OrdinaryDiffEq.jl tests (for testing native Julia versions vs the FORTRAN wrapper package), and the `radau` methods have no parallel in Julia as of right now, making them a strong contribution to the common interface.

It wasn't included in ODEInterface.jl directly because of the dependencies it adds, and so it keeps ODEInterface.jl a dependency-free library. For more discussion on this, see the following issue:

https://github.com/luchr/ODEInterface.jl/issues/9

### DIFF
--- a/ODEInterfaceDiffEq/url
+++ b/ODEInterfaceDiffEq/url
@@ -1,1 +1,1 @@
-https://github.com/JuliaDiffEq/ODEInterfaceDiffEq.git
+https://github.com/JuliaDiffEq/ODEInterfaceDiffEq.jl.git
--- a/ODEInterfaceDiffEq/url
+++ b/ODEInterfaceDiffEq/url
@@ -1,0 +1,1 @@
+https://github.com/JuliaDiffEq/ODEInterfaceDiffEq.git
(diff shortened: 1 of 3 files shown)


aviks:
`ODEInterfaceDiffEq` is a quite a mouthfull, with `DiffEq` at both ends. Are you sure there cannot be a better name? Final decision is of course upto you, but I'd at least try and see if a better name is possible.

ChrisRackauckas:
I think it makes it clear that it links `ODEInterface` and the `DiffEq` universe. The `DiffEq` solvers all have that appended to the end, so `ODEInterfaceDiffEq` is just following that convention to show that's a glue package. If it was a standalone library I wouldn't name it like this, but I think the name is pretty much stuck given what it is.

PR #7272, opened 2016-12-13
Register (and tag) PiecewiseLinear.jl

joehuchette:


tkelman:
Is this JuMP (optimization) specific? The name should probably reflect that, otherwise it sounds much more general than it is.

joehuchette:
That's how I intend to use it, but it could be useful otherwise. I'm open to name suggestions that aren't ridiculously long.

tkelman:
PiecewiseLinearOpt? PiecewiseLinearJuMP? PiecewiseLinearProg?

joehuchette:
I think I can live with PiecewiseLinearOpt

PR #7276, opened 2016-12-13
Register and tag Tokenize.jl [https://github.com/KristofferC/Tokenize.jl]

KristofferC:


Tokenize/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
```
tkelman: should be tested on 0.4 if supported there

tkelman:
Anything takrn from JuliaParser should be acknowledged in the top-level license. If JuliaParser does this, why not improve  the existing package?

KristofferC:
Added juliaparser license and removed support for 0.4.

This isn't an improvement over JuliaParser, it does a different thing. Please read the README in Tokenize.jl and compare what JuliaParser does if you are interested.

KristofferC:
Gonna try attobot.

PR #7281, opened 2016-12-13
Register DASKR.jl

ChrisRackauckas:
DASKR.jl is a wrapper for the DASKR FORTRAN DAE solver. it has a simple interface and an interface which matches JuliaDiffEq.


tkelman:
Don't use Pkg.dir, use `dirname(@__FILE__)`

Should acknowledge the daskr license in the top level license file.

tkelman:
If you use Compat, it needs to be listed in REQUIRE

ChrisRackauckas:
Put in a new patch v0.0.2 which makes those changes.

tkelman:
@aviks please leave new package registrations open at least 24 hours

aviks:
yes, sorry, i misread the date.

DASKR/versions/0.0.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+DiffEqBase
+Compat
```
ChrisRackauckas: Yes, future patch will have it.

PR #7282, opened 2016-12-13
Register ClimateTools.jl

Balinus:
First commit of ClimateTools package.


aviks:
Hi Balinus, did you not want to register this package any more?

Balinus:
Well, I think it might be best to register a new version 0.0.2 which will be available in some days. I hope I have not wasted your time on this? :/

aviks:
No, certainly not. We keep new package registrations open for a day or two before merging in case there are any comments on the package name (which is a global resource, and difficult to change later).

Balinus:
ok, thanks for the clarification. I'll do a pull request in the next couples of days. The package will be more advanced.

I was unsure about the package name. It's meant to be a diverse set of tools for people working in the climate fields, hence the name ClimateTools.

Balinus:
Well, it went faster than expected. I've submitted another pull request with version 0.0.2. Thanks!

PR #7295, opened 2016-12-14
Fix v0.7.1 of Optim (fast-track please)

pkofod:
I have no idea how the wrong SHA1 got in there, but I did have some PkgDev problems... Attobot is now installed, so I cannot mess it up again!


tkelman:
You really should not change sha of existing tags. How did Travis pass on the original PR?

tkelman:
Please don't do this. This was just a repeat of the existing tag, so it would have been far preferable to tag the right sha as 0.7.2 and not touch what had already been released. This makes the meaning of 0.7.1 depend on when METADATA was checked, which is not how versions are supposed to work.

tkelman:
You should actually still make an 0.7.2 tag for this. Anyone who upgraded to the original 0.7.1 might not see this update at all.

pkofod:
Im very sorry. Will tag when at a computer.

PR #7296, opened 2016-12-14
Tag Optim.jl v0.7.2

attobot:
Repository: [Optim.jl](https://github.com/JuliaOpt/Optim.jl)
Release: [v0.7.2](https://github.com/JuliaOpt/Optim.jl/releases/tag/v0.7.2)
Travis: [Travis Build Status](https://travis-ci.org/JuliaOpt/Optim.jl/builds/184001137)
Diff: [vs v0.7.1](https://github.com/JuliaOpt/Optim.jl/compare/88657f977cb4da51360b439653eb1ce8dfe65fe1...88657f977cb4da51360b439653eb1ce8dfe65fe1)
`requires` vs v0.7.1: no changes
cc: @pkofod


pkofod:
I get your point in #7295 @tkelman . My bad. This is the same sha as v0.7.1. I now have attobot doing the dirty work for me, so hopefully I will never mess up a `PkgDev.publish()` again!

ararslan:
Would it make sense to revert the changed SHA for the previous release that Tony mentioned?

pkofod:
> Would it make sense to revert the changed SHA for the previous release that Tony mentioned?

Maybe, but I'm not sure. The sha for v0.7.1 in Optim.jl's repo and in METADATA would then be different, and while I do appreciate the principles of the whole thing, this is going to be a problem for a very short time. As soon as v0.7.2 is merged, and people update, no one should be on different v0.7.1's any more.

So, I can go either way. If the sha is reverted, then I should probably change the tag in the repo as well. I just need to be told if it is reverted here, so the shas don't diverge.

PR #7297, opened 2016-12-14
Register EasyPhys.jl

julian-wolf:


ararslan:
Hi there! Nice package. I have a couple of questions/comments:

Do you have CI enabled? Looks like you have Julia 0.4 in your Travis test matrix but require at least Julia 0.5 for the package, which means that CI would always be red due to 0.4 failure.

If your package supports precompilation, it would be good to add `__precompile__()` above the `module` statement in src/EasyPhys.jl.

julian-wolf:
Hi @ararslan,

I do have CI enabled. I've changed the REQUIRE from 0.5 to 0.4; that was an error on my part. I've also added __precompile__(), which I hadn't previously been aware of—thanks for the tip!

julian-wolf:
Meanwhile, I've had some issues with my local copy of METADATA. I think it might be simplest to start over, so I'm going to close this request and give it a go with attobot.

PR #7315, opened 2016-12-15
Publish SpeedDate.jl

Ward9250:
Utility written using BioJulia to compute divergence times between DNA sequences.


aviks:
I'm not sure that tests cannot be written for this package. You could, for example, run the main method as the command line app. It will verify the integration with your dependencies.

Ward9250:
@aviks That's a great idea, thanks!

PR #7323, opened 2016-12-16
Tag StaticArrays v0.1.2 [https://github.com/JuliaArrays/StaticArrays.jl]

andyferris:


StaticArrays/versions/0.1.2/requires
```diff
@@ -0,0 +1 @@
+julia 0.5-
```
tkelman: when you're ready to drop support for prereleases, get rid of the trailing dash

PR #7328, opened 2016-12-16
Tag Mocha v0.1.3

pluskid:
A maintainance release.

Ref: https://github.com/pluskid/Mocha.jl/issues/211#issuecomment-266764184


Mocha/versions/0.1.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+HDF5 0.4.7
+JLD
+Logging
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

PR #7329, opened 2016-12-16
Register new package QuantumBayesian.jl v0.1.0

attobot:
Repository: [justindressel/QuantumBayesian.jl](https://github.com/justindressel/QuantumBayesian.jl)
Release: [v0.1.0](https://github.com/justindressel/QuantumBayesian.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/justindressel/QuantumBayesian.jl/builds/184515680)
cc: @justindressel


tkelman:
You should not be defining Base methods on only Base types like `Base.dot(a::AbstractArray, b::AbstractArray)` - this can change the behavior of other unrelated code depending on whether or not your package is loaded

justindressel:
Thanks for the tip - that makes sense.  (And thanks for checking over my code - I'm a relative newcomer to julia, so tips are very welcome.)  Do you have a recommendation for how to handle this case?  

I had been using type aliases to avoid over-specializing types, and this notion of dot is quite useful.

tkelman:
use a different (or module local, instead of extending from Base) name, or only define it for types that your package controls

PR #7332, opened 2016-12-16
Tag Mocha v0.2.0

pluskid:
Since Julia now requires a major version update for packages that change requirements for Julia with major version update. We now tag Mocha as v0.2.0 instead of v0.1.3.


aviks:
I think the v0.1.3 version should be removed from this commit.

pluskid:
Removed.

PR #7334, opened 2016-12-17
MarketData

milktrader:
New methods to download OS market data.


milktrader:
I'm experiencing weird behavior with `Pkg.update` where it says it cannot fast-forward so I'm just doing a `git pull` from the METADATA directory, which is causing all this cruft. Apologies, I'll see if I can track down what my issue is.

tkelman:
See https://github.com/JuliaLang/julia/pull/19624 and the linked discourse thread. Try `git pull --rebase` manually

PR #7346, opened 2016-12-19
Register SimpleParser [https://github.com/ordovician/SimpleParser.jl.…

ordovician:
This is a very simple parser, for people who want to learn and experiment with parsers. I got a module reading and writing NeXTSTEP style plist files using this parser I hope to add as a module as well later.

### DIFF
--- a/SimpleParser/url
+++ b/SimpleParser/url
@@ -1,0 +1,1 @@
+https://github.com/ordovician/SimpleParser.jl.git


ordovician:
Is it possible to give me some advice here? My last commit has tagged versions, but I see this fails due to earlier commits without it. What is the recommended way of dealing with this? Should I rewrite my git history, create a new repo or something?

tkelman:
Can you push the metadata tag (the requires and sha1 file for your first tagged version) to this same branch of your metadata pr? Another option is you could try out https://github.com/attobot/attobot

ordovician:
Ah I see the commit in this pull request, does not contain the latest commit. I originally asked the question under the assumption that it was, hence I didn't get why it wasn't accepted. I'll close this merge request than and attempt to create a new one manually from the last commit : 5a13ca22b3dda133b30614b82c548927bf256676

PR #7350, opened 2016-12-19
Tag Hiccup v0.1.0 [https://github.com/JunoLab/Hiccup.jl]

ranjanan:
Fixes https://github.com/JunoLab/Hiccup.jl/issues/17

cc: @andreasnoack


andreasnoack:
@ranjanan Thanks. There is a test error that suggests that you should tag a minor version instead of a patch version.

tkelman:
also should adjust travis so it runs on the supported release versions of Julia

tkelman:
looks like this won't work correctly unless MacroTools is tagged first

andreasnoack:
@ranjanan Are you able to tag MacroTools or is @MikeInnes the only person who can do that? It would be great if these things could be resolved fairly fast since this causes everything that depends on `DataFrames` (115 packages) to fail on master.

andreasnoack:
The changes hasn't taken effect here https://github.com/JunoLab/Hiccup.jl/releases. Probably simplest to retag with attobot

andreasnoack:
Superseded by https://github.com/JuliaLang/METADATA.jl/pull/7355

PR #7353, opened 2016-12-19
Tag MacroTools 0.3.3

ranjanan:


MacroTools/versions/0.3.3/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: declare a minimum version dependency that implements the features you need

tkelman:
This fails tests on 0.4. @andreasnoack please check these things before merging them, otherwise packages are going to break when people do Pkg.update. @ranjanan can you fix this asap? release versions of Julia are more important to keep stable than fixing issues that only appear on nightly.

andreasnoack:
Could such a check be made part of the `METADATA` tests?

tkelman:
attobot will display the badge and diff links, so everyone should be using that for a start. we can and plan to automate more tests here, but it's not always straightforward to do. suggestions welcome.

PR #7356, opened 2016-12-19
Tag Weave v0.3.0 [https://github.com/mpastell/Weave.jl]

mpastell:


kmsquire:
@tkelman: we generally encourage minimal version numbers for Compat.  Should this be true for all packages?

tkelman:
generally, yes - not listing a minimum version number is stating that your package works with every tagged version ever released of your dependency. if that's not true, then the REQUIRE specification is too loose.

kmsquire:
Okay, thanks.  I'll comment on that when I see it then.

tkelman:
though the most common response to such comments is adding bounds that match the most recent versions of everything, which is usually too tight - we should expand on the recommendations in the docs for this stuff, better explain best practices and how dependency requirements work in practice

PR #7364, opened 2016-12-20
Tag WriteVTK v0.4.5 [https://github.com/jipolanco/WriteVTK.jl]

jipolanco:
Fixes issue (and tests) in Julia 0.6.


tkelman:
if the only change is tightening a version dependency, it's actually better to modify the requires file of the existing tag (do not change the sha though). currently the previous version with the looser bound is still valid for dependency resolution to pick, so Pkg might downgrade your package to satisfy the bound.

PR #7366, opened 2016-12-20
Register and tag RegionTrees.jl

rdeits:
RegionTrees.jl is a small package designed to handle region quadtrees, octrees, and their N-dimensional equivalents. For more information, check out the [Readme](https://github.com/rdeits/RegionTrees.jl) and the [API demo](https://github.com/rdeits/RegionTrees.jl/blob/master/examples/demo.ipynb).


kmsquire:
Nice to have a package for this!

PR #7389, opened 2016-12-23
Pull request/3ba67554

danspielman:
As requested, changing the tag from version 0.0.3 to 0.1.0 because we dropped support for Julia 0.4 to support Julia 0.5


tkelman:
included in #7411

PR #7396, opened 2016-12-24
DiffEq Ecosystem Tag

ChrisRackauckas:
All of these got minor tags. Here's the changes which caused this:

1. A change in the nightly versions to macro hygiene standards broke a very common utility macro, so it was fixed but moved to DiffEqBase.
2. The abstract type `ParameterizedFunction` was moved out of ParameterizedFunctions.jl to DiffEqBase.jl so that way solvers / plot functions could type check and dispatch without requiring the entirity of the ParameterizedFunction library (which has lots of dependencies due to its symbolic computing macros).
3. Precompilation was enabled on libraries where it was now application (thanks to getting rid of conditional deps), and it was explicitly set to `__precompile__(false)` in libraries where it is known to not work. This explicit control should fix some errors.

These three changes spawned a minor version bump in DiffEqBase, and the other tagged libraries got minimum version requirements bumped to require this new version of DiffEqBase. 

Importantly, since this fixes the macro issue, it fixes the tests ecosystem-wide on the nightlies.

Other patches and new features follow, including huge improvements to the plot recipes and some performance improvements in StochasticDiffEq.jl, but these changes are small in comparison to the larger maitenance issue.


ChrisRackauckas:
This requires the tags from https://github.com/JuliaLang/METADATA.jl/pull/7395 .

PR #7410, opened 2016-12-26
Tag WAV v0.8.3 [https://github.com/dancasimiro/WAV.jl]

dancasimiro:
v0.8.3 is the same as v0.8.2, but it requires the correct version of Compat (0.9.5). These two minor versions add support for Julia v0.6.

### DIFF
--- a/WAV/versions/0.8.3/requires
+++ b/WAV/versions/0.8.3/requires
@@ -1,0 +1,3 @@
+julia 0.4
+Compat 0.9.5
+FileIO 0.2.0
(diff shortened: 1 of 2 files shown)


tkelman:
Past releases with bounds that are too loose either shouldn't be published or will need to be amended in-place (you can modify the requires file in this branch of your METADATA fork). `xor` support was added to Compat in 0.9.4, not 0.9.5

dancasimiro:
Ok, I removed the references to the invalid v0.8.2. I left the Compat required version at 0.9.5 because that is what I tested and there are changes in 0.9.5 related to xor ("Fix `$` in the test").

PR #7415, opened 2016-12-28
Register DiffEqPDEBase.jl

ChrisRackauckas:
DiffEqPDEBase.jl is an extension to DiffEqBase.jl which is specifically for PDEs. While DiffEqBase.jl is expected to be a stable library for holding the basic parts for the various types of ODEs, DiffEqPDEBase.jl is a continually growing library which contains problem types for as many PDEs as we can get to. It also includes meshing tools for unstructured finite element grids which to be used in component solvers.

Some of these parts were taken from FiniteElementDiffEq.jl to be placed in a Base library since they were so important that it caused FiniteElementDiffEq.jl  (a solver and thus a library high in dependencies) to be required throughout a lot of the testing suite (DiffEqDevTools.jl and DiffEqProblemLibrary.jl). With this structural change, the dependencies in the testing suite now only includes these two Base libraries and no component solvers, making it much easier to work with.

Note that the tests for this are dependent on https://github.com/JuliaLang/METADATA.jl/pull/7396 since it needs an abstract type which is defined in a newer version than the release of DiffEqBase.jl.


DiffEqPDEBase/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+RecipesBase
+VectorizedRoutines
+ChunkedArrays
```
tkelman: missing a DiffEqBase requirement

tkelman:
also missing DiffEqProblemLibrary in test/REQUIRE

ChrisRackauckas:
Thanks for the review. I made the requested fixes in new patch tag(s, I pushed the v0.2 right as the note about DiffEqProblemLibrary came in, sorry!).

tkelman:
Is the test failure `Mesh not defined` expected?

PR #7417, opened 2016-12-28
Tag PDE related DiffEq libraries

ChrisRackauckas:
This propagates the change of https://github.com/JuliaLang/METADATA.jl/pull/7415 to the rest of the ecosystem. The parts which moved from FiniteElementDiffEq.jl were removed, and FiniteElementDiffEq.jl was heavily upgraded to match the syntax of the "common interface" of ODEs/SDEs/DAEs. In addition, the test dependencies of DiffEqDevTools.jl and DiffEqProblemLibrary.jl were re-worked to now use DiffEqPDEBase.jl instead of FiniteElementDiffEq.jl, drastically reducing there dependencies. Lastly, DiffEqPDEBase.jl was incorporated into DifferentialEquations.jl, and default algorithm choices for the FEM Heat and Poisson equations were made.

Requires https://github.com/JuliaLang/METADATA.jl/pull/7415 and https://github.com/JuliaLang/METADATA.jl/pull/7396


ChrisRackauckas:
Bump. Users are reporting issues which are related to this tag not going through when the companion tags were merged. If this can't go through in a timely manner (I understand if no one is available), can I submit a PR with retroactive maximum version requirements of the previous version?

ChrisRackauckas:
Thanks @KristofferC!

KristofferC:
Using Attobot and opening PRs one by one would simplify review process significantly and probably speed at which things are merged.

ChrisRackauckas:
But the problem with opening PRs one by one is that these problems happen. Here it's because a type definition moved from a Base library, so the dependents need a tag for the change. I registered the new library where it moved separately (to keep registrations separate), but that's what lead to it being merged without the others. So when tags really do need each other, I like to put them together so this doesn't happen.

Would it be better to do one-by-one but also add intermediate PRs for new maximum version requirements? I'd have to keep those separate if I used AttoBot, and then if the maximum version requirements don't get merged at the same time then we have the same issue.

tkelman:
Sounds like you should be applying version bounds pre emptively then, and only loosening them once you're sure things work if these packages change so often in ways that break one another.

KristofferC:
If packages are so tightly coupled it seems weird to me to have them in separate packages at all. 
I realize there are reasons why to spread out packages (dependencies) but it significantly increases general overhead. I also feel like it must be harder for users when issues and code gets spread out everywhere.

ChrisRackauckas:
>If packages are so tightly coupled it seems weird to me to have them in separate packages at all.

They generally aren't tightly coupled. The only issue tends to be when something in DiffEqBase has a large change. For example, the abstract type `ParameterizedFunction` being moved from ParameterizedFunctions.jl into DiffEqBase.jl (so that way solvers can dispatch on it without having the dependency on ParameterizedFunctions and thus a dependency on SymEngine). That kind of change is rare  becoming rare since the API has essentially settled, so it is more of a special case that comes up every once in awhile.

Also, this set of PRs is a major step towards decoupling the packages. The real problem was that base-like features were in FiniteElementDiffEq (since all of the extra PDE stuff is pretty large and isn't necessary for non-PDEs which is what DiffEqBase is centered on), but this led to many packages which shouldn't have needed to require a solver to require FiniteElementDiffEq, causing dependency explosion and coupling.

So in a sense it's an issue that's showing up less and less, and part of it may be growing pains due to starting out as one package and not splitting everything "correctly" the first time. And since I usually know when something like this is going to occur, I can just follow @tkelman's suggest and preemptively apply maximum version requirements.

PR #7431, opened 2016-12-30
Tag Playground v0.2.0 [https://github.com/Rory-Finnegan/Playground.jl]

Rory-Finnegan:
I realize it has taken me a while to update Playground to 0.5, but I was hoping a couple of my dependencies would get updated. I've updated the minor version number as I've dropped support for DeclarativePackages.

### DIFF
--- a/Playground/versions/0.2.0/requires
+++ b/Playground/versions/0.2.0/requires
@@ -1,0 +1,5 @@
+julia 0.4
+ArgParse
+YAML
+Mocking
+Compat
--- a/Playground/versions/0.2.0/sha1
+++ b/Playground/versions/0.2.0/sha1
@@ -1,0 +1,1 @@
+d4ea10dccdeefeb1572cddb6acac25734cb1b114


andreasnoack:
You might want to consider https://github.com/attobot/attobot

Playground/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+ArgParse
+YAML
+Mocking
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

Rory-Finnegan: Good point. I'm guessing I can make those changes and tag a new patch release?

tkelman: no, tightening bounds requires adjusting the requires file of the previous tags in-place in metadata. making the change in require will be good before the next tag though.

Rory-Finnegan: Alright, will do. Thanks.

PR #7433, opened 2016-12-30
New release of GeneralizedMetropolisHastings for Julia 0.5

KrisDM:
New version which is compatible with Julia 0.5 (while backwards compatible with 0.4)
Minor bug fixes and improvements


andreasnoack:
You should consider https://github.com/attobot/attobot

GeneralizedMetropolisHastings/versions/0.1.0/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.4.5
+Compat 0.9.5
```
tkelman: are you actually requiring a feature that was first released in 0.9.5? minimum version requirements do not mean "this is the version I tested with," that's very likely to be too tight and rule out versions that do actually work

KrisDM: Thanks for the suggestion. Good point. I jumped without checking from 0.8.6 (which didn't have what I needed) to 0.9.5 (at that moment latest release). On a general note, how could I find out for sure without having to test each release in turn?

KrisDM:
@andreasnoack Thanks, I installed it, will use it next time.

GeneralizedMetropolisHastings/versions/0.1.0/requires
tkelman: best way is checking blame on the readme for the features you're using

KrisDM: Thanks, that worked. I'll bump it down again and use attabot to release a patch.

PR #7440, opened 2016-12-30
Register Memento.jl

Rory-Finnegan:
Registering and tagging an initial release of [Memento.jl](https://github.com/invenia/Memento.jl)


Memento/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
```
tkelman: if you support 0.4, should test against it on travis

PR #7450, opened 2017-01-01
Tag ODEInterface v0.0.11 [https://github.com/luchr/ODEInterface.jl]

luchr:
added Hairer/Wanner's RODAS solver (same lic as the other HW solvers)

Minor other things:
* testloop/testset for v0.4, v0.5 and trunk
* deprecation of `takebuf_string`

BUT:
* still not working for trunk, because of JuliaLang/julia/issues/19790, but I'm working on it, see issue JuliaLang/julia/issues/19805


ODEInterface/versions/0.0.11/requires
```diff
@@ -0,0 +1 @@
+julia 0.4
```
tkelman: if you still support Julia 0.4, you should test against it on Travis

luchr: I really really do support v0.4. I've added the the Travis-tests and Travis' julia v0.4 is happy :-)

PR #7451, opened 2017-01-02
Register new package MLLabelUtils.jl v0.0.1

attobot:
Repository: [JuliaML/MLLabelUtils.jl](https://github.com/JuliaML/MLLabelUtils.jl)
Release: [v0.0.1](https://github.com/JuliaML/MLLabelUtils.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaML/MLLabelUtils.jl/builds/188179683)
cc: @Evizero


Evizero:
ANN on Discourse: https://discourse.julialang.org/t/ann-mllabelutils-jl/1233

MLLabelUtils/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+StatsBase
```
tkelman: there's no particular minimum version of these that you're relying on, is there?

Evizero: Honestly I have not checked. What is the current default approach to establish this? Do people try out different version or just specify the current one which works?

tkelman: I assume you know what functionality you're using from StatsBase? Packages usually aren't super careful about listing what features they add in each version, but with a little bit of blame if you know what you're looking for you can usually find it. Or pin to a range until you find a minor version that doesn't work? If you don't list any minor version, you're saying "should work with any version that's ever been released" which probably isn't correct? If you list a minor version that's just the latest thing you've tested against, then that's almost certainly too conservative and would rule out version combinations that would actually work together, if anyone has pinned StatsBase to something a few versions old.

Evizero: Ok, yes, fair enough. I'll put a bit more effort into that aspect in the future.

PR #7453, opened 2017-01-02
Tag JuliaFEM.jl v0.2.1

attobot:
Repository: [JuliaFEM/JuliaFEM.jl](https://github.com/JuliaFEM/JuliaFEM.jl)
Release: [v0.2.1](https://github.com/JuliaFEM/JuliaFEM.jl/releases/tag/v0.2.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaFEM/JuliaFEM.jl/builds/188221275)
Diff: [vs v0.1.0](https://github.com/JuliaFEM/JuliaFEM.jl/compare/41796749ce35e04ecd0f610d075ae50ed49eed41...ff7c85742484ffce59e062bf853e2b480dc78679)
`requires` vs v0.1.0: 
```diff
--- 0.1.0/requires
+++ 0.2.1/requires
@@ -1,9 +1,11 @@
-julia 0.4
+julia 0.5 
+BaseTestNext
 ForwardDiff
+LightXML 0.4.0
+HDF5 0.7.2
+JLD 0.6.8
+Compat 0.10.0
+DataFrames
+Formatting
 Logging
-LightXML
-Docile
-Lexicon
-FactCheck
-HDF5
-Lint
+NLsolve
```
cc: @TeroFrondelius


JuliaFEM/versions/0.2.1/requires
```diff
@@ -0,0 +1,11 @@
+julia 0.5 
+BaseTestNext
+ForwardDiff
+LightXML 0.4.0
+HDF5 0.7.2
+JLD 0.6.8
+Compat 0.10.0
```
tkelman: Are you using a feature that was first released in 0.10.0? Minimum version requirements do not mean "the latest thing I tested with," that will likely rule out valid versions.

TeroFrondelius: You are right. We will improve in the future. The idea is to go through the source code and clean for any unneeded requirements. Then when the list is reduced, then we will need to somehow test with the old version of the packages. Can you point us some good practices, for example a package which would do the testing against old versions of the requirements? 

We know this isn't the perfect version of the package, but on the other hand if you never tag and release a version, then there never will be stable versions of our package.

tkelman: You can do Pkg.pin to test against older versions. Compat in particular lists all of its features on its readme, and you can check blame to see what version they were released in. For other packages, try to keep an eye out for what features you're actually using from them.

tkelman:
It's frowned upon to modify the behavior of another package's functions and types as you're doing in https://github.com/JuliaFEM/JuliaFEM.jl/blob/master/src/io.jl - if either the function name or at least one of the argument types is something defined within your own package then you're good, but otherwise you could be changing the behavior of unrelated code depending on whether or not your package is imported.

TeroFrondelius:
I will make another push to remove more unneeded requirements.

TeroFrondelius:
> It's frowned upon to modify the behavior of another package's functions and types as you're doing

Thanks for the pointer. We realized that this can cause unwanted functionality, bugs, for other users, who might not be aware of this. Thus we think with @ahojukka5 that we should change this for the v.0.3.0. I have opened issue about this https://github.com/JuliaFEM/JuliaFEM.jl/issues/72. 

Could this version still be added to Metadata? We planned to use this version in Production, thus it would be easier for our internal data management, if this will be published.

tkelman:
You should make a clear warning statement in your readme/documentation that loading your package changes the behavior of other packages, if you want to release this as is.

TeroFrondelius:
Type piracy should have been fixed now.

ahojukka5:
@tkelman Thanks for pointing out problems

tkelman:
You're still calling `FactCheck.getstats()` in your runtests.jl file without depending on FactCheck in either `REQUIRE` or `test/REQUIRE`

PR #7454, opened 2017-01-02
Pull request/4d203168

jkovacic:
I have implemented an algorithm that calculates incomplete gamma and beta functions (very important in statistics) and their inverses where applicable. It supports both real as well as complex arguments where applicable.

To be further discussed:
- the best tolerance for iterative algorithms
- formatting of mathematical expressions in docstrings
- error handling, i.e.what actually to do when a function is not defined for a given set of arguments or when an iterative algorithm does not converge fast
- etc., please note this is my first major Julia project


tkelman:
If you enabled Travis CI would your tests pass there? https://travis-ci.org/jkovacic/IncGammaBeta.jl

jkovacic:
Yes, I have enabled and triggered Travis CI. All tests have passed.

PR #7467, opened 2017-01-04
Tag RationalFunctions

aytekinar:
### Description

Tagging [`RationalFunctions 0.0.2`](https://github.com/aytekinar/RationalFunctions.jl/releases).

### Changelog

Added
- [x] Unit tests,
- [x] Function fitting (`funcfit`),
- [x] Plotting support (`plot` through `RecipesBase`), and,
- [x] Documentation (`?command`).

Sample plots for common platforms (as artifacts from unit tests) are also provided.

**Note.** Requires `Polynomials 0.1.2` (latest version as of now) due to the bugfix Keno/Polynomials.jl@b8a6bf4e.

### DIFF
--- a/RationalFunctions/versions/0.0.2/requires
+++ b/RationalFunctions/versions/0.0.2/requires
@@ -1,0 +1,4 @@
+julia 0.4
+Compat 0.8.0
+Polynomials 0.1.2
+RecipesBase
--- a/RationalFunctions/versions/0.0.2/sha1
+++ b/RationalFunctions/versions/0.0.2/sha1
@@ -1,0 +1,1 @@
+e859790c129c702fee6d9549fc1c8f23fbbded2f


tkelman:
It's maybe a bit questionable to be defining `/(p1::Poly, p2::Poly)` here, since there might be other comparable packages that would want to do the same and they'd have to fight with your package about it. Any two packages that tried to do that same method extension but returning different result types could never both be used in a way that relied on that functionality at the same time.

aytekinar:
@tkelman, I had thought about it. still, I thought any person who would like to use `/(p1:Poly, p2:Poly)` would not be `using RationalFunctions` and `using XYZ` at the same time. Do you think I would drop that constructor in the next version?

tkelman:
It's not just a single package issue though. If some set of packages decide to start depending on RationalFunctions and some other set of packages starts depending on XYZ, then any time you wanted to use packages from both those sets in the same Julia session, that method would behave unpredictably (would depend on loading order) and likely in a way that half the packages won't expect.

aytekinar:
I wish you had just requested a revision, then. Since no package is depending on `RationalFunctions` now, I will just drop support and tag it `v0.0.3` today.

tkelman:
It's mostly a long term concern. We should write a linter that checks type signatures for this automatically. I don't think Lint.jl has features for this yet, but I could be wrong.

PR #7478, opened 2017-01-05
Tag PyCall v1.8.0 [https://github.com/JuliaPy/PyCall.jl]

stevengj:
Closes JuliaPy/PyCall.jl#337

(Bumped the minor version because of added Float16 and qt5 support.)


stevengj:
whoops, need to fix Pkg.test, will resubmit

PR #7481, opened 2017-01-05
Tag NLsolve v0.9.0 [https://github.com/EconForge/NLsolve.jl]

KristofferC:
Tests fail on 0.6 because of changes in LineSearches.jl. No one really has time to figure out why right now. The previous tag is pretty much non functional.


tkelman:
The tests fail on all Julia versions.

KristofferC:
They passed on the PR: https://travis-ci.org/EconForge/NLsolve.jl/builds/189261875. The reason for failure here is due to changes in other packages (LinSearches.jl). It is not optimal but as I said, no one on the NLsolve side has time to look into that. The tests should probably just be disabled though.

PR #7482, opened 2017-01-05
Tag Pandas v0.4.0 [https://github.com/JuliaPy/Pandas.jl]

malmaud:


Pandas/versions/0.4.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
```
tkelman: this doesn't pass on 0.4, you're trying to use `unsafe_wrap` without Compat

PR #7484, opened 2017-01-06
Register new package Weber.jl

haberdashPI:
Release: [v0.2.0](https://github.com/haberdashPI/Weber.jl/tree/v0.2.0)


tkelman:
Are you sure you want to register all the 0.1.x versions too, or have this just start with a single version at 0.2.0?

tkelman:
Would your tests pass if you enabled Travis CI (or AppVeyor) ?

haberdashPI:
I'm happy to only include 0.2.0: the 0.1.x versions are used by some early pilot studies so I need them for my own purposes for reproducibility, but that doesn't need to be public. Pkg.publish automatically included these for me.

What's the recommended way of removing these version entries? Should I just update the pull request with a new commit that removes the directories for the 0.1.0 versions in METADATA???

Re Travis & AppVeyor: I haven't used them before, but I'll look into it and see if they pass. I suspect they may not, since some of my tests verify the precision of real-time events.

tkelman:
> Should I just update the pull request with a new commit that removes the directories for the 0.1.0 versions in METADATA???

Yep, that would do just fine. Just make the commit to this `pull-request/f21ec9ec` branch of your fork and it'll show up here.

tkelman:
good to go then?

haberdashPI:
Yes! Good to go. I've removed the 0.1.x versions.

As for CI: TravisCI doesn't work because I have yet to troubleshoot my linux installation (only Windows and Mac OS X), since I don't have a linux distro immediately available, and AppVeyor is giving me an access EXCEPTION_ACCESS_VIOLATION for PyObject_Call, I *believe* when Weber is first imported. I can't reproduce the error on my windows installation so I'm not sure what's happening.

tkelman:
Travis can run on mac, could try that. There's a way to remote desktop connect into a running appveyor VM if you really want to hunt it down. Maybe try setting the environment variable `PYTHON` to empty on appveyor to force PyCall to use a private python install?

haberdashPI:
Thanks for the advice! I'll take a look at that when I have chance.

PR #7488, opened 2017-01-07
Tag Tau v0.1.0

waldyrious:
This version restores compatibility with Base, after the MathConst --> Irrational change, and makes the tests pass again.


waldyrious:
Please squash the commits on merging. This is the first METADADA.jl update I'm making since I got commit rights in this package, and tried to follow the pattern of the past releases, which resulted in a few hiccups in the process.

Tau/versions/0.1.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.4
```
tkelman: do tests pass on 0.4? ref https://github.com/Aerlinger/Tau.jl/pull/12 looks like travis isn't enabled for pr's?

waldyrious: I thought I had tested that in https://github.com/Aerlinger/Tau.jl/pull/11/commits/378fa9f06673143cdcb4cdbe5462daeffe2566e1, but I clearly overlooked the definitions in .travis.yml. I'll test in my fork where I have enabled Travis testing (I need @Aerlinger's intervention to enable Travis in the main repo), and will report back.

waldyrious:
@tkelman confirmed, the tests don't actually pass in 0.4. I'll look into supporting it with Compat later (https://github.com/Aerlinger/Tau.jl/issues/13), but for now I'll change the requires file in this PR and in the repo (I could make a new tag and release 0.1.1, but since the travis.yml definitions in  doesn't actually test 0.4 in 0.1.0 I think that'd be overkill.)

Tau/versions/0.1.0/requires
waldyrious: @tkelman confirmed, the tests don't actually pass in 0.4. I'll look into supporting it with Compat later (https://github.com/Aerlinger/Tau.jl/issues/13), but for now I'll change the requires file in this PR and in the repo (I could make a new tag and release 0.1.1, but since the travis.yml definitions in  doesn't actually test 0.4 in 0.1.0 I think that'd be overkill.)

tkelman: Tightening version requirements requires modifying the existing tags in-place like you just did. Releasing a new tag that's 0.5-only wouldn't change the fact that the 0.1.0 tag (if you hadn't just fixed it) would be installable but non-functional on Julia 0.4.

waldyrious: I'm assuming you're talking about "tags" as in "version directories in METADATA.jl", not actually git tags in the package's repo, in which case, yeah, I see what you mean. Let me know if I misunderstood, though.

tkelman: > I'm assuming you're talking about "tags" as in "version directories in METADATA.jl"

Right, published versions in METADATA. Those can be a subset of the git tags in the package repo, some of the latter might go unpublished.

PR #7491, opened 2017-01-07
Add a bunch of maximum versions to ParameterizedFunctions

ChrisRackauckas:
This is to make sure DiffEqBase v0.5 won't be enabled until the ParameterizedFunction update.

Is this right? @tkelman

### DIFF
--- a/ParameterizedFunctions/versions/0.0.1/requires
+++ b/ParameterizedFunctions/versions/0.0.1/requires
@@ -1,4 +1,4 @@
 julia 0.5
 SymEngine 0.1.1
 DataStructures 0.4.6
-DiffEqBase 0.0 0.0
+DiffEqBase 0.0 0.5.0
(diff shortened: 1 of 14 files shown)


tkelman:
Should be added to every version. For the very early versions we should also add a dependency with lower bound of 0.0. It's not depended on by the package, but same issue - if there is a feasible version to downgrade to that doesn't have DiffEqBase upper bounded, that might happen to get DiffEqBase using thr latest version instead of holding it back.

ChrisRackauckas:
For the earliest ones I just set:

```
DiffEqBase 0.0 0.0
```

is that fine?

tkelman:
No, that isn't valid. Set 0.0 as the lower bound, but upper bounds consistent with the rest.

ChrisRackauckas:
There, made the requested changes.

PR #7499, opened 2017-01-08
PyTest.jl register and tag v0.0.1

pdobacz:
PyTest.jl is "trying to give Julia pytest-like testing abilities"


tkelman:
It's not correct to use `Pkg.dir` to refer to a package's own location, since it could be installed elsewhere. Better to use `dirname(@__FILE__)`

pdobacz:
Thank you @tkelman, fixed as suggested

PR #7501, opened 2017-01-08
Register LowRankApprox [https://github.com/klho/LowRankApprox.jl.git]…

klho:
…: v0.0.1


tkelman:
Ken Ho! Long time no see!

tkelman:
What's the reason for needing local copies of `geqrf, gelqf, orgqr, orglq` here?

klho:
These are just slight variants that allow you to pass in the work array--useful, for example, in the context of a loop to avoid repeated memory allocation. (To be fair, this doesn't really make things faster, but it does clean up the @time memory allocation report--to make tracking real memory issues easier.)

PR #7506, opened 2017-01-09
Tag ParameterizedFunctions.jl v0.6.0

attobot:
Repository: [JuliaDiffEq/ParameterizedFunctions.jl](https://github.com/JuliaDiffEq/ParameterizedFunctions.jl)
Release: [v0.6.0](https://github.com/JuliaDiffEq/ParameterizedFunctions.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/ParameterizedFunctions.jl/builds/190156765)
Diff: [vs v0.5.1](https://github.com/JuliaDiffEq/ParameterizedFunctions.jl/compare/525fb3e8eb6ab246afc52d6cde248cdd0db4d6ad...19eabb49f1f510e9e6ecd44837ffa88c5587b359)
`requires` vs v0.5.1: 
```diff
--- 0.5.1/requires
+++ 0.6.0/requires
@@ -1,5 +1,5 @@
 julia 0.5
 SymEngine 0.1.2
 DataStructures 0.4.6
-DiffEqBase 0.4.0 0.5.0
+DiffEqBase 0.5.0
 SimpleTraits 0.1.1
```
cc: @ChrisRackauckas

PR #7507, opened 2017-01-09
Tag DiffEqDevTools.jl v0.6.0

attobot:
Repository: [JuliaDiffEq/DiffEqDevTools.jl](https://github.com/JuliaDiffEq/DiffEqDevTools.jl)
Release: [v0.6.0](https://github.com/JuliaDiffEq/DiffEqDevTools.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqDevTools.jl/builds/190157098)
Diff: [vs v0.5.0](https://github.com/JuliaDiffEq/DiffEqDevTools.jl/compare/4f5485fe45195d038fc4c8f07d5ae819a21db499...778d0e4fe872b77f4cd948a9c4adbfee147ac017)
`requires` vs v0.5.0: 
```diff
--- 0.5.0/requires
+++ 0.6.0/requires
@@ -1,5 +1,5 @@
 julia 0.5
 RecipesBase 0.1.0
-DiffEqBase 0.4.0
+DiffEqBase 0.5.0
 RecursiveArrayTools 0.1.2
 DiffEqPDEBase
```
cc: @ChrisRackauckas


ChrisRackauckas:
Reset the tests since the DiffEqBase tags went through.

ChrisRackauckas:
Tests fail because:

```
ERROR: unsatisfiable package requirements detected: no feasible version could be found for package: DataStructures
```

Not sure why since the only deps which require it (Ordinary and Stochastic DiffEq) both need DataStructures 0.4.6...

and as always, tests pass locally (just Pkg.update'd to make sure it wasn't due to some recent update in some dependency that put a maximum version on there)

tkelman:
Can you please come up with a minimal test case for that? I'm fairly sure there's a bug and that error message is blaming the wrong package but have never gotten a reproducible test case to prove it.

ChrisRackauckas:
Yeah, it's definitely blaming the wrong package because on AppVeyor I instead get:

```
INFO: Computing test dependencies for DiffEqDevTools...
ERROR: unsatisfiable package requirements detected: no feasible version could be found for package: DiffEqPDEBase
```

Nothing has a minimum version requirement on DiffEqPDEBase.

ChrisRackauckas:
Looks like when I re-run the tests now they work and this will be safe to merge. I may have happened because it didn't see

https://github.com/JuliaLang/METADATA.jl/pull/7506
https://github.com/JuliaLang/METADATA.jl/pull/7505

Maybe the tests were just run before those were found? (But I thought I re-ran them "just after", must've been too soon)

ChrisRackauckas:
AppVeyor tests passed because it runs a subset of the tests which does not include the API change. However, the `tab` to `tableau` fieldname change also needs to be reflected in OrdinaryDiffEq for the test to pass, so the test requirement was bumped to v1.0.0 which is:

https://github.com/JuliaLang/METADATA.jl/pull/7508

Together this shows its safe to merge, with this being able to go live but its tests without user disruption (it's not a user-facing repository anyways, it's for the testing tools).

ChrisRackauckas:
Thanks!

PR #7515, opened 2017-01-09
Tag Pandas v0.4.0 [https://github.com/JuliaPy/Pandas.jl]

malmaud:


Pandas/versions/0.4.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+PyCall 1.7.2
+PyPlot 2.2.4
+Lazy 0.11.4
+Compat 0.10.0
```
tkelman: `unsafe_wrap` was implemented in 0.8.0

malmaud: Is it problematic to still merge this, though?

tkelman: this is too conservative. it can be loosened in a future release, but this would hold your package back unnecessarily and incorrectly if anyone had Compat pinned to something between 0.8.0 and 0.9.x

malmaud: I'm willing to take that risk - all these lower bounds are just approximate; they could probably all be lowered by a few minor version numbers but I can't realistically check exactly when each feature I'm using was introduced in each package.

tkelman: I hope you know what features you're using. This seems like "I don't want to" rather than "I can't."

What was failing on 0.4 before you added Compat was due to `unsafe_wrap`. It doesn't look like you're relying on anything more recent than that.

You also have a distracting number of ambiguities on 0.4, but that's an unrelated issue.

malmaud: Yes, I meant 'I don't want to', since it seems very labor-intensive to check every method of every package I'm using to find the version it was introduced in. If that's the policy now, is it documented somewhere?

tkelman: It's not policy but it is a recommendation, and it's what lower bounds are supposed to be used for.

PR #7516, opened 2017-01-09
Register Espresso.jl - package for expression matching and transformation

dfdx:


Espresso/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Einsum
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

tkelman:
You shouldn't be defining `Base.getindex{T}(::UniformScaling{T}, ::Int64)`, that is a method and types owned by Base, and it can change the behavior of other unrelated code depending on whether or not your package has been imported.

dfdx:
> You shouldn't be defining Base.getindex{T}(::UniformScaling{T}, ::Int64)

Ah, I thought I had removed it. Thanks for noticing. 

I fixed issues you mentioned but while retagging completely corrupted my METADATA, so I will post back / create new PR as soon as I get it fixed.

dfdx:
Finally, I had to re-submit this PR as #7581.

PR #7502, opened 2017-01-08
Tag Compat.jl v0.11.0

attobot:
Repository: [JuliaLang/Compat.jl](https://github.com/JuliaLang/Compat.jl)
Release: [v0.11.0](https://github.com/JuliaLang/Compat.jl/releases/tag/v0.11.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaLang/Compat.jl/builds/190066153)
Diff: [vs v0.10.0](https://github.com/JuliaLang/Compat.jl/compare/8c6b04c747279e1cd9479a14bad32af739f76d3f...8e198e977db5cadd173d9111da2a1fa1013c898c)
`requires` vs v0.10.0: no changes
cc: @TotalVerb

PR #7517, opened 2017-01-10
Upper bound Compat on packages at risk of breaking with 0.11

TotalVerb:
Compat 0.11 introduces the ∘ operator from Base. Out of caution, we should ensure that packages that depend on Compat are not broken by this change.

## Cliffords

cc @marcusps @blakejohnson 
Since your package no longer depends on Compat, you don't need to take any action. It also looks like you have already fixed the ∘ operator.

## CoordinateTransforms

cc @c42f @andyferris 
Your package also no longer depends on Compat so there is no rush to tag a new version. But if the package will depend on Compat in the future again, it should probably avoid defining the ∘ operator locally if possible. I don't think it will break this package, but...

## Geodesy

Geodesy imports ∘ from CoordinateTransforms, so if it also uses Compat, then there will likely be an error due to importing the same name from multiple packages. Geodesy also does not currently depend on Compat, but this should be addressed for the future.

cc @tkelman 
related #7502

### DIFF
--- a/Cliffords/versions/0.2.0/requires
+++ b/Cliffords/versions/0.2.0/requires
@@ -1,3 +1,3 @@
 julia 0.3-
 Iterators
-Compat
+Compat 0.0 0.11
--- a/Cliffords/versions/0.2.1/requires
+++ b/Cliffords/versions/0.2.1/requires
@@ -1,3 +1,3 @@
 julia 0.3-
 Iterators
-Compat
+Compat 0.0 0.11
--- a/Cliffords/versions/0.2.2/requires
+++ b/Cliffords/versions/0.2.2/requires
@@ -1,3 +1,3 @@
 julia 0.3-
 Iterators
-Compat
+Compat 0.0 0.11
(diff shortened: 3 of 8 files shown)


tkelman:
Their tests do error if you have Compat imported previously, which it often will be:
```
WARNING: both Cliffords and Base export "∘"; uses of it in module Main must be qualified
Error During Test
  Test threw an exception of type UndefVarError
  Expression: X * Y == im ∘ Z
  UndefVarError: ∘ not defined
```

They are all failing on PkgEval nightly, so I suspect they'll still fail on release now even if versions of them that do not depend on Compat are installed. I think we should extend these bounds to all tagged releases of these packages.

TotalVerb:
That looks like a conflict with Base.

andyferris:
Thanks for the heads-up. I had expected to work on this by v0.6 final, but I forgot that `Compat` can also do the same. 

On that note, why do we have releases of Compat for development versions of Julia? Shouldn't we need Compat master for unreleased Julia features?

TotalVerb:
It's to allow package authors to get their packages working with 0.6 without checking out Compat. Since packages cannot depend on master, if we didn't do this, PackageEvaluator would fail on every package depending on Compat.

PR #7518, opened 2017-01-10
Upper bound Julia version for past ACME versions

martinholters:
ACME 0.1.0 and 0.1.1 were broken by JuliaLang/julia#19595

### DIFF
--- a/ACME/versions/0.1.0/requires
+++ b/ACME/versions/0.1.0/requires
@@ -1,3 +1,3 @@
-julia 0.3
+julia 0.3 0.6.0-dev.1553
 Compat 0.7.9
 ProgressMeter 0.2.1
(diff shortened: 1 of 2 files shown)


tkelman:
Will you have a new version that works soonish, or should this package not be installable at all on nightly?

martinholters:
I already have; you might have missed that some @tkelman merged #7461 :smile:

tkelman:
Ah whoops. They weren't visible at https://github.com/martinholters/METADATA.jl/tree/2ed38f25de7671ea4042905a16690a0237558f53/ACME/versions, this branch must have been created from a parent commit before that.

martinholters:
You're right, and that actually confused me, too, for a second :smile:

PR #7546, opened 2017-01-12
Tag BandedMatrices v0.2.0 [https://github.com/JuliaMatrices/BandedMat…

dlfivefifty:
…rices.jl]

• Supports negative bandwidths
• Use BLAS for views of banded matrices
• Adds SymBandedMatrix


dlfivefifty:
• Removes support for Julia 0.4

PR #7550, opened 2017-01-12
Tag PositiveFactorizations v0.0.4

timholy:


tkelman:
https://github.com/timholy/PositiveFactorizations.jl/pull/7 look more appropriate for Compat to have than here

timholy:
Not sure I understand (was that sentence missing a word?)

tkelman:
Those are examples of type piracy, extending Base methods on Base types. In this case since the language added these exact methods and you're just backporting them, it would be best to add them in Compat.jl, not somewhat hidden in PositiveFactorizations.

timholy:
It's not an exact match, though; to avoid a huge slew of ambiguity warnings I had to restrict the `*` methods more than Base (`SubArray` rather than `AbstractArray`). That's why I didn't think they were Compat-worthy. The `Diagonal` one is more general, of course, but since I was adding some private methods I just got lazy.

So, do you want me to put all of them in Compat, some of them in Compat, or is this OK?

tkelman:
Hrm okay well it helps a little with a subset of the same problem, and could be more widely usable even with the signature as you have it right?

timholy:
It definitely broadens the applicability of those methods. Sounds like I should add them to Compat.

PR #7553, opened 2017-01-12
Tag PowerModels v0.2.3 [https://github.com/lanl-ansi/PowerModels.jl]

ccoffrin:
- Added an upper bound on JuMP version as a quick fix for Julia 0.5 builds
- Added a few new features for Matpower file parsing (see CHANGELOG.md)

### DIFF
--- a/PowerModels/versions/0.2.3/requires
+++ b/PowerModels/versions/0.2.3/requires
@@ -1,0 +1,5 @@
+julia 0.4.6
+JSON 0.6.0
+MathProgBase 0.5.4
+JuMP 0.14 0.15-
+Compat 0.7.9
--- a/PowerModels/versions/0.1.0/requires
+++ b/PowerModels/versions/0.1.0/requires
@@ -1,3 +1,3 @@
 julia 0.4.6
-JSON
-JuMP
+JSON 
+JuMP 0.14 0.15-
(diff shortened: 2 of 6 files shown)


PowerModels/versions/0.2.3/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4.6
+JSON 0.6.0
+MathProgBase 0.5.4
+JuMP 0.14 0.15-
```
tkelman: note that upper bounds should be applied on all past tagged versions in order to do what you expect - otherwise the Pkg resolver can choose to keep your package at an old version in order to use the latest JuMP

ccoffrin: Like so?

tkelman: yep! did they all have 0.14 as a minimum?

ccoffrin: I didn't start using JuMP until roughly v0.14, so I don't know how far back compatibly might go.

Is it possible to give a version upper bound and let the lower bound be free?

tkelman: You can lower bound to 0.0, but that's likely way too low. If you're only using the recent JuMP syntax, they're pretty good about listing their major changes like that in their release notes.

ccoffrin: Was able to relax the early versions a little bit.  😄

PR #7556, opened 2017-01-13
Tag Compat v0.11.1

timholy:


tkelman:
we should stop tagging feature additions to Compat as patch releases if we're going to follow semver

PR #7569, opened 2017-01-14
Pull request/02398656

montyvesselinov:
Switching to 0.5 support only.


Mads/versions/0.3.0/requires
```diff
@@ -0,0 +1,27 @@
...
+Distributions
+DataStructures
+DataFrames
+StatsBase
+ODE
+Klara 0.7
+Optim 0.7
+BlackBoxOptim
+Documenter 0.8.10
+Compat 0.12
```
tkelman: unless you're relying on a feature that was first released in 0.12, this is too tight a bound

tkelman:
contained in https://github.com/JuliaLang/METADATA.jl/pull/7641, closing this one

PR #7581, opened 2017-01-16
Register Espresso [https://github.com/dfdx/Espresso.jl.git]: v0.1.0

dfdx:
Fixed and re-submitted version of #7516.


Espresso/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Einsum 0.1
+Compat 0.10
```
tkelman: are you relying on transcode or `.=` ?

dfdx: Nope, I don't think so.

tkelman: given you added it in https://github.com/dfdx/Espresso.jl/commit/c0a799ba602504378c788d6eb1ed0d0707089e15, unless you're now using it to support some 0.6 syntax/deprecations, it looks like you could get away with removing it (and all instances of `using Compat` and `@compat` calls) for now.

dfdx: Yes, it's here to support 0.6 in nearest future. The trick is that there's another related package (namely, [XDiff.jl](https://github.com/dfdx/XDiff.jl)) that I need to finish synchronization with at least on 0.5 before moving both packages to 0.6. And, to be honest, I don't feel it makes much sense to remove dependency on this version just to re-introduce it in a couple of weeks.

tkelman: Were the 0.6 features you're going to need introduced in Compat before or after 0.10? Blame on the readme is usually a pretty good indication of when things were introduced: https://github.com/JuliaLang/Compat.jl/blame/7eeef71f1db0e70f3c19402975535f208ad4ebcc~/README.md

dfdx: Ah, I see where you are driving at. Seems like I have quite outdated version of Compat in REQUIRE (just copied it from `Pkg.status()`) that doesn't actually support new broadcasting syntax that I may need. Do you think setting it to the current 0.12 is the right way to go?

tkelman: The broadcasting syntax was a while ago, 0.7.15 - https://github.com/JuliaLang/Compat.jl/commit/03fd3eb662fbf766ad4b8aaaccebabe10cdcd059, `String(take!())` replacing `takebuf_string` was newer in 0.9.5 if you think you'll be using that

dfdx: No, I don't expect to do any work with buffers in this package.

tkelman:
The travis check is complaining that the git tag doesn't match the sha here.

dfdx:
I've pushed the latest tag to GitHub, so Travis should stop complaining. I also [tried](https://discourse.julialang.org/t/how-to-update-pr-to-metadata-jl/1550) to rebase commits and push back to PR to make git history look better, but with no luck. So if it matters, I can create one more PR (this time using AttoBot) or we can just retrigger Travis job and merge it as is if there are no more comments.

tkelman:
We can squash this on merge, so no need to worry about the history. Your local copy of metadata could possibly hit a bug where libgit2 won't rebase properly, in case you hit that you may need to manually reset your metadata repo.

PR #7584, opened 2017-01-17
Upper bounds for ApproxFun v0.4.1

dlfivefifty:
• Plots v0.10 removed precompiling, which breaks ApproxFun v0.4.1
• BandedMatrices v0.2 made many changes, which breaks ApproxFun v0.4.1


tkelman:
You'll need to apply these to past versions too, otherwise Pkg could just downgrade ApproxFun to an older release in order to use newer Plots and BandedMatrices.

dlfivefifty:
Done.

tkelman:
And just to prevent downgrading to the very very old 0.0.x versions, I think we should julia upper bound those to 0.4 (since there are no BandedMatrices tags that support anything earlier). When there's no existing julia lower bound it can be set to 0.0.

dlfivefifty:
Done.

tkelman:
looks like travis is going to complain about missing git tags. we should come up with a way to grandfather/exempt versions from that check. maybe just ci skip for now.

dlfivefifty:
OK, I lowered the lower bound.  I'm not sure if "just ci skip for now" was meant for me, as I don't know how to do that.

ApproxFun/versions/0.4.1/requires
```diff
@@ -1,9 +1,9 @@
-julia 0.4
+julia 0.4 0.6
```
tkelman: you still have a fair amount of time, but be aware that if you don't release a version either without an upper bound or with an upper bound that allows for 0.6, then once the final release is out the package won't be installable there

tkelman:
> I'm not sure if "just ci skip for now" was meant for me, as I don't know how to do that.

It was, but let's see if the very specific exception I just put in will be enough for now.

ApproxFun/versions/0.4.1/requires
dlfivefifty: I'm about to tag ApproxFun v0.5, which will be Julia 0.5 and above.

dlfivefifty: And also use BandedMatrices v0.2, and replaces Plots dependency on RecipesBase

tkelman: sounds good

tkelman:
Double-check that the commit(s) I pushed here make sense to you. One was just a workaround for checks on really old tags, the other more important one is making the corresponding bound adjustments for matching versions of SingularIntegralEquations.

dlfivefifty:
Yes it makes sense, thanks!

PR #7586, opened 2017-01-17
Tag RobotOS v0.4.2 [https://github.com/jdlangs/RobotOS.jl]

jdlangs:


tkelman:
This fails its tests on 0.4 due to using AsyncCondition: https://travis-ci.org/jdlangs/RobotOS.jl/jobs/192834602

Could make the package 0.5-only?

jdlangs:
Looks like `Compat` has it covered. Can I move the `0.4.2` tag to the new fixed commit or will that break something?

PR #7589, opened 2017-01-17
Tag RobotOS v0.4.2 [https://github.com/jdlangs/RobotOS.jl]

jdlangs:
New `0.4.2` tag with fixed testing on julia `0.4`.

@tkelman


tkelman:
Nice work fixing the callbacks issue. I know some people who are still on 0.4 because of that.

jdlangs:
Like most stuff I've implemented, after the fact it seems like such an obvious and trivial fix that I'm kicking myself for not having done it earlier...

PR #7601, opened 2017-01-19
Tag ApproxFun v0.5.0 [https://github.com/JuliaApproximation/ApproxFun…

dlfivefifty:
….jl]

Drops 0.4 support and now depends on RecipesBase instead of Plots.


ApproxFun/versions/0.5.0/requires
```diff
@@ -0,0 +1,11 @@
+julia 0.5
+FastGaussQuadrature 0.1
+FastTransforms 0.0.7
+RecipesBase 0.1
+Compat 0.12
```
tkelman: are you relying on https://github.com/JuliaLang/Compat.jl/pull/302 ?

dlfivefifty:
No. On second thought I shouldn't need Compat at all since this only supports 0.5, but as soon as I update for 0.6 it will be needed again.

dlfivefifty:
Can this be merged?

tkelman:
I would remove the bounds on the Compat dependency if you aren't using it at all right now.

dlfivefifty:
I don't know how to retag without creating a new pull request, so I'll do that.

PR #7610, opened 2017-01-19
Tag Clipper v0.2.0 [https://github.com/Voxel8/Clipper.jl]

maxvoxel8:


tkelman:
`import WinRPM` is likely to fail unless you add `@windows WinRPM` to your require file

SimonDanisch:
Yeah sorry, I think I'm to blame here... There were a couple of weird issues with this, and I run out of time to investigate... Let me see if I can come back to this today!

maxvoxel8:
Ok, I added the requirement.

PR #7613, opened 2017-01-20
Tag TrafficAssignment.jl v0.3.1

chkwon:
fixed a bug


tkelman:
Upper bounds aren't generally going to achieve what you want unless you apply them to all past versions, otherwise Pkg can just downgrade your package and install the last version that doesn't have an upper bound

chkwon:
@tkelman Ah.. you are right. So what is the best way?

tkelman:
We can modify the requires files of the older versions to add bounds as well: https://github.com/chkwon/METADATA.jl/tree/pull-request/a071afc6/TrafficAssignment/versions

chkwon:
Thanks. Always learning new things. 

Except for the last working versions (v0.2.3 and v0.3.1), can I delete older versions?

chkwon:
Oh it seems like I can fix easily. Never mind.

tkelman:
> can I delete older versions?

No! Old versions need to remain in metadata for reproducibility's sake.

chkwon:
I've updated upper bounds.

tkelman:
Cool, that's good so far. One thing to note is that since old versions depended on Graphs but not LightGraphs, one feasible resolution would be to downgrade to the most recent version of TrafficAssignment that didn't depend on LightGraphs in order to use the latest LightGraphs. While that's valid, you probably want to avoid it?

You can add "artificial" dependencies `LightGraphs 0.0 0.8` (or maybe 0.7?) on the very early versions and it'll avoid that. Though you will probably want to fix your code to be compatible with the latest lightgraphs reasonably soon, otherwise anyone with your package installed will stay held back to non-latest lightgraphs.

chkwon:
Great suggestion. I added artificial dependencies `LightGraphs 0.0 0.7` on very early versions.

tkelman:
Travis is complaining about some missing tags. Easiest would be to recreate and push those, or we can figure out how to make an exception.

tkelman:
When Julia 0.6 comes out it'll downgrade to v0.2.1, unless you put out a newer release with an upper bound higher than 0.6 by then. You've got some time.

chkwon:
@tkelman Thanks! I'll prepare for v0.6.

tkelman:
Bump, be aware that with your current bounds, when 0.6.0 final gets released, users of this package will get downgraded to 0.2.1 of it.

sbromberger:
Why is LightGraphs now being required for TrafficAssignment < 0.2.0 ? It wasn't used by the package until version 0.2.0. (before that, it was Graphs.)

PR #7618, opened 2017-01-20
Tag NormalizeQuantiles v0.2.1

oheil:
Some small changes to comply with julia 0.6 syntax changes/deprecations


tkelman:
metadata has been frozen for package versions that support julia 0.3. could drop support for 0.3 and/or 0.4 and re-tag as package version 0.3.0

oheil:
I drop 0.3 support

PR #7630, opened 2017-01-20
Tag IncrementalInference v0.2.3 [https://github.com/dehann/Incrementa…

dehann:
…lInference.jl]

[IncrementalInference.jl](https://github.com/dehann/IncrementalInference.jl)
v0.2.2 -> v0.2.3
[Build Status](https://travis-ci.org/dehann/IncrementalInference.jl)


tkelman:
If you're using ForwardDiff, you should list it explicitly in REQUIRE

dehann:
Hi @tkelman, thanks for picking that up. Although, I'm not explicitly using ForwardDiff.jl in the package just yet. This is slow development in that direction. Would you still prefer the dependency to be added at this point (won't be actively used for another several weeks)?

I think you may be referring to [this line](https://github.com/dehann/IncrementalInference.jl/blob/master/src/dev/autodiff.jl#L3)... Notice I'm keeping intermittent code in the ```src/dev``` folder (convenience thing).

Alternatively I could move that code elsewhere if you prefer?

tkelman:
Ah okay. I guess I'd keep not-currently-used code on a different branch, but that's just me.

dehann:
Yeah branches are better... Can I delete tag v0.2.3, retag and push the new changes to this branch?

tkelman:
Sure, it hasn't been published to metadata yet, so go ahead if you'd like to.

dehann:
To tricky, will just create a new pull request.

PR #7632, opened 2017-01-20
Tag AstroLib v0.0.7

giordano:


AstroLib/versions/0.0.7/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

giordano: I knew that you'd asked that ;-)

However, it's a bit of a pain discovering which is the minimum version of Compat required as there is no simple changelog to read (i.e., like `NEWS.md` for Julia).  I use only `Compat.String`, is 0.8.0 sufficient?

tkelman: 0.7.15 is enough if `Compat.String` is all you use - https://github.com/JuliaLang/Compat.jl/commit/4c8c62d275fd11273124fcf4d76f2513eafa0ccc

The best way is usually checking git blame on the Compat readme, though you might have to click back past any commits that just reworded things.

giordano: Yep, I did try blaming the README, but there have been a few rewording commits which defeat the purpose of blaming it, then I blamed Compat.jl directly.  Thank you, I'll update the PR.

tkelman: As of a few days ago the github blame has a new, extremely useful, "view blame prior to this commit" button

giordano: Updated.  I use git gui locally to blame files (or even better Magit inside Emacs), which has that feature since ever ;-)

PR #7656, opened 2017-01-24
Delete Currencies v0.3.1

TotalVerb:
Currencies 0.3.1 was accidentally tagged on a branch instead of master, and that branch has since been deleted. This means that the commit `e269f452c2e09fd91f06fb282ac5068ac4b681d2` is now invalid.

This creates the following warning when adding Currencies:

```
WARNING: unknown Currencies commit e269f452, metadata may be ahead of package cache
```

I do not know how to deal with this issue. In this PR I have proposed simply deleted the offending tag, but I don't know if this will cause problems.


tkelman:
Past versions should not be deleted. That warning is harmless, but you can create a branch for it if you want.

PR #7669, opened 2017-01-25
Register and tag LPVSpectral, tag DifferentialDynamicProgramming

baggepinnen:
I have apparently made a huge mess of my METADATA and git settings and `PkgDev.publish()` is not working at all (see error below). Anyway, I hope this manually created pull request is OK.

```
julia> PkgDev.publish()
ERROR: GitError(Code:ENOTFOUND, Class:OS, Failed to resolve path '': No such file or directory)
 in (::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String})(::Base.LibGit2.GitRepo) at /local/home/fredrikb/.julia/v0.5/PkgDev/src/entry.jl:82
 in with(::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String}, ::Base.LibGit2.GitRepo) at ./libgit2/types.jl:638
 in publish(::String, ::String) at /local/home/fredrikb/.julia/v0.5/PkgDev/src/entry.jl:64
 in publish() at /local/home/fredrikb/.julia/v0.5/PkgDev/src/PkgDev.jl:70
```


LPVSpectral/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
```
tkelman: do you still need to support prereleases of 0.5? if not, drop the trailing dash

tkelman:
You'll need to manuall push the git tags.

baggepinnen:
@tkelman  I updated the requirements to from 0.4 to 0.5 for DDP, got rid of the 0.5- for LPVSpectral and pushed the tags to the individual repos. Thanks for your help, I'm still far from comfortable with git at this level :/

PR #7682, opened 2017-01-26
Tag BenchmarkTools v0.0.7 [https://github.com/JuliaCI/BenchmarkTools.jl]

jrevels:


BenchmarkTools/versions/0.0.7/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat 0.9.5
```
tkelman: it doesn't look like you're using `take!` or `numerator` or `denominator`, so I think this is too tight a bound

PR #7672, opened 2017-01-25
Tag TravelingSalesmanHeuristics v0.0.4 

evanfields:
Simple update to remove deprecated Graphs.jl dependency. https://github.com/evanfields/TravelingSalesmanHeuristics.jl

PR #7685, opened 2017-01-26
Make TravelingSalesmanHeuristics 0.0.4 require julia 0.5

tkelman:
and retag as 0.1.0

ref https://github.com/JuliaLang/METADATA.jl/pull/7672, cc @evanfields this will require you to push the same commit as a git tag v0.1.0. Or you can restore 0.4 support using Compat 0.8.4 for view and tag a new release as 0.0.5.


evanfields:
@tkelman Apologies for breaking 0.4 compatibility, and thanks for the catch. If I want to remove 0.4 support (seems reasonable at this point), shouldn't I edit `REQUIRE` and tag _that_ commit as `v0.1.0`?

tkelman:
that works too, but if you aren't going to make any other changes then it works to have the change in metadata first, as long as you make it in the package before your next tag

evanfields:
@tkelman https://github.com/JuliaLang/METADATA.jl/pull/7691

PR #7686, opened 2017-01-26
Tag Compose v0.5.0 [https://github.com/GiovineItalia/Compose.jl]

tlnagy:
Fixes https://github.com/GiovineItalia/Compose.jl/issues/232


Compose/versions/0.5.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.4
+Colors
+Compat 0.9.0
```
tkelman: it looks like you aren't using anything newer than unsafe_wrap which was 0.8.0

tlnagy: I'll lower the min version and resubmit this.

PR #7693, opened 2017-01-27
Tag Weber v0.2.4 and v0.3.0

haberdashPI:
Release: [v0.2.4](https://github.com/haberdashPI/Weber.jl/tree/v0.2.4) and [v0.3.0](https://github.com/haberdashPI/Weber.jl/tree/v0.3.0)

release v0.2.4
TravisCI: [Build Status](https://travis-ci.org/haberdashPI/Weber.jl)
AppVeyor: [Build status](https://ci.appveyor.com/project/haberdashPI/weber-jl/branch/release-0.2.x)

release v0.3.0
TravisCI: [Build Status](https://travis-ci.org/haberdashPI/Weber.jl)
AppVeyor: [Build status](https://ci.appveyor.com/project/haberdashPI/weber-jl/branch/release-0.3.0)


tkelman:
since you do `using MacroTools` in 0.3.0, you should depend on it explicitly in REQUIRE

haberdashPI:
Fixed! Sorry about that.

haberdashPI:
Hold on for a bit... seems I've messed something up in my branches....

haberdashPI:
Okay, I think the sha1's should be fixed. 

I'm getting a very odd error when testing on both TravisCI and AppVeyor where Images v0.4.50 is installed, which seems to fail in Julia v0.5.0. In the preivous builds on TravisCI and AppVeyor, v0.5.14 of Images is installed. I don't understand why this version would have changed simply by adding MacroTools to require...

haberdashPI:
Ahh.... now the build seems to failing for v0.2.4 as well, now I'm very confused.

tkelman:
That's exactly what I was warning @timholy about in https://github.com/JuliaLang/METADATA.jl/pull/7698 and why that shouldn't have been merged without doing something about the earlier tags.

haberdashPI:
Okay. I'm not sure I fully understand the problem, reading through that conversation. Is the issue that I depend on some package which depends on Images, which now has an upper bound??? That doesn't seem quite right, because the upper bounds in the commit I read in #7698 are all 0.6.0 for Images, and the problem here is that Images v0.4.50 instead of v0.5.14 is installed. Is there some way I can solve this for my package? Adding an upper (or lower???) bound to Images in my REQUIRE field?

tkelman:
We need to complete #7698 the rest of the way. Since early Images versions didn't depend on FixedPointNumbers, Pkg is downgrading images in order to get the latest FixedPointNumber (or avoid installing it entirely).

haberdashPI:
Ah okay. Thanks for explaining.

timholy:
Sorry for the hassle, @haberdashPI. Unless someone sits down and figures out how to resolve the dependency chain, perhaps the easiest approach will be to revert #7698 once all those packages have new versions available. See that issue for more info.

haberdashPI:
Not a problem @timholy. It's only a minor inconvenience at the moment.

tkelman:
Should be better now

PR #7696, opened 2017-01-27
Tag GLPlot.jl v0.1.0

attobot:
Repository: [SimonDanisch/GLPlot.jl](https://github.com/SimonDanisch/GLPlot.jl)
Release: [v0.1.0](https://github.com/SimonDanisch/GLPlot.jl/releases/tag/v0.1.0)
Diff: [vs v0.0.5](https://github.com/SimonDanisch/GLPlot.jl/compare/3042923c0c9281b0ff6d7bccb44f3ad3c161d9b7...870f9b6a1d5e57e8d3889c3d38b82f2d1e809837)
`requires` vs v0.0.5: 
```diff
--- 0.0.5/requires
+++ 0.1.0/requires
@@ -1,14 +1,3 @@
-julia 0.3-
-ImmutableArrays
-Images
-Reactive
-Mustache
-GLFW
-GLAbstraction
-ModernGL
-GLWindow
-GLText
-Quaternions
-Compat
-Color
-FixedPointNumbers
+julia 0.5
+GLVisualize 0.2.0
+Iterators
```
cc: @SimonDanisch


dpsanders:
I think you may want to update the Julia version you allow in REQUIRE!

SimonDanisch:
you mean in travis?

dpsanders:
Sorry, I misread on my phone. +1, by the way ;)

tkelman:
Any way to get Travis to run here? xvfb maybe?

GLPlot/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+GLVisualize 0.2.0
+Iterators
```
tkelman: `src/GLPlot.jl` has
```
using GLVisualize, GLWindow, ModernGL, Reactive, GLAbstraction, Colors
using FixedPointNumbers, FreeType, SignedDistanceFields, Images, Packing
using GeometryTypes, GLFW, FileIO, FixedSizeArrays, Quaternions
```
so you need all of those in REQUIRE

SimonDanisch: Yeah, but these are GLVisualize dependencies... Since you recommended to introduce bounds, I figured I should not duplicate these bounds. GLVisualize will pretty much be the package to dictate them.

tkelman: is every version of glvisualize guaranteed to always depend on every one of those packages? if not, then this version of glplot will not alway work

SimonDanisch: Yes! GLPlot is pretty much guaranteed to only use the packages glvisualize is using. Let me sleep over this, but intuitively it seems to be less error prone to only have glvisualize dictate the dependencies.

tkelman: that has to be an all-to-all version mapping though, unless you tie the version of GLVisualize very tightly for each GLPlots release

SimonDanisch: I need to do that anyways, since the breaking api changes in GLVisualize will be much more frequent than any removal of dependencies in GLVisualize...
GLPlot is really just a small layer over GLVisualize and shall be allowed to include a couple more dependencies that I wouldn't like to have in GLVisualize.
But the ones you listed are all for making the basic visualization pipeline work, and if they ever change, glplot won't need them separately from GLVisualize (or if so, it must be an explicit refactor with the goal of moving functionality to GLPlot).
Actually I should clean up these imports as well. I'm pretty sure some are not even used in GLPlot itself.

tkelman: If you use them directly you should either depend on them, or have exact pinned bounds to know GLVisualizeVersions that do depend on them. If you don't use them directly in GLPlot, then they shouldn't be imported by this package.

SimonDanisch:
I tried `xvfb` multiple times without success :( Even if we get it running, I worry that it won't offer an high enough opengl version :( I was talking with @maleadt  if we can share his gpu server for this!

SimonDanisch:
for reference concerning testing:

https://github.com/travis-ci/travis-ci/issues/5645

tkelman:
So if you're using the new FixedPointNumbers name, you need a corresponding FixedPointNumbers minimum bound. Anything this package directly imports it should directly depend on, I don't see how that's a requirement that's controversial.

SimonDanisch:
It's not a huge problem, but in my mental model, GLPlot should only depend on GLVisualize. GLVisualize dictates what versions are needed, which is why I'd like to have them there! That GLPlot uses some packages on top of it is just because of 1-3 functions... I guess the cleanest way would be to get rid of the 1-3 function calls and just not depend on these packages anymore. But that will need some refactoring.
It's a simple trade off for now, to keep GLPlots REQUIRE simple and simpler should mean less room for errors.
Btw, I cleaned up all the unused import statements as well ;)

SimonDanisch:
>So if you're using the new FixedPointNumbers type names, you need a corresponding FixedPointNumbers minimum bound.

I made it backwards compatible on purpose. I don't want to mess with people who still rely on the old range of packages (there are some).

tkelman:
> GLVisualize dictates what versions are needed

Any particular version of GLVisualize has its set of requirements. GLPlot has a dependency on GLVisualize, but the transitive dependencies are the union of all allowed GLVisualize versions. GLVisualize 0.2 and higher hardly bound dependencies at all. This would be the case if you both upper and lower bounded the GLVisualize requirement to an existing version, which had its own bounds appropriate for both packages.

tkelman:
> I made it backwards compatible on purpose.

How? The `N0f8` name isn't going to work at all for FixedPointNumbers prior to 0.3.

SimonDanisch:
Good point, I think I should have bumped the minimum version:
https://github.com/JuliaGL/GLVisualize.jl/blob/master/src/GLVisualize.jl#L29

SimonDanisch:
Okay, does this look better to you?

tkelman:
upper bounds are exclusive - you can do something like `GLVisualize 0.2.1 0.2.1+` to pin to that specific version

SimonDanisch:
...knew it!

SimonDanisch:
Okay, GLAbstraction is now in charge of FixedPointNumber compat for all packages. 
I guess this won't pass, since it refers to a tag from GLVisualize which doesn't exist yet?

SimonDanisch:
Tests pass :)

SimonDanisch:
It seems very unreasonable to force me to tag a new GLPlot version for every bug fix GLVisualize version.
So I loosened up the constraint to accept 2.2 : < 3.0.

tkelman:
either add requirements on all dependencies, or guarantee those requirements are enforced by depending only on versions that exist. if you allow dependencies on versions that don't exist, then you need to declare dependencies on everything you import.

SimonDanisch:
Yes, that's a good theoretical point. But in practise, GLPlot only uses 1%
of the functions, while GLVisualize uses 100%. GLPlot should only use the
stable part of these APIs. Also, I'm the sole developer of those packages
and I really don't want GLPlot to use different versions compared to
GLVisualize. If that happens, it should be treated as a bug. Finally, it
should ideally be up to the user what GLVisualize version they use with
GLPlot. I will have to see if the API is stable enough, but if not I simply
tag a non 2.x version.

MikeInnes:
Please check that:
- [ ] The package's CI is green for this tag (if applicable).
- [ ] The version bounds are up to date.

Then reply with `+ready` so I can merge.

tkelman:
> I really don't want GLPlot to use different versions compared to GLVisualize.

As long as GLPlot depends on GLVisualize, it is impossible to ever have both installed at the same time and using different versions of their common dependencies. The issue here is if the list of dependencies of GLVisualize has changed over time (or may change in the future), then Pkg may choose to use an older version of GLVisualize (or some future one that doesn't exist yet) that does not depend on everything that GLPlot imports. If it's imported directly by GLPlot, it should be depended on directly in GLPlot's REQUIRE file.

aviks:
Hi @SimonDanisch sorry this has languished.  Is this still relevant? The METADATA tests fail with `ERROR: LoadError: packages with unsatisfiable requirements found:
    GLPlot v0.1.0 – no valid versions exist for package Reactive`

Can you take a look?

PR #7698, opened 2017-01-28
Place upper bounds on users of FixedPointNumbers and Images

timholy:
In preparation for great disruption. Only touches packages I have installed on my machine so that I can push the change to the `REQUIRE` file to each package.


timholy:
I should add that I'm not proposing to update all of those packages, only about 12 of them. Would it be better to only put bounds on the ones that I plan to update? Otherwise Images might get "pinned" by packages that don't update.

tkelman:
Do you expect the changes you're going to make to result in breakage in packages other than the ones you'll be updating?

timholy:
Haven't checked. Hard to imagine there won't at least be depwarns, some possibly performance-crippling (inner loop).

GLWindow/versions/0.3.0/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.4
```
tkelman: you shouldn't be changing this?

timholy: Hmm, don't know how that happened. Will have to look at my script.

GLVisualize/versions/0.2.0/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.4
```
tkelman: or this, or the compat version here

GLAbstraction/versions/0.3.0/requires
```diff
@@ -1,11 +1,11 @@
-julia 0.5
+julia 0.4
```
tkelman: or this

GLWindow/versions/0.3.0/requires
timholy: Aha. It's a difference between the package REQUIRE file and what's registered in metadata.

tkelman:
Something to keep in mind here - if there are packages for which some past tags depend on (Images) or (FixedPointNumbers) or (both), but not all, then Pkg resolution could do unexpected things in response to this with the way it is now. In order to get the latest version of Images or FixedPointNumbers once those are released, Pkg may choose to downgrade one of these other packages to a very old version that predates depending on Images/FixedPointNumbers. A way to fix that is to add artificial dependencies to the very old versions, so for each of these packages (all of the past tags depend on Images) or (all of the past tags depend on FixedPointNumbers) or (all of the past tags depend on both) with the same upper bounds throughout.

timholy:
Sounds complicated and likely error prone---what if some old version of some old package genuinely didn't depend on either? And I'm certainly not going to check that manually.

I think I'm going to revise this so that it only affects packages that I intend to update myself.

timholy:
Oh, ugh. Ancient history is coming back to bite:
```sh
ERROR: LoadError: LoadError: The following tags for ImageView have not been pushed to upstream repository: 0.0.0, 0.0.1, 0.0.10, 0.0.11, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 0.0.6, 0.0.7, 0.0.8, 0.0.9, 0.1.12.
```
I've already fixed Netpbm, ImageMagick, and NRRD, but without automated tools to help fix such problems this is becoming faintly ridiculous.

timholy:
OK, fixed. Because the travis queue is short and I want to get on with other Travis-intensive work, I'm merging. Will try to fix any problems later.

timholy:
Lovely:
```julia
julia> PkgDev.publish()
INFO: Validating METADATA
    Compose v0.0.0 – no valid versions exist for package Color

 in check_metadata(::Set{String}) at /home/tim/.julia/v0.5/PkgDev/src/entry.jl:322
 in (::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String})(::Base.LibGit2.GitRepo) at /home/tim/.julia/v0.5/PkgDev/src/entry.jl:93
 in with(::PkgDev.Entry.##4#9{String,Dict{String,Array{String,1}},String}, ::Base.LibGit2.GitRepo) at ./libgit2/types.jl:638
 in publish(::String, ::String) at /home/tim/.julia/v0.5/PkgDev/src/entry.jl:64
 in publish() at /home/tim/.julia/v0.5/PkgDev/src/PkgDev.jl:84
```
Some ancient version of a package (a version no longer in use) that depends on a package that is no longer in use is now blocking me from fixing problems that actually matter. I can't tag new stuff until I figure out some chain of dependencies that is not remotely obvious---all previous versions of `Color` have an upper bound of 0.1 on `FixedPointNumbers`, so why the heck is this happening? Is this a bug in `resolve` or is there some conflict in the multidimensional space of interlocking package versions?

This is exactly the kind of thing I was worried about---until we have better tools for resolving these problems, I think it's counterproductive to ask package authors to resolve problems by adding upper bounds.

timholy:
https://travis-ci.org/JuliaImages/ImageAxes.jl/jobs/196147164

`ShowItLikeYouBuildIt` doesn't have any version requirements other than `julia 0.5`. Resolver bug?

tkelman:
It's packages other than the ones that you're immediately making new releases that I was more concerned about. I've done this bounding several times now to deal with similar disruptive changes in other packages, I would have helped get this serving that purpose correctly if you'd have let me.

tkelman:
Yes the error message blames the wrong package there. Have seen it happen several times but never with a reliable test case that I could use to track it down and fix it.

timholy:
As far as the problems this might cause elsewhere, one good option is to revert it once I have new versions of all these packages in place. I managed to get pretty close: the only ones left are to tag NRRD, figure out one last test failure in ImageMagick and tag it, then tackle Images itself, then it's basically done. I was really hoping to use the fact that Travis/AppVeyor tend to be very quiet early in the mornings on the weekends, and indeed I put that time to good use, but now they're pretty bogged down and it's too inefficient to work on anymore until tomorrow.

The other option is to try to figure out how to solve the dependency chain yourself. Given that my energies are focused on getting through a really complex multi-package merge, I'm not so interested in solving that particular problem (though I know you saw the tools I did contribute to PkgDev). But I certainly won't object if others do so. If no one comes to the rescue, rather than untangle this Gordian knot I'm personally putting my hopes on reverting this PR.

PR #7699, opened 2017-01-28
Tag MXNet v0.2.0 [https://github.com/dmlc/MXNet.jl]

vchuravy:


MXNet/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+Compat 0.9.1
```
tkelman: `take!` requires 0.9.5

tkelman:
you have `BaseTestNext` in a lot of test files (under an if that won't get hit since you're 0.5-only now), but it wasn't listed in `test/REQUIRE`

vchuravy:
Just addressed your comments. Should I delete the tag and release a new one or include these changes in an immediate `v0.2.1`?

tkelman:
tightening version requirements needs to be done in place

PR #7701, opened 2017-01-28
Tag FixedPointNumbers v0.3.0

timholy:
Lots of renaming. Note that there is a shell script in `contrib/` to help fix deprecation warnings in packages. Many of the packages in the "Colors" and "Images" ecosystem already have branches that fix these warnings, so wait a day or two for the dust to settle before submitting PRs for those packages.


tkelman:
It's probably not a good idea to have this version still supporting Julia 0.4, since many packages won't be updated there any more. I wish you hadn't rushed this through while I was asleep.

tkelman:
I'm going to propose making this tag 0.5-only to avoid making the user experience much worse on 0.4.

PR #7717, opened 2017-01-28
jmor

valerdi:


dpsanders:
Hi, it looks like you're having some issues with registering your package. 

Currently, the src directory seems to be empty.
Also, it seems to me that the name "jMOR" is not descriptive enough (and there is no information in the README to know what the package is about).
It would also be good to add a couple of tests.

Finally, it may be easier to use AttoBot to do the package registration: https://github.com/attobot/attobot

Don't give up, you're almost there!

PR #7721, opened 2017-01-28
Bump IndirectArrays and Netpbm

timholy:


IndirectArrays/versions/0.1.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5-
```
tkelman: do you still need to support prereleases?

timholy: Good catch, will fix.

tkelman:
Netpbm is `using ColorTypes` in its tests but not directly depending on it in `REQUIRE` or `test/REQUIRE`

PR #7722, opened 2017-01-28
Maximum versions on DiffEqBase for StochasticDiffEq

ChrisRackauckas:
There will be a tag in DiffEqBase and a corresponding update in StochasticDiffEq which changes the noise interface to allow for in-place RNGs. No previous version will be compatible with the defaults in the new version, so this will block the release until a compatible version of StochasticDiffEq.jl is available.

### DIFF
--- a/StochasticDiffEq/versions/0.0.1/requires
+++ b/StochasticDiffEq/versions/0.0.1/requires
@@ -2,3 +2,4 @@ julia 0.5
 ParameterizedFunctions 0.1.0
 Parameters 0.5.0
 ChunkedArrays 0.1.0
+DiffEqBase 0.0.0 0.7.0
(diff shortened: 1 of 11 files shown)


tkelman:
This doesn't matter for any package other than StochasticDiffEq?

It looks like StochasticDiffEq 0.0.1 didn't depend on DiffEqBase, but in order to avoid letting Pkg downgrade StochasticDiffEq when you have a new DiffEqBase tag, it would be best to add an artificial dependency `DiffEqBase 0.0.0 0.7.0` to StochasticDiffEq 0.0.1.

ChrisRackauckas:
>This doesn't matter for any package other than StochasticDiffEq?

Yes. Only StochasticDiffEq-related. There's one other bug fix, but it's a bug fix for something that was broken before anyways.

>It looks like StochasticDiffEq 0.0.1 didn't depend on DiffEqBase, but in order to avoid letting Pkg downgrade StochasticDiffEq when you have a new DiffEqBase tag, it would be best to add an artificial dependency DiffEqBase 0.0.0 0.7.0 to StochasticDiffEq 0.0.1.

Oh, I thought I did do that. Thanks for catching it.

ChrisRackauckas:
Updated.

PR #7723, opened 2017-01-28
Pull request/f5716d14

abelsiqueira:


CUTEst/versions/0.1.0/requires
```diff
@@ -0,0 +1,6 @@
+BinDeps
+julia 0.4
```
tkelman: If you support Julia 0.4, you should be testing against it on Travis

abelsiqueira: Forgot to change it.

After the correction, should I update the tag (0.1.1) or do I force update this?

tkelman: Up to you, but if you want to do it as 0.1.1 then we should remove the publishing of 0.1.0 from this branch

abelsiqueira:
Force pushed

abelsiqueira:
This is ready to be reviewed again.

tkelman:
Don't download `https://raw.githubusercontent.com/abelsiqueira/linux-cutest/master/install.sh` from master, that won't be reproducible over time. Either check that script into this repository, or download a specific tag or sha of it.

tkelman:
Using environment variables with spaces in the variable name is a bit unusual.

abelsiqueira:
Updated with the changes here: https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/118

abelsiqueira:
Thank you.

PR #7727, opened 2017-01-28
Fix bounds issues with Images and FixedPointNumbers

tkelman:
completing #7698 the rest of the way

This will likely complain about un-pushed git tags. I'll probably work out a way to disable that check when only modifying requires of an existing tag.

### DIFF
--- a/ColorTypes/versions/0.0.1/requires
+++ b/ColorTypes/versions/0.0.1/requires
@@ -1,1 +1,2 @@
 FixedSizeArrays
+FixedPointNumbers 0.0- 0.3
--- a/Images/versions/0.0.0/requires
+++ b/Images/versions/0.0.0/requires
@@ -1,3 +1,4 @@
 Color
 Tk
 Cairo
+FixedPointNumbers 0.0- 0.3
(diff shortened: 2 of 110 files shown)


timholy:
So it was mostly a consequence of having an undeclared (indirect) dependency? Very interesting. Should we be testing that each package listed in a `using` or `import` statement gets listed explicitly?

Thanks for unraveling this! :heart:

tkelman:
> Should we be testing that each package listed in a using or import statement gets listed explicitly?

Yes, but that's not what was happening here exactly. I don't think these versions depended on this package (could be wrong, would have to check) - the issue here is trying to prevent Pkg from doing unintuitive things with version resolution in response to adding upper bounds on some but not all past versions.

tkelman:
Hm, this didn't help quite as much as I hoped it might. It looks like now if you install some package that depends on Images, for example ImageRegistration, on Julia 0.5 it will now install Images 0.4.44 instead of 0.4.50. That appears to be in order to avoid installing ColorVectorSpace, which is strange. I'll add Julia 0.5 upper bounds to those versions that wouldn't load there.

PR #7728, opened 2017-01-29
Add Julia 0.5 upper bound to old Images versions

tkelman:
that do not load at all in 0.5

ref https://github.com/JuliaLang/METADATA.jl/pull/7727#issuecomment-275885268


tkelman:
I'm using this branch name on my fork as a test case for https://github.com/JuliaLang/julia/issues/20313, though I've reverted this commit there to reproduce the odd resolver behavior of installing 0.4.44

PR #7734, opened 2017-01-29
New versions of ImageMagick and NRRD

timholy:


tkelman:
you need to take Compat out of ImageMagick's deps/build.jl

PR #7737, opened 2017-01-29
Add Images 0.6 upper bound on ColorSchemes, ImageQuilting, VideoIO

tkelman:
to postpone upcoming breakage, ref https://github.com/timholy/Images.jl/pull/577#issuecomment-275918132

for the 3 earliest versions of VideoIO, adding artificial Images.jl dependency
to avoid letting Pkg downgrading VideoIO in order to upgrade Images

cc @cormullion, @juliohm, @kmsquire

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -60,13 +60,28 @@ function get_local_tags(dir)
 end
 
 # Get files modified in the PR as a diff against origin
-changed = split(readchomp(`git -C $BUILD_DIR diff --name-only origin/metadata-v2 HEAD`), "\n")
+diff_list(x) = split(readchomp(`git -C $BUILD_DIR diff --name-only
+    --diff-filter=$x origin/metadata-v2 HEAD`), "\n")
+
+changed = diff_list("CDMRTUXB")
+for file in changed
+    if endswith(file, "sha1")
+        # policy 8, do not modify existing published tag sha1's
+        error(string("Do not modify existing published tag sha1 files (policy 8).\n",
+                     "Ask for an exception if absolutely necessary, but tag commits\n",
+                     "should be immutable once published for reproducibility."))
+    end
+end
+
+# Get files added in the PR as a diff against origin
+# (only check tags match for newly-added files)
+added = diff_list("A")
 
 const RGX = r"^[^/]+/versions/([\d.]+)"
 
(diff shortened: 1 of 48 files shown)


juliohm:
@tkelman I also need to reflect the changes in my repository correct? So that in the future when attobot submits the pull request to METADATA there is no conflict? Thanks.

tkelman:
You don't immediately need to do anything, just be aware that Images 0.6 is coming out soon and in its current state it's going to cause your package to break.

Before your next release tag you should either add this upper bound to the package's REQUIRE file, which means anyone who has your package installed will have Images.jl held back to the 0.5.14 release; OR you should test your package against Images 0.6 and fix whatever's wrong.

.test/check-pr.jl
```diff
@@ -60,13 +60,28 @@ function get_local_tags(dir)
...
+    if endswith(file, "sha1")
+        # policy 8, do not modify existing published tag sha1's
+        error(string("Do not modify existing published tag sha1 files (policy 8).\n",
+                     "Ask for an exception if absolutely necessary, but tag commits\n",
+                     "should be immutable once published for reproducibility."))
+    end
+end
+
+# Get files added in the PR as a diff against origin
+# (only check tags match for newly-added files)
```
tkelman: @timholy this fixes your "ancient history" issue, sorry I didn't do it earlier - at least you had commit access on all the packages where you were modifying bounds (and so did I on my earlier PR, I pushed a few even earlier ones - you wrote some Julia code for that PkgDev PR, I did a messy bash one-liner loop)

cc @ararslan as well since this is tweaking the test you wrote

cormullion:
@tkelman Thanks for the heads-up. I'm hoping to look at 0.6 compatibility as soon as I can, but am waiting until that [ColorTypes issue](https://github.com/JuliaGraphics/ColorTypes.jl/issues/65) is fixed...

tkelman:
Sorry this is referring to Images.jl version 0.6, not Julia 0.6. Images 0.6 will be released supporting Julia 0.5, so `Pkg.update()` will upgrade to it once it's available if upper bounds of other installed packages don't prevent it.

.test/check-pr.jl
ararslan: The change looks good to me. 👍

PR #7738, opened 2017-01-29
Delaunay Meshes - Constrained Delaunay triangulations

matthias-vigelius:
This package provides constrained Delaunay-Triangulations. It implements an edge-flipping algorithm to restore the Delaunay property and to remove intersecting edges. The user can define constraints as closed polygons and the package automatically marks faces interior to constrained regions. The package provides various convenience methods, such as exporting triangulations as FEM-meshes or computing the centroid of faces (Voronoi vertices)


tkelman:
Defining `convert(::Type{GeometricalPredicates.Point2D}, x::Vector{Float64})` can change the behavior of other unrelated code that happens to use the `GeometricalPredicates.Point2D` type. Wouldn't it be preferable to define that method in the GeometricPredicates package itself, so all of its users would get consistent behavior?

tkelman:
Also note that doing things like
```julia
local seed = rand(UInt32)
local fs = open("FindConstrainedTriangles.seed", "w")
write(fs, string(seed))
close(fs)
```
in your tests isn't a great idea if the file you're modifying is checked into git. That will make the repository "dirty" when run, and Pkg would not update the package again if a new release were made to avoid overwriting the local modifications.

matthias-vigelius:
@tkelman fair points. I'm not sure if having a `convert` function in `GeometricalPredicates` would be the best way to go about it so I'll just get rid of it altogether for the time being. Neither the .seed nor the .svg files should be checked in, I'll fix the .gitignore.

matthias-vigelius:
@tkelman I fixed the code as requested. As per the guidelines I bumped the minor version as well (I did add two more API functions while waiting for the PR to come through).

SimonDanisch:
Btw, have you considered to use StaticArrays or GeometryTypes? I see a lot of code that uses Vector for the 2/3D static vectors, potentially leading to slow and awkward code ;)

matthias-vigelius:
@SimonDanisch yes, I did see the GeometryTypes but haven't yet found the time to work them out - would be happy about any pointers :) Is there a fixed-size array type in Julia? the linear algebra docu only refers to the standard vector type so I assumed that would be the canonical type. The package is not (yet?) performance-optimized, it's just a pre-requisite for my proof of concept solver.

tkelman:
It would be worthwhile to add a package description other than "No description, website, or topics provided." That will be populated into the listing at pkg.julialang.org

PR #7741, opened 2017-01-30
Pull request/17fb5000

montyvesselinov:
MathProgBase added; many other Type changes.


tkelman:
If you still support Julia 0.4 in ReusableFunctions, then you shouldn't use String without getting it from Compat

montyvesselinov:
A new pull request was made

-monty

PR #7751, opened 2017-01-30
Pull request/9b2393e1

montyvesselinov:
Compat added in ReusableFunctions


tkelman:
Not yet, no. You need to do either `import Compat.String` or `using Compat.String` to avoid deprecation warnings from using `String` on Julia 0.4 in ReusableFunctions. e.g. the travis log https://travis-ci.org/madsjulia/ReusableFunctions.jl/jobs/196610865

tkelman:
The Mads tags look good, so if those work with ReusableFunctions v0.1.8 and you're in a hurry, we could separate the Mads tags into their own PR.

tkelman:
#7779 looks good

PR #7735, opened 2017-01-29
Tag QML.jl v0.3.0

attobot:
Repository: [barche/QML.jl](https://github.com/barche/QML.jl)
Release: [v0.3.0](https://github.com/barche/QML.jl/releases/tag/v0.3.0)
Travis: [Travis Build Status](https://travis-ci.org/barche/QML.jl/builds/196317306)
Diff: [vs v0.2.0](https://github.com/barche/QML.jl/compare/d661a430617111b365c5d7579b2ae0d9547ff310...7905f31800cf8ec0ef285464da1aaf0af8fa1321)
`requires` vs v0.2.0: 
```diff
--- 0.2.0/requires
+++ 0.3.0/requires
@@ -1,4 +1,4 @@
 julia 0.5
-BinDeps
-CxxWrap 0.1.8
+BinDeps 0.4.2
+CxxWrap 0.2.2 0.3-
 @osx Homebrew
```
cc: @barche

PR #7755, opened 2017-01-30
Tag QML.jl v0.3.0 [https://github.com/barche/QML.jl/releases/tag/v0.3.0]

barche:
Supersedes PR #7735


tkelman:
Hm, you used to be `using Compat` without having it listed your `REQUIRE`. That wasn't right, but at least you've fixed it now.

PR #7763, opened 2017-01-31
Register FWT [https://github.com/arsenal9971/FWT.jl.git]

arsenal9971:
This package serve as an alternative to the usual wavelet library Wavelets.jl, it is also necessary for a bigger package that I also developed for filter-based Shearlet Transform, Shearlab.jl

### DIFF
--- a/FWT/url
+++ b/FWT/url
@@ -1,0 +1,1 @@
+https://github.com/arsenal9971/FWT.jl.git


kmsquire:
Hi @arsenal9971, you'll need to tag a version for this to be registered.  

Also, have you read through the [guidelines for naming a package](http://docs.julialang.org/en/stable/manual/packages/#guidelines-for-naming-a-package)?  Generally packages with acronyms are discouraged (with some exceptions).

Cheers!

kmsquire:
Ah, I see that this is included (with a tag) in #7765.  Closing this one.

arsenal9971:
Hey @kmsquire the version is already tagged, I am aware of the guidelines for naming a package, but the acronymous follows the name Fast Wavelet Transform, as FFT is corresponding to the Fast Fourier Transform, is possible to make an exception in this one. I developed this library since the other wavelet implementation in Julia (Wavelets.jl) was not sufficiently clear for me to use it, I also need this library to the principal one called Shearlab.jl that I alread did pull request.

arsenal9971:
There are unmerged commits? How can I solve that?, If I change some file, can I send pull request again with PkgDev.publish( ), or ask again for pull request in metadata?

dpsanders:
Do you use `Wavelets.jl` in your package? It's listed in `REQUIRE`.

arsenal9971:
@dpsanders Yes I did, I used just the filter design function, because is really helpful, but I could also design the filters by hand; I don't know what is more convenient

dpsanders:
OK, that's fine.

PR #7764, opened 2017-01-31
Tag AffineTransforms v0.2.0

timholy:


timholy:
It's a little confusing to know how to specify the `REQUIRE` file here since any dependency on Images version is through Requires.jl. ImageMetadata should only be loaded if it's the new Images, so I think [this](https://github.com/timholy/AffineTransforms.jl/commit/5289aa84cb020ced0cc34fce3901f5e58fb370ac#diff-3c5f4c2343e17b35d7aad555d9dcf529L97) is fine.

PR #7766, opened 2017-01-31
Tag LowRankModels v0.2.1 [https://github.com/madeleineudell/LowRankMo…

cstjean:
…dels.jl]

cc. @madeleineudell


tkelman:
Do you expect this to be failing its tests on Travis?

cstjean:
I'm just the tagger, but @madeleineudell told me that the current master was fine to tag.

tkelman:
That doesn't really answer the question, should it be passing its tests?

Since this is importing things from DataArrays, it should depend on DataArrays directly in REQUIRE. There's also a use of FactCheck in the tests without a corresponding entry in test/REQUIRE.

PR #7779, opened 2017-02-01
Pull request/995288c0

montyvesselinov:
ReusableFunctions bug fix and Compat.String added.


montyvesselinov:
Thank you very much!

PR #7782, opened 2017-02-01
Tag NPZ v0.2.0 [https://github.com/fhs/NPZ.jl]

fhs:
Fixes fhs/NPZ.jl#15


NPZ/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+ZipFile
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

fhs:
I've added versions to requires file.

NPZ/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+ZipFile 0.3.0
+Compat 0.8.4
```
tkelman: are you using `Compat.view` ? `Compat.String` was 0.7.15

fhs: I don't think so. However, one of the dependency (ZipFile) requires Compat 0.8.4

tkelman: there could conceivably be future versions of ZipFile where that is no longer the case though

fhs: I've changed Compat version to 0.7.15

PR #7784, opened 2017-02-01
Max versions to prepare for next release

ChrisRackauckas:
The following max versions are required:

- OrdinaryDiffEq.jl, StochasticDiffEq.jl, and DelayDiffEq.jl need to max version DiffEqBase.jl because the next version will allow for an improved interpolations API, which requires new arguments and will thus error on older versions.
- DelayDiffEq.jl needs to version match OrdinaryDiffEq.jl's updates because of its design of essentially bootstrapping off of it.
- DifferentialEquations.jl needs to max version OrdinaryDiffEq.jl and StochasticDiffEq.jl because their default algorithm choices changed to have some type parameters. This actually doesn't make a difference for usage (it'll work), but it's for the tests.

### DIFF
--- a/DifferentialEquations/versions/0.1.2/requires
+++ b/DifferentialEquations/versions/0.1.2/requires
@@ -7,3 +7,4 @@ IterativeSolvers
 GenericSVD
 Plots
 OrdinaryDiffEq 0.0.0 1.1.0
+StochasticDiffEq 0.0.0 0.5.0
--- a/DifferentialEquations/versions/0.1.2/requires
+++ b/DifferentialEquations/versions/0.1.2/requires
@@ -6,3 +6,4 @@ EllipsisNotation
 IterativeSolvers
 GenericSVD
 Plots
+OrdinaryDiffEq 0.0.0 1.1.0
--- a/DifferentialEquations/versions/0.1.2/requires
+++ b/DifferentialEquations/versions/0.1.2/requires
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.5
 Parameters
 GrowableArrays
 ChunkedArrays
@@ -6,5 +6,3 @@ EllipsisNotation
 IterativeSolvers
 GenericSVD
 Plots
-OrdinaryDiffEq 0.0.0 1.1.0
-StochasticDiffEq 0.0.0 0.5.0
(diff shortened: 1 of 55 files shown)


DifferentialEquations/versions/0.3.0/requires
```diff
@@ -9,3 +9,5 @@ GenericSVD
 Compat
 Plots
 InplaceOps
+OrdinaryDiffEq 0.0.0 1.1.0
+StochasticDiffEq 0.0.0 0.5.0
```
tkelman: since all versions of these are julia-0.5-only, this would make this package not installable on 0.4. better to upper bound the julia version in these cases.

ChrisRackauckas: Just `julia 0.4 0.4`?

ChrisRackauckas:
Set all of the v0.4 ones to max version v0.4.

tkelman:
the upper bounds are exclusive, so should be 0.5

DifferentialEquations/versions/0.1.2/requires
```diff
@@ -6,3 +6,5 @@ EllipsisNotation
 IterativeSolvers
 GenericSVD
 Plots
+OrdinaryDiffEq 0.0.0 1.1.0
+StochasticDiffEq 0.0.0 0.5.0
```
tkelman: missed this one

ChrisRackauckas:
I think I botched all of these then, along with https://github.com/JuliaLang/METADATA.jl/pull/7722/commits/02ee4aad085c70e516e7b14a09f7808e4784225b . I want all of these to use the current release until they are all together. I'll update this all in one sec.

ChrisRackauckas:
Alright, I changed the maximum versions so that way everything is bounded, so only when I do the new release of DiffEqBase and DifferentialEquations will everything begin to use OrdinaryDiffEq.jl v1.2.0 (tagging soon) and StochasticDiffEq v1.0.0 (tagging soon as well).

tkelman:
You want some of the diffeqbase upper bounds to be 0.7 and some to be 0.8?

ChrisRackauckas:
>You want some of the diffeqbase upper bounds to be 0.7 and some to be 0.8?

Yes, those v0.7 ones were the older ones from a previous release. They should be fine.

PR #7800, opened 2017-02-02
Bump ImageView, minor bump GtkUtilities

timholy:
GtkUtilities' dependence on Images is fake, but it serves to synchronize a behavior change.


timholy:
Needs another bugfix before tagging

PR #7826, opened 2017-02-04
Pull request/7afec46f

montyvesselinov:
Ipopt not required; difficult to install behind a firewall; more fixes.


andreasnoack:
Please consider using https://github.com/attobot/attobot when tagging your package. It makes the process simpler.

PR #7827, opened 2017-02-04
Tag PowerModels v0.3.0 [https://github.com/lanl-ansi/PowerModels.jl]

ccoffrin:


tkelman:
You shouldn't use `Pkg.dir` to refer to files within a package, that's not correct if the package gets loaded from elsewhere on LOAD_PATH. Use `dirname(@__FILE__)`.

ccoffrin:
I think I am missing something.  In the case of the runtests.jl I can use `dirname(@__FILE__)` to find the testing directory.  However, this does not work during an interactive session in REPL.  How should I tell a novice user to find a file in the test dir when they are using REPL?

tkelman:
In terms of documentation examples, `Pkg.dir` is close enough to tell users. If you want to get robust to the package being installed elsewhere in a way users can leverage at the REPL, then best would be defining a module-level constant for where your examples are located, `PowerModels.examples` or something, set based on `dirname(@__FILE__)` at module load time. Tests should always use the more careful version though.

ccoffrin:
Ok, I made the fix.  https://github.com/lanl-ansi/PowerModels.jl/pull/81

Remind me, how do I update the version to point to this commit?  `PkgDev.tag("PowerModels", v"0.3.0")` throws an error since the tag already exits.

tkelman:
Replace the git tag then modify the sha1 file in this branch of your metadata fork

ccoffrin:
Should be good now.

PR #7834, opened 2017-02-05
Upper bound CRlibm 0.4 in all tagged versions of ValidatedNumerics

dpsanders:


tkelman:
Looks right. What's broken?

tkelman:
Wait no, there aren't any CRlibm versions available for Julia versions prior to 0.4, so adding this dependency would make those early ValidatedNumerics versions not installable on Julia 0.3 if anyone's still using that. Maybe better to upper-bound the Julia dependency to 0.4 for those early versions instead of adding an unsatisfiable dependency.

dpsanders:
The upper bound is on the version of CRlibm.jl, not on the version of
julia. Or have I misunderstood?

dpsanders:
Ah OK I see.

ValidatedNumerics/versions/0.0.3/requires
```diff
@@ -1 +1,2 @@
 julia 0.0 0.3
+CRlibm 0.0 0.4
```
tkelman: 0.0.1, 0.0.2, and 0.0.3 won't need the CRlibm dependency

dpsanders:
Thanks, think I've got it right now!

dpsanders:
CRlibm v0.4 no longer automatically defines the functions that ValidatedNumerics requires.

ValidatedNumerics/versions/0.6.0/requires
```diff
@@ -1,5 +1,5 @@
-julia 0.4
-CRlibm 0.2.2
+julia 0.4 
+CRlibm 0.3 0.4
```
tkelman: is raising the minimum also intended?

dpsanders: Yes, thanks.

dpsanders:
Could this be merged, please?

PR #7837, opened 2017-02-05
prepare for MathProgBase 0.6 release

mlubin:
MPB 0.6 is [coming soon](https://github.com/JuliaLang/METADATA.jl/pull/7838). We dropped support for Julia 0.4 and removed functionality for default solvers, so it warranted a new minor release. Nevertheless most solver interfaces will not be affected.


tkelman:
Should propagate this out to master of all of these.

PR #7843, opened 2017-02-06
Register new package EnglishText.jl v0.1.0

attobot:
Repository: [TotalVerb/EnglishText.jl](https://github.com/TotalVerb/EnglishText.jl)
Release: [v0.1.0](https://github.com/TotalVerb/EnglishText.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/TotalVerb/EnglishText.jl/builds/198693998)
cc: @TotalVerb


TotalVerb:
## List of initial features

- indefinite for finding the right indefinite article to use
- english for converting a positive integer to an English-language wordy representation, and the inverse function - unenglish to convert the wordy representation back.
- pluralize to pluralize a noun
- ItemList to create a list of objects
- ItemQuantity to associate a noun with a number
- sentences to iterate over the sentences of a text

## Motivation

Many applications display information to readers in prose format instead of tabular format. It is often important to generate human-readable, grammatically correct prose. However, taking care of grammatical special cases is tedious. For instance, the following code:

```julia
julia> string(ItemList([ItemQuantity(2, "dog"), ItemQuantity(1, "cat")]))
"2 dogs and 1 cat"
```

Behind the scenes, a variety of special cases are taken care of.

## Approach

EnglishText.jl uses established algorithms where available. The precise methods used are documented in the modules themselves.

EnglishText.jl uses a modular approach. Applications not requiring all the exports may use a submodule, such as `EnglishText.ItemLists` instead of the entire package.

## Goals

- To provide a convenient, universally useful approach to abstracting away grammatical special cases
- To be self-documenting where possible, but well-documented nevertheless
- To not have unnecessary performance bottlenecks

## Non-Goals

This is not an NLP package and does not aim to include an English parser.

## Use in the Wild

I am using this package in numerous personal projects, including a static site generator I hope to release this year.

## Organizations

This package is probably a good candidate for inclusion in JuliaText, though I do not know any member of that organization.

tkelman:
Don't use `Pkg.dir` to refer to data files within a package, that won't be correct if the package is loaded from elsewhere on LOAD_PATH. Use `dirname(@__FILE__)` instead.

TotalVerb:
Thanks for the comments on this and the license. These issues have been fixed.

EnglishText/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+
+JSON 0.8.2
+Compat 0.8.5
```
tkelman: `@__DIR__` was not added to Compat until 0.9.1

TotalVerb:
Thanks again; I've updated the Compat version requirement.

PR #7844, opened 2017-02-06
Register new package MultiScaleModels.jl v0.0.1

attobot:
Repository: [JuliaDiffEq/MultiScaleModels.jl](https://github.com/JuliaDiffEq/MultiScaleModels.jl)
Release: [v0.0.1](https://github.com/JuliaDiffEq/MultiScaleModels.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/MultiScaleModels.jl/builds/198712293)
cc: @ChrisRackauckas


ChrisRackauckas:
Note that this is the start of a new naming scheme in DiffEq-land, which are packages build around making it easy to build domain-specific types of models which are compatible with the scientific computing stack. This is the first in the series, for building multi-scale heirarchical models, which lets them "look" like an array. It's not DiffEq-specific at all, and the v0.1.0 release will make sure it actually has no dependency on the DiffEq stack at all (right now it has one on DiffEqBase.jl, which I don't know if it's necessary but I'd like to test it more). The type structures defined by the package could have many uses outside of DiffEq (some have already expressed interest), and so I am keeping this separate.

A linker package MultiscaleDiffEq.jl will be setup which will need DiffEqBase.jl, and may need some larger dependencies like OrdinaryDiffEq.jl and StochasticDiffEq.jl, and will add DiffEq-specific functionality (like further enhancements to the integrator interface when solving multiscale models).

Next in this lineup will be FinancialModels.jl, which is for defining the financial models. Since these are always SDEs, PDEs, or Markov Chains (such discrete equations can be solved using the latest DifferentialEquations.jl), it ends up being pretty DiffEq-specific. I am wondering if it should be renamed to FinancialDiffEq.jl for this reason, but it'll be coming up soon.

Then the next one in this series will be BiologicalModels.jl or BiologicalDiffEq.jl, for building systems biological models for simulation via the DiffEq stack. Naming will match the financial parts.

ViralBShah:
I like the idea of having the DiffEq suffixes in the package names, because without it, they sound very broad. With DiffEq in the name, at least I personally got a very clear idea of what the package might be about. It also enhances the DiffEq identity.

ChrisRackauckas:
Okay, I'll stick to that and submit the registration PRs. 

MultiScaleModels.jl is different for the reasons noted though, different enough that I think it and it's docs should be kept separate from the rest of DiffEq, since it's about creating array types (which just so happen to be solvable in DiffEq, but in reality can be used anywhere that was smart enough to ask for an `AbstractVector` instead of a `Vector`).

ChrisRackauckas:
Speaking of naming though, would things be allowed to be re-named during the Pkg3 transition? I think ParameterizedFunctions.jl should be ParameterizedDiffEq.jl. It was just named and registered before the whole JuliaDiffEq setup, but it's DiffEq-specific and so it should probably follow the naming scheme.

tkelman:
If this package is about a set of multi-resolution tree array types, I wouldn't call it "Models" as that can mean very different things in different fields.

ChrisRackauckas:
Do you have a suggestion for a better name? I'm willing to take suggestions. I would've thought this was the obvious name since I am using it to build multi-scale models.

tkelman:
MultiScaleTypes? MultiScaleBase ?

ChrisRackauckas:
I see where you're going. MultiScaleArrays?

ChrisRackauckas:
Alright. Let me update the name of the basic abstract type to MultiScaleArray, along with the name.

KristofferC:
With the rate new packages, +1 for including `DiffEq` in the name as much as possible. It works well as a namespace.

ChrisRackauckas:
Alright, it's MultiScaleArray.jl. This PR will close and a new one will open for that. The DiffEq-specific modeling packages were renamed

- DiffEqFinancial.jl (PR open)
- DiffEqBiological.jl (waiting on DiffEqJump.jl)

DiffEqJump.jl (for defining jump equations) will have a registration PR open, and DiffEqBiological.jl will depend on that. That will complete what was discussed all the way back in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59 .

ChrisRackauckas:
This PR should close. I don't think attobot closed it because of the name change?

PR #7853, opened 2017-02-07
Tag TaylorSeries.jl v0.3.1

lbenet:


TaylorSeries/versions/0.3.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat 0.15.0
```
tkelman: what are you relying on that was first released in 0.15.0?

lbenet: I don't not relying in too much; I simply included the current version I have for Julia v0.4. 

Do you think I should revert it to what it was before? If so, should I re-tag another version, or should I delete the current tag and apply it to another commit? 

Thanks for your help and time.

tkelman: Minimum version requirements should reflect the lowest version that implemented features or bug fixes you absolutely rely on, not what you happened to have installed when testing (the latter will almost always be too tight, preventing installation of version combinations that would actually work fine together). Ideally you would test the former too but we'll hopefully automate that before too long.

Could modify the requires file in place in this branch of your metadata fork as long as you make the corresponding change on master before your next tag, if there aren't any code changes required.

lbenet: I guess I should do it correctly in the package first, and then make the PR in METADATA, or update the current one. 

On the package, should I revert (https://github.com/JuliaDiff/TaylorSeries.jl/commit/b3729df5500dae166567ff64d2106f1d19227989), find the proper version, update, and retag? What is unclear to me is whether I should delete the previous tag or simply add a new one? 

Thanks for the time, and sorry for my lack of git literacy.

tkelman: It's not an absolute requirement that the package's version of REQUIRE exactly matches the metadata requires file. They can differ, the metadata version is the one that actually matters for release tags of registered packages. So if the only thing you're going to change in the package is the minimum version, you don't actually need to modify the sha1 or git tag to point to a new release that has that change. You can just modify the metadata requires file in place for the new tag without changing anything else, and it's not a problem to have them be a little different.

tkelman: if you go that route you should make the corresponding change in the package too prior to the next time you plan on making a tag, but it can be separate and doesn't require immediately modifying the git tag or release sha

lbenet:
Updated the requires files; I'll revert back to the same Compat.jl version in the package soon, and certainly before tagging a new version where we drop support to Julia v0.4

mlubin:
I don't think it's reasonable to ask package authors to verify which was the precise version of Compat that introduced the features that they're using. A known working version should be sufficient, and if someone is bothered by it they should be free to make a PR to address the issue.

lbenet:
Thanks @mlubin for your comment.

As I mentioned, the idea is to tag a last version with support to Julia v0.4. The reason is that solving some deprecations in Julia 0.6, I ended up having tests not passing in Julia 0.4 (see [this PR](https://github.com/JuliaDiff/TaylorSeries.jl/pull/82)).

I am open to any option; please let me know how to proceed.

tkelman:
@mlubin not annotating this carefully causes real problems when you've pinned package versions for reproducibility. Given a package author should know what features they're using I can always assist with determining the appropriate versions.

lbenet:
Thanks! So I'll set back in `REQUIRE` the version of Compat.jl, and try to be more careful about this the next time.

dpsanders:
It would be great if there were an automated way to discover the required `Compat.jl` version.

PR #7859, opened 2017-02-07
Register new package SpecialFunctions.jl v0.0.1

attobot:
Repository: [JuliaMath/SpecialFunctions.jl](https://github.com/JuliaMath/SpecialFunctions.jl)
Release: [v0.0.1](https://github.com/JuliaMath/SpecialFunctions.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaMath/SpecialFunctions.jl/builds/199128757)
cc: @ararslan


martinholters:
With the `julia 0.5` requirement, it will be nearly impossible to maintain a package that uses any of the functions being removed in Julia 0.6 in a way that it runs on Julia 0.6 and also still on Julia 0.4. Is that an issue? (We will mandate `julia 0.5` as a minimum in METADATA shortly after Julia 0.6 is released, right?)

ararslan:
@martinholters In user code (not in a package since we don't have conditional dependencies), I guess one could do
```julia
if VERSION >= v"0.5.0"
    using SpecialFunctions
    import SpecialFunctions: bessely, ...
end
```
Though that seems like awful practice. However, with that it should work just fine for any Julia version in which the special functions are exported from Base. Packages that need these functions will need a minimum requirement of 0.5 though. I didn't bother verifying/maintaining 0.4 support in the package since this uses the code as defined in Base for 0.6 (pre-removal), which may not be 0.4-compatible.

I don't know about version requirements in METADATA once 0.6 is released. That sounds like a question for @tkelman.

dpsanders:
Can't you put that kind of check directly in the package, and re-export the Base functions for 0.4 and 0.5?

ararslan:
@dpsanders I suppose, though then 0.4 code wouldn't be able to benefit from any enhancements that have been made to the special functions code since 0.4 was released. Just seems like more effort than it's worth to actively support 0.4 at this point.

dpsanders:
Good point. So then could just redefine the special functions that are available in 0.4 and warn the user that they will have a lot of warnings when doing so. Though I agree that it seems reasonable to just drop support for 0.4 now.

martinholters:
One option we could consider is to tag a SpecialFunctions version requiring `julia 0.4 0.5` which is empty otherwise. Then packages could safely require `SpecialFunctions` but would need to do a `VERSION` check with conditional `using`/`import` like above. Not pretty, but workable.

ararslan:
I maintain that it isn't really worth the effort to maintain a version of this package that supports 0.4. If another package that needs special functions uses 0.4, they can find a way to work that out through branches, tags, or what have you. As a 0.6 release approaches, I've found that dropping 0.4 support is becoming less and less controversial, especially due to other incompatibilities.

martinholters:
Just to be clear: I have no problem with SpecialFunctions requiring Julia 0.5. I just wanted to raise awareness and point out a possible solution. If no-one thinks this is a real issue, I'm absolutely fine with leaving this as is.

tkelman:
We won't know the version number at which these get removed from base until that gets merged, but is the plan to have these extend Base on versions new enough where the Base version would otherwise just throw an error?

ararslan:
Yes. Once the Base PR is merged, I'll update the package to import and export the Base symbols. In the interim I figured not (im|ex)porting at all was easiest.

PR #7860, opened 2017-02-07
Tag Images v0.7.0

timholy:
This version choice is me beginning to think about adhering to [semver](http://semver.org/), which means that if you add one new exported function, then it should be a minor version bump. Since there are still many missing algorithms from Images, this will mean that we'll be up to 0.38.0 before you know it. I guess there's nothing really wrong with that, but it's quite different from the way base julia is released and so might be perceived as weird.

Specifically, this bump is for the addition of `feature_transform` and `distance_transform`.


tkelman:
Nope, version numbers are cheap, this is the right way to do it. As long as nobody was expecting the next minor version to have breaking changes and pre-emptively bounding it, this doesn't hurt anything.

Have you looked into attobot yet?

timholy:
I've glanced at the page but haven't sat down to use it yet. On the agenda.

tkelman:
Should probably update your NEWS file accordingly

timholy:
Done. I'll retag on that release (later).

PR #7865, opened 2017-02-07
Tag Images.jl v0.7.0

attobot:
Repository: [JuliaImages/Images.jl](https://github.com/JuliaImages/Images.jl)
Release: [v0.7.0](https://github.com/JuliaImages/Images.jl/releases/tag/v0.7.0)
Diff: [vs v0.6.1](https://github.com/JuliaImages/Images.jl/compare/728216c493a054e75b12401c07d9b4389a00bf46...f408a21365c6b70c93252bdcb6911a507ccc4918)
`requires` vs v0.6.1: no changes
cc: @timholy


KristofferC:
And suddenly, tagging packages aint so bad ;)

timholy:
Still rough when some change requires a synchronized release across ~10 packages. Even when I already have branches pushed for all 10 packages that pass their tests locally, I've learned to carve out an entire weekend for that kind of merge.

simonbyrne:
It would be good to think about how we can reduce that sort of busywork.

tkelman:
Staging things into a specific branch of your metadata fork then making temporary package branches for Travis with modified `Pkg.init` lines would address that part, though it could use a lot more automation there.

timholy:
Relevant: https://github.com/tbreloff/MetaPkg.jl. My most recent escapade started before I knew about that, and it requires a `git purge` that made me nervous. But in the future I'll likely try it.

PR #7866, opened 2017-02-07
Register new package MultiScaleArrays.jl v0.0.1

attobot:
Repository: [JuliaDiffEq/MultiScaleArrays.jl](https://github.com/JuliaDiffEq/MultiScaleArrays.jl)
Release: [v0.0.1](https://github.com/JuliaDiffEq/MultiScaleArrays.jl/releases/tag/v0.0.1)
cc: @ChrisRackauckas


tkelman:
Seems you might not want to define a `*(m::AbstractMultiScaleArray,y::AbstractMultiScaleArray)` method defaulting to elementwise, depending what the matrix size semantics of these are supposed to be.

ChrisRackauckas:
That's a good point. They are supposed to act like a Vector for all math, so it should just error like a vector.

ChrisRackauckas:
I implemented that change, and it'll be in the v0.1.0. I actually have a v0.1.0 which adds a bunch more features, at the cost of a DiffEqBase dependency and requires tags to DiffEqBase, OrdinaryDiffEq, and StochasticDiffEq for full effect. I am arbitrarily planning to wait a little bit on that just to give you all a break. Thanks for helping/dealing with round of tags, you do good work as always.

PR #7867, opened 2017-02-07
Register new package DiffEqJump.jl v0.0.1

attobot:
Repository: [JuliaDiffEq/DiffEqJump.jl](https://github.com/JuliaDiffEq/DiffEqJump.jl)
Release: [v0.0.1](https://github.com/JuliaDiffEq/DiffEqJump.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqJump.jl/builds/199268243)
cc: @ChrisRackauckas


tkelman:
It's a bit odd that none of your tests actually check that the results are meaningful

ChrisRackauckas:
It's much easier to do with plots, but that works well for developing but not so well for unit tests. I'm not sure of a good way to do it with numbers, other than to try and derive some mean statistics and do a Monte Carlo experiment of sorts. So that's actually what I am doing right now. I just wanted to get the registration start, but will be following up with a v0.1.0 which adds the variable rate jump implementation mentioned in the docs, (along with hopefully some test which is verified by some hand calculation somehow, if I can get something like that. If not, at least a regression test).

PR #7883, opened 2017-02-09
Register GraphGLRM [https://github.com/mihirparadkar/GraphGLRM.jl.git]

mihirparadkar:


tkelman:
will need a tagged version - you may want to use https://github.com/attobot/attobot, makes things a bit easier

mihirparadkar:
Thanks, the attobot seems to take care of the version tagging.

PR #7885, opened 2017-02-09
Tag OrdinaryDiffEq.jl v1.3.1

attobot:
Repository: [JuliaDiffEq/OrdinaryDiffEq.jl](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl)
Release: [v1.3.1](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/releases/tag/v1.3.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/builds/199909649)
Diff: [vs v1.3.0](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/compare/cc1f795d9a743ff6ddf3aa87fe409ea48a2942f0...066cb78ce282919845258b49e55a0afcb9062886)
`requires` vs v1.3.0: no changes
cc: @ChrisRackauckas


ChrisRackauckas:
The tests will fail here until https://github.com/JuliaLang/METADATA.jl/pull/7884 is merged, but that release is not required in any way.

tkelman:
It's not required, except to make the tests pass? Huh?

ChrisRackauckas:
Well it's required to make the `DEDataArray` functionality available, and there is a test for that now. 

http://docs.juliadiffeq.org/latest/features/data_arrays.html

However, `DEDataArray` is only mentioned in the test itself, and the patches for compatibility do not make reference to it at all. So if a user got this version without getting the updated DiffEqBase then all that would happen is that new test would fail, but everything else would be fine.

tkelman:
Then it sounds like your test/REQUIRE needs a minimum version requirement.

ChrisRackauckas:
But the change is in DiffEqBase which is in REQUIRE, and doesn't need a higher minimum version requirement.

tkelman:
right, so this is one of the few cases where listing something in both REQUIRE and test/REQUIRE serves a purpose - the REQUIRE dependency can go without listing a minimum version if it's not needed for library functionality, but test/REQUIRE can list a minimum version to ensure the tests will pass

ChrisRackauckas:
I see. I didn't know you could do that.

PR #7888, opened 2017-02-09
Register RNG.jl

sunoru:
https://github.com/sunoru/RNG.jl/issues/17


tkelman:
We try to discourage acronym names when they aren't wrapping an underlying library.

ViralBShah:
+1. I would prefer to pick a slightly more descriptive name due to the global namespace, and also because people may want to use RNG as a type in their own codes. I love this package, but I hope we can find a slightly different name.

sunoru:
Hmm.. So how about just `Random.jl`? Because we aim to replace `Base.Random`..

sunoru:
Hum.. I was considering "Random.jl" because we aim to replace `Base.Random`. But this package only includes several different random number generators instead of the various random functions. Actually I think RNG is a very common acronym and imo it's the most descriptive name. Something like "RandomNumberGenerator.jl" is to long for a name while others are not as descriptive as it.

dpsanders:
RandomNumbers.jl?

sunoru:
I don't know... Perhaps it's the best compromise because `RNG.jl` and `RandomNumberGenerators.jl` are neither suitable.

dpsanders:
Personally I don't see the problem with RandomNumberGenerators.jl. 
It makes the code more readable, and tab completion means that you almost never have to type the whole thing.

ViralBShah:
I like RandomNumbers.jl. Or RNGCollection.jl too. @sunoru pick what you like of all the suggestions and let's go with it.

sunoru:
Fine. Then maybe I just choose RandomNumbers.jl.

ViralBShah:
Thank you. I will close this, and we can then merge the new one.

PR #7893, opened 2017-02-09
Register new package Tensors.jl v0.4.0

attobot:
Repository: [KristofferC/Tensors.jl](https://github.com/KristofferC/Tensors.jl)
Release: [v0.4.0](https://github.com/KristofferC/Tensors.jl/releases/tag/v0.4.0)
Travis: [Travis Build Status](https://travis-ci.org/KristofferC/Tensors.jl/builds/199999483)
cc: @KristofferC


KristofferC:
This is a renamed package from https://github.com/KristofferC/ContMechTensors.jl. The old name is not suitable because it has nothing really to do with Continuum Mechanics. After this is merged I will push a tag to ContMech saying it is renamed and how to Pkg.pin to remove the deprecation warning.

tkelman:
How does this compare to TensorOperations or other related existing packages?

ViralBShah:
I usually don't like to suggest package names, but this seems like a very general name. Is it that all Julia tensor development will center around this package?

KristofferC:
TensorOperations provide a macro to do index notation on Julia arrays. 

Regarding the name:

"Tensors are geometric objects that describe linear relations between geometric vectors, scalars, and other tensors. Elementary examples of such relations include the dot product, the cross product, and linear maps."

The types in this packages maps exactly to these geometric objects and the functions defined maps exactly to these relations.

KristofferC:
Can I merge this?

tkelman:
Still of the opinion that this is an overly generic name.

tkelman:
LowOrderTensors perhaps, if this doesn't implement anything beyond order 4?

KristofferC:
You never shared your opinion that you thought this was overly generic. Anyway, support for higher order tensors could be added with not much effort. The focus is just not on that right now. If anyone is interested in that, PRs would be welcome.

I think the naming requirements are quite arbitrary, we already have for example `JuliaFEM`, `Calculus`, `EllipticFEM`, `Images`, `Monads`, `Plots` etc. Considering all the other generic names in METADATA that have gotten approved, my opinion is that this is a perfectly fine name.

tkelman:
JuliaFEM and Plors were both recommended against, Calculus/Images/Monads are all quite old. EllipticFEM seems fairly specific to me, though the acronym isn't ideal. This package is already registered under a more specific name, you're just suggesting it's too specific and not applicable any more?

KristofferC:
What happens after a recommendation and the author still want to use the name? Is a recommendation the same as a blanket ban on that name (it can't be since the other packages got registered)?

Regarding the specifier of the name (Continuum mechanics) it doesn't make any sense because you could use it just as well for something else, like electromagnetism. Having a limiter on the package name that is non sensible is harmful for the package. Imagine a `GreenColors` package. 

There are other ways to define "tensors" than how they are defined in the package (the ML crowd have stolen the tensor term and turned it into something completely different). However, this package operations and types are defined from the mathematical / physical meaning of tensor. If you google tensor you get links like https://www.grc.nasa.gov/www/k-12/Numbers/Math/documents/Tensors_TM2002211716.pdf, http://zeus.plmsc.psu.edu/~manias/MatSE447/03_Tensors.pdf which are EXACTLY the types and operations that the package provide.

KristofferC:
Bump

KristofferC:
Bump.

tkelman:
It's a shared namespace, at least until Pkg3, so by claiming a general name you're effectively saying your package is the definitive one and only possible approach for a tensor package. And taking 2 names for the same package.

KristofferC:
I can delete the previous one if that is preferable or transfer it to someone else if the name is wanted. Other people have renamed their packages.

> by claiming a general name you're effectively saying your package is the definitive one and only possible approach for a tensor package

Definitely not. There is a package called "DifferentialEquations" and yet there are a lot of more differential equation packages. There is a package called "Equations" and yet, a lot of people have stuff relating to equations in their packages. There is a package called "Graphs" and yet there are more packages dealing with graphs. 

In that case every name is overly general because having something (ugly) like "LowRankTensors" indicate that this is the only possible approach for low rank tensors which is obviously not true. Same with "HttpServer", it is of course not the only way to have a http server. It is ONE of the way.s

tkelman:
deleting a package is not good for reprodibility, metadata is not designed for deletions. it's also difficult to rename package smoothly, so it's been very rare.

KristofferC:
I know, which is not why I am deleting the package. You brought up the fact that I "occupied" two names and my response was that I could transfer it if someone wanted that name.

If Pkg3 solves this issue with namespaces and is destined to be released in a year or so, can we just merge this and get on with our lives?

ViralBShah:
I am merging this for now. We already have other generic package names - and I only wanted to point it out. I assume that unmaintained packages can empty the namespace in the future, until Pkg3 comes about, should the need arise.

tkelman:
And this males the problem worse.

ViralBShah:
Well, maybe it might and maybe not - I don't know. You definitely have more foresight and experience with the package ecosystem. We need a general solution, which is in the works and let's help accelerate that. I know that not everyone agrees on this, but I personally don't feel the need to be rigid about naming beyond a simple discussion.

PR #7903, opened 2017-02-10
Tag MacroTools.jl v0.3.5

attobot:
Repository: [MikeInnes/MacroTools.jl](https://github.com/MikeInnes/MacroTools.jl)
Release: [v0.3.5](https://github.com/MikeInnes/MacroTools.jl/releases/tag/v0.3.5)
Travis: [Travis Build Status](https://travis-ci.org/MikeInnes/MacroTools.jl/builds/200392462)
Diff: [vs v0.3.4](https://github.com/MikeInnes/MacroTools.jl/compare/8ded92884b20d6c356868b5de5c379e63d1b9902...23ca18070eee5d831abe04318ad666c2b3c651af)
`requires` vs v0.3.4: no changes
cc: @MikeInnes


tkelman:
This has started causing segfaults in packages that depend on DataFrames on Julia 0.5.

abelsiqueira:
Also segfaults for packages depending on Lazy.jl

timholy:
It doesn't appear to cause the problem on `release-0.5` (`0.5.1-pre+31`). Pretty please, can we have a julia 0.5.1?

tkelman:
I'm working on it, but yes, it does. It's intermittent.

tkelman:
My mistake, I was reproducing via Mimi.jl which was actually shelling out to whatever `julia` happened to be on the path, which wasn't the one I thought I was testing against. Even if I get 0.5.1 released soon, 0.5.0 is still widely used and this release should be reverted for now. Either marked 0.6-only, or replaced with an 0.3.6 that has the same sha as 0.3.4.

aviks:
PR #7921 reverts this.

tkelman:
it shouldn't just be reverted, but superseded

tkelman:
or have its requirements adjusted (which would possibly require a test-list exemption, depending which direction we do it in)

tkelman:
We have 3 immediate options here, the faster we do one of them the better.

1. tag a MacroTools 0.3.6 using the same commit as 0.3.4 - this needs someone with commit access to create the git tag to pass travis, though we could temporarily disable the check if @MikeInnes or anyone else with commit access don't help here
2. upper bound MacroTools 0.3.5 to be 0.4-only
3. lower bound MactoTools 0.3.5 to be 0.6-only

Option 1 would be easiest IMO. It may be possibly to do both 2 and 3 but it would be the first time as far as I know that that union of version ranges feature was used in metadata, it's not a widely tested capability.

PR #7910, opened 2017-02-10
Tag Weber v0.3.3 [https://github.com/haberdashPI/Weber.jl]

haberdashPI:
Release: [v0.3.3](https://github.com/haberdashPI/Weber.jl/tree/v0.3.3)

TravisCI: [Build Status](https://travis-ci.org/haberdashPI/Weber.jl)
AppVeyor: [Build status](https://ci.appveyor.com/project/haberdashPI/weber-jl/branch/master)


tkelman:
A few of the tests use DataFrames and VideoIO, should those be listed in test/REQUIRE ?

haberdashPI:
Those tests are used to empirically evaluate run-time latency: you run the script and record sound or video, and then analyze it to see what the timing is actually like. They aren't run when you call Pkg.test("Weber"). (Comments in those scripts explain some of this). They are more of a work in progress rather than a user-facing set of functions. When I clean them up and provide some better documentation for them, then it make make sense to add them to REQUIRE (but in that case my guess is that they will be a part of src rather than tests).

PR #7912, opened 2017-02-10
Tag Optim.jl v0.7.5

attobot:
Repository: [JuliaOpt/Optim.jl](https://github.com/JuliaOpt/Optim.jl)
Release: [v0.7.5](https://github.com/JuliaOpt/Optim.jl/releases/tag/v0.7.5)
Travis: [Travis Build Status](https://travis-ci.org/JuliaOpt/Optim.jl/builds/200508576)
Diff: [vs v0.7.4](https://github.com/JuliaOpt/Optim.jl/compare/f29374a69b59c470a4b23d99f6db541efec4a233...c667a35d96ac5ff8615c57411e04bf9385e5d061)
`requires` vs v0.7.4: no changes
cc: @pkofod


tkelman:
This broke several dependent packages, see http://pkg.julialang.org/pulse.html

pkofod:
I see. This appears to be a deprecation issue. I am not near a computer, but will have a look when I am (tomorrow local time)

I haven't followed pkg3 development but are there any ideas out there for how to easily run tests of all packages in Metadata that currently has a require entry for your package? Some packages are very creative with their use of optim (manually using features that are not part of the user facing API) and as a result it can be hard to predict if packages will break given a change, if you don't know the code of all the packages.

tkelman:
That would be independent of Pkg3, we are working on some automated reverse-dependency testing that could run on demand.

pkofod:
> That would be independent of Pkg3, we are working on some automated reverse-dependency testing that could run on demand.

Cool.

So how would you prefer I fix this? Upper limits on the current versions of celeste and nlsolve + PRs, or?

tkelman:
something like that. I can reproduce the NLsolve failure if I use a clean JULIA_PKGDIR

pkofod:
Okay, I may just have to put versions in require first then.

pkofod:
So I can see that one of the deprecations fail to work in NLsolve, but how do I just put an upper limit in a require? Should I put `0.0 0.7.5` if 0.7.5 is the first version I know doesn't work? I have no idea to what extent the current NLsolve is compatible with other versions of Optim. And should I do it to all NLsolve versions or just the current? (sorry, I have had next to none computer time this week)

tkelman:
yeah, just an upper bound is fine if that's all the information you have. needs to be done to all past versions, unless the older versions somehow work correctly with latest Optim

pkofod:
Happy to help our @jeff-regier et al if it's a problem, but as you say, it is my impression that they're on master (or something not tagged at least) to run their code on master Julia.

jeff-regier:
Shouldn't be a problem, we just haven't a release recently. Thanks for the concern.

PR #7915, opened 2017-02-11
Tag Mimi v0.2.2 [https://github.com/anthofflab/Mimi.jl]

davidanthoff:
This is a very minimal bugfix release. Would be great if it could be merged quickly, there is a whole group of researchers waiting for this :)


tkelman:
you need a minimum version dependency on Distributions, your tests fail against older versions

davidanthoff:
I'm tempted to just let it be for now, v0.2.2 was really just a one line bug fix release, and v0.3.0 is around the corner and should have all the right minimum versions etc. Plus figuring out reasonable lower bounds would take some time, and this week is insane for me.

PR #7919, opened 2017-02-11
Reclaim deprecated HTTP.jl package

quinnj:


tkelman:
> `Base.take!(io::Base.AbstractIOBuffer) = takebuf_array(io)`

Please let Compat do these.

tkelman:
> `Base.getindex{T}(A::Vector{T}, i::http_parser_url_fields)`

instead of using a type for this is strange. I still think duplicating URIParser inside a monolithic package that's a bigger and more complicated dependency than many consumers need, instead of using/improving it, is a mistake.

tkelman:
top-level license file should mention that src/parser.jl was derived from separately copyrighted code

tkelman:
> `function Base.print(io::IO, sch::String, userinfo::String, hostname::String, port::String, path::String, query::String, fragment::String)`

type piracy, changes the behavior of anything else that happens to call `print` with 7 strings

quinnj:
Responses:
  * I can always pull in Compat later
  * I'm definitely not planning on keeping the URI code in HTTP long-term (per our previous discussion). For now though, I think it's prudent to keep it exclusive to HTTP.jl while things beta and iron out over a release or two, and then update URIParser. 
  * I did put license info in src/parser.jl itself, is that not sufficient? Happy to put it top-level as well, along w/ the note about the go cookies code as well
  * More like "more-specific-method" piracy, right? 😄  I'll figure out a better setup there.

tkelman:
> I can always pull in Compat later

The problem is both you and Compat trying to override the exact same method. Will almost certainly give warnings if anything tries to use both your package and Compat.

tkelman:
Anywhere you're using an index type to indicate how a vector of bytes should be interpreted would be made clearer by actually using a dedicated wrapper type.

quinnj:
> Anywhere you're using an index type to indicate how a vector of bytes should be interpreted would be made clearer by actually using a dedicated wrapper type.
I don't follow your concern here; I'm just defining getindex for a enum type that is guaranteed to be 1 <= i <= 8. I don't see any harm in that.

tkelman:
You're right, that one's not as odd as I thought it was. `Base.getindex(A::Vector{UInt8}, o::Offset) = String(A[o.off:(o.off + o.len - 1)])` is though, indexing with a specific type turns the bytes into a string?

quinnj:
Alright, changed the `Offset` getindex method to return bytes and added the Compat dependency. Waiting on travis before I redo the commit here.

quinnj:
Alright, I think we're all squared away here.

HTTP/versions/0.4.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+MbedTLS
```
tkelman: is there a minimum version that you're relying on for any types or methods?

quinnj: hmmm, perhaps, let me check.

quinnj: Might as well use 0.4.0, the 0.5 only release. Updated.

tkelman: didn't some of the earlier releases also work on 0.5? mainly asking about whatever features you're relying on

quinnj: I think the features I needed were 0.3.1, which was the last 0.4/0.5 release, and then 0.4.0 was the 0.5-only release.

tkelman: then sounds like 0.4.0 is a bit too tight? and/or the package wasn't following semver all that well

tkelman:
This is effectively a new package and should have stayed open the same waiting period.

quinnj:
Ah, I didn't get that impression from our discussion of preserving old git tags and such from ye olde HTTP.jl. Got a little excited!

tkelman:
yes technically it's a tag and url move, but the code's totally replaced

PR #7921, opened 2017-02-11
Revert "Tag MacroTools.jl v0.3.5"

aviks:
Reverts JuliaLang/METADATA.jl#7903

Causing segfaults. Is it safe to revert this?

PR #7927, opened 2017-02-12
make MacroTools 0.3.5 julia-0.4-only

tkelman:
closes https://github.com/JuliaLang/julia/issues/20581

ref #7903

replaces #7921


MikeInnes:
It would make more sense to make it 0.6-only, given that the patch release was only to fix depwarns there anyway.

tkelman:
Yes, but that would require introducing a versioning policy exception that I didn't want to make. You're welcome to tag new versions that are 0.6-only. And given this package is highly depended on and it took you 2 days to respond here, could you please give a few other people commit access?

PR #7948, opened 2017-02-13
Bump Interpolations and OffsetArrays

timholy:


tkelman:
> `Base.fill(x, inds::Tuple{UnitRange,Vararg{UnitRange}})`

that method really shouldn't be extending `Base.fill` - would be fine as `OffsetArrays.fill` though

tkelman:
If you're going to use DualNumbers in Interpolations, it should be in REQUIRE

timholy:
Thanks. Hold off on this, looks like we weren't running Travis tests on 0.4, and it's not working.

> that method really shouldn't be extending Base.fill - would be fine as OffsetArrays.fill though

It's complicated. Methods are designed to dispatch by `indices` type. Base arrays handle `OneTo`, and packages handle non-1 arrays. OffsetArrays is the one package designed to work with `UnitRange` indices, all others specialize for their own internal indices. Since `UnitRange` is the fallback for `AbstractUnitRange` indices, it seems necessary to have a fallback corresponding array type available.

timholy:
OK, I think this is fixed.

PR #7951, opened 2017-02-14
Tag CoordinateTransformations.jl v0.3.3

attobot:
Repository: [FugroRoames/CoordinateTransformations.jl](https://github.com/FugroRoames/CoordinateTransformations.jl)
Release: [v0.3.3](https://github.com/FugroRoames/CoordinateTransformations.jl/releases/tag/v0.3.3)
Diff: [vs v0.3.2](https://github.com/FugroRoames/CoordinateTransformations.jl/compare/529d9501be719490f54a503d8130e0ed10822cac...67491e4df08a66ef23373b6b63ebbee921246a54)
`requires` vs v0.3.2: no changes
cc: @andyferris


CoordinateTransformations/versions/0.3.3/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
```
tkelman: do you still need to support prereleases?

tkelman:
if you're adding new features that other packages will depend on, then it's not really a bugfix release according to semver

andyferris:
Ahh... so can you please explain what is the word with semver before v1.0? 

Plenty of packages make breaking changes and use a minor tag (take Images, for example) and similarly add new features with a patch tag... The levels get pushed down, no?

tkelman:
Technically semver says anything goes under 0.x, but that's not very useful. If anyone will need to depend on a new feature or avoid a breaking change, it's worth using at least a minor version IMO.

andyferris:
(Sorry this got a bit long...).

OK... I am just wondering if that is optimal for users. Generally, semver is useful because it provides *information*. The primary information I want is "is there a breaking change". Having minor mean both breaking changes and new features means I can't unambiguously predict breaking changes - that is, it reduces the amount of information I have about breaking changes. OTOH, new features are not meant to be breaking, so typically nothing goes *badly* if I use a larger set of features, so I don't need that information as strongly. (And there is still a mechanism for setting a minimal patch level for required new features, it just takes a little more investigation for figuring out when exactly the feature I need is introduced). Given that for 0.x.y releases, we are more-or-less "compressing" 3 numbers (pieces of information) into 2 numbers, I think it is important to recognize which piece of information is of primary consideration and which remaining two pieces of information must become secondary. IMO breaking changes are the clear winner (i.e. I'd rather be able to figure out which is the maximum compatible version of something quickly then figure out which the minimal working *bugfix* level is). Note that your suggestion here is allowing full (uncompressed) information about bugfixes while compressing/mixing breaking changes and new features into a single number, which when said like that seems like the least desirable situation to me.

Maybe you can also clear up how the Julia community has operated in the past (as you would be the best person placed to know)? Maybe I'm incorrect, but I'd be a bit surprised if *Images* has only released 7 rounds of new features or that *DataFrames* has only released new features 6 times. Perhaps they have been conservative and release new features together *en mass* and only on occasion, but for e.g. *StaticArrays* I've been creating and releasing a trickle of small features all the time (using patch levels unless breaking - though with *StaticArrays* in particular it's not so clear because "new" features really means "implement a specialized method for a function from `Base.LinAlg` that was previously either broken or slow when taking a `StaticArray` as input").

Anyway - this isn't to start an argument or to derail this process. I just wanted to share an opinion with you and see what you thought. Please say "just do it" if you're too busy to discuss, and we're happy to comply. 😄

tkelman:
Most package authors have more or less ignored semver. I'm trying to get people to pay closer attention to their dependencies, including what features they're relying on. Version numbers are cheap and I see no reason minor version changes should be made less frequent so you can use them only for breaking changes. If you want to do that, just go to 1.0 and be prepared for a 2.0 the next time you want to break something. Mixing bug fixes and feature additions in the same patch level is also losing information, and is more harmful in my mind. Minor versions are functionally different, patch versions should be functionally equivalent - if code works with 0.3.8, it should also work with 0.3.2 unless you hit a bug  that has been fixed. If you've introduced features between patch releases and people are relying on those features, that won't work.

andyferris:
Thanks for your point of view. 

I think there is a good reason why semver has three numbers - having either "breaking vs non-breaking" or "feature change vs bugfix" isn't  really sufficient. 

I think going to 1.0.0 and just following plain semver would be a good move - but we all wait for Julia's lead here and the official docs do ask us to "be conservative". I feel you will have less trouble getting compliance with semver when everyone's package is v1+. I would suggest changing that official recommendation to "use vanilla semver" in the docs for Julia v0.6 since it is touted as the last version before v1.0, and as you say, numbers are cheap - even with "prerelease" programming languages. :)

For now I think we will follow your suggestion - the most important thing (to me) here is consistent community expectations and you are really leading the charge here. (Note that I'll be busy till tomorrow).

tkelman:
I don't interpret "be conservative" as "use the same level of versioning for features and bugfixes" at all. As long as there's a way to distinguish them, what's the issue with using the same level of versioning for breaking changes and feature additions? And as long as it isn't the most granular level of versioning, it leaves open the possibility of backporting bug fixes to other functionally different release series. Using patch for everything doesn't.

tkelman:
Yeah, we should revise that and just say "semver is strongly encouraged" with a link and a brief explanation of what that means, if there isn't one yet

andyferris:
Please close this in favor of #7964

PR #7947, opened 2017-02-13
Tag ImageCore.jl v0.1.2

attobot:
Repository: [JuliaImages/ImageCore.jl](https://github.com/JuliaImages/ImageCore.jl)
Release: [v0.1.2](https://github.com/JuliaImages/ImageCore.jl/releases/tag/v0.1.2)
Travis: [Travis Build Status](https://travis-ci.org/JuliaImages/ImageCore.jl/builds/201238497)
Diff: [vs v0.1.1](https://github.com/JuliaImages/ImageCore.jl/compare/8b01c7e1f3437e2c7e75e43ac7701f3eee2d0430...b31e927e9a7591ea5ead2e8d0ff33acb59381f1e)
`requires` vs v0.1.1: 
```diff
--- 0.1.1/requires
+++ 0.1.2/requires
@@ -1,5 +1,6 @@
 julia 0.5
 FixedPointNumbers 0.3.0
+ColorTypes
 Colors
 MappedArrays 0.0.3
 Graphics
```
cc: @timholy

PR #7963, opened 2017-02-14
Tag ImageCore.jl v0.1.2 [https://github.com/JuliaImages/ImageCore.jl/…

tkelman:
…releases/tag/v0.1.2]

plus some version bounds on ~~FixedPointNumbers for~~ very early versions of Color.jl and Gadfly, which helps with the resolution issues seen in (closes #7947), at least locally for me on julia 0.6-dev. Let's see what it does on other Julia versions. cc @timholy


tkelman:
Had to push some tags to Color.jl (I thought I fixed the travis check to only be picky about git tags if you're adding new sha1 files, but apparently it didn't work correctly), but looks like I fixed 0.6-dev compared to #7947. Now will see how to fix Mads on 0.5.

tkelman:
By faking VERSION, I've learned that JuliaLang/julia#20319 fixes the remaining issue. I will see if that's backportable, but in the meantime I'll also want to see what bounds tweaks I could make the get things happy without needing to modify JULIA_PKGRESOLVE_ACCURACY.

tkelman:
Well that felt a lot like shuffling around the deck chairs. Anyway, good enough for now I think.

timholy:
Thanks a bunch, @tkelman! And of course @carlobaldassi.

timholy:
Should say, this now gives the whole Images suite a fighting chance of getting through CI on 0.6, so it should have quite a cascade.

tkelman:
I'm glad we isolated the odd resolver behavior and uninformative failure messages enough to improve them for 0.6 - should make future debugging of this kind of situation much easier.

timholy:
Yes. And while you know of my skepticism about adding upper bounds (based on early attempts at doing similar things locally), as usual the right answer is not to avoid it, but to fix it. So, here's a new picture of me:
image

PR #7976, opened 2017-02-15
Place an upper bound on ASTInterpreter's use of AbstractTrees

Keno:


tkelman:
this needs to be done on all past tags, or Pkg can just downgrade ASTInterpreter.

please use branches in your fork whenever you can for pr's

PR #7992, opened 2017-02-16
Tag a new version of StructIO

Keno:
Renames the `@struct` macro to `@io` since struct is now a keyword on 0.6. All users need to get upper bounds to keep working. I've tagged all the ones that I have. Don't think there's anybody else.

### DIFF
--- a/DWARF/versions/0.0.0/requires
+++ b/DWARF/versions/0.0.0/requires
@@ -1,3 +1,3 @@
-julia 0.2-
+julia 0.2- 0.5
 ELF
 StrPack
--- a/DWARF/versions/0.0.1/requires
+++ b/DWARF/versions/0.0.1/requires
@@ -1,4 +1,4 @@
 julia 0.5-
-StructIO
+StructIO 0.0.1 0.1.0
 ObjFileBase
 AbstractTrees
--- a/ELF/versions/0.0.2/requires
+++ b/ELF/versions/0.0.2/requires
@@ -1,5 +1,5 @@
 julia 0.5-
-StructIO
-DWARF
-ObjFileBase
+StructIO 0.0.1 0.0.2
+DWARF 0.0.2 0.0.3
+ObjFileBase 0.0.2 0.0.3
 FileIO
--- a/Gallium/versions/0.0.3/requires
+++ b/Gallium/versions/0.0.3/requires
@@ -1,10 +1,10 @@
 julia 0.5-
 TerminalUI
-DWARF
+DWARF 0.0.3 0.2.0
 ObjFileBase
-ELF
-MachO
+ELF 0.0.3 0.1.0
+MachO 0.0.3 0.0.4
 ASTInterpreter
 AbstractTrees
-COFF
+COFF 0.0.2 0.1.0
 CRC
--- a/MachO/versions/0.0.2/requires
+++ b/MachO/versions/0.0.2/requires
@@ -1,3 +1,3 @@
 julia 0.5-
-ObjFileBase
-StructIO
+ObjFileBase 0.0.2 0.0.3
+StructIO 0.0.1 0.0.2
--- a/StructIO/versions/0.1.0/requires
+++ b/StructIO/versions/0.1.0/requires
@@ -1,0 +1,2 @@
+julia 0.5
+Compat 0.17.0
--- a/COFF/versions/0.0.1/requires
+++ b/COFF/versions/0.0.1/requires
@@ -1,4 +1,4 @@
 julia 0.5-
-ObjFileBase
-StructIO
-DWARF
+ObjFileBase 0.0.2 0.0.3
+StructIO 0.0.1 0.1.0
+DWARF 0.0.2 0.0.3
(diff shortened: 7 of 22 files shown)


ELF/versions/0.0.2/requires
```diff
@@ -1,5 +1,5 @@
 julia 0.5-
-StructIO
-DWARF
-ObjFileBase
+StructIO 0.0.1 0.0.2
```
tkelman: since the first tags of ELF had different requirements without bounds, Pkg could choose to downgrade to very old ELF in order to get latest StructIO

MachO/versions/0.0.2/requires
```diff
@@ -1,3 +1,3 @@
 julia 0.5-
-ObjFileBase
-StructIO
+ObjFileBase 0.0.2 0.0.3
```
tkelman: MachO 0.0.1 will also need bounds to prevent downgrading

ELF/versions/0.0.2/requires
Keno: That seems in principle fine. I guess I should add some bounds to Gallium to force a recent ELF

tkelman:
DWARF also depends on StructIO

Keno:
So it does. Forgot to add those to the commit.

ELF/versions/0.0.2/requires
tkelman: will those early ELF versions even load on recent Julia?

Keno: Probably not

tkelman: ELF 0.0.1 has the same StructIO, DWARF, ObjFileBase requirements so it should really be included here

StructIO/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Compat 0.18.0
```
tkelman: abstract type renaming is 0.17, you don't need 0.18 unless you have parametric type aliases

Keno: I don't think it matters that this is accurate.

tkelman: it does when people apply bounds for reproducibility

tkelman: Applying bounds that are too conservative rules out version combinations that would actually work fine together.

TotalVerb:
Would it be more gentle to `@eval` a `@struct` macro for deprecations? This syntax still parses fine on 0.6:

```julia
julia> dump(:(@struct))
Expr
  head: Symbol macrocall
  args: Array{Any}((1,))
    1: Symbol @struct
  typ: Any
```

DWARF/versions/0.0.1/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.5-
-StructIO
+StructIO 0.0.1 0.1.0
```
tkelman: DWARF 0.0.0 should have a bound of some kind, possibly a julia version upper bound if it doesn't load at all, otherwise Pkg can downgrade to it in order to get the latest StructIO

Gallium/versions/0.0.3/requires
```diff
@@ -1,10 +1,10 @@
...
+DWARF 0.0.3 0.1.0
 ObjFileBase
-ELF
-MachO
+ELF 0.0.3 0.1.0
+MachO 0.0.3 0.0.4
 ASTInterpreter
 AbstractTrees
-COFF
+COFF 0.0.2 0.1.0
```
tkelman: there isn't a COFF 0.1 yet, but once there is, the fact that the earlier versions of Gallium did not depend on it means Pkg could downgrade to one of those Gallium versions in order to get COFF 0.1

Keno: At that point there'll be a new version of Gallium that works with COFF 0.1.0

DWARF/versions/0.0.1/requires
Keno: They probably don't work on 0.5, so I'll just bound them like that.

DWARF/versions/0.0.0/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.2-
+julia 0.2- julia 0.5
```
tkelman: `julia 0.2- 0.5`

Keno: Clearly I need to go get some food.

Gallium/versions/0.0.3/requires
tkelman: okay, as long as you remember to do that

this is unsatisfiable if COFF 0.0.2 requires DWARF >= 0.1.0 while Gallium 0.0.3 requires DWARF < 0.1.0

Keno: Ok, the Gallium version bound should be looser. Will fix.

tkelman: 0.0.3 <= ELF < 0.1.0 combined with 0.0.2 <= COFF < 0.1.0 is unsatisfiable
since COFF 0.0.2 requires DWARF >= 0.1.0 while ELF 0.0.3 requires DWARF < 0.1.0

Keno:
I'll submit a new PR for this

PR #8018, opened 2017-02-18
Tag ComputationalResources.jl v0.0.2

attobot:
Repository: [timholy/ComputationalResources.jl](https://github.com/timholy/ComputationalResources.jl)
Release: [v0.0.2](https://github.com/timholy/ComputationalResources.jl/releases/tag/v0.0.2)
Travis: [Travis Build Status](https://travis-ci.org/timholy/ComputationalResources.jl/builds/202918972)
Diff: [vs v0.0.1](https://github.com/timholy/ComputationalResources.jl/compare/054650dd4c9442686b51f577abe6b8512f724f07...b26c16e9af456cb1a379449963cb6987b56270b1)
`requires` vs v0.0.1: 
```diff
--- 0.0.1/requires
+++ 0.0.2/requires
@@ -1 +1,2 @@
 julia 0.4
+Compat 0.18
```
cc: @timholy


ComputationalResources/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat 0.18
```
tkelman: 0.17 for abstract type

timholy: Will keep that in mind for the future, but doesn't really hurt to demand more?

tkelman: yes it does, it would prevent upgrading this package unnecessarily if anyone has Compat pinned for reproducibility

timholy: Fix coming. But I think it's a little too picky; ImageFiltering is going to require 0.18 anyway and it's currently the only consumer. A lower bound shouldn't have to be the absolutely lowest possible lower bound; it should merely need to be sufficient to make the package work.

tkelman: known to work is too tight and will prevent version combinations that work fine from being installed together. bounds too loose allows installation of version combinations that don't work together. both are wrong, but with different consequences.

PR #8019, opened 2017-02-18
Fix link to manual in README

w1res:
The link to the Julia's manual section about packages doesn't work with the trailing slash. I tried in both Firefox and MS Edge and had the same issue. Removing the trailing slash fixed the problem in both.


tkelman:
Thanks!

PR #8035, opened 2017-02-20
Register and tag DataTables 0.0.1

nalimilan:


nalimilan:
I'm not sure what's the Juno error, but if that indicates a real problem we can merge https://github.com/JuliaData/DataTables.jl/pull/16 and re-tag.

PR #8040, opened 2017-02-20
Register new package SVR.jl v0.1.0

attobot:
Repository: [madsjulia/SVR.jl](https://github.com/madsjulia/SVR.jl)
Release: [v0.1.0](https://github.com/madsjulia/SVR.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/madsjulia/SVR.jl/builds/203488664)
cc: @montyvesselinov


montyvesselinov:
Hope this can be added soon; we have dependencies to this repo. Thank you very much!

tkelman:
Hi Monty. The package naming guidelines suggest avoiding acronyms, it would be clearer if spelled out. Did you try using any of the existing bindings to libsvm?

tkelman:
It's also better to use `dirname(@__FILE__)` rather than `Pkg.dir` to support installing the package elsewhere on LOAD_PATH.

montyvesselinov:
libsvm does not work for SVR:

https://github.com/simonster/LIBSVM.jl/issues/14

It is working only for classification problems. It is also outdated and not supported.

SupportVectorRegression is a mouthful for a module name. SVR is much better. SVR is a popular math term for this type of analyses.

montyvesselinov:
> It's also better to use dirname(@__FILE__) rather than Pkg.dir to support installing the package elsewhere on LOAD_PATH.

Are you saying that:

`libpath = joinpath(Pkg.dir("SVR"), "deps", "libsvm-3.22")`

should be

`libpath = joinpath(dirname(@__FILE__), "deps", "libsvm-3.22")`

My impression was that `Pkg.dir("SVR")` is the preferred way to go.

tkelman:
`joinpath(dirname(@__FILE__), "..", "deps", "libsvm-3.22")` is preferred. `Pkg.dir` is not correct when the package is loaded from elsewhere on `LOAD_PATH`, it's really only suitable for interactive use.

There is tab completion on module names. Having never heard of SVR (Google results show the Russian foreign intelligence service...) a more explicit name would be more obvious for potential users to know what the package is for. It's up to you, but longer and obvious is a better choice IMO than saving a few keystrokes and using a cryptic acronym that people outside the field won't recognize.

montyvesselinov:
I think we will prefer to go with SVR. We will use the github tag options to describe better what were are doing in SVR. In our field SVR is pretty widely used.

dirname(@__FILE__) version pushed as a new patch.

Thank you very much!

tkelman:
replaced by #8052

montyvesselinov:
Thank you very much!

-monty

PR #8044, opened 2017-02-21
Register FatGraphs [https://github.com/CarloLucibello/FatGraphs.jl.git]

CarloLucibello:
This a derivative of LightGraphs, the main differences being the support to custom graph types (inheriting from an abstract graph type, and defining a basic interface) and many under the hood improvements specified in the changelog.

Bye,
Carlo

### DIFF
--- a/FatGraphs/url
+++ b/FatGraphs/url
@@ -1,0 +1,1 @@
+https://github.com/CarloLucibello/FatGraphs.jl.git


dpsanders:
I don't think the name is appropriate.

CarloLucibello:
>I don't think the name is appropriate.

I agree. Graphs is taken. What about:
- FastGraphs
- Erdos (in honor of the matematician)
- Grafi (the italian for graphs, running out of ideas here)

I think Erdos is growing on me. Any advise against it? some "googlability" problems?

CarloLucibello:
Also, should I tag a version before registration?

tkelman:
> should I tag a version before registration?

At the same time.

So much of the code here is a straight unmodified fork of LightGraphs, would it make any sense to trim this package down to primarily consist of the pieces where it differs?

CarloLucibello:
> So much of the code here is a straight unmodified fork of LightGraphs, would it make any sense to trim this package down to primarily consist of the pieces where it differs?

not really: all methods' signatures of LightGraphs have been relaxed here to accept G<:AGraph, AGraph being an abstract graph type. This is one of the reasons that pushed me to fork LightGraphs a few months ago

tkelman:
replaced by #8085

PR #8062, opened 2017-02-22
Removed v0.1.1 of IntervalTrees

Ward9250:
Whilst version 0.1.1 of IntervalTrees appears to work fine, it introduces problems downstream in other packages and was therefore premature.


tkelman:
versions should not be deleted once published. either the packages where this is causing issues can add upper bounds, or can make a 0.1.2 tag to supersede this (can be the same as 0.1.0 if you want)

PR #8065, opened 2017-02-22
Fixed Bio require files

Ward9250:
Retroactively fixes Bio require files to avoid issues Bio.Intervals has when using IntervalTrees v0.1.1. @tkelman

### DIFF
--- a/Bio/versions/0.1.0/requires
+++ b/Bio/versions/0.1.0/requires
@@ -1,7 +1,7 @@
 julia 0.4-
 BufferedStreams
 Colors
-IntervalTrees
+IntervalTrees 0.0.1 0.1.1
 Iterators
 Libz
 Switch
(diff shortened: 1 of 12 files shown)


tkelman:
yep, perfect

though is the change that broke in IntervalTrees intended to be permanent? so will there be a future Bio release that is adjusted to work with it?

Ward9250:
I'm not completely sure right now, I initially thought my few simple additions to IntervalTrees - convenience outer constructors somehow broke things, but actually there are other changes to IntervalTrees which are more "invasive" - performance optimisations and the like which also could be the cause, and so I don't know yet if the fix will be a change to IntervalTrees or Bio.jl.

tkelman:
Okay. We'll merge this for now to fix the immediate issue, but if the fix ends up in IntervalTrees and 0.1.2 or 0.2.0 there makes this unnecessary, we can revert it later.

Ward9250:
That sounds good, thanks @tkelman this has been puzzling me all day!

PR #8085, opened 2017-02-23
Register Erdos [https://github.com/CarloLucibello/Erdos.jl.git]

CarloLucibello:
changed the name of the repo from FatGraphs #8044


CarloLucibello:
bumpity bump

ViralBShah:
After merging - I realize that at least calling it ErdosGraphs would have been better. If you like the idea, I'll be happy to revert.

CarloLucibello:
Erdos is ok, maybe not as explicative and searchable, but it sounds better and Paul Erdős is almost a synonym of graph theory. Thanks!

PR #8097, opened 2017-02-24
Tag PiecewiseLinearOpt v0.0.2

joehuchette:
https://github.com/joehuchette/PiecewiseLinearOpt.jl


tkelman:
just checking, is

> 2,479 changed files with 121,133 additions and 29 deletions.

intentional?

joehuchette:
Yep, I added a bunch of test instances. Any reason it might cause problems?

tkelman:
git doesn't handle large numbers of small files very well, if you don't expect to ever change the data it might be simpler to put them all in an hdf5 file

joehuchette:
OK, I'll look into doing that in the future.

mlubin:
The package doesn't load without JuMP 0.16 (https://github.com/joehuchette/PiecewiseLinearOpt.jl/issues/5), may want to update requires and wait for JuMP 0.16 to be released before merging.

PR #8124, opened 2017-02-25
Register CollisionDetection [https://github.com/krcools/CollisionDete…

krcools:
…ction.jl.git]: v0.0.1


CollisionDetection/versions/0.0.1/requires
```diff
@@ -0,0 +1 @@
+julia 0.4
```
tkelman: need to list `Compat 0.18` for Compat.Iterators

PR #8139, opened 2017-02-27
Register new package Convertible.jl v0.1.0

attobot:
Repository: [helgee/Convertible.jl](https://github.com/helgee/Convertible.jl)
Release: [v0.1.0](https://github.com/helgee/Convertible.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/helgee/Convertible.jl/builds/205749059)
cc: @helgee


tkelman:
This globally changes the behavior of `Base.convert` for all pairs of types depending on whether or not the package is imported, which is a bit fishy. What about making this opt-in instead of extending the global `Base.convert(::Type{<:Any}, ::Any)` - do something like the `@views` macro that locally rewrites or  replaces calls to `convert` with `Convertible.convert` ?

helgee:
I see your point. You now need to wrap calls to `convert` with `@convert` to make use of the new behaviour.

helgee:
Why is it you always find the errors after tagging a release? 👎 Sorry for the noise.

PR #8153, opened 2017-02-28
Tag RoME v0.0.4 [https://github.com/dehann/RoME.jl]

dehann:
[RoME.jl](https://github.com/dehann/RoME.jl)
v0.0.3 -> [v0.0.4](https://github.com/dehann/RoME.jl/releases/tag/v0.0.4)
[Build Status](https://travis-ci.org/dehann/RoME.jl)
diff: [vs v0.0.3](https://github.com/dehann/RoME.jl/compare/0f105f4...f687b6c)

REQUIRE
```
julia 0.5
IncrementalInference 0.2.5
Graphs 0.7.1
TransformUtils 0.0.5
KernelDensityEstimate 0.2.3
Distributions
Colors
Gadfly
JLD
HDF5
```

AWS seems to be down today, resulting in Travis issues, but the tests do pass. (For example, Travis instance is not able to download julia-0.5 stable from aws)


tkelman:
You're missing Optim as a test requirement.

Are you relying on a feature that was first released in Graphs 0.7.1?

dehann:
Ah, thanks. No, not explicitly relying on Graphs 0.7.1 here -- I can loosen that requirement here if you want? The related package IncrementalInference.jl does require 0.7.1 though, and will like move to 0.7.2 once I finish some of the new features in Graphs.jl.

tkelman:
ah, fair enough, the transitive dep here means you won't be able to use anything lower anyway

PR #8156, opened 2017-03-01
Update REQUIRE for Query

davidanthoff:
Not sure this is the recommended way to handle this, but NamedTuples v2.0.0 (just tagged in JuliaLang/METADATA.jl#8154) breaks Query, so my intention here is to set an upper limit.

I'll try to release a patch release of Query soon that will then work with NamedTuples v2.0.0, but that might take a couple of days and I don't want folks to be stuck with a non-working version until then.

### DIFF
--- a/Query/versions/0.0.1/requires
+++ b/Query/versions/0.0.1/requires
@@ -2,6 +2,6 @@ julia 0.5-
 NamedTuples 1.0.0
 FunctionWrappers 0.0.1
 DataStructures 0.4.5
-Requires 0.3.0
+Requires 0.3.0 2.0.0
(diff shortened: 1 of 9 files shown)


tkelman:
why are you upper-limiting Requires?

davidanthoff:
Because Query doesn't work with NamedTuples 2.0.0.

tkelman:
then you should upper limit NamedTuples, not Requires

davidanthoff:
Argh, I shouldn't try to fix these things after a long day. Thanks for catching this, I'll prep a new PR shortly.

PR #8173, opened 2017-03-01
Named tuples

mdcfrancis:
Includes both v3.0.0 and v3.0.1 - only change in v3.0.1 are the travis definitions


NamedTuples/versions/3.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5.0-
```
tkelman: do you still need to support prereleases of 0.5? if not, drop the trailing dash

mdcfrancis: I have dropped, this will require a re-tag to 3.0.2, correct ?

tkelman: or you can edit this file in place in your metadata fork. it's not critical, can wait until next tag. up to you.

mdcfrancis:
Updated version here and also in the NamedTuples package. Will be included in the next tag automatically.

mdcfrancis:
thanks

PR #8178, opened 2017-03-02
MP3.jl 0.0.1 does not work with SampledSignals 0.3

tkelman:
ref http://pkg.julialang.org/detail/MP3.html, http://pkg.julialang.org/detail/SampledSignals.html

cc @ssfrr @jongwook


ssfrr:
This is weird, I think it should work... I'll look into what's going on and push/tag versions that work. I need to do some 0.6 fixes anyways.

PR #8185, opened 2017-03-02
Fix missing dependency of SingularIntegralEquations on Plots

tkelman:
cc @dlfivefifty @MikaelSlevinsky since ApproxFun 0.5 came out this has been failing due to an undeclared dependency http://pkg.julialang.org/detail/SingularIntegralEquations.html


dlfivefifty:
Thanks, I'll tag a new version next week for 0.5

PR #8191, opened 2017-03-02
Tag SugarBLAS v0.0.3 [https://github.com/lopezm94/SugarBLAS.jl]

lopezm94:


SugarBLAS/versions/0.0.3/requires
```diff
@@ -0,0 +1 @@
+julia 0.4
```
tkelman: if you still support 0.4, you should still test there
https://github.com/lopezm94/SugarBLAS.jl/commit/fa61677bc7f0f7d2b135062a9da6ece65e21192e

lopezm94:
Included 0.4 version testing and specified the exact version of Julia when the parser changed for `=>`.

PR #8193, opened 2017-03-02
Tag FaSTLMM v0.0.2 [https://github.com/sens/FaSTLMM.jl]

sens:
Now requires Julia version 0.5 (instead of 0.4).  Fixed a major problem related to use of the optimizer.


sens:
There are still some problems.  Still working.

PR #8206, opened 2017-03-04
Register new package CMakeWrapper.jl v0.0.1

attobot:
Repository: [rdeits/CMakeWrapper.jl](https://github.com/rdeits/CMakeWrapper.jl)
Release: [v0.0.1](https://github.com/rdeits/CMakeWrapper.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/rdeits/CMakeWrapper.jl/builds/207607297)
cc: @rdeits


rdeits:
This package provides a download of cmake 3.7.2 on windows, mac, and linux, and two BinDeps-compatible providers: CMakeProcess (the high-level analogue to BinDeps.Autotools) and CMakeBuild (analogous to BinDeps.AutotoolsDependency). It's already been pretty useful for me in https://github.com/rdeits/LCMCore.jl/pull/6 https://github.com/rdeits/DrakeVisualizer.jl/pull/43 and https://github.com/rdeits/BotCoreLCMTypes.jl/pull/1

I'm not entirely sold on the package name--I'm open to suggestions.

ararslan:
Out of curiosity, why not just include the functionality as part of BinDeps? Or does this have functionality outside the scope of BinDeps?

CMakeWrapper/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+BinDeps 0.4.7
```
tkelman: What's in 0.4.7 that you rely on? BinDeps hasn't introduced any new features for a long time, this is almost certainly too conservative.

rdeits: Nothing as far as I know. I'll test against 0.4.0 and see if I can loosen that bound.

rdeits:
@ararslan bringing this into BinDeps would also be totally reasonable. I didn't choose to do that for a couple of reasons: 

1. This is all pretty new and likely to change as I find issues. Having it in a registered package means that my other packages can start using it right away, even if I need to change the API later on. 
1. A lot of the complexity of this package comes from the fact that it downloads (and possibly compiles) its own copy of cmake. Most BinDeps users won't need that, so it's nice to have a separate tool to handle that download. 

Actually, I was originally *only* going to include the cmake download in this package. The BinDeps integration was an afterthought that turned out to be easier than I'd expected.

CMakeWrapper/versions/0.0.1/requires
rdeits: Ok, 0.4.0 works, so I'll change the requirement to 0.4

rdeits:
Ok, @tkelman I think I've addressed all of your comments. 

1. I've changed the windows install to copy instead of symlink
1. I've added an architecture check on linux and a fallback source build for any architecture which is not x64. 
1. I've added tests of that fallback build on Travis (I can't actually test on a non-x64 architecture there, but at least I can verify that the source build is set up correctly).

PR #8215, opened 2017-03-05
Tag LightGraphs.jl v0.7.3

attobot:
Repository: [JuliaGraphs/LightGraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl)
Release: [v0.7.3](https://github.com/JuliaGraphs/LightGraphs.jl/releases/tag/v0.7.3)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGraphs/LightGraphs.jl/builds/207950759)
Diff: [vs v0.7.2](https://github.com/JuliaGraphs/LightGraphs.jl/compare/0bbc5f63ebf06a89dc4ab25e4b96e601167c12b9...c1fdcc0159499b7be2aa7aaa8bba7f90ab3fa70e)
`requires` vs v0.7.2: 
```diff
--- 0.7.2/requires
+++ 0.7.3/requires
@@ -1,7 +1,8 @@
 julia 0.5-
 GZip 0.2.20
-LightXML 0.4.0
+EzXML 0.3.0
 ParserCombinator 1.7.11
 JLD 0.6.3
 Distributions 0.10.2
 StatsBase 0.9.0
+DataStructures 0.5.0
```
cc: @sbromberger


LightGraphs/versions/0.7.3/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.5-
```
tkelman: do you still need to support prereleases?

sbromberger: Good question. Does this imply that 0.5 release isn't supported? I confess that the min/max versioning in REQUIRES has always confused me.

tkelman: no, this implied any prerelease of 0.5, or newer, is supported. if you drop the trailing dash it means any release of 0.5, or newer, is supported

sbromberger: Ah, well, we do support the 0.5 prereleases. We stopped supporting any 0.4 code in 0.6.3. So I guess this is "most correct" but might not be desirable from a METADATA perspective?

tkelman: 0.5 prereleases includes the very first commit after 0.4 branched though.

sbromberger: (Any objections to keeping this stable for now at `0.5-`? We're getting ready to do some big refactoring and that will probably coincide with the end of 0.5 support in general.)

sbromberger: Ah, I see what you mean. OK. Let me see if I can make the change via GitHub Releases. (It's my first time through a release process outside of `Pkg`.)

LightGraphs/versions/0.7.3/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.5 0.6-
```
tkelman: i think we've had this conversation before, all this upper bound will do, unless it's applied to all past versions, will keep people back to the previous version on 0.6

sbromberger: Argh. Yes. You're right - we have had this conversation before. Sorry. I though I was being clever.

sbromberger:
As always, thank you, @tkelman, for the guidance.

PR #8224, opened 2017-03-06
Register Defer [git://github.com/abrewster/Defer.jl.git]: v0.0.1

adambrewster:


tkelman:
If you enabled Travis CI, would your tests pass? Are you intending for this to support Julia 0.4?

adambrewster:
I think so, but I don't know much about Travis.  I'll try to enable it.

I thought it would work on 0.4, but it apparently doesn't.  I'll try to fix it or change the requirement to 0.5.

adambrewster:
0.4 is fixed.  I do still use it for a couple of things, and I don't know of any reason to not support it for something like this.

Like I said, I don't know much about Travis, but from https://travis-ci.org/adambrewster/Defer.jl/jobs/208295335 it seems to be working.

tkelman:
Should probably elaborate on the package description since `A proposal for #7721` without any context isn't all that clear about what the package does.

PR #8238, opened 2017-03-07
Register new package Rx.jl: v0.1.0

yuehhua:
https://github.com/yuehhua/Rx.jl


Rx/url
```diff
@@ -0,0 +1 @@
+https://github.com/a504082002/Rx.jl.git
```
ararslan: Did you recently change your username? This redirects to https://github.com/yuehhua/Rx.jl. To be on the safe side, this should be updated to reflect the current URL.

ararslan:
Looks like you don't have continuous integration (i.e. Travis or AppVeyor) set up for your repository. It's always a good idea to run tests in that way. Do you know whether the tests would pass if you set it up?

dpsanders:
The name `Rx.jl` is not very clear; please review the guidelines for naming registered Julia packages in the manual. Would it be possible to come up with a more self-explanatory name?

tkelman:
It's discouraged to extend Base methods on signatures that are made up entirely of Base types, like your `map(f::Function)` and `filter(f::Function)`, as that can change the behavior of unrelated code

ararslan:
For reference, the manual section to which @dpsanders is referring is http://docs.julialang.org/en/latest/manual/packages#Guidelines-for-naming-a-package-1.

Rx/url
yuehhua: Yeah, I have change my username, but I should have update to the current URL.

yuehhua:
@ararslan @dpsanders Thanks for suggestions. I'm new to collaborate a opensource project.
I wish to establish a project like [ReactiveX](https://github.com/ReactiveX) and everyone just call it "Rx".
So, ReactiveExtensions.jl could be a good one?

About CI, I haven't establish it, could you tell me how to do it or where can I find some resources?
I think I would use Travis.

@tkelman OK, I think that won't be a good solution, and I will try to come up with other signatures.

Thank you all for help.

yuehhua:
Move to [ReactiveExtensions.jl](https://github.com/yuehhua/ReactiveExtensions.jl)

PR #8249, opened 2017-03-08
Adding PolynomialMatrices

neveritt:
`PolynomialMatrices` aims at supporting univariate polynomial matrix calculations.
The package extends `Polynomials`, and tries to provide a set of basic mathematical
functionality between `Number`s, `Poly`s and `PolyMatrix` objects.


neveritt:
Is it ok to register and tag at the same time?

blegat:
Yes it is ok :)

tkelman:
Would be a good idea to fill in the repo description, as that'll show up on pkg.julialang.org.

PR #8253, opened 2017-03-08
Tag MATLAB v0.5.1 [https://github.com/JuliaInterop/MATLAB.jl]

ChrisRackauckas:


musm:
is the ci failure here expected

ChrisRackauckas:
Only because I was a dope and forgot to push the tags.

musm:
you are dope :P

PR #8271, opened 2017-03-10
Register new package xkpasswd.jl v0.0.1

attobot:
Repository: [freeboson/xkpasswd.jl](https://github.com/freeboson/xkpasswd.jl)
Release: [v0.0.1](https://github.com/freeboson/xkpasswd.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/freeboson/xkpasswd.jl/builds/209610905)
cc: @freeboson


tkelman:
Note the typical Julia style / naming conventions (http://docs.julialang.org/en/release-0.5/manual/style-guide/#use-naming-conventions-consistent-with-julia-s-base, http://docs.julialang.org/en/release-0.5/manual/packages/#guidelines-for-naming-a-package) recommend giving modules, and hence packages, capitalized names. We only have a handful of exceptions to that out of the current 1300ish registered packages.

tkelman:
Command-line git also isn't a dependency of Julia, so using submodules for data isn't always going to work.

freeboson:
I can rename it 'XKPasswd' but since it's based on 'xkcd' it looks terrible that way.

I can make `Git.jl` a dependency to clean that up.

freeboson:
@tkelman I've renamed the repo and module to `XKPasswd(.jl)`. I did change `deps/build.jl` to use `Git.jl`, but I'm not sure what the best way to fetch and extract the data would be without using it, as probably I would need command line tools like `wget`/`curl`/`zip`/etc.

tkelman:
Okay. Should update the url, sha, and requires here.

freeboson:
Hmm I thought @attobot would close this when I deleted the old release, but it did create #8334 for me

PR #8318, opened 2017-03-14
Register RequiredKeywords.jl

adamslc:
This package provides the `@required_keywords` macro which can be used to make a keyword argument in a function mandatory. It is useful for self-documenting function calls in non-performance critical code because the meaning of each value is labeled.

https://github.com/adamslc/RequiredKeywords.jl


adamslc:
Done

PR #8322, opened 2017-03-14
Pull request/30163934

Alexander-Barth:
PhysOcean is a module with some utility function for physical oceanography and includes a library of empirical functions to compute a.o. the density of seawater and air-sea fluxes.


Alexander-Barth:
Sorry, I forget the require file. I will make a new pull request.

PR #8336, opened 2017-03-15
Move Interpolations.jl repository to JuliaMath

Evizero:
ref: https://github.com/JuliaMath/Interpolations.jl/issues/140#issuecomment-286672127

cc: @timholy


sglyon:
Looks good.

Do we know if CI is activated at the new home?

Evizero:
Looks like it: https://travis-ci.org/JuliaMath/Interpolations.jl

PR #8343, opened 2017-03-15
Upper bounds on Parameters.jl in DiffEq

ChrisRackauckas:
Avoiding 0.7.0 due to escaping issues. See https://discourse.julialang.org/t/nested-macros-bypass-package-dependencies/2683/4

### DIFF
--- a/AlgebraicDiffEq/versions/0.0.1/requires
+++ b/AlgebraicDiffEq/versions/0.0.1/requires
@@ -1,4 +1,4 @@
 julia 0.5
-Parameters 0.5.0
+Parameters 0.5.0 0.7.0
 Sundials 0.3.0
 DiffEqBase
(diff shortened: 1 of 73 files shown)


tkelman:
Since early versions of DiffEqBase did not depend on Parameters, this could result in downgrading to very old DiffEqBase to get the latest Parameters. Could add an artificial bounded dependency to avoid this.

ChrisRackauckas:
You mean, remove the latest Parameters.jl tag? I think that's a good idea too. Let me know if you want me to do that.

tkelman:
No, I mean supercede it with a tag that works.

tkelman:
should be made unnecessary by https://github.com/JuliaLang/METADATA.jl/pull/8348

PR #8344, opened 2017-03-15
Upper bounds on Parameters.jl

ChrisRackauckas:
Due to escaping issues in Parameters.jl, these packages fail to precompile after `Pkg.update()`. See https://discourse.julialang.org/t/nested-macros-bypass-package-dependencies/2683/4

@rdeits
@cstjean

### DIFF
--- a/CMakeWrapper/versions/0.0.1/requires
+++ b/CMakeWrapper/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.5
 BinDeps 0.4
-Parameters 0.5
+Parameters 0.5 0.7.0
--- a/CMakeWrapper/versions/0.0.2/requires
+++ b/CMakeWrapper/versions/0.0.2/requires
@@ -1,3 +1,3 @@
 julia 0.5
 BinDeps 0.4
-Parameters 0.5
+Parameters 0.5 0.7.0
--- a/ScikitLearn/versions/0.0.1/requires
+++ b/ScikitLearn/versions/0.0.1/requires
@@ -1,5 +1,5 @@
 julia 0.4
 PyCall 1.4.0
 SymDict 0.0.2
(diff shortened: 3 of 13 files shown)


rdeits:
:+1: thank you!

cstjean:
That means that we'll have to update in sync when Parameters.jl is fixed. Otherwise if I update ScikitLearn to `Parameters 0.8.0`, it'll cause a Pkg.update conflict with `CMakeWrapper`'s `Parameters 0.5.0 0.7.0`.

I don't know if that's ever done, but could we instead undo Parameters.jl's latest tag, until this is fixed?

ChrisRackauckas:
>That means that we'll have to update in sync when Parameters.jl is fixed. Otherwise if I update ScikitLearn to Parameters 0.8.0, it'll cause a Pkg.update conflict with CMakeWrapper's Parameters 0.5.0 0.7.0.

It will resolve itself. That just means that when you update to use Parameters 0.8.0 (which will be necessary for Julia v0.6), users will not get the latest version of ScikitLearn.jl until CMakeWrapper.jl also updates. There would be no breakage, just a delay in the release. Also note that DiffEq has this same upper bound.

Yes, another way to solve this is to just undo the Parameters.jl tag. Since it's just for v0.6-compatibility, that would probably be fine. @mauro3 @tkelman ?

mauro3:
I should be able to finish Chris' fix in Parameters.jl tomorrow and tag a new version.  So if things can wait until then, then that would be easiest. Undoing the last tag of Parameters.jl would be fine by me too.

rdeits:
I'm happy to wait.

cstjean:
Likewise.

tkelman:
should be made unnecessary by https://github.com/JuliaLang/METADATA.jl/pull/8348

PR #8346, opened 2017-03-15
Tag DiffBase v0.0.5 [https://github.com/JuliaDiff/DiffBase.jl]

jrevels:


tkelman:
This broke ForwardDiff as you probably know by now, but since this tag still supports Julia 0.4 it also broke things there. http://pkg.julialang.org/detail/ForwardDiff.html

jrevels:
Yeah, the ForwardDiff "breakage" (really, inducing a test failure) was intentional, because it adds a test for a known bug that was getting fixed. It should cause ReverseDiff tests to fail as well, until a new version gets tagged containing https://github.com/JuliaDiff/ReverseDiff.jl.

PR #8355, opened 2017-03-16
Fix NLsolve URL

pkofod:


tkelman:
LineSearches as well

pkofod:
fixed, thanks

NLsolve/url
```diff
@@ -1 +1 @@
-git://github.com/EconForge/NLsolve.jl.git
+git://github.com/JuliaNLSolvers/NLsolve.git
```
tkelman: needs to keep the .jl

PR #8357, opened 2017-03-16
Tag ForwardDiff v0.4.1 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:


tkelman:
Can you also tag a version that will pass on Julia 0.4?

PR #8377, opened 2017-03-17
Tag Nemo version 0.6.0

thofma:


tkelman:
You should be using specific sha or tag versions of all the repositories that you're cloning in `deps/build.jl`, otherwise external changes can break previously-working versions of your package.

thofma:
We are maintaining the upstream repositories and we make sure not to break anything.

tkelman:
That helps, but doesn't change the fact that what you install with Nemo version X shouldn't depend on when you install it.

thofma:
We have fixed upstream versions.

thofma:
Is there something else to address?

StefanKarpinski:
Can we reopen this and get it merged? @tkelman: please do what's needed to move this forward.

thofma:
Waiting for one more PR in Nemo. Then I will tag it.

PR #8378, opened 2017-03-17
Tag TimeZones v0.3.0 [https://github.com/JuliaTime/TimeZones.jl]

omus:
Now supports Julia 0.6 via the nightly


tkelman:
If the content of the windows xml file ever changes due to a download, it'll make the package dirty and prevent it from upgrading

omus:
Interesting point. I guess the solution is to avoid overriding the committed file.

omus:
Fixed the issue with the windows xml file making the package dirty. Updated the SHA1 and pushed an updated tagged version.

tkelman:
Can you fix the issue where this downloads that file repeatedly and never cleans up after itself?

omus:
I'll make a patch release for that.

PR #8382, opened 2017-03-18
Tag TextWrap v0.2.0 and FastaIO v0.3.0; cap Julia versions in older FastaIO releases

carlobaldassi:
https://github.com/carlobaldassi/TextWrap.jl
https://github.com/carlobaldassi/FastaIO.jl

In both cases: drop support for Julia < 0.5, removes deprecations on 0.6, switch docs from readthedocs to Documenter and docstrings.

FastaIO was actually broken on 0.6 so I also capped the older versions requirements.


tkelman:
how broken? would be better to set the upper bound to a release, unless it doesn't even load at the very first prerelease you've listed here

carlobaldassi:
Broken as in unusable.
Setting the cap at a release wouldn't work, since then it would get installed also on, say, 0.6-alpha. It's been broken since months, so I don't think there is much point in finding out exactly which 0.6-dev release  caused it.

tkelman:
since there's a newer version available, the upper bound isn't strictly necessary. I'd rather put it at something known to be bad (the alpha would be fine) than something that's "probably bad"

carlobaldassi:
But why? I don't understand the point.

tkelman:
Unless it's broken at the very first commit after a release branched, then anything with a trailing dash is too tight of a bound

tkelman:
It's better to allow installation of a possibly-broken package version, especially if there are newer replacement versions available that should work fine for the same Julia version, than to disallow installation of a working package version. The latter would result in resolution failures if you tried to install that specific package version (to reproduce an old result, for example).

carlobaldassi:
Thanks for the explanation. I'm not convinced, but it doesn't really matter and it's 3AM here, so I just set the caps at the releases.

tkelman:
Should probably deprecate `readall(::FastaReader)`, looks like BiomolecularStructures is using it under that name

carlobaldassi:
I tried to do that, but when I export it there's a conflict with GZip also exporting it, which means that its use must be qualified code will be broken anyway. If I import if from GZip first, then code will break again when GZip will remove it (which will be soon, it's already gone from master).

carlobaldassi:
Also, the current code of BiomolecularStructures will break anyway on 0.6 and will need updating, e.g. it uses ASCIIStrings.

PR #8385, opened 2017-03-18
Tag WAV v0.8.4 [https://github.com/dancasimiro/WAV.jl]

dancasimiro:
This addresses two minor deprecation warnings from julia v0.6:

- remove usage of `typealias`
- change style of array creation


WAV/versions/0.8.4/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+Compat 0.18.0
```
tkelman: none of your type aliases are parametric, so you actually don't need at-compat for them and can leave this min version what it was at the last tag

dancasimiro: I'll make those changes and tag a different revision.

PR #8391, opened 2017-03-18
Tag FlatBuffers and upper bound Feather

quinnj:


Feather/versions/0.2.0/requires
```diff
@@ -1,5 +1,5 @@
 julia 0.5
-FlatBuffers 0.1.1
+FlatBuffers 0.1.1 0.2.0
```
tkelman: unless these bounds are added to all past versions, Pkg may choose to downgrade Feather to a 0.1.x release in order to upgrade FlatBuffers.

quinnj: but not for Julia 0.5, right? All other past Feather versions were 0.4

tkelman: they have a min julia version but not a max julia version right? so the 1.x versions are allowed to be installed on 0.5

but yes if no FlatBuffers versions support 0.4 then we can't do the "fake dependency" trick

quinnj:
Alright, updated all past Feather versions to have the upper bound on FlatBuffers 0.2.0

Feather/versions/0.1.2/requires
```diff
@@ -1,5 +1,5 @@
 julia 0.4
-FlatBuffers
+FlatBuffers 0.1.1 0.2.0
```
tkelman: did these first versions need a lower bound?

quinnj: yes

tkelman: Feather was registered Jun 20th, but the 0.1.1 tag of FlatBuffers was not until August, how does the former depend on the latter if it didn't exist yet?

quinnj: Updated

PR #8080, opened 2017-02-23
Tag Convex.jl v0.5.0

attobot:
Repository: [JuliaOpt/Convex.jl](https://github.com/JuliaOpt/Convex.jl)
Release: [v0.5.0](https://github.com/JuliaOpt/Convex.jl/releases/tag/v0.5.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaOpt/Convex.jl/builds/204447766)
Diff: [vs v0.4.0](https://github.com/JuliaOpt/Convex.jl/compare/31960219657a17d2b4d826282c00935a9f2f8e97...e0c1267d4a8e7c19dc2373b806a7eaa28605fa88)
`requires` vs v0.4.0: 
```diff
--- 0.4.0/requires
+++ 0.5.0/requires
@@ -1,3 +1,3 @@
 julia 0.4
-MathProgBase 0.5 0.6
+MathProgBase 0.5 0.7
 DataStructures
```
cc: @madeleineudell

PR #8400, opened 2017-03-19
Tag Convex v0.5.0 [https://github.com/JuliaOpt/Convex.jl]

Ayush-iitkgp:
This is an improvement over the PR  #8080 
Please close the previous PR and merge this one.


Convex/versions/0.5.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
```
tkelman: if you don't need to support prereleases, drop the trailing dash

Ayush-iitkgp: Not sure. We would want to keep support for prereleases. We are planning to completely drop support only after stable julia v0.6 release.

tkelman: prereleases of julia 0.5, which was released in september?

ViralBShah: Yes, the - should be dropped, since you don't want to support 0.5 pre-releases.

Ayush-iitkgp:
All suggestions incorporated in the latest new PR by attobot. So sorry for the mess!

PR #8402, opened 2017-03-19
Tag GLPK v0.4.1 [https://github.com/JuliaOpt/GLPK.jl]

carlobaldassi:
Fix deprecation warnings on julia 0.6.

### DIFF
--- a/GLPK/versions/0.4.1/requires
+++ b/GLPK/versions/0.4.1/requires
@@ -1,0 +1,4 @@
+julia 0.5
+BinDeps
+@osx Homebrew
+Compat 0.18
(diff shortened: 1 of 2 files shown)


GLPK/versions/0.4.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps
+@osx Homebrew
+Compat 0.18
```
tkelman: abstract type only needs 0.17

carlobaldassi: I had looked at the typealias change, which was done in 0.18, but I guess I don't need that for non-parametric types. I'll fix it and resubmit later.

PR #8406, opened 2017-03-19
new package: PEG.jl

wdebeaum:
I wrote a new Julia package for writing Parsing Expression Grammars.


wdebeaum:
Oops, the tag didn't get pushed. Had to "git push --tags" in PEG.

tkelman:
Sorry for the delay here. If you enabled Travis CI, would your tests pass there?

It's up to you, but it's recommended to avoid acronyms in package names, for the sake of discoverability by anyone unfamiliar with the term. Ref http://docs.julialang.org/en/stable/manual/packages/#guidelines-for-naming-a-package

wdebeaum:
I just tried TravisCI, it passed (after adding a blank line to the end of README.md to get it to run).

I did read the naming guidelines, but the rule about acronyms didn't say anything about discoverability, only avoiding the possibility of confusion. I thought that "PEG" was closer to "USA" than to "PMA" in that respect, since software libraries named "PEG" are usually about Parsing Expression Grammars and not, say, Percutaneous Endoscopic Gastrostomy.

tkelman:
It's more about ubiquity of the acronym, is it a common enough term that anyone who sees PEG would know what that means and does even if they aren't familiar with the term parsing expression grammars?

wdebeaum:
OK, well, it would be nice if the documentation was more explicit about all this. To me, non-confusion, discoverability, and ubiquity are different things, and it's not clear that one implies the others. But maybe I'm just being pedantic because my day job involves making fine linguistic distinctions like that :)

Anyway, would it be better if I renamed PEG to ParsingExpressions? The guidelines say plurals are for packages that mostly define a type. While PEG technically doesn't do that, conceptually you could say it defines a subtype of Function that could appropriately be called ParsingExpression.

OTOH, I could follow guideline 5 and rename it something else entirely. But that seems an odd response to your concern about discoverability.

"There are only two hard things in Computer Science: cache invalidation and naming things." --Phil Karlton

ararslan:
Package naming can indeed be tough, as it's equal parts marketing, self-expression, and "a name I won't forget." 😉 To those familiar with formal grammars, the acronym PEG is most likely obvious. If I may offer my opinion here, I actually think the name ParsingExpressions would be somewhat misleading, as it suggests parsing Julia `Expr`s. To that effect, I think PEG is actually better than ParsingExpressions. I have no further suggestions though.

wdebeaum:
That's a good point, it would be misleading. So let's just stick with PEG.

PR #8411, opened 2017-03-20
Added package: BiologicalSymbols

Ward9250:
This new package from BioJulia marks the first step in a concerted effort to make the BioJulia codebase in Bio.jl more modular. In time Bio.jl will be a meta-package that re-exports functionality in more specific BioJulia packages.


BiologicalSymbols/versions/1.0.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

bicycle1885:
I think we need more discussion about the design of new modular packages. Personally, I think this package is tightly connected to biological sequences and should shipped be in a package.

Ward9250:
Discussion over this has been going on over the weekend on Gitter, so do jump in there. My impression from it is we should make things as modular as possible, but not too modular - much as @ChrisRackauckas does with juliadiffeq. In this specific case of this proposed package, I can think of cases in which scripts or packages would want to use nucleotides or amino acids, without needing to install and entire Sequence package. Working with just mutations for example.

BiologicalSymbols/versions/1.0.0/requires
Ward9250: Thanks, I've added Compat 0.20.0.

tkelman: you aren't using Compat.readline are you? abstract type was 0.17. be sure to also do this on master.

Ward9250: @tkelman Thanks, I've set it to 0.17.0. This package does not use the readline function or define any readline methods.

Ward9250:
A rename to BioSymbols has been requested.

Ward9250:
Ok rename should be complete.

bicycle1885:
Please suspend this for a while. This change is very important and we need some time to reach a consensus on the design (discussions here: https://github.com/BioJulia/Bio.jl/issues/425).

BioSymbols/versions/1.0.0/sha1
```diff
@@ -0,0 +1 @@
+87a1cd45b3e777baa5bc4f0597d53040c559647c
```
tkelman: at this commit the module in https://github.com/BioJulia/BioSymbols.jl/blob/87a1cd45b3e777baa5bc4f0597d53040c559647c/src/BioSymbols.jl still says `BiologicalSymbols`

Ward9250: Fixed!

tkelman:
LGTM then, @bicycle1885 do you still have concerns about this?

Ward9250:
In our discussion you will see BioSymbols has been included as a package in our dependency graph, and I think in that thread by and large people have agreed on the new organisation and the naming now, what do you say @bicycle1885?

Ward9250:
Several days have passed without any proposals to change the modular design we've settled on over in BioJulia/Bio.jl#425, so I think this is ready for release now. Since this is a very basic package it will allow other parts of Bio.jl to be modularised subsequently.

PR #8412, opened 2017-03-20
Tag LineSearches.jl v1.0.0

attobot:
Repository: [JuliaNLSolvers/LineSearches.jl](https://github.com/JuliaNLSolvers/LineSearches.jl)
Release: [v1.0.0](https://github.com/JuliaNLSolvers/LineSearches.jl/releases/tag/v1.0.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/LineSearches.jl/builds/212843857)
Diff: [vs v0.1.5](https://github.com/JuliaNLSolvers/LineSearches.jl/compare/007640e240e4e8a2a2d38636d2c6f22540c69b3c...624820d99090648a37ee232c33fa2876c1973139)
`requires` vs v0.1.5: 
```diff
--- 0.1.5/requires
+++ 1.0.0/requires
@@ -1 +1,2 @@
-julia 0.4
+julia 0.5
+NLSolversBase
```
cc: @anriseth


anriseth:
Many of the tests have been temporarily disabled, as they require https://github.com/JuliaNLSolvers/Optim.jl/pull/388, which again depends on this tag.

tkelman:
Will this be breaking for the existing versions of Optim or NLSolve?

anriseth:
Yes. We no longer return the number of function and gradient calls, so all
calls to the linesearches methods break.

tkelman:
Okay, then to avoid breaking those other packages when users do `Pkg.update()`, we should apply upper bounds to their past versions' dependencies on LineSearches so people won't get the breaking changes until there's a new version of Optim and/or NLSolve ready that can work with it.

anriseth:
Ah, sure. Will adding an upper cap `v1.0.0-` to REQUIRE solve it? https://github.com/JuliaNLSolvers/Optim.jl/pull/389 (If I understand correctly, this will put a strict upper bound to VERSION < v1.0.0?

tkelman:
Partially. It needs to be done in the METADATA requires files of the past tags as well.

anriseth:
Thank you!

PR #8415, opened 2017-03-20
Add upper cap on LineSearches

anriseth:
Due to breaking changes in https://github.com/JuliaNLSolvers/LineSearches.jl/pull/37 
Ref #8412


tkelman:
So far so good. Though there's another complicated issue here. Since the early versions of NLSolve and Optim did not depend on LineSearches, Pkg may choose to downgrade to an old version of either or both of those packages in order to get a newer LineSearches, depending on what the resolver decides to do.

There are a few things we could do about that. Either we could add "fake" LineSearches dependencies to those earlier versions with an upper bound, or we could apply Julia upper bounds to them to avoid making those old versions installable. We should only do the latter on package + julia version combinations that we know don't work at all.

anriseth:
I'm in favour of option 1, and just add fake dependencies. @pkofod?

anriseth:
@tkelman I have added fake LineSearches dependencies with upper bounds to all NLsolve and Optim versions that require julia 0.4 (none require 0.5). Is that sufficient, or would you like me to do it for all versions (that will create trouble for anyone who for some reason were using julia 0.3 or older)

pkofod:
I agree with opt 1

tkelman:
We should probably do a combination of both, and upper bound the julia version on the 0.3-era tags. Let me take a closer look and see what versions do and do not load at all on 0.5 when I've got less of a headache.

NLsolve/versions/0.9.0/requires
```diff
@@ -4,5 +4,5 @@ Calculus
 Distances
 ForwardDiff 0.3.0
 Compat 0.8.4
-LineSearches
+LineSearches 0.1.2 1.0.0-
```
tkelman: was this relying on 0.1.2 without saying so?

pkofod:
> We should probably do a combination of both, and upper bound the julia version on the 0.3-era tags. Let me take a closer look and see what versions do and do not load at all on 0.5 when I've got less of a headache.

Once again, thank you for helping out. Let me ask, could we have done this transition in a different (read:easier) way? Or is this just how it goes?

tkelman:
Would it be possible to more smoothly deprecate the old API, or change the downstream packages to be able to work against either new or old? Currently you can't really have both old and new versions of the same package loaded in Julia at the same time. Could give the new API a new package name which would be messy but less complicated from a version resolution standpoint.

In Pkg3 we probably need some way independent of bounds to adjust what the resolver will do - mark old versions as unmaintained perhaps, so the resolver will only pick them if it finds no solution from just maintained versions of all packages.

NLsolve/versions/0.9.0/requires
pkofod: Hm, I don't think 0.1.1 would fail, but 0.1.2 does include a bug fix that came up in NLsolve.

anriseth: For reference: https://github.com/JuliaNLSolvers/LineSearches.jl/pull/22

tkelman:
Okay I ran through and out of these pre-0.4 versions of Optim and NLsolve, only Optim 0.4.3 and 0.4.4 can be imported in Julia 0.5 without an error. So those are pretty safe upper bounds.

tkelman:
Oh, annoying, SloanDigitalSkySurvey depended on Optim 0.4.4 specifically.

Okay, I have a slightly odd request of you. Could you do, in a checkout of Optim,
```
git checkout v0.4.4
git tag v0.4.6
git checkout v0.4.5
git tag v0.4.7
git push --tags
```
See the commit message of my last commit, I'm making a duplicate tag of 0.4.4 as 0.4.6 but with a minimum Julia version of 0.4, and a duplicate tag of 0.4.5 as 0.4.7 so any package that had an Optim 0.5 upper bound will still be using the same code.

anriseth:
> Okay, I have a slightly odd request of you. Could you do, in a checkout of Optim,

Can you do this please, @pkofod ? I don't think i have write access to Optim

pkofod:
should be done, now @anriseth  just needs to open/close

tkelman:
Cool, that worked. Everyone okay with this version?

pkofod:
> Cool, that worked. Everyone okay with this version?

lgtm

anriseth:
Yes, all good

PR #8427, opened 2017-03-20
Tag Gadfly v0.5.4 [https://github.com/GiovineItalia/Gadfly.jl]

tlnagy:
* Dramatically speed up precompilation by removing old, duplicate code
  * Add `Geom.abline`
  * Add `Geom.density2d`


ararslan:
Nice, Gadfly is getting better all the time.

Looks like this should be a minor release rather than a patch release though, as support for Julia 0.4 has been dropped. That's why the Travis check is failing. For reference, the exact message is:
> ERROR: LoadError: New tag 0.5.4 of package Gadfly requires julia 0.5.0, but version 0.5.0 of Gadfly requires julia 0.4.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.6.0 using `PkgDev.tag("Gadfly", :minor)`.

tlnagy:
Thanks for pointing this out. Will tag and try again.

PR #8436, opened 2017-03-21
Tag ForwardDiff v0.3.5 [https://github.com/JuliaDiff/ForwardDiff.jl]

jrevels:


ForwardDiff/versions/0.3.5/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4 0.6
```
tkelman: why the upper bound here?

PR #8431, opened 2017-03-21
Tag RigidBodyDynamics.jl v0.1.0

attobot:
Repository: [tkoolen/RigidBodyDynamics.jl](https://github.com/tkoolen/RigidBodyDynamics.jl)
Release: [v0.1.0](https://github.com/tkoolen/RigidBodyDynamics.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/tkoolen/RigidBodyDynamics.jl/builds/213259232)
Diff: [vs v0.0.6](https://github.com/tkoolen/RigidBodyDynamics.jl/compare/b782a82dffa4c0d89ca1f8697dbcc409dd478b2f...dfbf7aac31875c69ea8def33400baaa387822bbb)
`requires` vs v0.0.6: 
```diff
--- 0.0.6/requires
+++ 0.1.0/requires
@@ -4,3 +4,4 @@
 DataStructures 0.4.6
 LightXML 0.4.0
 DocStringExtensions 0.3.1
+Compat 0.18.0
```
cc: @tkoolen

PR #8444, opened 2017-03-22
add upper bounds on RigidBodyTreeInspector's dependency on RigidBodyDynamics

tkelman:
ref #8431, breaks due to 'TreeDataStructure not defined'

cc @rdeits @tkoolen


rdeits:
👍  thank you!

PR #8448, opened 2017-03-22
Tag RoME v0.0.5 [https://github.com/dehann/RoME.jl]

dehann:
UPDATED, and UPDATED again to use ```getfield``` over ```eval```.

[RoME.jl](https://github.com/dehann/RoME.jl)
v0.0.4 -> [v0.0.5](https://github.com/dehann/RoME.jl/releases/tag/v0.0.5)
[Build Status](https://travis-ci.org/dehann/RoME.jl)
diff: [vs v0.0.4](https://github.com/dehann/RoME.jl/compare/6bb6b52...f91331f)

REQUIRE
```
julia 0.5
IncrementalInference 0.2.7
Graphs 0.7.1
TransformUtils 0.0.5
KernelDensityEstimate 0.2.3
Distributions
Colors
Gadfly
JLD
HDF5
Optim
ProgressMeter
```

[IncrementalInference.jl](https://github.com/dehann/IncrementalInference.jl)
v0.2.6 -> [v0.2.7](https://github.com/dehann/IncrementalInference.jl/releases/tag/v0.2.7)
[Build Status](https://travis-ci.org/dehann/IncrementalInference.jl)
diff: [vs v0.2.6](https://github.com/dehann/IncrementalInference.jl/compare/cb47af420...559a915)

REQUIRE
```
julia 0.5
Graphs 0.7.1
GraphViz
Gadfly
Colors
NLsolve
Optim
Distributions
KernelDensityEstimate 0.2.3
HDF5
JLD
ProgressMeter
```


tkelman:
Is there a way to avoid overwriting the methods from IncrementalInference here?

dehann:
Hi, yes I hope so. I was not easily able to expose the methods from a child package though. Maybe you could point me in the right direction here please? What is happening now is ```RoME``` requires ```IIF``` as a general ````REQUIRE```. But the ```IIF.savejld()``` requires access to types and convert functions defined in ```RoME``` when called from further down in the dependency chain.

I have similar methods exposed in the other direction, by using ```import IncrementalInference: somefnc``` in ```RoME```. The issue in this case I think I would need to create a dependency cycle by doing ```import RoME: somefnc, sometype``` in ```IIF```. Which we'd obviously like to avoid.

The abstraction I'm going for is an inference/optimization library ```IIF```, gets extended with more advanced manifold constraint functions etc. elsewhere, such as in ```RoME```.

tkelman:
define the methods for abstract types in the dependency package, subtype those abstract types for the extended methods in the dependent package?

dehann:
I'm sure you can, but I'm having some difficulty getting the abstract right -- sure hindsight of the layers will make perfect sense. I think the problem [originates here in ```IIF```](https://github.com/dehann/IncrementalInference.jl/blob/v0.2.6/src/FGOSUtils.jl#L22)

dehann:
The real issue is I'm storing the type information as a string in a Database use-case of the code, and now trying to reuse those same packing and unpacking functions to circumvent a limitation in ```JLD.jl```. DBs dont store Julia types, but we can do ```T = eval(parse(strval))```.

```JLD.jl``` does not seem to like functor types:
```julia
type MyType <: Function
...
end

@JLD.save MyType(...) # fails
```

dehann:
@tkelman , sorry, could you just restart the check on this PR please. I forgot to push updated tag ```v0.0.5``` to ```dehann/RoME.jl```.

I have to move ```IncrementalInference``` forward to fix this issue. Please see basic example of how fix the above code replication issue [here](https://github.com/dehann/IncrementalInference.jl/blob/v0.2.7/test/saveconvertertypes.jl)

dehann:
Great, thanks again @tkelman !

PR #8463, opened 2017-03-23
Register new package IEEEMinMax

ararslan:
This package provides the pre-0.6, IEEE 754-2008 compliant behavior of `min` and `max` as `minnum` and `maxnum` (named as per IEEE). Pretty lightweight, useful for Graphics and Winston.


ararslan:
@tkelman If you think this should be part of NaNMath that's fine--just close this and I'll open a PR there instead.

ararslan:
Should also ask @mlubin what he thinks of having it in NaNMath.

tkelman:
ref https://github.com/mlubin/NaNMath.jl/issues/2

ararslan:
Ah yeah.

PR #8469, opened 2017-03-23
Tag Distributions.jl v0.12.2

attobot:
Repository: [JuliaStats/Distributions.jl](https://github.com/JuliaStats/Distributions.jl)
Release: [v0.12.2](https://github.com/JuliaStats/Distributions.jl/releases/tag/v0.12.2)
Travis: [Travis Build Status](https://travis-ci.org/JuliaStats/Distributions.jl/builds/214461698)
Diff: [vs v0.12.1](https://github.com/JuliaStats/Distributions.jl/compare/f4535a6f7d275a718c159452c86fa76aa9ac773b...2d0cc12f44672a3a646ae9258019a3e0fde20626)
`requires` vs v0.12.1: no changes
cc: @ararslan


tkelman:
This may have caused a test failure in KernelDensityEstimate http://pkg.julialang.org/logs/KernelDensityEstimate_0.5.log

ararslan:
I'm inclined to think that the KernelDensityEstimate failures are unrelated to Distributions.

* This appears to be the extent to which KernelDensityEstimate uses Distributions: https://github.com/dehann/KernelDensityEstimate.jl/commit/bf7e7bd7308dde115c9f7e03ddb8e6480e533cfe

* This is the extent of the changes made in this version of Distributions: https://github.com/JuliaStats/Distributions.jl/compare/f4535a6f7d275a718c159452c86fa76aa9ac773b...2d0cc12f44672a3a646ae9258019a3e0fde20626

Looking through the KernelDensityEstimate code, I don't see how any of the changes here could have caused those failures.

tkelman:
it's either Distributions, Compat, or an intermittent failure. here's what changed in the pkgeval logs: https://gist.github.com/anonymous/881084ec64fbc8696d96c3e237379698

tkelman:
http://pkg.julialang.org/detail/KernelDensityEstimate.html has been consistent for over 6 months on release julia so intermittent failure seems unlikely?

ararslan:
I just installed KernelDensityEstimate and ran the failing tests locally and they all passed. I have the same versions for the installed packages as listed in the log. So I have no idea what could be causing the failure.

tkelman:
Ping @dehann in case you have any ideas. Seems to be intermittent, but much less likely to happen on Distributions 0.12.1 than on 0.12.2.

dehann:
Hi thanks, I'm investigating...

dehann:
I'm reluctant to blame ```Distributions v0.12.2``` for this test failure. I don't see the failures when using ```v0.12.2``` my side either. However, I am careful of other recent changes. I'm looking closely at a [change here in ```Compat v0.21.0```](https://github.com/JuliaLang/Compat.jl/compare/2d0f59007ca318c76d1c9007774cfc13d7134ebc...86bd62b86a40bdf243e544a1933518c8c9aea02e#diff-f34b50ad70663b5a85fd643930f44ec4R1161), but unlikely since my tests pass fine with the new ```Compat v0.21.0``` also.

My first concern is that there may be a dispatch problem somewhere -- only suspicion at this point. I do indeed [overload some functions such as ```Base.rand```](https://github.com/dehann/KernelDensityEstimate.jl/blob/c9ede5d2d0b66a7a6661bc69b70b2ab2443b440f/src/KernelDensityEstimate.jl#L8). Aside, the ```KernelDensityEstimate``` code is not that complicated. The [trickiest part is a circular reference](https://github.com/dehann/KernelDensityEstimate.jl/blob/c9ede5d2d0b66a7a6661bc69b70b2ab2443b440f/src/BallTreeDensity01.jl#L216) which I'd like to remove, but hasn't been a show stopper thus far.

I should add, there was a brief period when the ```rand functions```---it think it was---where producing weird output on OSx systems. This was during the a move to new LLVM. The tests on OSx cleared soon enough and I left it there.

Background, ```KernelDensityEstimate.testProds``` does quality assurance tests on an internal Gibbs MCMC procedure. By nature this is stochastic, and therefore quality tests are on the basis of confidence: The result should at least, 50% of the time, be in the ball park of the correct result, a fairly weak test. So expected output is something like:
```julia
[testProds(...) for i = 1:10] = Bool[true,true,true,true,false,true,true,true,false,false]
[testProds(...) for i = 1:10] = Bool[true,true,true,true,true,true,true,true,true,true]
[testProds(...) for i = 1:10] = Bool[true,true,true,true,true,false,true,true,true,false]
```
However this event in the log you found is very, very unlikely (but not impossible).
```julia
v = [testProds(D=D,M=M,N=N,n=n,dev=dev,MCMC=MCMC) for i = 1:10] = Bool[false,false,false,false,false,false,true,false,true,true]
```
I'd ask if anyone sees this continue please let me know. I'm watching out for it also. Regardless, I will loosen the tolerance of the quality assurance tests in the next tag of ```KernelDensityEstimate```. For reference, I'd expect these test to fail very very seldom, but I don't want to make the tolerances so loose that other problems slip through.

tkelman:
Distributions definitely isn't at fault, just had one on 0.4: https://github.com/JuliaCI/pkg.julialang.org/blob/8fda89864aa41bb53bee49cadc2011e6f67a625e/logs/KernelDensityEstimate_0.4.log

PR #8475, opened 2017-03-24
Tag ApproxFun v0.6.0 [https://github.com/JuliaApproximation/ApproxFun…

dlfivefifty:
….jl]

Supports Julia 0.6


tkelman:
Will SingularIntegralEquations 0.2.2 need an upper bound?

dlfivefifty:
Yes, I was going to add it when I tagged a new version

tkelman:
better to do now, otherwise Pkg.update() will break SingularIntegralEquations for any users

dlfivefifty:
https://github.com/JuliaLang/METADATA.jl/pull/8495

PR #8476, opened 2017-03-24
register and tag IndexedTables.jl

shashi:
https://github.com/JuliaComputing/IndexedTables.jl


tkelman:
`map` and `foreach` on NamedTuple should be defined in that package, and on Tuple should not be globally modified here.

shashi:
Closing in favor of #8634

PR #8496, opened 2017-03-25
New package Factors.jl

hamzaelsaawy:


timholy:
Is the name "Factors" a bit too generic for something specifically targeted at Bayesian and Markov networks? I was expecting it to be about polynomials or matrices.

quinnj:
I thought it was another `CategoricalArrays`/`PooledDataArray` type of implementation (since they're called factors in R)

hamzaelsaawy:
Would renaming to `PGMFactors` be more appropriate?

tkelman:
That would be an improvement, though the PGM acronym isn't necessarily common knowledge.

MikeInnes:
@hamzaelsaawy the choice of name is yours; if you want to stick with this, let me know and I can merge.

PR #8501, opened 2017-03-26
Tag Wavelets v0.6.0 [https://github.com/JuliaDSP/Wavelets.jl]

gummif:


tkelman:
Renaming files in case only is asking for trouble on case-sensitive file systems. Anyone who has the package already installed on Windows or Mac would likely have it stop working after this upgrade.

gummif:
Any suggestions? Rename the files to a temp name and back?

tkelman:
I guess it would actually be okay to have the case not match, but it's making me think if there's something it would break. Or maybe code paths in libgit2 that don't get exercised often.

Could move them into a subfolder of src for a little while if that's any less ugly than a temporary file name that differs in something more than case?

tkelman:
Let me know what you want to do, if anything. Maybe we just merge this and keep our fingers crossed that nothing breaks.

gummif:
OK, I think you can just merge this now and I will fix the file names somehow in the next revision (soon).

PR #8529, opened 2017-03-28
Tag AWSCore et all

samoconnor:


InfoZIP/versions/0.0.10/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

samoconnor: Fair enough...

But there is no change to `requires` vs 0.0.9: https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/InfoZIP/versions/0.0.9/requires

Is this a METADATA.jl-update blocking issue?

tkelman:
Compat is missing from SymDict's REQUIRE, and it should not be modifying Base methods on Base types.

samoconnor:
I've added Compat to SymDict's REQUIRE: https://github.com/samoconnor/SymDict.jl/blob/master/REQUIRE  See #8533

Re: modifying Base methods on Base types...
Is this a general comment on SymDict? or has this inadvertently changed since v0.0.5 ?

samoconnor:
See #8533

tkelman:
General comment, but was not noticed before. It's bad practice now and was bad practice before.

samoconnor:
As far as I can tell, none of the `Base.merge` or `Base.get` methods in SymDict.jl are modifying base methods insofar as commenting out any of them causes the tests to throw a `MethodError`. i.e. they all seem to be new methods. Am I missing something?

tkelman:
They're adding methods that are not present in Base, with signatures that consist only of Base types. That means unrelated code that may have been expecting an error, or using a different code path, can have its semantics subtly changed by importing your package because of the global modification of method tables you're doing.

samoconnor:
Ok, so adding methods is bad practice in addition to modifying methods.
I've added a new issue to SymDict: https://github.com/samoconnor/SymDict.jl/issues/2

PR #8533, opened 2017-03-28
Tag AWSCore et all

samoconnor:
Update REQUIRES julia 0.5 per https://github.com/samoconnor/XMLDict.jl/pull/4#issuecomment-289667167

(obsoletes prior PR #8529)


samoconnor:
Travis says:

New tag 0.0.12 of package XMLDict requires julia 0.5.0, but version 0.0.3 of XMLDict requires julia 0.4.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.1.0 using `PkgDev.tag("XMLDict", :minor)`.

Stand by...

PR #8534, opened 2017-03-28
Tag AWSCore et all

samoconnor:
Drop v0.4 support.


samoconnor:
Note that https://github.com/samoconnor/SymDict.jl/issues/2 was found in a review of a previous version of this PR, but it identifies a problem that has always been in SymDict and is not a recent regression.
Request that this PR be accepted to allow users to have access to the bug fixes in AWS*.

tkelman:
would be better to separate into independent pr's, if the tags aren't interdependent

AWSLambda/versions/0.1.0/requires
```diff
@@ -0,0 +1,14 @@
+julia 0.5
+AWSCore
```
tkelman: needs minimum version for AWSConfig to be defined

samoconnor: This is all [major version zero](http://semver.org/#spec-item-4) stuff.
At this point I don't think the AWS* users will gain much from a fine-grained AWS* version interdependency map, so I don't intend to have version tags in `REQUIRES` . Users know to do `Pkg.checkout()` to get the latest versions before reporting issues. Most of my support issue discussions are like this: "it's broken", "I think I fixed that, try `Pkg.checkout(...)`", "that works, thx, can you tag it?".

tkelman: Without minimum version dependencies, it takes considerably more effort to pin or constrain to an previous version and have things work correctly. With version dependencies, Pkg will do the right thing and downgrade other packages if needed to satisfy user-imposed constraints.

samoconnor: I understand the benefits of version dependencies, but I don't intend to provide them at this point.

AWSS3/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+AWSCore
```
tkelman: missing dependencies on HTTPCommon and Requests, if they're imported directly they should be listed as direct dependencies

AWSSDB/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+AWSCore
+SymDict
```
tkelman: missing dependency on Compat

AWSSQS/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+AWSCore
```
tkelman: also missing Compat (which you can probably just drop the `using` of)

AWSS3/versions/0.1.0/requires
samoconnor: https://github.com/samoconnor/AWSS3.jl/commit/b73bc9c83b3084517b2d78ba8ef223da13eea60c

AWSSDB/versions/0.1.0/requires
samoconnor: https://github.com/samoconnor/AWSSDB.jl/commit/cacca4fe9c7d022c087b55aef27f0e0b84a43a91

AWSSQS/versions/0.1.0/requires
samoconnor: https://github.com/samoconnor/AWSSQS.jl/commit/0a5c0301f8ebdb992e58d5b868c30b31a016c36c

samoconnor:
@tkelman 
Do you intend to reject this PR based on the `REQUIRES` issues above (fixes committed to master but not yet tagged) ?

tkelman:
No, this is mostly okay, the missing lower bounds can be added retroactively and most other things can be fixed by re-tagging. Look into attobot if PkgDev is too much trouble.

samoconnor:
stand by, I'm attempting some by-hand-gitting to update this PR to updated tags

samoconnor:
Notes on manual gitting for future reference:

`git rebase -i origin/metadata-v2` -- delete intermediate tags

`git remote add samgithub https://github.com/samoconnor/METADATA.jl.git` -- tell git about PR fork

`git push --force samgithub metadata-v2:pull-request/375d2b14` -- push to PR fork

tkelman:
and `git push --tags` in each of the packages

PR #8544, opened 2017-03-28
Register new package GPUArrays.jl v0.0.1

attobot:
Repository: [JuliaGPU/GPUArrays.jl](https://github.com/JuliaGPU/GPUArrays.jl)
Release: [v0.0.1](https://github.com/JuliaGPU/GPUArrays.jl/releases/tag/v0.0.1)
cc: @SimonDanisch


SimonDanisch:
Needs: Transpiler, Matcha & Sugar to be tagged first!

tkelman:
The conditional importing of dependencies means this shouldn't be enabling precompilation, it's not going to work reliably.

SimonDanisch:
I tried to make sure that it's static after `Pkg.build`, so it shouldn't be a problem! Or am I missing something?

tkelman:
Pkg.build won't re-trigger if the set of available dependencies changes

SimonDanisch:
Thats fine by me and i can add that to the readme as a gotcha!

tkelman:
Right but if you try and load from a precompiled file and the dependencies aren't there, it's not going to load. Conditional dependencies cause more problems than they solve at the moment.

SimonDanisch:
They solve the very real issue, that the backends only work on certain hardware and are hard to install sometimes, which shouldn't disable the whole package.
All conditionally loaded packages are in the REQUIRE (actually not CUDAnative, since it's 0.6 only), so  you shouldn't be able to delete a dependency by accident.

On the other hand, I greatly care about compile times, especially since I have some people interested in this package coming from Python. I don't want their first impression to be: `Oh wow, this takes a couple of seconds EVERY TIME I load it`... And likewise I don't want to wait those seconds myself.

I trade a `might happen for someone unlucky and can easily be clarified in the install instructions` with an everyday improvement any day!

tkelman:
> All conditionally loaded packages are in the REQUIRE

Ah okay, though CUDAdrv, GLAbstraction, GLWindow, GLFW, ModernGL, ColorTypes, and DataStructures are not

PR #8548, opened 2017-03-28
Refactor METADATA tests to work outside of Travis CI

omus:
I adjusted the testing code so that the code works with other CIs.

### DIFF
--- a/.test/ci.sh
+++ b/.test/ci.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+set -e
+cd $(dirname $0)/..
+git checkout -b localbranch
+cd ..
+ln -s $PWD/METADATA.jl METADATA
+for ver in 0.4 0.5 0.6; do
+  mkdir -p ~/.julia/v$ver/.cache julia-$ver
+  ln -s $PWD/METADATA.jl ~/.julia/v$ver/METADATA
+  if [ $ver = 0.6 ]; then
+    url="julianightlies/bin/linux/x64/julia-latest-linux64"
+  else
+    url="julialang/bin/linux/x64/$ver/julia-$ver-latest-linux-x86_64"
+  fi
+  curl -A "Travis-CI for METADATA tests $(curl --version | head -n 1)" \
+    -L --retry 5 https://s3.amazonaws.com/$url.tar.gz | \
+    tar -C julia-$ver --strip-components=1 -xzf - && \
+    julia-$ver/bin/julia -e 'versioninfo(); include("METADATA/.test/METADATA.jl")' && \
+    touch success-$ver &
+done
+wait
+if ! [ -e success-0.4 -a -e success-0.5 ]; then # add success-0.6 once there are any 0.6-only package versions
+  exit 1
+fi
--- a/.test/ci.sh
+++ b/.test/ci.sh
@@ -1,21 +1,29 @@
 #!/bin/sh
 set -e
 cd $(dirname $0)/..
+BUILD_DIR=$PWD
+[ -z "$CI_TMP_DIR" ] && CI_TMP_DIR=$BUILD_DIR.tmp
+[ -z "$JULIA_PKGDIR" ] && JULIA_PKGDIR=$HOME/.julia
+export BUILD_DIR JULIA_PKGDIR
+
+# Need to be on a local or remote branch for check_metadata
+git checkout -q --detach HEAD && git branch -D localbranch 2>/dev/null || true
 git checkout -b localbranch
-cd ..
-ln -s $PWD/METADATA.jl METADATA
+
+mkdir -p $CI_TMP_DIR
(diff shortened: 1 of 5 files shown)


omus:
Messed up how I set `env` in the `.travis.yml` file.

tkelman:
Would be good to verify that this still flags the same set of problems correctly, and won't silently succeed or otherwise ignore things due to the changed set of env vars and relative directories.

omus:
I'll try to get to this tomorrow. I guess the way to test this to make some PRs based upon this version with various failure scenarios?

tkelman:
That sounds like the most effective way to verify it's still working as intended, yes.

omus:
Made a bunch of PR tests which are working so far (waiting on #8772). Let me know of any other tests I should run.

.test/ci.sh
```diff
@@ -0,0 +1,32 @@
...
+[ -z "$CI_TMP_DIR" ] && CI_TMP_DIR=$BUILD_DIR.tmp
+[ -z "$JULIA_PKGDIR" ] && JULIA_PKGDIR=$HOME/.julia
+export BUILD_DIR JULIA_PKGDIR
+
+# Need to be on a local or remote branch for check_metadata
+git checkout -q --detach HEAD && git branch -D localbranch 2>/dev/null || true
+git checkout -b localbranch
+
+mkdir -p $CI_TMP_DIR
+cd $CI_TMP_DIR
```
tkelman: why is this running under a separate .tmp ?

tkelman:
git tag sha mismatch and unpushed git tag should be tested

.test/ci.sh
omus: If we download versions of Julia into the `PWD` when the METADATA tests run they think folders such as "julia-0.4" are invalid packages.

```julia
ERROR: LoadError: Package julia-0.4 has no tagged versions
 in macro expansion; at .../METADATA.jl/.test/METADATA.jl:44
```
The `CI_TMP_DIR` allows users to control where the julia binaries are downloaded to instead of just defaulting to `$PWD/..`

tkelman: it shouldn't be downloading inside metadata

omus: > it shouldn't be downloading inside metadata

Agreed. The original code just downloaded it in the parent directory. The new code makes a sibling directory to download to.

> why is this running under a separate .tmp ?

Do you want an alternative default for the tmp directory?

omus:
Git tag/SHA mismatch has been successfully tested.

.test/ci.sh
tkelman: that would be less of a change, and would make the cd in the .jl file unnecessary in the default case, right?

omus: > would make the cd in the .jl file unnecessary in the default case, right?

That is correct. I'll change the default to be `CI_TMP_DIR=$PWD/..`

omus:
Rebased all METADATA test PRs. Results are the same.

.test/ci.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+set -e
+cd $(dirname $0)/..
+BUILD_DIR=$PWD
+[ -z "$CI_TMP_DIR" ] && CI_TMP_DIR=$PWD/..
+[ -z "$JULIA_PKGDIR" ] && JULIA_PKGDIR=$HOME/.julia
+export BUILD_DIR JULIA_PKGDIR
+
+# Need to be on a local or remote branch for check_metadata
+git checkout -q --detach HEAD && git branch -D localbranch 2>/dev/null || true
```
tkelman: why's the `checkout --detach HEAD` necessary?

omus: If the work tree HEAD is currently a checked out branch then `git checkout -q HEAD` won't detach from the branch and is a essentially a no-op.

```bash
$ git checkout -b foobar
Switched to a new branch 'foobar'

$ git checkout -q HEAD

$ git branch -D foobar
error: Cannot delete branch 'foobar' checked out at '.../repo'

$ git checkout -q --detach HEAD

$ git branch -D foobar
Deleted branch foobar (was f2a4d6c).
```

omus: In this specific case I'm doing to ensure that `localbranch` is deleted in the extremely unlikely case that the branch is currently checked out.

tkelman: guess if the CI were to run in a persistent environment like Jenkins or buildbot that would be more likely

omus: The GitLab CI I use these changes under is persistent.

omus:
@tkelman should I close all the METADATA test PRs?

tkelman:
Hah, I was going to say yeah may as well close them. I found the apparently super efficient way of doing it, deleting the branch they were targeting.

omus:
That's a nice solution.

PR #8554, opened 2017-03-28
Register and tag ParticleFilters.jl

zsunberg:
https://github.com/JuliaPOMDP/ParticleFilters.jl


tkelman:
Looks good. I'm curious why the `alias_sample!` function had to be copied from StatsBase.

PR #8570, opened 2017-03-29
Tag WriteVTK v0.5.1 [https://github.com/jipolanco/WriteVTK.jl]

jipolanco:
Fixes Julia 0.6 deprecation warnings.


WriteVTK/versions/0.5.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.4
+LightXML 0.4.0
+Libz 0.2.1
+BufferedStreams 0.2.0
+Compat 0.19.0
```
tkelman: 0.17 for abstract type

jipolanco: Right... I'll adjust the requirement and then update this PR

PR #8576, opened 2017-03-29
Tag MLLabelUtils.jl v0.1.0

attobot:
Repository: [JuliaML/MLLabelUtils.jl](https://github.com/JuliaML/MLLabelUtils.jl)
Release: [v0.1.0](https://github.com/JuliaML/MLLabelUtils.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaML/MLLabelUtils.jl/builds/216358036)
Diff: [vs v0.0.1](https://github.com/JuliaML/MLLabelUtils.jl/compare/272c7ec61bd6a2ebb40e263f32786e91f82d3a00...90ce402b0126fe7c0ceeed186ee0ad1f103fcdb7)
`requires` vs v0.0.1: 
```diff
--- 0.0.1/requires
+++ 0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
-StatsBase
-LearnBase
-MappedArrays
+julia 0.6-
+StatsBase 0.13
+LearnBase 0.1.3 0.2.0
+MappedArrays 0.0.6
```
cc: @Evizero

PR #8579, opened 2017-03-29
Upperbound old MLLabelUtils version

Evizero:
ref: https://github.com/JuliaLang/METADATA.jl/pull/8576#discussion_r108697686


MLLabelUtils/versions/0.0.1/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.5
 StatsBase
-LearnBase
+LearnBase 0.1.0 0.2.0
```
tkelman: 0.0.1 and 0.0.2 would not have worked?

Evizero: no. I double checked just now. The min version should be `0.1.1`, since there was functionality introduced in LearnBase to make MLLabelUtils work

PR #8580, opened 2017-03-29
Register new package PhantomJS.jl v0.0.1

attobot:
Repository: [fredo-dedup/PhantomJS.jl](https://github.com/fredo-dedup/PhantomJS.jl)
Release: [v0.0.1](https://github.com/fredo-dedup/PhantomJS.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/fredo-dedup/PhantomJS.jl/builds/216388106)
cc: @fredo-dedup


PhantomJS/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+BinDeps
+Compat 0.8
```
tkelman: looks like you might be able to get away without Compat entirely?

fredo-dedup: You're right. I removed it. BinDeps requires Compat but I don't need to mention it here.

tkelman: also note https://github.com/fredo-dedup/PhantomJS.jl/pull/1, packages shouldn't use `Pkg.dir` to refer to their own locations

tkelman: should also remove `using Compat` from `deps/build.jl`, it's not guaranteed that BinDeps will always depend on Compat

PR #8590, opened 2017-03-30
Tag Temporal v0.2.5 [https://github.com/dysonance/Temporal.jl]

dysonance:
Added some function documentation to I/O methods and fixed some bugs I had introduced trying to prepare for Julia 0.6 that were causing the package to break for Julia 0.5. Still need to revisit how to make that transition smoothly with some of the big changes in syntax.

### DIFF
--- a/Temporal/versions/0.2.5/requires
+++ b/Temporal/versions/0.2.5/requires
@@ -1,0 +1,2 @@
+julia 0.5
+Requests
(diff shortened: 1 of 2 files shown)


tkelman:
JSON should be listed in REQUIRE if used directly

should not be defining `getindex(s::String, b::BitArray{1})` or `getindex(s::String, b::Vector{Bool})` here, those are Base types

dysonance:
Ahh, apologies, thanks for reminding me, I forgot to drop the JSON dependency in the REQUIRE.

And sorry about the getindex redefinition, I realize now how that could cause instabilities, will re-write the code to avoid having to do that.

I will close this PR for now and re-open after making the changes and retagging.

PR #8597, opened 2017-03-30
Tag FixedPointNumbers.jl v0.3.6

attobot:
Repository: [JuliaMath/FixedPointNumbers.jl](https://github.com/JuliaMath/FixedPointNumbers.jl)
Release: [v0.3.6](https://github.com/JuliaMath/FixedPointNumbers.jl/releases/tag/v0.3.6)
Travis: [Travis Build Status](https://travis-ci.org/JuliaMath/FixedPointNumbers.jl/builds/216668367)
Diff: [vs v0.3.5](https://github.com/JuliaMath/FixedPointNumbers.jl/compare/f3935899fa1507580bb4564d1f0f27eecc7961c2...3d86ace04a987ac84b90a03e18d936303e4fc88f)
`requires` vs v0.3.5: no changes
cc: @timholy


timholy:
I just noticed https://github.com/JuliaLang/julia/pull/21223; this change isn't strictly necessary anymore, though neither does it hurt anything. I can revert if desired.

tkelman:
Doesn't look like this would hurt anything? Maybe also add some mock test to base that could have caught such an ambiguity to prevent it from accidentally being added in the future?

timholy:
Yeah, it definitely shouldn't hurt anything, so feel free to merge.

Adding a meaningful test is hard, because there are an infinite number of such tests that could be added. `f(a::TypeA, b)` defined in Base might be ambiguous with some package `f(a, b::TypeB)` method. I think the best we can do right now is try to be consistent about which arguments we specialize. http://docs.julialang.org/en/latest/manual/methods.html#man-method-design-ambiguities-1

PR #8528, opened 2017-03-28
Tag Polynomials.jl v0.1.4

attobot:
Repository: [JuliaMath/Polynomials.jl](https://github.com/JuliaMath/Polynomials.jl)
Release: [v0.1.4](https://github.com/JuliaMath/Polynomials.jl/releases/tag/v0.1.4)
Travis: [Travis Build Status](https://travis-ci.org/JuliaMath/Polynomials.jl/builds/215782783)
Diff: [vs v0.1.3](https://github.com/JuliaMath/Polynomials.jl/compare/85e2bccd234ad345a9146325117555a4849da875...101d574deaf26ad30f2c713b3def9c4c186f04c5)
`requires` vs v0.1.3: no changes
cc: @ararslan

PR #8605, opened 2017-03-30
Upper bound Polynomials to 0.1.4 for ControlCore, PolynomialFactors, …

tkelman:
…PolynomialMatrices, RationalFunctions, and Roots

There was a change in `eltype`, ref https://github.com/JuliaLang/METADATA.jl/pull/8528#issuecomment-290297570.

cc @jverzani @aytekinar @neveritt 

Having ControlCore, PolynomialMatrices, RationalFunctions, or Roots installed will hold Polynomials back to 0.1.3 until you tag a new version that works with Polynomials 0.1.4 (this was already done for PolynomialFactors, so the upper bound is only applied to the older versions of that package).

### DIFF
--- a/ControlCore/versions/0.0.1/requires
+++ b/ControlCore/versions/0.0.1/requires
@@ -1,2 +1,2 @@
 julia 0.4
-Polynomials
+Polynomials 0.0- 0.1.4
--- a/PolynomialFactors/versions/0.0.1/requires
+++ b/PolynomialFactors/versions/0.0.1/requires
@@ -3,3 +3,4 @@ Compat 0.7.15
 Iterators 0.1.0
 Combinatorics 0.2.1
 Primes 0.1.1
+Polynomials 0.0- 0.1.4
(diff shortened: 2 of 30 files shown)


tkelman:
It should be noted that the earliest version of RationalFunctions seemed to work okay with Polynomials 0.1.4, but I'm assuming it would be preferable to hold Polynomials back a little rather than downgrade to a much older tag of RationalFunctions.

The early tags of Roots.jl are kind of hard to test these days, even on Julia 0.4 it looks like ForwardDiff 0.1.8 (which is the version used for some of the older Roots tags due to upper bounds) doesn't load at all.

aytekinar:
@tkelman, thanks for your help! in the next version, should we require `Polynomials 0.1.5`, then?

tkelman:
If https://github.com/JuliaMath/Polynomials.jl/pull/117 goes through and reverts the change, then this wouldn't be needed.

neveritt:
Thanks @tkelman for using your superpowers.

cc @aytekinar

jverzani:
We merged JuliaMath/Polynomials.jl#117 to roll back the defn of `eltype` for polynomials. Hopefully this will stop the breakage in `PolynomialMatrices`.

PR #8616, opened 2017-03-31
Tag SingularIntegralEquations v0.2.3 [https://github.com/JuliaApproxi…

dlfivefifty:
…mation/SingularIntegralEquations.jl]

Update for ApproxFun 0.6

### DIFF
--- a/SingularIntegralEquations/versions/0.2.3/requires
+++ b/SingularIntegralEquations/versions/0.2.3/requires
@@ -1,0 +1,6 @@
+julia 0.5
+ApproxFun 0.6
+BandedMatrices 0.2.2
+Compat 0.19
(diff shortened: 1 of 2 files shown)


SingularIntegralEquations/versions/0.2.3/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.5
+ApproxFun 0.6
+BandedMatrices 0.2.2
+Compat 0.19
```
tkelman: doesn't look like you're using `IndexStyle`, so should be 0.18

dlfivefifty:
Replace with https://github.com/JuliaLang/METADATA.jl/pull/8632

PR #8621, opened 2017-03-31
Register new package TextParse.jl v0.1.0

attobot:
Repository: [JuliaComputing/TextParse.jl](https://github.com/JuliaComputing/TextParse.jl)
Release: [v0.1.0](https://github.com/JuliaComputing/TextParse.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaComputing/TextParse.jl/builds/217081100)
cc: @shashi


tkelman:
Isn't `src/lib/date-parse.jl` unnecessary on 0.6? And on 0.5 does it change behavior?

shashi:
date-parse.jl is not loaded on `v0.6`

On `v0.5` it doesn't change the default behavior. `TextParse.DateFormat` and `Base.Dates.DateFormat` are different types, I define `DateTime(str, ::TextParse.DateFormat)` but that still won't affect existing use of the constructor.

tkelman:
In that case this looks okay. It would be good to describe briefly in the readme how this differs from CSV.jl, when people might want to use one vs the other. If it's strictly for CSV files then TextParse is arguably a bit overly broad of a name, but maybe you plan on adding non-CSV features later?

shashi:
@tonykelman I have updated the README accordingly. This package can be used as a framework to write parsers for other formats. I plan to add non-CSV features when I need it (eventually I will).

shashi:
Thanks!

PR #8627, opened 2017-03-31
Tag Celeste v0.3.0 [https://github.com/jeff-regier/Celeste.jl]

jeff-regier:


Celeste/versions/0.3.0/requires
```diff
@@ -0,0 +1,13 @@
+julia 0.6-
+Compat 0.18.0
+DataFrames
+DataStructures 0.5.2
+Distributions
+FITSIO 0.8.4
+DiffBase 0.0.3
+ForwardDiff 0.3.4 0.5
```
tkelman: should this upper bound apply to past tags too? Pkg could choose to downgrade Celeste in order to upgrade ForwardDoff, unless past versions also have upper bounds

jeff-regier:
Hi Tony, I think the past tags are ok without it. There is no 0.5 of ForwardDiff even, so we're probably overly cautious here.

Celeste/versions/0.3.0/requires
jeff-regier: Hi Tony, I think the past tags are ok without it. There is no 0.5 of ForwardDiff even, so we're probably overly cautious here.

dpsanders: The best thing to do, in that case, is not to put any upper bound.

jeff-regier: @jrevels Do we need the upper bound?

jrevels: Since Celeste reaches under the API and uses `Dual`s directly, and [`Dual` is getting overhauled in ForwardDiff v0.5.0](https://github.com/JuliaDiff/ForwardDiff.jl/pull/213), it's almost certainly the case that ForwardDiff v0.5.0 will break Celeste until Celeste updates to support it. So I think the upper bound makes sense.

jeff-regier: @tkelman Is it OK to merge this PR as it is?

tkelman: yep, sorry for the delay

tkelman: now that there is a ForwardDiff 0.5.0, Pkg might downgrade Celeste to 0.2.1 in order to upgrade to it

PR #8640, opened 2017-04-01
Script to populate and save test dependencies of registered packages

tkelman:
Since the contents of `test/REQUIRE` aren't saved in METADATA or any other central location, I wrote this script to collect and save them all. Turns out for the nested dictionary here, JLD is slower and takes much more space than saving the data as a round-trippable Julia syntax string. Using the Julia serializer would be faster than parsing the string representation, but take a few times more space and be much worse to track in version control (being binary).

At least running on Linux with a good connection, the script takes a little under a minute to run the first time, and the saved data is just under a megabyte. Updating a previously saved copy takes just a few seconds, since it only needs to check newly-tagged versions.

I plan on using this data for some automated reverse-dependency testing for proposed new tags, and it'll be good to include test-only reverse dependencies as well as standard dependencies. Since the data is a little large, not needed for normal usage of Pkg, and it would be a fair amount of git noise to continuously track it on the metadata-v2 branch that everyone uses, I'll probably set up a separate branch on my personal METADATA and a Travis cron job to update it nightly. Will see how that goes. It might make things slightly easier if the script itself is committed here though. It's not that long and hopefully once things are up and running I won't have to change it much (at least until Pkg3, but this makes for a pretty good reason to save test dependencies within the Pkg3 registry format anyway).

### DIFF
--- a/.test/testdeps.jl
+++ b/.test/testdeps.jl
@@ -1,0 +1,124 @@
+#!/usr/bin/env julia
+# script to populate test dependencies of registered packages
+# in the same format as metadata-v2
+
+using ProgressMeter
+cd(Pkg.dir("METADATA")) do
+    old_branch = readchomp(`git rev-parse --abbrev-ref HEAD`)
+    old_upstream = readchomp(ignorestatus(`git config --get remote.upstream.url`))
+    try
+        old_upstream == "" || run(`git remote rm upstream`)
+        run(`git remote add -f upstream https://github.com/JuliaLang/METADATA.jl`)
+        sha_upstream = readchomp(`git rev-parse upstream/metadata-v2`)
+        startpoint = strip(readchomp(ignorestatus(`git log -1 --format=%B testdeps --`)))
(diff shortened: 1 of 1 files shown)


tkelman:
I guess Jeff has more front-end performance work to do, saving as JSON is faster even accounting for the bit of translation needed. https://gist.github.com/1bab78b690bc0cba7944ca3f04f5f1d9

tkelman:
I'll need to refactor which Pkg functions and types this goes through, turns out this version does not preserve platform annotations.

tkelman:
current output of this script, which contains the entire content of metadata plus all of the test-only dependencies in a 2 MB json file: https://gist.github.com/3fc1c5039911f8f1b3fcd3d1a5ef20ef

tkelman:
> I'll probably set up a separate branch on my personal METADATA and a Travis cron job to update it nightly. Will see how that goes.

will be running this at https://github.com/tkelman/METADATA.jl/commits/dailyjson for now

PR #8658, opened 2017-04-03
Tag Laplacians v0.1.2 [https://github.com/danspielman/Laplacians.jl]

danspielman:
Including a new Laplacian solver, and improving the numerics on some others.


tkelman:
If you use Compat and Conda in deps/build.jl, you need to list them in REQUIRE

danspielman:
Turns out that the right thing to do was remove deps/build.jl.
I'll do that, and then create a new pull request.

PR #8668, opened 2017-04-03
Tag PGFPlots v1.5.0 [https://github.com/sisl/PGFPlots.jl]

tawheeler:
Dropping 0.4 support.
Added `xbins` option for `Plots.Linear(function, domain)`.


PGFPlots/versions/1.5.0/requires
```diff
@@ -0,0 +1,12 @@
+julia 0.5
+TikzPictures 0.3.4
+Images 0.5.2
```
tkelman: ImageMeta requires 0.6, and IndirectArrays is missing from REQUIRE

PGFPlots/versions/1.5.0/sha1
```diff
@@ -0,0 +1 @@
+6b9755457caf87a8295c1d928286ec3891d45ace
```
tkelman: will need to replace the git tag too

tawheeler:
Thanks! Updated with Images 0.6 and IndirectArrays

PR #8699, opened 2017-04-05
Tag Expect v0.2.0

wavexx:


tkelman:
Do you expect the test failure on mac travis?

wavexx:
I disabled testing on OSX. This is something down in libuv failing to call cfmakeraw appropriately. Since I do not have a mac, it's hard for me to debug this particular issue even though I submitted the patch for cfmakeraw myself.

tkelman:
Do you know if your past releases worked on mac? Would this be a regression on that platform?

wavexx:
Expect.jl was never tested on OSX in earlier releases. I couldn't test the libuv patch on OSX either when I submitted the PR.

If there's a way to get access to an interactive shell, I'd be happy to debug it myself.

wavexx:
Maybe I should add that, in any case, the previous versions are for sure broken.

tkelman:
I think if you contact Travis support you can get temporary ssh access to a build worker

wavexx:
On Wed, Apr 05 2017, Tony Kelman wrote:
> I think if you contact Travis support you can get temporary ssh access to a
> build worker

That's nice to know. Any particular pointer or just the normal support?

But given this is likely something that I'll likely have to fix in
libuv, can we move this PR forward meanwhile?

I'd like to get Polyglot.jl back into a usable state. And fix some
deadlocks in the previous Expect code as well.

[looking at the code in libuv, I don't see anything that could break on
OSX actually. Must be something minor]

PR #8701, opened 2017-04-05
Tag Gasp.jl v0.1.0

kpamnany:
Now tested and working on up to 1K Xeon nodes and up to 4K Xeon Phi nodes. :)


tkelman:
Any idea why Travis started failing at https://github.com/kpamnany/Gasp.jl/commit/6a7260af1e9ffda5ed0364a0f51996bf91106e5e ?

kpamnany:
I'm trying to figure it out. Travis runs two tests, one of Dtree, and the other of Garray, both started the exact same way (with `mpiexec.hydra -np 4 ...`). The first fails, the second passes. But both tests pass on three different machines I have tested on locally. The failure is a segfault in the MPI process management daemon (`hydra_pmi_proxy`), which makes no sense. I'll figure it out and fix it eventually.

kpamnany:
Ahh, it was the added direct call to `__init__()`, needed for the static compiles we've been doing of Celeste. It shouldn't be a problem in general.

tkelman:
Good find. Do you want the tagged release to include https://github.com/kpamnany/Gasp.jl/pull/2/commits/5b9e47302235e020e45fc4db4bcfe0a0abfeecf6 and pass on Travis then, or neither?

kpamnany:
It's fine to release this commit. I may be getting rid of `__init__()` altogether and forcing users to explicitly initialize, but there are other implications to be considered there so for now this is good.

PR #8703, opened 2017-04-06
Tag GSL v0.3.6 [https://github.com/jiahao/GSL.jl]

yuyichao:


tkelman:
You probably shouldn't be introducing deprecations in a patch release that still supports 0.4, where many packages aren't updating any more

PR #8710, opened 2017-04-06
Tag NLSolversBase.jl v2.0.0

attobot:
Repository: [JuliaNLSolvers/NLSolversBase.jl](https://github.com/JuliaNLSolvers/NLSolversBase.jl)
Release: [v2.0.0](https://github.com/JuliaNLSolvers/NLSolversBase.jl/releases/tag/v2.0.0)
Diff: [vs v1.0.2](https://github.com/JuliaNLSolvers/NLSolversBase.jl/compare/4b9b0221f1ad4dd8d92417bfa6eb91aa57611a2d...368feba0ae66d9e3d8f0a5015fb2e1e05962a8f4)
`requires` vs v1.0.2: no changes
cc: @pkofod


pkofod:
This allows us to address https://github.com/JuliaNLSolvers/Optim.jl/issues/156 in Optim and NLsolve. It is going to be breaking a lot of code, but it seems worth it to be consistent with other Julia code, and I'd rather do it now, than after Julia v1.0 is released.

KristofferC:
I feel this need some type of warning or something. Perhaps also a PkgEval run with this and file issues in packages that break?

pkofod:
> I feel this need some type of warning or something.

You mean on `using` ? We could do that.

> Perhaps also a PkgEval run with this and file issues in packages that break?

I doubt that would tell us anything at this point, as nothing will actually require this package v2.0.0. The only packages that depend on NLSolversBase have upper limits in their REQUIRE.

KristofferC:
> The only packages that depend on NLSolversBase have upper limits in their REQUIRE.

Okay, cool, then merging this is no problem.

Perhaps we could try to coordinate a release that bumps the upper limit of NLSolversBase on the relevant packages and run PkgEval on that PR and file issues for breaking packages?

pkofod:
Yes, that is certainly a good idea. I plan to do discourse, blogs, etc to make sure everyone who is following Julia knows :)

tkelman:
LineSearches depends on NLSolversBase but doesn't have an upper bound

pkofod:
> LineSearches depends on NLSolversBase but doesn't have an upper bound

Okay. Then let this sit until LineSearches is patched and tagged. I believe we'll have to change a few lines to conform to this change (kind of the point of relying on the NLSolversBase API in the first place).

(relevant https://github.com/JuliaNLSolvers/LineSearches.jl/commit/e76e5e2f2f16db95e31e7b67cb26643dfab9b81c )

tkelman:
or could add the upper bound in metadata to the existing tag, without waiting.

PR #8716, opened 2017-04-06
Tag SpeedDate.jl v0.2.2

attobot:
Repository: [Ward9250/SpeedDate.jl](https://github.com/Ward9250/SpeedDate.jl)
Release: [v0.2.2](https://github.com/Ward9250/SpeedDate.jl/releases/tag/v0.2.2)
Travis: [Travis Build Status](https://travis-ci.org/Ward9250/SpeedDate.jl/builds/219261420)
Diff: [vs v0.2.1](https://github.com/Ward9250/SpeedDate.jl/compare/578d4fcf294f22f3d0318739278a80b923703543...0f218e44765cf2842cfad13c74bf09e598e4b1c1)
`requires` vs v0.2.1: 
```diff
--- 0.2.1/requires
+++ 0.2.2/requires
@@ -1,6 +1,6 @@
 julia 0.5
 DataFrames
 Gadfly
-Bio 0.4.3
+Bio
 ArgParse 0.4
 Gtk 0.10.4
```
cc: @Ward9250


SpeedDate/versions/0.2.2/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.5
+DataFrames
+Gadfly
+Bio
```
tkelman: does it actually work against any released version of Bio.jl then?

SpeedDate/versions/0.2.2/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.5
+DataFrames
+Gadfly
+Bio 0.4.7
```
tkelman: what's in 0.4.7 that wasn't in anything earlier? `distance` or `flagmutations` or `EachWindow` ?

Ward9250: Bio 0.4.7 was a release in which I added (with this package in mind actually) a method of distance which utilises a sliding window.

tkelman: looked a little odd because there didn't seem to be any calls to those here? Pkg auto-upgrades, only time you have to worry about bounds is if something else is pinning or bounding. And when that happens, setting bounds too tight prevents installation of version combinations that could work together, maybe with a bit less functionality but that's probably preferable to resolution failures?

Ward9250: The windowed function added to 0.4.7 is called in the package in file dating.jl, line 118: `dists, vars, windows = distance(model, sequences, args["width"], args["width"])`. This line was called in previous versions of the package too, when it used the lower version number of Bio, but doing so would crash with a method error, I only had the function on my Bio master, which I never noticed I had checked out. 0.4.7 fixes all of that.

PR #8717, opened 2017-04-06
Tag QuartzImageIO v0.3.0 [https://github.com/JuliaIO/QuartzImageIO.jl]

rsrock:


QuartzImageIO/versions/0.2.1/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5 0.6-
```
tkelman: what is it that doesn't work on 0.2.0 and 0.2.1 with julia 0.6?

rsrock: This line here is the culprit
https://github.com/JuliaIO/QuartzImageIO.jl/commit/58fc90c2b0b1e18e27fe9efbd6080053e614c199#diff-626a7c38e3d063b142ed47cbd941c5dfR418

Without that change, we run into "ERROR: LoadError: TypeError: ccall method definition: expected Type, got TypeVar".  It turns out we don't need the more general version in this package, a fixed ccall return type of `Ptr{Void}` is fine.

tkelman: is that at load time, unconditionally? I'd use a known version as the upper bound, like release or alpha, since early 0.6-dev versions wouldn't have had this issue (could possibly come up if ever needed to bisect on base to hunt down a bug triggered by this package)

rsrock: It is at load time, unconditionally.  Honestly, I'm not sure when this problem appeared in 0.6-dev.

Closing off all of 0.6- seems to be the most conservative choice. v0.2.0 and v0.2.1 are broken on some good part of 0.6-.  But the change here in v0.3.0 should work on all of 0.6-dev because it gets rid of the typevar altogether.  So any testing on 0.6- should pull in this fix and should pass in a bisect. I wish I could say that I planned it that way, but I have to admit that I hadn't even considered that problem.

Should I try to narrow the window down any further?

tkelman: it's probably Jameson's PR that changed how ccall works. I'd rather only do upper bounds to known broken versions, instead of ruling out installation of version sets that could work together

rsrock: I assume that's JuliaLang/julia#18754.  So I closed off at 0.6-pre.alpha, which occurred about a month later.  Good enough for government work?  I think to pin it down to that commit exactly, I'd have to put a version check in the code, because REQUIRES only handles tags.

tkelman: I forget, was there a reason the upper bound was only added on 0.2.x versions here?

rsrock: I don't recall any.  This PR will close off the rest: #9820

PR #8723, opened 2017-04-06
Pull request/58f1e26dd

emoszkowski:


DSGE/versions/0.1.6/requires
```diff
@@ -4,6 +4,6 @@ DataStructures
 Distributions
 FredData
 HDF5
-Optim 0.5
+Optim 0.7
```
tkelman: isn't​ this also needed on 0.1.5? should also make this change on master's REQUIRE file

emoszkowski: Yup, good point. I think the easiest thing is for me to just push a single version update. I'll close this PR and open a new one.

PR #8729, opened 2017-04-07
Tag DiffEqBase.jl v0.14.0

attobot:
Repository: [JuliaDiffEq/DiffEqBase.jl](https://github.com/JuliaDiffEq/DiffEqBase.jl)
Release: [v0.14.0](https://github.com/JuliaDiffEq/DiffEqBase.jl/releases/tag/v0.14.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqBase.jl/builds/219585936)
Diff: [vs v0.13.0](https://github.com/JuliaDiffEq/DiffEqBase.jl/compare/f84c545d2cc739ae03a6b0df68c105d98ed680ef...6146d23c21fd44329fb405546317a24e7eba7ed5)
`requires` vs v0.13.0: no changes
cc: @ChrisRackauckas


ChrisRackauckas:
Tests are expected to fail. But this is safe if merged after https://github.com/JuliaLang/METADATA.jl/pull/8728 . Some libraries in the tests will require updates in addition.

tkelman:
And you'll also need to fix
```
both ParameterizedFunctions and DiffEqBase export "ParameterizedFunction"; uses of it in module Main must be qualified
Constructed Parameterized Functions: Error During Test
  Got an exception of type LoadError outside of a @test
  LoadError: UndefVarError: ParameterizedFunction not defined
```

ChrisRackauckas:
Yes, that's why yesterday I merged an upperbound on ParameterizedFunctions.jl (whose name is now slightly off, should be something about macros or DSL, but that could be addressed with Pkg3) which makes the current release not able to be used with DiffEqBase 0.14.0. Then I am releasing this, along with a new version of ParameterizedFunctions.jl which requires 0.14.0. Both of these tags are needed together to fix the error.

PR #8730, opened 2017-04-07
Tag JFVM v1.0.0 [https://github.com/simulkade/JFVM.jl]

simulkade:
Compatible with version 0.6


JFVM/versions/1.0.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
```
tkelman: if you still support julia 0.4, should still be testing it on travis

simulkade: You are right. I don't support it anymore. I changed it and opened a new pull request.

PR #8748, opened 2017-04-08
Register new package PackageGenerator.jl v0.0.1

attobot:
Repository: [bramtayl/PackageGenerator.jl](https://github.com/bramtayl/PackageGenerator.jl)
Release: [v0.0.1](https://github.com/bramtayl/PackageGenerator.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/bramtayl/PackageGenerator.jl/builds/219981334)
cc: @bramtayl


ararslan:
How is this different from the functionality provided by PkgDev?

bramtayl:
> Generate a package named `package` with some nice bells and whistles.
> These include:
> 
> - creating a matching github repository
> - activating the repository on travis
> - creating working documentation that
>   - automatically syncs
>   - tests doctests as part of your package testing suite

ararslan:
Seems like it would be worth it to submit a PR to PkgDev to incorporate those features rather than having a separate package.

bramtayl:
I have have tried twice, first here https://github.com/JuliaLang/PkgDev.jl/pull/92 then here https://github.com/JuliaDocs/Documenter.jl/pull/416

bramtayl:
I can keep pushing for the documenter pull request, but

- This is quite a large chunk of code to review and maintain
- This package combines the functionality from PkgDev.generate and Documenter.generate. It turns out this is much safer and efficient; in the two phase model, Documenter.generate had to use heuristics to figure out exaxtly what was done in PkgDev.generate

bramtayl:
What do you think

ararslan:
I'll leave it up to @tkelman, master of all things METADATA.

bramtayl:
Is the issue here a lack of tests? It can't figure out a good way to test this (I can do it myself; in fact I generated the container for this package using its own code).

tkelman:
Why doesn't this include anything related to appveyor?

tkelman:
Default generated "no tests" should fail so you don't get a free pass unless you actually test something. https://github.com/JuliaLang/PkgDev.jl/pull/45

bramtayl:
> Why doesn't this include anything related to appveyor?

I thought about including appveyor. The default set up now is just to test on Linux; I've found testing on Windows and OSX much slower and redundant for the simple macro packages I write. I was considering including a `multiplatform = true` keyword in a future version. If this is a priority, though, I can do this now.

> Default generated "no tests" should fail so you don't get a free pass unless you actually test something. 

Ok, I can change this. It's set up right now to build a simple documentation page out of the box to show off functionality and documentation isn't built if tests fail.

ChrisRackauckas:
>I thought about including appveyor. The default set up now is just to test on Linux; I've found testing on Windows and OSX much slower and redundant for the simple macro packages I write. I was considering including a multiplatform = true keyword in a future version. If this is a priority, though, I can do this now.

Most users don't use Linux (probably the vast majority don't). Many things can be OS dependent and the only way to make sure there isn't anything OS dependent is to test (for example, Optim.jl had errors on 32-bit Windows for awhile because it left off AppVeyor tests which ended up showing that hardcoded Int64 dispatches were causing issues). Tests should make sure the package works for most users.

bramtayl:
Ok, the new version includes

- Full appveyor support
- Testing on all three platforms
- Tests that fail by default

bramtayl:
And also a much more straight-forward way of dealing with `ssh_keygen_path`

PR #8751, opened 2017-04-08
Tag Diversity.jl v0.3.0

attobot:
Repository: [richardreeve/Diversity.jl](https://github.com/richardreeve/Diversity.jl)
Release: [v0.3.0](https://github.com/richardreeve/Diversity.jl/releases/tag/v0.3.0)
Travis: [Travis Build Status](https://travis-ci.org/richardreeve/Diversity.jl/builds/219877437)
Diff: [vs v0.2.8](https://github.com/richardreeve/Diversity.jl/compare/48abb39279e8ad14b94ed08d8740319bba6258ca...c5d28ea4b191e050994e25b892dbc8e74d8239bb)
`requires` vs v0.2.8: 
```diff
--- 0.2.8/requires
+++ 0.3.0/requires
@@ -1,4 +1,3 @@
-julia 0.3
-Docile 0.3.1-
-Lexicon
-Compat 0.7.9-
+julia 0.5
+Compat 0.21.0
+DataFrames 0.9.0
```
cc: @richardreeve


Diversity/versions/0.3.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+Compat 0.21.0
```
tkelman: `@compat abstract type` only requires 0.17, are you relying on anything newer?

richardreeve:
I couldn't find the list there used to be of what was defined in which version of Compat. Does it really matter?

tkelman:
https://github.com/JuliaLang/Compat.jl/blame/master/README.md is the closest thing. If you set bounds too tight, then you prevent installation of version combinations that would work together.

richardreeve:
Right, but if I'm not being stupid (which is likely!) blame doesn't actually say the version that you need – do you mean I should track through the commits and find out which version it gets included in? I could have sworn there used to be a nice list saying which update belonged to which version. In any event I understand your general point, but it scarcely seems to apply to Compat...

tkelman:
Do you know what features you're using from Compat? Another relatively simple option is
```
ENV["JULIA_PKGDIR"] = joinpath(pwd(), "jlpkgtmp")
Pkg.init()
Pkg.clone("Diversity")
Pkg.pin("Compat", v"0.18.0")
Pkg.test("Diversity")
```
which would fail due to indirect dependencies on DataFrames if you tried with Compat 0.17.0, but works fine with Compat 0.18.0 on Julia 0.5. You're also setting a patch release requirement for Distances in test/REQUIRE, are you relying on a bugfix or did that package introduce new features in a patch?

tkelman:
> blame doesn't actually say the version that you need

If you click the commit on github there's a little indicator underneath the message which says what tags that commit is included in.

richardreeve:
Ah, I didn't realise that, thanks. As far as Distances is concerned, the most important code was added in 0.3.2, but only 0.4.1 stops all of the deprecation warnings. Likewise for StatsBase that I've now added to the test REQUIRE. Anyway, I've switched to Compat 0.18.0 as you suggest.

tkelman:
Fixing deprecation warnings is nice to have (and I imagine those are on 0.6 more so than 0.5) but not mandatory for the package to function, right? Pkg will generally try to use the newest versions of things it's allowed to, lower bounds are to express the lowest version that would work correctly and upper bounds are to avoid known-broken API changes in dependencies.

richardreeve:
Okay, I've done what you required now.

richardreeve:
Minor tidy while it waits to go into METADATA.

PR #8752, opened 2017-04-08
Tag Hexagons v0.1.0 [https://github.com/GiovineItalia/Hexagons.jl]

tlnagy:
Julia 0.6 compat


Hexagons/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
+Compat
```
tkelman: best to declare a minimum version dependency that implements the required features

tkelman: 0.17 for `@compat abstract type`

tlnagy:
Thanks @tkelman, that slipped by me during review

PR #8756, opened 2017-04-09
Tag Vega v0.7.0

randyzwitch:
Update to 0.5 compatibility, before moving to 0.6 compatibility


tkelman:
Looks like you removed Compat from REQUIRE but not src/Vega.jl?

randyzwitch:
Can I get away with this change, just modifying the SHA1? Removing it was my intention, so I just pushed the fix to the main Vega.jl library.

tkelman:
That should have failed Travis, unless you also update the git tag in the repo to match

randyzwitch:
I'll just close it and do it the right way then.

PR #8757, opened 2017-04-09
Register new package DataFmtExt.jl v0.0.3

attobot:
Repository: [strickek/DataFmtExt.jl](https://github.com/strickek/DataFmtExt.jl)
Release: [v0.0.3](https://github.com/strickek/DataFmtExt.jl/releases/tag/v0.0.3)
cc: @strickek


strickek:
Extension for Julia readdlm() and writedlm(): Uses semicolon as default delimiter, supports decimal comma and Date/DateTime types. The standard behavior of the functions is not changed. A new Bool argument (true or false), which has to be added after the other positional arguments, activates and controls the additional functionality for reading and writing textfiles.

tkelman:
It would be better not to extend Base's readdlm, but use a package local readdlm instead. Otherwise any other package that had the same idea would be overwriting methods with this one, and couldn't be used at the same time.

strickek:
@tkelman : Im am quite new in Julia. Could you give me a hint what I shold change in the head of the module or where in the documentation I found the answer for local useing packages to be installed with Pkg.add. Here is the actual head of the module:
```
module DataFmtExt
using Base.Dates,
    Base.DataFmt.readdlm_string, Base.DataFmt.val_opts
import Base.DataFmt.readdlm, Base.DataFmt.writedlm
export readdlm, writedlm
```

tkelman:
If you leave off the `import Base.DataFmt.readdlm, Base.DataFmt.writedlm` and `export readdlm, writedlm`, then the package can have its own private `DataFmtExt.readdlm` and `DataFmtExt.writedlm` that won't globally modify the Base version. Any users could do `using DataFmtExt: readdlm, writedlm` locally in their modules if they wanted to use yours with short names.

strickek:
Ok, thank you. My generall idea was to keep it simple and as near as possible to standard readdlm / writedlm. Therefore the special using syntax is to complex for my usability goal. So I am thinking about rewriteing the module and use the new names readdlm2 / writedlm2 instead. This would have also the advantage, that I dont need the additional Bool argument to switch between the standard and the new methods. What do you think?

tkelman:
"2" isn't all that descriptive to distinguish it from the usual defaults, could maybe call the functions `readdlmext` to mirror the package name if there aren't any other ideas

strickek:
I like ..2 because it is short and its quite the same in R with csv2. The new module name will be "ReadWriteDlm2".
Do I have anything to do for deleting this "Register new package DataFmtExt.jl v0.0.3 #8757" ?

tkelman:
we can just close this if that's what you'd like to do.

the package naming guidelines recommend being descriptive in your package names, if there's a way to concisely convey how your versions differ in the naming that would help in discoverability

PR #8766, opened 2017-04-10
Tag NamedArrays v0.5.4 [https://github.com/davidavdav/NamedArrays.jl]

davidavdav:
Possibly the final version for Julia 0.5, before moving to 0.6 development.


tkelman:
Since you've dropped support for Julia 0.4 here, this needs to use a new package minor version

PR #8768, opened 2017-04-10
METADATA Tests: Missing SHA

omus:
Testing changes in #8548


omus:
Works as intended

> ERROR: LoadError: LoadError: The following tags for Example have not been pushed to upstream repository: 0.0.3

PR #8769, opened 2017-04-10
diffeq upper bounds

ChrisRackauckas:
The next DiffEqBase.jl is going to have a very breaking API change, and each of these will have a release to follow it.

### DIFF
--- a/Sundials/versions/0.3.0/requires
+++ b/Sundials/versions/0.3.0/requires
@@ -1,3 +1,4 @@
 julia 0.4
 BinDeps 0.4.3
 Compat 0.9.0
+DiffEqBase 0.0.0 0.15.0
--- a/Sundials/versions/0.3.0/requires
+++ b/Sundials/versions/0.3.0/requires
@@ -1,4 +1,3 @@
 julia 0.4
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase 0.0.0 0.15.0
--- a/Sundials/versions/0.4.0/requires
+++ b/Sundials/versions/0.4.0/requires
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase 0.0.0 0.1.3
+DiffEqBase 0.0.0 0.2.0
--- a/Sundials/versions/0.4.0/requires
+++ b/Sundials/versions/0.4.0/requires
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase
+DiffEqBase 0.0.0 0.15.0
--- a/Sundials/versions/0.4.0/requires
+++ b/Sundials/versions/0.4.0/requires
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase 0.0.0 0.15.0
+DiffEqBase 0.0.0 0.1.3
--- a/DASKR/versions/0.0.1/requires
+++ b/DASKR/versions/0.0.1/requires
@@ -1,2 +1,2 @@
 julia 0.5
-DiffEqBase
+DiffEqBase 0.0.0 0.15.0
--- a/DASKR/versions/0.0.2/requires
+++ b/DASKR/versions/0.0.2/requires
@@ -1,3 +1,3 @@
 julia 0.5
-DiffEqBase
+DiffEqBase 0.0.0 0.15.0
 Compat
--- a/DelayDiffEq/versions/0.1.0/requires
+++ b/DelayDiffEq/versions/0.1.0/requires
@@ -1,5 +1,5 @@
 julia 0.5
-DiffEqBase 0.8.0
+DiffEqBase 0.8.0 0.15.0
 OrdinaryDiffEq 1.2.0
 DataStructures 0.4.6
 RecursiveArrayTools 0.2.0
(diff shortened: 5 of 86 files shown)


Sundials/versions/0.4.0/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase
+DiffEqBase 0.0.0 0.15.0
```
tkelman: this looks like it needs an even lower bound, it doesn't load with DiffEqBase 0.14, SundialsAlgorithm not defined

Sundials/versions/0.3.0/requires
```diff
@@ -1,3 +1,4 @@
 julia 0.4
 BinDeps 0.4.3
 Compat 0.9.0
+DiffEqBase 0.0.0 0.15.0
```
tkelman: this is impossible to satisfy on julia 0.4

ChrisRackauckas: Should I not add the upper bound to ones which require an older version of Julia?

Sundials/versions/0.4.0/requires
ChrisRackauckas: `SundialsAlgorithm` is defined in Sundials

Sundials/versions/0.3.0/requires
tkelman: can't add the dependency period, unless you make an empty no-op release of DiffEqBase that supports those versions of Julia

Sundials/versions/0.4.0/requires
tkelman: not at 0.4.0 it wasn't https://github.com/JuliaDiffEq/Sundials.jl/compare/v0.4.0...v0.5.0

ChrisRackauckas:
Is this what you were asking for? Any 0.4 were fixed to not introduce a DiffEqBase dep. The old Sundials was given the correct bound.

Sundials/versions/0.4.0/requires
```diff
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.3
 Compat 0.9.0
-DiffEqBase 0.0.0 0.15.0
+DiffEqBase 0.0.0 0.1.3
```
tkelman: this should probably be 0.2, not 0.1.3 https://github.com/JuliaDiffEq/DiffEqBase.jl/compare/v0.1.2...v0.1.3

ChrisRackauckas: oh yeah, forgot it was exclusion there.

tkelman:
You may have the issue of Pkg downgrading to an old ODE or Sundials that doesn't depend on DiffEqBase in order to upgrade to a newer DiffEqBase.

ChrisRackauckas:
I'll make sure those tags are not the last ones then. Would that be fine?

tkelman:
Sure, but it'll be best to prepare DiffEqBase-0.15-compatible tags of those two packages quickly.

ChrisRackauckas:
Yes. This is a pretty sweeping interface change, but

```
StochasticDiffEq
OrdinaryDiffEq
DiffEqDevTools
DiffEqProblemLibrary
DiffEqPDEBase
DelayDiffEq
FiniteElementDiffEq
ODEInterfaceDiffEq
DiffEqUncertainty
DiffEqSensitivity
DiffEqWebBase
```

Are all updated on master, and

```
Sundials
ODE
LSODA
```

Already have the PRs completed (I am waiting for this and the DiffEqBase PR to go in to merge them, and am looking for someone to review if needed).

```
DASSL
DASPK
```

will have PRs in a moment.

Tests on each of the JuliaDiffEq packages show that this is sufficient. DiffEqBase.jl will get this update, and a bunch of releases will pour in soon after.

ChrisRackauckas:
All of the PRs are set. When this and the DiffEqBase changes are merged, everything can just get tagged.

ChrisRackauckas:
Here's the setup:

When this is merged, the DiffEqBase v0.15.0 can be merged. Right after that, DiffEqProblemLibrary.jl can be tested on CI and merged (its test is just compatibility with the interface for problem definitions). At the same time, DiffEqDevTools.jl will merge too. You're going to have to take that one on faith: its tests need OrdinaryDiffEq.jl and StochasticDiffEq.jl tests, which need DiffEqDevTools.jl's update. This is the issue because it's the library where all of the testing functionality goes... but if we just accept that one due to local testing, we can test it after the other releases to make sure it's okay.

Then 

```
StochasticDiffEq
DiffEqPDEBase
FiniteElementDiffEq
ODEInterfaceDiffEq
Sundials
ODE
LSODA
DASSL
DASPK
```

should be able to have their tests pass. After those are released,

```
OrdinaryDiffEq
```

which uses ODEInterfaceDiffEq in its tests, and then after that

```
DelayDiffEq
DiffEqUncertainty
DiffEqSensitivity
```

which all use OrdinaryDiffEq.jl in its tests. Then re-test DiffEqDevTools.jl to make sure its tests pass, and it's all done. All of the master branches + PRs are ready for this.

So essentially, the only issue here is that all of the packages have integration tests, and when there is a huge internal API change (here, a change to all of the abstract types to only have the essential parameters), this breaks the testing suite. The testing packages are usually rock solid and don't change much, and allow all of the solvers to run the same tests. But they themselves need a solver to test their testing functionality...

But yes, we're ready.

tkelman:
okay sounds good, I'll double check this one last time in an hour or two then we can get them going

ChrisRackauckas:
I'm at least going to start getting them in motion, so it's just clicking reset on Travis.

PR #8771, opened 2017-04-10
METADATA Test: Success

omus:
Testing changes in #8548


tkelman:
This should have failed due to the minimum Julia version being too low, but it's in the grandfathered list. Different package maybe.

omus:
@tkelman you want me to test with another package?

tkelman:
yeah, which error path were you trying to trigger here?

omus:
I was actually trying to validate that checks could still pass.

PR #8772, opened 2017-04-10
METADATA Test: Minimum Julia version (policy 5)

omus:
Not sure if modifying "requires" files is allowed.

Testing changes in #8548


martinholters:
> ERROR: LoadError: METADATA/TimeZones/versions/0.3.0/requires: oldest allowed julia version 0.3.0 too old (>= 0.4.0 needed)

This is what you were after, right?

omus:
Yes it was

PR #8776, opened 2017-04-11
Tag RLEVectors v0.2.10 [https://github.com/phaverty/RLEVectors.jl]

phaverty:


RLEVectors/versions/0.2.10/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5-
```
tkelman: if you don't need to support prereleases, don't use a trailing dash here

tkelman:
since this drops support for julia 0.4, it needs to use a new package minor version

tkelman:
closing in favor of #8777

PR #8787, opened 2017-04-11
Tag Rotations.jl v0.4.0

attobot:
Repository: [FugroRoames/Rotations.jl](https://github.com/FugroRoames/Rotations.jl)
Release: [v0.4.0](https://github.com/FugroRoames/Rotations.jl/releases/tag/v0.4.0)
Travis: [Travis Build Status](https://travis-ci.org/FugroRoames/Rotations.jl/builds/221000246)
Diff: [vs v0.3.5](https://github.com/FugroRoames/Rotations.jl/compare/3ef1cce96f25dc604fc4dcb5c84b51b4c84cd488...aeb9c30e4a24f14b12be59c5a751c5ee6d3f8428)
`requires` vs v0.3.5: 
```diff
--- 0.3.5/requires
+++ 0.4.0/requires
@@ -1,2 +1,2 @@
-julia 0.5
-StaticArrays 0.3.0
+julia 0.6-
+StaticArrays 0.4.0 0.5-
```
cc: @tkoolen


tkoolen:
The upper bound on StaticArrays is in anticipation of further breaking changes in StaticArrays, expected in a week or so.

tkelman:
What do you expect those breaking changes to do w.r.t. the previous releases of Rotations? Unless the upper bound is applied to past versions, Pkg could choose to downgrade Rotations in order to upgrade StaticArrays.

tkoolen:
You're right; for 0.6 users, I guess it would indeed be possible for Pkg to choose to downgrade to an earlier version of Rotations that is not even compatible with Julia 0.6.

I suppose this could be fixed by setting a StaticArrays upper bound of 0.4.0- for versions 0.3.x (0.2 and earlier used FixedSizeArrays). We could also use 0.6- as the Julia upper bound for all Rotations versions up to this tag. Which of these has your preference (or maybe both)?

tkelman:
The StaticArrays upper bound on the past tags seems more accurate, if you think the breaking changes would apply to the way in which those versions were using StaticArrays

tkoolen:
Please close in favor of https://github.com/JuliaLang/METADATA.jl/pull/8793.

PR #8793, opened 2017-04-11
Tag Rotations.jl v0.4.0

tkoolen:
Repository: [FugroRoames/Rotations.jl](https://github.com/FugroRoames/Rotations.jl)
Release: [v0.4.0](https://github.com/FugroRoames/Rotations.jl/releases/tag/v0.4.0)
Travis: [Build Status](https://travis-ci.org/FugroRoames/Rotations.jl)
Diff: [vs v0.3.5](https://github.com/FugroRoames/Rotations.jl/compare/v0.3.5...v0.4.0)
requires vs v0.3.5:

--- 0.3.5/requires
+++ 0.4.0/requires
@@ -1,2 +1,2 @@
-julia 0.5
-StaticArrays 0.3.0
+julia 0.6-
+StaticArrays 0.4.0 0.5-


tkoolen:
CC: @c42f @andyferris (sorry I forgot to do that with the previous PR).

tkelman:
Is the intent that StaticArrays 0.5.0 is going to be Julia-0.6-only? Since the early versions (0.0.1, 0.1.0, and 0.2.0) of Rotations didn't depend on StaticArrays, it may be better to upper-bound the Julia version of those to 0.6 (release, or a known-broken alpha or beta, rather than disallowing even the early prereleases). The same downgrading concern applies for them, but since those supported Julia 0.4 we can't add a fake dependency on StaticArrays to them (unless you want to make a placeholder no-op dummy version of StaticArrays that supports 0.4, just for this purpose)

tkoolen:
StaticArrays 0.4.0 is already 0.6-only, and I don't expect @andyferris to want to re-add 0.5 support. I think it makes more sense for Rotations versions before 0.4.0 to also explicitly have a 0.6-something upper bound than to do the dummy-StaticArrays thing.

As for what upper bound to use: 0.6 would not be useful to the people waiting for this tag right now, who would still have to check out master. I think that kind of thing could lead to fewer users with code that depends on Rotations upgrading to the latest nightly right now, perhaps causing fewer bugs in 0.6 to be found before its release. Rotations and StaticArrays were broken way before alpha and beta releases of 0.6; the https://github.com/JuliaLang/julia/issues/265 fix was the main cause. So I think the upper bound has to be 0.6-.

Really though, all this is working around the fact that both FixedSizeArrays and StaticArrays don't have properly specified Julia version upper bounds for their old releases. But for the sake of moving this PR along, I prefer to add 0.6- upper bounds to all earlier versions of Rotations.

c42f:
@tkelman - yes, StaticArrays 0.4.0 and above will probably be julia 0.6 only.  (It was more or less a package rewrite to make StaticArrays work on julia-0.6. I don't think we have the resources to support both versions.)

@tkoolen I'm happy for us to go ahead and fix the StaticArrays julia version upper bounds if they're broken and this is causing you problems. Please go ahead and do this here if it helps - I'm happy to review it.

I'm a bit concerned about the generous `julia-0.6-` upper bound for the very early versions of Rotations - could this result in the package system deciding to downgrade Rotations to a very early version?  IIRC the API there was rather different and less refined, and I think we really don't want people going back to those old versions.

tkelman:
Those early Rotations versions do load on 0.5, so they'd also load on the early 0.6 prereleases. Upper bounds are exclusive, so the trailing dash is usually not a good idea on an upper bound as it unnecessarily prevents installation of version combinations that would actually work together.

c42f:
Right, that makes sense for the 0.6- upper bounds.

tkoolen:
I'm confused now. Maybe I'm just misunderstanding both of you.

@tkelman, you're saying you want looser upper bounds than 0.6- on the previous releases, correct? 

An upper bound of 0.6 is not possible for the reason I mentioned in https://github.com/JuliaLang/METADATA.jl/pull/8793#issuecomment-293429843. So if we want upper bounds on the previous releases that include early 0.6 prereleases, we'd have to figure out a very specific version number where things went wrong (a bound like `julia 0.5 0.6.0-dev.1553` (note: completely made up)), which we can do, but it requires some effort and is surprising to me because a quick grep of the `requires` files in METADATA.jl shows that only ACME.jl has ever done that. 

It just seems to me like there is very little to gain in supporting early 0.6 prereleases in combination with old Rotations versions in this way. If things don't work with some 0.6 nightly version from three months ago, the natural thing for a user is to get the latest nightly or 0.6 release, I would think.

c42f:
@tkoolen I think I'm also confused, perhaps I didn't understand your comment https://github.com/JuliaLang/METADATA.jl/pull/8793#issuecomment-293429843. IMHO we should just fix the StaticArrays julia version upper bounds, if this would sort out the complications here.

tkelman:
> 0.6 would not be useful to the people waiting for this tag right now, who would still have to check out master

Why do you say that? Even with an inactive upper bound that's looser than the current nightly of 0.6-pre.beta, Pkg would by default try to upgrade to latest StaticArrays and Rotations, if it can. Could set the upper bound to 0.6-pre.alpha if that's known to be broken.

tkoolen:
> 0.6 would not be useful to the people waiting for this tag right now, who would still have to check out master

Sorry, that wasn't right; I was conflating the lower bound on 0.4.0 with the upper bound on the earlier releases.

However, if the upper bound on the pre-StaticArrays Rotations versions were 0.6, then these earlier Rotations versions would be compatible with Julia 0.6 prerelease versions. Suppose that a future StaticArrays 0.5.0 uses 0.6- as its lower bound. Then, for 0.6 prerelease versions , Pkg could choose to downgrade Rotations to a pre-StaticArrays version in order to upgrade StaticArrays to 0.5.0.

> Could set the upper bound to 0.6-pre.alpha if that's known to be broken.

OK, I'll do that.

@c42f, I think this solution is sufficient for Rotations. I'll leave the StaticArrays version bounds up to you guys.

c42f:
Here's what I get when pinning the early versions of Rotations using julia-0.5.0:

```
v0.0.1 - package loads, but tests fail 
v0.1.0 - package doesn't load, tests fail
v0.2.0 - tests pass
```

so the first two should probably have an upper bound of 0.5.0.

tkoolen:
Done.

PR #8798, opened 2017-04-12
Tag JSON.jl v0.9.0

attobot:
Release: [v0.9.0](https://github.com/JuliaIO/JSON.jl/releases/tag/v0.9.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaIO/JSON.jl/builds/221240384)
Diff: [vs v0.8.3](https://github.com/JuliaIO/JSON.jl/compare/a145d09fd507c4f590d003606ae68a98649b111f...6ea48684d47faebf97077c78bd923744b4987225)
`requires` vs v0.8.3: no changes
cc: @TotalVerb


TotalVerb:
It should be noted that this version removes deprecations that have been in place since JSON 0.6.0. A GitHub search turned up only old, unmaintained forks of Bokeh.jl and Plots.jl that use removed deprecated functionality, so no packages should be broken by this.

tkelman:
It might be worthwhile to go through and add upper bounds to old versions of some downstream packages that we know were using the now-removed functions, but that can be done later.

TotalVerb:
Bokeh is easy to do, but Plots is a little tougher because it never actually required JSON except as a conditional dependency when Plotly (which did require JSON) is also installed.

A possible solution is to add an unnecessary JSON dependency to all versions of Plots before 0.9.1.

PR #8804, opened 2017-04-12
Tag Loess v0.2.0 [https://github.com/JuliaStats/Loess.jl]

tlnagy:


Loess/versions/0.2.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Compat 0.8.4
```
tkelman: need 0.17 for `@compat abstract type`

tlnagy: Thanks for catching that.

PR #8810, opened 2017-04-12
Tag Compose v0.5.0 [https://github.com/GiovineItalia/Compose.jl]

bjarthur:
1. stripped out old precompilation code
2. dropped support for 0.4 and added same for 0.6
3. upgraded to new documentation system


Compose/versions/0.5.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5
+Colors
+Compat 0.19.0
```
tkelman: only need 0.18 for parametric aliases

PR #8803, opened 2017-04-12
Tag DiffEqProblemLibrary.jl v0.5.0

attobot:
Repository: [JuliaDiffEq/DiffEqProblemLibrary.jl](https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl)
Release: [v0.5.0](https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/releases/tag/v0.5.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqProblemLibrary.jl/builds/221423395)
Diff: [vs v0.4.0](https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/compare/fe37c4b3f2fb4d808f5b681d3cf1fbd079fb7f67...0930b72b2f6abdcdc96fedb7d128d7da4f664954)
`requires` vs v0.4.0: 
```diff
--- 0.4.0/requires
+++ 0.5.0/requires
@@ -1,6 +1,6 @@
 julia 0.5
 ParameterizedFunctions 0.5.0
-DiffEqBase 0.2.0
+DiffEqBase 0.15.0
 FiniteElementDiffEq 0.0.5
 JLD 0.6.5
 DiffEqPDEBase
```
cc: @ChrisRackauckas

PR #8815, opened 2017-04-12
sensible minimum versions for BioEnergeticFoodWebs

ChrisRackauckas:
To fix https://github.com/JuliaLang/METADATA.jl/pull/8803#issuecomment-293709507


tkelman:
"sensible" in what way? Is @tpoisot using something that didn't exist in 1.6.0 or earlier?

ChrisRackauckas:
He is using the default algorithm handling and allows for use with some of the stiff methods. Technically that can work with an earlier version, but the tests in DifferentialEquations.jl for the stuff he uses were updated at this release.

ChrisRackauckas:
Actually, it looks like this won't help...

ChrisRackauckas:
```
julia> PkgDev.Entry.check_metadata()
    BioEnergeticFoodWebs v0.3.0 – no valid versions exist for package PolynomialFactors
    BioEnergeticFoodWebs v0.3.1 – no valid versions exist for package PolynomialFactors
```

still happens on the merge commit. I am not sure why I had one check work.

ChrisRackauckas:
Disreguard that last statement, was testing on the wrong branch.

tkelman:
Don't quite get why tests being updated should determine the minimum version a downstream consumer needs to use some features, but okay.

ChrisRackauckas:
This is the earliest commit that would ensure he got the same behavior:

https://github.com/JuliaDiffEq/DifferentialEquations.jl/commit/3fe213e6d9a6e3e86f61634c6419340d1517cbd0

1.7.0 updates a test here for what types are returned, but since he's not actually using the type, that won't actually affect his behavior. Any commit before this though would change the algorithm choice in his code. I need to find out how to pin down the tag for this.

ChrisRackauckas:
v1.5.0 is the first with the change of that default to give the `Vern7` algorithm, which is likely the default algorithm given in his case. Earlier should work, but that would mean it would change up the ODE algorithm he would get, which could cause problems? I'll put this down to v1.5.0 since that's the safe bet.

ChrisRackauckas:
I put it at v1.5.0 for the reason above. 

https://github.com/JuliaDiffEq/DifferentialEquations.jl/commit/4cd4f310e095f655c10b33c4a6b99f56ed1c89f0

If you wouldn't care about switching up the algorithm that is given to him, v1.3.0 is the first version where this algorithm defaulting existed. Anything before that would error. Let me know if you want it changed.

ChrisRackauckas:
Not sure why Travis was saying it failed. So I merged it locally, I check it, and we'll see what it does next....

tkelman:
I think the resolver's wrong here. Though interestingly it passes with the Julia 0.5.0 resolver code, and fails with 0.5.1 unless you set `JULIA_PKGRESOLVE_ACCURACY` to 2 or higher. cc @carlobaldassi if you're around.

tkelman:
This probably isn't strictly necessary any more, but whatever was giving the resolver a hard time has gone away at least for now.

PR #8817, opened 2017-04-12
add upper bounds to lightgraphs versions

sbromberger:
per https://github.com/JuliaGraphs/LightGraphs.jl/issues/573

cc @tkelman

### DIFF
--- a/LightGraphs/versions/0.1.0/requires
+++ b/LightGraphs/versions/0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.2.6
 GZip 0.2
 StatsBase 0.6 0.9.0
--- a/LightGraphs/versions/0.1.1/requires
+++ b/LightGraphs/versions/0.1.1/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.2.6
 GZip 0.2
 StatsBase 0.6 0.9.0
--- a/LightGraphs/versions/0.1.10/requires
+++ b/LightGraphs/versions/0.1.10/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.2.6
 GZip 0.2
 StatsBase 0.6 0.9.0
--- a/LightGraphs/versions/0.1.11/requires
+++ b/LightGraphs/versions/0.1.11/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.4
 GZip 0.2
 Distributions 0.6.6
--- a/LightGraphs/versions/0.1.12/requires
+++ b/LightGraphs/versions/0.1.12/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.4
 GZip 0.2
 Distributions 0.6.6
--- a/LightGraphs/versions/0.1.13/requires
+++ b/LightGraphs/versions/0.1.13/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.4
 GZip 0.2
 Distributions 0.6.6
--- a/LightGraphs/versions/0.1.14/requires
+++ b/LightGraphs/versions/0.1.14/requires
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.3 0.4
 Compat 0.4
 GZip 0.2
 Distributions 0.6.6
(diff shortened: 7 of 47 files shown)


sbromberger:
Not sure these errors are related to LG.

tkelman:
TrafficAssignment is. It has upper bounds on its LightGraphs dependency, the bounds you're adding here make the intersection empty and infeasible.

sbromberger:
TA v0.5.0 doesn't even require LG but it's in the failure log. How do we resolve this? There should be no gaps in our coverage anywhere.

tkelman:
Ref https://github.com/JuliaLang/METADATA.jl/pull/7613, the artificial dependencies were added there to prevent Pkg from downgrading TrafficAssignment in order to upgrade LightGraphs

sbromberger:
How can I fix this?

tkelman:
Need looser bounds. TrafficAssignment has a LightGraphs 0.7 upper bound, so at least one version of LightGraphs < 0.7 needs to remain installable on all Julia versions that TrafficAssignment supports.

sbromberger:
Not sure I understand. LightGraphs supports all Julia versions between 0.2 and 0.6. Here are the cutoffs:

| LG Version | Lower Bound | Upper Bound |
|---|---|---|
|0.1.0 through 0.3.7|0.3|0.4|
|0.4.0 through 0.6.3|0.4|0.5|
|0.7.0 through 0.7.2|0.5-|0.6|
|0.7.3 through 0.7.5|0.5|0.6|
|0.8.0|0.6-|

With what version of Julia is LG not compatible?

tkelman:
TrafficAssignment's early tags do not have Julia upper bounds, so they claim to support 0.4, 0.5, 0.6, and in the first few, also 0.3. If you upper bound all LG < 0.7, then those TrafficAssignment versions are no longer installable on the versions of julia above your upper bound. Those early versions of TrafficAssignment are the only ones that claim to support julia 0.6, so these LG bounds make that package no longer installable on 0.6 at all.

sbromberger:
Well, I don't think that's a problem LightGraphs should solve. Even if we were to solve it, the early versions of TrafficAssignment won't work in any case, since LightGraphs <= 0.7 won't work in Julia 0.6.

Can we get TrafficAssignment to correctly specify the version of Julia/LightGraphs it requires?

tkelman:
We can do something. Either adding Julia upper bounds to the versions of TrafficAssignment that currently don't have any, which would effectively deprecate the package and make it not installable on 0.6 until a new version of it gets tagged. Or we could loosen some of the bounds on LightGraphs, if the compatibility range is actually wider than what the bounds currently say.

sbromberger:
> Or we could loosen some of its bounds on LightGraphs, if the compatibility range is actually wider than what the bounds currently say.

It's not, as far as I'm aware. We've moved to the minor versions when we've stopped support for a specific version of Julia.

sbromberger:
@tkelman it seems to me that placing artificial constraints on TrafficAssignments may not have been the right way of handling this, as it's now preventing us from putting correct bounds on LightGraphs. What can we do to merge this PR?

sbromberger:
Also: if you look at the `REQUIRE` file in the TrafficAssignment.jl repo, LightGraphs wasn't even required until v0.2.0. It appears they retroactively required LightGraphs for all versions going back to the beginning. This is a problem.

tkelman:
It was necessary to avoid downgrading TrafficAssignment in order to upgrade LightGraphs. Having both installed should prefer keeping LightGraphs back. If there's another way of accomplishing that, we can adjust things to do so.

sbromberger:
> Having both installed should prefer keeping LightGraphs back.

OK, but this is now preventing LightGraphs from tagging appropriate upper bounds, right? Forgive me since many of the subtleties around versioning are still a mystery to me. I still don't know how we resolve the situation, and it appears that it became an issue because TrafficAssignment is specifying bounds for LightGraphs that we know won't work.

sbromberger:
Is it possible to move forward with this? I'd really like to get the upper bounds in on LG prior to v0.9.0. Thanks.

tkelman:
Would need to try alternate solutions to accomplish what TrafficAssignment set the bounds to do, which is avoid letting TrafficAssignment be downgraded to very old versions in order to upgrade LightGraphs.

sbromberger:
I'm not sure what problem we're actually trying to solve here, so closing this out.

chkwon:
Sorry that old versions of TA weren't set properly up. It seems that it would be best to make a new TA version that works with Julia 0.6 and LightGraphs 0.8, without upper bounds?  If you agree,  I will work on it as soon as possible.

chkwon:
@tkelman @sbromberger 

I'll make a new version of TrafficAssignment with the following `REQUIRE`:

```
julia 0.6-
LightGraphs 0.8
Optim 0.7
```

Would this be okay?

PR #8820, opened 2017-04-12
Bound JSON on packages that used deprecated _print

TotalVerb:
See #8798.

I obtained the list of packages from https://github.com/JuliaIO/JSON.jl/issues/158. The maximum tag with the bound in each case is:

- Plots 0.9.0
- Escher 0.3.2
- Plotly 0.3.1 (note versions 0.2.0 are earlier were already bounded by 0.6.0)
- Bokeh 0.2.0 (note this is the latest tag; the problem is fixed on master but there has been no tag since for this unmaintained package)

Plots is special because it never depended on JSON; instead it used it through its conditional dependency Plotly. I think it is not harmful to add an unnecessary JSON dependency to old versions of Plots.

### DIFF
--- a/Bokeh/versions/0.0.1/requires
+++ b/Bokeh/versions/0.0.1/requires
@@ -1,2 +1,2 @@
 Mustache
-JSON
+JSON 0.0 0.9
--- a/Bokeh/versions/0.1.0/requires
+++ b/Bokeh/versions/0.1.0/requires
@@ -1,5 +1,5 @@
 julia 0.3
 Mustache
-JSON 0.4
+JSON 0.4 0.9
 Dates 0.3.2
 Compat
--- a/Bokeh/versions/0.2.0/requires
+++ b/Bokeh/versions/0.2.0/requires
@@ -1,5 +1,5 @@
 julia 0.3
 Mustache
-JSON 0.4
+JSON 0.4 0.9
 Dates 0.3.2
 Compat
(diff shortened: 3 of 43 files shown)


tkelman:
Thanks, you rock.

Looks like Plots.jl prior to https://github.com/JuliaPlots/Plots.jl/commit/d38803660181cf38efb7ca06171ad38f4803abe8 (released in 0.5.2) wouldn't technically need this, but probably better to hold back JSON than downgrade Plots if someone asked for both JSON and some version of Plots between 0.5.2 and 0.9.0.

@tbreloff any objections to what this does?

tkelman:
Looks like Tom's been offline for a couple weeks. Don't think this should hurt anything, especially as most people are hopefully using 0.10.x by now anyway.

tkelman:
LanguageServer.jl was using `_writejson` and could use the same treatment, cc @ZacLN
https://github.com/JuliaCI/pkg.julialang.org/blob/e67ffedaeab9edd240cd1a84f4d8352eef9fff86/logs/LanguageServer_0.5.log

TotalVerb:
See https://github.com/JuliaEditorSupport/LanguageServer.jl/pull/51 for a fix, and https://github.com/JuliaLang/METADATA.jl/pull/8846 for the related METADATA PR.

It's unfortunate that internal details of JSON are so frequently relied upon even when preceded by an underscore. Hopefully improvements to the documentation will help.

tkelman:
Agreed, though with 344 dependents, and not a lot of language-level support keeping you from foot gunning, bound to happen. Just glad PkgEval finally went through for the first time in a week, and is able to flag this kind of thing. (And that PkgEval itself wasn't using any of the internal bits of JSON that changed here.)

PR #8835, opened 2017-04-13
Tag DASKR v0.1.0 [https://github.com/JuliaDiffEq/DASKR.jl]

ChrisRackauckas:


DASKR/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+DiffEqBase 0.15.0
+Compat 0.9.0
```
tkelman: does this even need to be using Compat right now?

ChrisRackauckas: Looks like it doesn't.

PR #8847, opened 2017-04-14
Tag LSODA v0.1.0 [https://github.com/rveltz/LSODA.jl]

ChrisRackauckas:
@rveltz

### DIFF
--- a/LSODA/versions/0.1.0/requires
+++ b/LSODA/versions/0.1.0/requires
@@ -1,0 +1,5 @@
+julia 0.5
+BinDeps 0.4.3
+Compat 0.9.0
+Parameters 0.5.0
+DiffEqBase 0.15.0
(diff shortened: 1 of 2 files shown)


LSODA/versions/0.1.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+BinDeps 0.4.3
+Compat 0.9.0
```
tkelman: seems like Compat isn't actually being used for anything at the moment?

ChrisRackauckas: Yeah, I'll setup Compat on the abstract types right now though.

ChrisRackauckas:
Superseded by https://github.com/JuliaLang/METADATA.jl/pull/8849

PR #8878, opened 2017-04-16
Register new package IterableTables.jl v0.0.1

attobot:
Repository: [davidanthoff/IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl)
Release: [v0.0.1](https://github.com/davidanthoff/IterableTables.jl/releases/tag/v0.0.1)
cc: @davidanthoff


davidanthoff:
This is essentially a lot of stuff moved from [Query.jl](https://github.com/davidanthoff/Query.jl). Once ``IterableTables`` is registered and tagged, I'll tag the corresponding Query.jl release as well.

tkelman:
FlatBuffers and PooledArrays are missing from test/REQUIRE

davidanthoff:
> FlatBuffers and PooledArrays are missing from test/REQUIRE

Done. I actually only have explicit ``using`` statements for them to work around this [bug](https://github.com/MikeInnes/Requires.jl/issues/30). And if someone has an idea how to get @MikeInnes to fix that bug, I'd be grateful :)

davidanthoff:
Actually, now that the Requires bug is fixed, let me update the tag here once more and remove some dependencies that we don't need explicitly. Will be done in a minute.

davidanthoff:
Ok, I've shortened both the ``using`` and the ``REQUIRE`` list, ready to be merged from my side.

davidanthoff:
@tkelman travis seems stuck? It has looked like this for the last 10 hours. Is there something you can do about that?

tkelman:
Can this package work without overwriting other packages' methods? That's much more direct than the usual type piracy. Whether or not this package has been loaded shouldn't dramatically change the behavior of otherwise identical code.

davidanthoff:
@tkelman I don't think the package overwrites any method at this point, it just adds methods that are not defined in the packages. It is still clearly type piracy. The goal is to move these methods (and all integrations) into the packages that are being integrated (davidanthoff/IterableTables.jl#20) once things have stabilized, and at that point there won't be any type piracy anymore.

I know this is not ideal, but I think it is the most practical way... I hope to resolve all of this (i.e. move things out and get rid of all type piracy) by juliacon.

One thing I could do in the meantime is to just spit out a warning at load time for this package that says something like ``WARNING: The package IterableTables.jl was just loaded. This package might modify the behavior of other packages." or something like that?

davidanthoff:
@tkelman what is the next step here?

tkelman:
`DataFrame(::Any)` is a method overwrite, as are several others. What's the path from having the overwrites to not having them? You may have a hard time convincing maintainers of that many packages to add an indirect dependency on Requires.

Is this proof of concept version something you expect people to rely on and use? Or would it be cleaner to put the temporary parts on a branch or something for experimentation purposes?

davidanthoff:
> DataFrame(::Any) is a method overwrite, as are several others.

I don't think that is correct, I don't think there are any method overrides in the package.

I'm undecided between two paths for the future migration, but neither would require folks to take a dependency on ``Requires`` (because I don't want to inflict that pain on anyone).

Option 1 is that once things have stabilized, I will remove *all* integrations from ``IterableTables`` (and the dependency on ``Requires``), open PRs in all the various repos with the integration code that moves that code into these packages. Those PRs would essentially come with a message a la "I will remove all integrations from ``IterableTables`` on date X, at that point your package will only be integrated with ``IterableTables`` and ``Query`` if you merge this PR". I would try to rope maintainers of these packages in well in advance so that this comes across more friendly :)

Option 2 is that I create another package ``IterableTablesBase`` that has no dependency on ``Requires`` and no integration code. The migration of integration code would then essentially be that a package takes a dependency on ``IterableTablesBase`` and moves the integration code from ``IterableTables`` into the package. At the end of that process ``IterableTables`` would be empty.

I guess a third option would be to have ``IterableTables`` only hold the core definitions but no integrations (and no dependency on ``Requires``), and have another package ``IterableTablesIntegrations`` that holds the various integrations until they are moved to the respective packages. With that option we would end up with an empty ``IterableTablesIntegrations`` package at the end.

If you have another idea, I'd really welcome it. I've been thinking how to make this smooth, and all of these seem somewhat involved...

> Is this proof of concept version something you expect people to rely on and use? Or would it be cleaner to put the temporary parts on a branch or something for experimentation purposes?

No proof of concept, this is real stuff that I think would be super useful to lots of folks. The next version of Query also completely relies on this.

davidanthoff:
@tkelman Bump, any chance we can move this forward?

tkelman:
NullableArrays is being used in a test without being explicitly depended on

davidanthoff:
> NullableArrays, DataStreams, and others are being used in tests without being explicitly depended on

I think I fixed all of these now.

tkelman:
> I guess a third option would be to have IterableTables only hold the core definitions but no integrations

There's a lot of desire for something to the effect of a lightweight abstract tables interface package right now. Separating out the integrations, even to temporary unregistered package[s] for a little while, would make the interface part here simpler to adopt right away for this purpose. The timeline for each integration is going to depend on specific pairwise package maintenance, so lumping them all together in one place and putting them on the same timeline seems more harmful than helpful.

PR #8881, opened 2017-04-16
Update requires to guard against CxxWrap API change

jstrube:


tkelman:
could you do these all in a single PR so it's possible to look at the changes all at once? just make multiple edits within a single branch

jstrube:
Wasn't sure what level of granularity was preferred.
I've edited all versions requiring julia 0.5 in this request and closed the others.

tkelman:
Thanks, this is preferable for requires changes. Ref

> be aware that for upper bounds to be effective, they'll likely need to be retroactively added to past versions' require files in metadata - otherwise Pkg could choose to downgrade LCIO in order to upgrade CxxWrap

this also applies to the earlier tags

jstrube:
OK. I was mis-interpreting the `julia 0.4` req as `julia 0.4 0.5-`, and CxxWrap 0.1.5 was the last one compatible with julia 0.4, so I thought that was safe, but of course I was wrong.
"Fixed" now.
What's good practice for removing outdated releases?

tkelman:
I'd avoid the trailing dash on the upper bounds, since upper bounds are exclusive and that would rule out the very first prerelease after 0.4 branched.

jstrube:
ok. makes sense.
Thanks for the careful checking. It's reassuring to know how much quality checking the julia packages go through.

PR #8884, opened 2017-04-16
move OffsetArrays to JuliaArrays

Evizero:
https://github.com/JuliaArrays/OffsetArrays.jl

This needs confirmation from @timholy @alsam, because I noted that this wasn't a transfer per se, more of a "manual" fork, so the original repository is still at alsam/OffsetArrays.jl. (was this on purpose?) Anyway, this means that there is no automatic forwarding to the the new home at JuliaArrays, and the link in METADATA has to be replaced for the move to take effect

ref: https://github.com/alsam/OffsetArrays.jl/pull/26#issuecomment-294231805

### DIFF
--- a/OffsetArrays/url
+++ b/OffsetArrays/url
@@ -1,1 +1,1 @@
-git://github.com/alsam/OffsetArrays.jl.git
+git://github.com/JuliaArrays/OffsetArrays.jl.git


tkelman:
Doing it in this way is bad, since people's remotes will not automatically change. Please delete the current JuliaArrays/OffsetArrays.jl copy, and actually do a "Transfer repository" otherwise github will not do redirects and no one who has previously installed the package would see new content.

alsam:
ok @tkelman just killed https://github.com/JuliaArrays/OffsetArrays.jl but looks transfer doesn't work for me - don't have permissions for the operation only for import.

tkelman:
Someone needs to give you admin privileges on JuliaArrays, at least to do the transfer

alsam:
ok thanks @tkelman just have wrote to https://github.com/alsam/OffsetArrays.jl/pull/26

tkelman:
this time worked

PR #8902, opened 2017-04-18
Update requires

JeffreySarnoff:


tkelman:
This looks right. Just to be clear, FastRounding 0.0.2 would not work at all with ErrorfreeArithmetic 0.0.1 or 0.0.2?

JeffreySarnoff:
FastRounding 0.0.2 would not work *correctly* with ErrorfreeArithmetic 0.0.1 or 0.0.2.

PR #8918, opened 2017-04-19
Tag OdsIO v0.1.0 [https://github.com/sylvaticus/OdsIO.jl]

sylvaticus:
Main feature: Added ods_write() to "complete" the package with data export to ods.


tkelman:
DataStructures is missing from REQUIRE

sylvaticus:
Thank you... I edited my local copy of the REQUIRE file, committed and pushed it to GitHub.

Which is now the suggested workflow ? Should I make a new tag e.g. `PkgDev.tag(pkg, v"0.1.1")` and publish it, or delete that tag and tag 0.1 the last commit ?

tkelman:
Would be best to fix on the 0.1.0 tag. Don't necessarily need to modify the sha or git tag though, you could just modify the requires file in place on this branch of your metadata fork.

For next time, I recommend using https://github.com/attobot/attobot

PR #8927, opened 2017-04-20
Tag AWS* and deps

samoconnor:
AWSSNS - add SMS sending
Misc v0.6 tweaks.
Fix https://github.com/samoconnor/SymDict.jl/issues/2


AWSSQS/versions/0.1.4/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.5
+AWSCore 0.1
+SymDict
```
tkelman: looks like this needs 0.1.2 for the lowercase function name

tkelman:
Evidently the git tags don't match the sha values any more

samoconnor:
@tkelman, do you understand why travis is failing now?

tkelman:
You're not allowed to modify the sha of existing tags.

samoconnor:
got it, fixed

PR #8948, opened 2017-04-21
Register COESA [https://github.com/danielmatz/COESA.jl.git]

danielmatz:
Repo: https://github.com/danielmatz/COESA.jl.git


tkelman:
Funny, I recognize the numbers from having done aerospace simulation work a long time ago, but knew it by a different acronym. Maybe it would be more descriptive to call this StandardAtmosphere.jl if it could conceivably grow to have the options of multiple model variants eventually?

danielmatz:
Actually, I did originally call it StandardAtmospheres.jl with that intention, but decided to just focus on this one model.  I figured I'd just make separate packages for any other models that I get around to implementing.

I went with COESA for the name because Matlab's corresponding function is [atmoscoesa](https://www.mathworks.com/help/aerotbx/ug/atmoscoesa.html).

PR #8952, opened 2017-04-22
tag FixedEffectModels 0.3.1

matthieugomez:


FixedEffectModels/versions/0.3.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5
+Compat
```
tkelman: should list a minimum version that implements the features you need

PR #8955, opened 2017-04-22
Tag DynamicalBilliards.jl v1.1.0

attobot:
Repository: [Datseris/DynamicalBilliards.jl](https://github.com/Datseris/DynamicalBilliards.jl)
Release: [v1.1.0](https://github.com/Datseris/DynamicalBilliards.jl/releases/tag/v1.1.0)
Travis: [Travis Build Status](https://travis-ci.org/Datseris/DynamicalBilliards.jl/builds/224739634)
Diff: [vs v1.0.0](https://github.com/Datseris/DynamicalBilliards.jl/compare/672037455e605854323976a40e57eefbee6cf3d9...9a0c5d8c8a60405899eef92aa19e1c83807d5b16)
`requires` vs v1.0.0: 
```diff
--- 1.0.0/requires
+++ 1.1.0/requires
@@ -1,3 +1,2 @@
-julia 0.5
-StaticArrays
-PyPlot
+julia 0.5
+StaticArrays
```
cc: @Datseris


Datseris:
`PyPlot` is being removed from REQUIRE because all plotting functionality was moved into a sister package `DynamicalBilliardsPlotting` (soon to be submitted to METADATA)

aytekinar:
There seems to be a problem in `METADATA` related to this package, @tkelman.

In my systems (Arch Linux), I get problems when `Pkg.update()`ing in `julia 0.4.7`. When I `git status`, I see `DynamicalBilliardsPlotting/versions/0.1.0/requires` as modified. Then I run `git diff`, I see this
```diff
warning: CRLF will be replaced by LF in DynamicalBilliardsPlotting/versions/0.1.0/requires.
The file will have its original line endings in your working directory.
diff --git a/DynamicalBilliardsPlotting/versions/0.1.0/requires b/DynamicalBilliardsPlotting/versions/0.1.0/requires
index 7b0122153..9de6e804c 100644
--- a/DynamicalBilliardsPlotting/versions/0.1.0/requires
+++ b/DynamicalBilliardsPlotting/versions/0.1.0/requires
@@ -1,3 +1,3 @@
-julia 0.5
-PyPlot
-DynamicalBilliards
+julia 0.5
+PyPlot
+DynamicalBilliards
```

Is this a problem concerning only me, or is this a known issue?

Datseris:
That sounds unsettling! However, I do not have this issue and I got the package installed as well..
`Pkg.update()` throws errors sometimes recently, but people in Gitter say it has something to do with the updater, not the package itself...

aytekinar:
Sorry, @Datseris ! I didn't mean to offend you, nor this package, by the way :) Thank you for the explanation. I have had this problem before, but `git reset --hard HEAD~10` and `git fetch --all --prune` followed by `git merge --ff-only origin/metadata-v2` solved my problems before.

I am on the terminal right now, and I run these commands directly in the `~/.julia/v0.4/METADATA` folder. I have not been able to solve it myself, even by using `dos2unix <file>` to convert `CRLF`s to `LF`s.

Apparently, this is a common issue :(

Datseris:
you did not offend me in any way!!!
In the end of the day, if my package has problems I am definitely the one at fault, so I shouldn't be getting offended in the first place.

I am not sure whether your issue is a common issue though, don't take my word for granted! I am a newbie after all... I just said what I heard in gitter.

tkelman:
It's because @Datseris has an unusual git config setting for line endings. Normally git shouldn't commit files with windows line endings, but these last couple tags from @Datseris have them. We should actually be checking and enforcing to prevent that from happening.

tkelman:
See also https://github.com/JuliaLang/METADATA.jl/pull/8964

Datseris:
So it **is** my fault. Well that sounds bad, how can I change this so that It doesn't happen again?

I haven't changed my git config settings ever, so I thought I was using "default" settings...

aytekinar:
@Datseris something like [this](https://help.github.com/articles/dealing-with-line-endings/) should help you. the problem is, under windows, line endings are treated as CR+LF (carriage return + line feed), whereas in *NIX systems, it is LF (line feed, only). That's the reason why I see your files as `modified`, even though it is only the line endings in 3 of the lines.

Thank you, @tkelman, for your prompt response!

Datseris:
Okay, thank you. I did what the guide said. Funny enough, my `git config core.autocrlf` was already set to `true`. I hope the command `--global` made the magic happen.

tkelman:
Actually since this came from Attobot we can partially fix it there. It's copying the REQUIRE file as-is, including line endings. We should have it normalize to unix line endings just in case. Will open an issue on attobot.

PR #8963, opened 2017-04-23
Tag Gadfly v0.6.1 [https://github.com/GiovineItalia/Gadfly.jl]

bjarthur:
1. fixed line_style
2. fix dep warns for julia 0.6
3. fix normalization of histogram
4. lots of fixes for Date aesthetics
5. simplified Geom.abline interface
6. fix jitter for categorical data
7. constructor refactoring


Gadfly/versions/0.6.1/requires
```diff
@@ -0,0 +1,16 @@
+julia 0.5
+Colors 0.3.4
+Compat 0.8.5
```
tkelman: need minimum of 0.18 for parametric aliases

tkelman: need to update the metadata requires file for this

ararslan:
I edited the Compat version in this PR to match @bjarthur's update to the package. Does 0.6.1 need to be retagged to include the Compat change, @tkelman?

bjarthur:
what would be the consequences of simply merging this PR without the second commit that updates the Compat requirement?  Gadfly's dependency on Compose ensures the correct version of Compat will be used.  and the commit to Gadfly post tag v0.6.1 means it will be fixed eventually.

ararslan:
> what would be the consequences of simply merging this PR without the second commit that updates the Compat requirement?

Probably nothing

tkelman:
Tightening requirements needs to be done in place, a future tag with tighter requirements can't fix the existence of a past tag with requirements that are too loose. It's not guaranteed that all future versions of Compose will have a sufficiently high minimum Compat requirement (they may at some point not depend on Compat at all).

ararslan:
Okay, so @bjarthur should delete the v0.6.1 tag and retag it?

tkelman:
not necessarily, having the commit here is equivalent since metadata's require file takes priority for tagged versions of registered packages

ararslan:
Okay, so this is fine as-is?

tkelman:
This caused a BoundsError in Mads.jl https://github.com/JuliaCI/pkg.julialang.org/blob/ec1913d13f8bff8600af1fcef7431cb022e947eb/logs/Mads_0.5.log
cc @montyvesselinov

montyvesselinov:
Thank you very much for checking in all of these. I also noticed this
failure.

-monty

montyvesselinov:
Is this going to be fixed soon?

tkelman:
You should open an issue on Gadfly with a small self contained test case

PR #8966, opened 2017-04-23
Tag ArgParse v0.5.0 [https://github.com/carlobaldassi/ArgParse.jl]

carlobaldassi:
Mostly, add support for julia 0.6 and drop julia 0.4.

Previous versions would not load on julia 0.6, so I added an upper bound to all of them.


carlobaldassi:
Fixed the Compat issue and updated this PR.

PR #8967, opened 2017-04-23
Pull request/4c4c9dce

tawheeler:
Fixed 0.6 deprecation warnings.


tkelman:
There has been a PGFPlots 1.5.0, so this wouldn't usually get installed as far as I can tell

tawheeler:
Hmm. Not sure why the old stuff is in there or why PGFPlots is behind. I used PkgDev.tag(PGFPlots, :minor)...
Will re-pull METADATA

PR #8972, opened 2017-04-24
Tag ERFA v0.2.1 [https://github.com/JuliaAstro/ERFA.jl]

nirinA:


tkelman:
If `using Compat` is still present in `deps/build.jl`, it should still be listed in `REQUIRE`

nirinA:
thank you for the review. `deps/build.jl` is corrected, we removed `Compat`.

PR #8976, opened 2017-04-24
Register CoupledFields [https://github.com/Mattriks/CoupledFields.jl.…

Mattriks:
See https://github.com/GiovineItalia/Gadfly.jl/issues/791#issuecomment-292705828 for an example of what `CoupledFields.jl` can do!


CoupledFields/versions/0.0.1/requires
```diff
@@ -0,0 +1,3 @@
+Compat 0.8.5
```
tkelman: abstract type requires 0.17

Mattriks:
I found the small section on ["Fixing Package Requirements"](https://docs.julialang.org/en/latest/manual/packages.html#Fixing-Package-Requirements-1), but the info is a bit sparse for me to know if I've done this correctly.  I will now fix the Require file in the `CoupledFields` repo.

tkelman:
are you using parametric aliases or Compat.Iterators? Why 0.18 instead of 0.17?

Mattriks:
Not currently, but the idea here is to add `CoupledFields` to `Gadfly` (see [comment](https://github.com/GiovineItalia/Gadfly.jl/issues/791#issuecomment-292680904)) so I made the Require file the same as is in the [current master](https://github.com/GiovineItalia/Gadfly.jl/blob/a691fd5f88e167e480567fefe4ed2cb6cb123299/REQUIRE#L3) of Gadfly.

tkelman:
That's unnecessary, packages' dependencies are independent and should reflect what they use themselves.

Mattriks:
I thought about adding a parametric alias or a Compat.Iterators to `CoupledFields` 😃 , but instead I changed the require file to Compat 0.17.0.

PR #8982, opened 2017-04-24
Tag GLVisualize.jl v0.2.3

attobot:
Repository: [JuliaGL/GLVisualize.jl](https://github.com/JuliaGL/GLVisualize.jl)
Release: [v0.2.3](https://github.com/JuliaGL/GLVisualize.jl/releases/tag/v0.2.3)
Diff: [vs v0.2.2](https://github.com/JuliaGL/GLVisualize.jl/compare/26b34816b7762f12bc2f885fa37f1aa6d9717c04...fcf8a555562c3e42ca0b5bf25869e46425aed1da)
`requires` vs v0.2.2: 
```diff
--- 0.2.2/requires
+++ 0.2.3/requires
@@ -25,3 +25,5 @@
 Iterators
 Packing
 UnicodeFun
+
+Compat 0.18
```
cc: @SimonDanisch


GLVisualize/versions/0.2.3/requires
```diff
@@ -0,0 +1,29 @@
...
+FileIO
+MeshIO
+
+SignedDistanceFields
+FreeType
+Iterators
+Packing
+UnicodeFun
+
+Compat 0.18
```
tkelman: this was already listed above but at 0.9.5, should modify the existing minimum instead of listing it twice

SimonDanisch:
Woops...

SimonDanisch:
seems like I need to restrict the reactive version!

GLVisualize/versions/0.2.3/requires
```diff
@@ -0,0 +1,28 @@
+julia 0.5
+
+GLWindow 0.3
+GLAbstraction 0.3.1
+GeometryTypes 0.2
+
+GLFW
+ModernGL
+FixedSizeArrays
+Reactive 0.3.3 0.3.7
```
tkelman: does this upper bound need to apply to all past versions?

SimonDanisch: What do you mean? But I think the range is indeed wrong, since it won't include 0.3.7, no?
GLVisualize works with 0.3.3 - 0.3.7 right now...

tkelman: Upper bounds are exclusive. And what I mean is if you have an upper bound only on this newest version of GLVisualize, Pkg would be allowed to keep GLVisualize held back to 0.2.2 or earlier and upgrade Reactive to the latest. Would things work if it did that?

SimonDanisch: Oh... No it wouldn't.

SimonDanisch: What can be done about this?

tkelman: If the latest Reactive version broke all past tags of GLVisualize, then best thing to do is open a separate PR that adds Reactive upper bounds to all the past GLVisualize tags' require files. Don't touch the sha or the git tags, just the metadata copies of the require files.

SimonDanisch: There was so much restructuring going on with Reactive, that I didn't realize that it was simply a bug: https://github.com/JuliaGizmos/Reactive.jl/pull/136
So I guess this moves the problem to tagging reactive? We need to switch out the bugged tag with the bugfix tag?

tkelman: If prior versions work, and newer versions are going to work after the bugfix gets tagged, then I don't think you should have bounds. You would want to avoid the particular single bugged version, and that is possible to express with bounds but it's a bit unintuitive and not many people have been doing disjoint ranges.

SimonDanisch: that's why I asked if it's appropriate to just rewire Reactive's tag, so that the bugged version isn't in METADATA anymore...

tkelman: The buggy tag has been published for a while now, it should be superseded with a new tag but not modified in-place.

SimonDanisch:
Okay, so there shouldn't be any harm to merge this...
Now I need to open a PR to change the other tags as well, right?
How far should I go back?

GLVisualize/versions/0.2.3/requires
```diff
@@ -0,0 +1,28 @@
+julia 0.5
+
+GLWindow 0.3
+GLAbstraction 0.3.1
+GeometryTypes 0.2
+
+GLFW
+ModernGL
+FixedSizeArrays
+Reactive 0.4.1
```
tkelman: do 0.3.x versions work?

> If prior versions work, and newer versions are going to work after the bugfix gets tagged, then I don't think you should have stricter bounds for it.

SimonDanisch: Not all!

tkelman: what's the minimum version then? what feature are you relying on? or bugfix before which **none** of the releases would work?

tkelman: If there are versions prior to 0.3.7 that would work, then you shouldn't rule out installation of them.

SimonDanisch:
@timholy supplied that code.. I believe the rational behind it was, that these packages are not needed for the old images, but when Images uses them, it will also require them. 
Not sure if putting them directly in the REQUIRE will mess with the old images

tkelman:
> when Images uses them, it will also require them

That's only enforceable if it's guaranteed that all allowed Images versions do so. With no upper bound, there's no way to ensure that.

SimonDanisch:
Well, it doesn't need to, since all code that is using this is in a similar conditional. I believe @timholy used this as a switch to verify if one is one the new version of Images that works with those packages.

MikeInnes:
Please check that:
- [ ] The package's CI is green for this tag (if applicable).
- [ ] The version bounds are up to date.

Then reply with `+ready` so I can merge.

tkelman:
specifically w.r.t Reactive,

> If there are versions prior to 0.3.7 that would work, then you shouldn't rule out installation of them.

SimonDanisch:
So how do I do a version range from `0.3.0 - 0.3.7 &&  0.4.0` ?

tkelman:
`Reactive 0.3.0 0.3.7 0.4.0`

SimonDanisch:
Thanks! ... Too easy. This is actually even pretty clearly stated in the docs...

SimonDanisch:
should be closed in favor of: https://github.com/JuliaLang/METADATA.jl/pull/9436

PR #8986, opened 2017-04-24
Register & Tag POMDPToolbox and POMDPModels (Pull request/7097886d)

zsunberg:
POMDPToolbox includes a lot of functionality that is not part of the core POMDPs.jl interface, but makes working with it much easier. POMDPModels is a collection of simple examples. It is needed to test POMDPToolbox. The tests for these packages are failing right now because they REQUIRE each other, but they should pass as soon as they are registered.


tkelman:
POMDPModels is missing a minimum julia version requirement

tkelman:
And its `gridworld.jl` test that runs the notebook depends on an unregistered package, GenerativeModels

tkelman:
StatsBase and Distributions are missing from POMDPToolbox's REQUIRE, and it has the same unregistered test dependency issue with GenerativeModels

zsunberg:
Ok, I think I fixed everything you mentioned. I apologize for the sloppiness of the original commit. Are there automatic tools for helping to catch sloppy errors like that, or do you just grep for "import", etc.? Are there quality-checking tools in attobot?

zsunberg:
From the travis log, it appears that I can't fix the lack of a julia version... starting over in a new PR

PR #8999, opened 2017-04-25
Tag StaticArrays.jl v0.5.0

attobot:
Repository: [JuliaArrays/StaticArrays.jl](https://github.com/JuliaArrays/StaticArrays.jl)
Release: [v0.5.0](https://github.com/JuliaArrays/StaticArrays.jl/releases/tag/v0.5.0)
Diff: [vs v0.4.0](https://github.com/JuliaArrays/StaticArrays.jl/compare/a2eaeb427211980ae2f595c4416c3629747095cb...aa5606993863b90dbdbe341443ae4d00d3f0e495)
cc: @c42f


jrevels:
The Travis failure is quite unfortunate given that the only dependency listed in the REQUIRE here is `julia 0.6-`...

tkoolen:
Possibly related to the Travis failure, should this PR incorporate the upper bound changes for previous versions as discussed in https://github.com/JuliaLang/METADATA.jl/pull/8793?

c42f:
Yikes, it took hours for me to reproduce this locally.  I suspect it all came down to getting the exact status of the METADATA repo right cf https://github.com/JuliaLang/METADATA.jl/blob/cb7816bcc0efebce60deffbcb0df1f90d6914665/.test/ci.sh#L9

@tkoolen It could be this, I don't really understand what went wrong yet.  I should probably fix those anyway - will need to open a new PR manually though, I don't think I can modify the attobot one.

c42f:
I spent quite some time nailing down the upper bounds for StaticArrays, but this only made things slightly worse.

In all, I can't see how the breakage here is due to this particular StaticArrays release.  I more suspect 
this is fallout from https://github.com/JuliaLang/julia/pull/21485

c42f:
But then again, other PRs should be seeing the same issues if that was the case.  In all, very confusing.

c42f:
@tkelman Running the METADATA.jl tests locally, the latest julia master (built locally) works fine, but the one I downloaded to match the CI run which failed does not.  Can you restart CI please?

c42f:
Ok, it seems the the somewhat complex dependencies of Geodesy, Rotations, CoordinateTransformations and StaticArrays may have brought out a bug in Pkg.resolve, which has since been fixed in julia master, helped along by some upper bound patching in https://github.com/JuliaLang/julia/issues/21430

andyferris:
I'm late to the party here, but what is the status here? Can I help somehow?

c42f:
As far as I can tell, there's nothing else we can do until the julia version used by METADATA CI is upgraded - perhaps have to wait until the next beta (See https://github.com/JuliaLang/METADATA.jl/blob/ae47b51d684fca0c5ced7ac102748f902f896952/.test/ci.sh#L21)

The latest version of julia master built locally doesn't produce the errors we're seeing here.

andyferris:
OK, can we merge this for the meanwhile? I think there is sufficient demand.

tkelman:
Merging this would then cause every other PR to metadata to fail on Travis.

andyferris:
I see... How can we solve this? (Unless I misunderstand something, it appears that the solution lies somewhat outside the scope of this PR...).

tkelman:
#9013 worked

tkelman:
This caused new failures in
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/CompScienceMeshes_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/ConcaveHull_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/IntervalArithmetic_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/IntervalConstraintProgramming_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/IntervalRootFinding_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/NearestNeighbors_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/QuantumInfo_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/QuantumTomography_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/RandomQuantum_0.6.log
- https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/ValidatedNumerics_0.6.log

KristofferC:
Seems like many of these are from the same failure:

```
WARNING: An error occurred during inference. Type inference is now partially disabled.
TypeError(func=:apply_type, context="Vararg count", expected=Int64, got=Any)
```

tkoolen:
See also https://github.com/FugroRoames/Rotations.jl/pull/32, looks like
the same failure.

martinholters:
```julia
julia> Base._methods_by_ftype(Tuple{Type{StaticArrays.SArray{Tuple{L}, T, 1, L}} where T where L, Any}, 4, typemax(UInt))
ERROR: TypeError: apply_type: in Vararg count, expected Int64, got Type{Any}
Stacktrace:
 [1] _methods_by_ftype(::Any, ::Int64, ::UInt64, ::Array{UInt64,1}, ::Array{UInt64,1}) at ./reflection.jl:521
 [2] _methods_by_ftype(::Any, ::Int64, ::UInt64) at ./reflection.jl:507
```

tkelman:
thanks for reducing that @martinholters

c42f:
Hah, I put this in the too hard basket last night, but this morning it looks like @martinholters has already come to the rescue in #21614.  Thanks so much for looking into this, it's very much appreciated!

martinholters:
The remaining failures in ValidatedNumerics and IntervalArithmetic are: "invalid subtyping in definition of IntervalBox" because `StaticVector` now expects two type parameters. So I guess these need upper bounds on StaticArrays and new versions tagged with that fixed.
Failures in IntervalConstraintProgramming and  IntervalRootFinding are simply caused by them using IntervalArithmetic.

PR #9001, opened 2017-04-25
Tag BasisMatrices v0.3.0 [https://github.com/QuantEcon/BasisMatrices.jl]

sglyon:
## New features

- Introduction of a custom sparse matrix type `SplineSparse`. Allows for compact representation of a Spline BasisMatrix as well as efficient computation of `row_kron` and `funeval`
- Smolyak interpolation (See b29fb7bebcc19c864804f0521b0d24bdc4068b02. Also 65ee30b9b867933fe2d97b8e1ef6b6ec39454975, 469cf63fbc5422ecda69669ebbc49eb9866f1cdb, a5ff91d2cc9c705e6d6cc34b48d055b1335eed8e)
- Updates to complete polynomials  (thanks @cc7768  -- see #26)
    - 1st order derivatives
    - Specialized methods for evaluation at one point (a vector)
- Julia 0.6 compatibility 
- New demo and example (thanks @vgregory757 ) (bd92d9f81cc174f1936906e69fac59a023af2260, adf9e7ee9f13f4d4abbcc8bc2c893591e8b5944c)
- Consistent API for all subtypes of `BasisParams`
- Can now pass the types `Expanded`, `Direct`, or `Tensor` when constructing a `BasisMatrix` instead of only being able to pass instances (e.g. `Expanded` now works where before you needed `Expanded()`)
- New method for `::Basis  \ ::AbstractArray` to do fitting. 
- `LinParams` and `SplineParams` now have a type parameter for the type of abstract vector representing breakpoints. Helps with memory efficiency inside that type, in the `nodes` function, and leaves room for performance enhancements in routines like `lookup` and `evalbase` (although these improvements haven't been implemented yet).
- 

## Breaking changes

- Rename `BasisStructure` --> `BasisMatrix`
- Only field of Basis is a tuple of BasisParams objects. Constructors and methods are backwards compatible, so it shouldn't impact standard usage
- `evaluate(::Interpoland, ...)` has been renamed to call. So if you had an `Interpoland` object `itp` you evaluate at a point x by calling `itp(x)` instead of `evaluate(itp, x)`

## Improvements

- `>100x` performance improvement in calling `funeval` with a basis containing only `SplineParams` or `LinParams`
- Chebyshev evalbase now faster.
- Type stability throughout the library -- especially in the `funeval` family of methods
- No longer squeeze trailing dimensions of returns from `funeval`. Instead we have many more specialized methods that do the trimming in a type stable way. Things just "just work" how you'd expect them to work.
- The interpolated type now uses the `Tensor` form internally for (re-)fitting
- Spline evaluation now more efficient as row and column indices can be reused for all requested derivative orders
- Add `@inferred` to many tests to ensure we don't get type instability regressions.


BasisMatrices/versions/0.3.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+QuantEcon
+Compat 0.9
```
tkelman: needs 0.17 for abstract type

should also use compat for anonymous type parameter syntax, or be more specific about the VERSION cutoffs for it

sglyon: Thanks for checking. I'll bump compat for that.

What do you mean by anonymous type param syntax? This works just fine on 0.5 and 0.6 for me...

tkelman: `Foo{<:Integer}`

sglyon: The usage of that is restricted to be inside blocks that look like

```julia
@static if VERSION >= v"0.6-"
```

so I think I'm ok . I know that's messy, but I couldn't get some fancy type parameters to work with the same syntax on 0.5 and 0.6 so I just have a couple of those blocks in the library right now.

It will eventually get removed once I drop support for julia 0.5 , but I don't see that happening in the next few months.

I actually found a couple minor things I need to clean up in addition to the bump on the version number for compat so I will update this PR pretty soon (hopefully today or tomorrow)

tkelman: I'm saying Compat should support simple cases of that. Will look up what version. If not, at least using the type system rewrite PR merge commit as the VERSION cutoff would be more accurate

sglyon: I couldn't ever get it to work with compat and didn't have any more time to spend trying to figure it out so I just did the ugly thing for now.

Good call on making the version number more precise. I've done that and once I'm finished with those other changes I mentioned I'll update here.

Since I opened this I got attobot set up for the repo. I'm going to use that to update the tag -- that might open a new PR. If so I will close this one.

Thanks again -- I still don't know how you do it. You're role in the community is invaluable

tkelman: Yep attobot will open a separate PR since you opened this one the PkgDev way. An initially-attobot-opened PR can be amended in place though, ref. https://github.com/attobot/attobot#updatingdeleting-a-release

sglyon: OK cool. Sounds like I should just close this.

Will open a new one through attobot soon

PR #9007, opened 2017-04-25
Register RRRMC [https://github.com/carlobaldassi/RRRMC.jl.git]

carlobaldassi:
https://github.com/carlobaldassi/RRRMC.jl.git


tkelman:
This is a pretty opaque acronym name, would be better for discoverability and more in line with the naming guidelines to spell it out a bit.

carlobaldassi:
I had that thought too, my problem is that it's already published under that name in the accompanying paper (I didn't think I would register it as a package initially). At least the acronym is so bad that it's fairly unambiguous...

PR #9009, opened 2017-04-25
Register new package POMDPModels.jl v0.1.1

attobot:
Repository: [JuliaPOMDP/POMDPModels.jl](https://github.com/JuliaPOMDP/POMDPModels.jl)
Release: [v0.1.1](https://github.com/JuliaPOMDP/POMDPModels.jl/releases/tag/v0.1.1)
cc: @zsunberg


tkelman:
3-arg convert is kind of strange...

zsunberg:
Yeah, the three arg convert is a little weird, but the way to convert a state to a vector might depend on the problem domain... maybe we should call it something else? @etotheipluspi

PR #9011, opened 2017-04-26
Upper bound StaticArrays versions

c42f:
Unfortunately `StaticArrays` has been very sensitive to julia versions which have sometimes led people to insert workarounds for the lack of `StaticArrays` upper bounds.  Here, I've added upper bounds for all of the older versions, which should prevent any of the borken ones being installed in julia-0.6.

I did this as part of an attempt at understanding the CI errors in #8999. This PR will likely suffer from spurious (I think) CI failures of the same kind.

Notes:

* broadcast() changes broke some things early in 0.6 - I didn't track
  these down in detail, and have used 0.6- instead.

* Version 0.6.0-dev.1671 (44d7677) introduced the world changes to fix
  julialang/julia#265, which broke most of the StaticArrays generated
  functions.

* StaticArrays-0.3.x is the maintainence branch for julia-0.5
  compatibility, and doesn't work with 0.6.0-dev.1671.  I didn't track
  down the upper bound here in detail.

CC @tkoolen @andyferris


andyferris:
Thank you Chris!

carlobaldassi:
The EEG `v"0.0.3"` error is spurious and is fixed on latest Julia master.

The Rotations errors however are not, those versions require StaticArrays < `v"0.3.0-"` or `v"0.4.0-"`, which are made unavailable in latest Julia by this PR. The solution would be to cap all Rotations versions `v"0.3.*"` to julia `v"0.6-"` as well.

Note: in order to see this, try doing e.g. `Pkg.add("Rotations", v"0.3.0", v"0.3.0+")` on latest Julia, the error message will provide the explanation:

```
julia> Pkg.add("Rotations", v"0.3.0", v"0.3.0+")
ERROR: Unsatisfiable requirements detected for package StaticArrays:
└─version range [0.0.3,0.3.0-) required by package Rotations, whose allowed version range is [0.3.0,0.3.0+):
  └─version range [0.3.0,0.3.0+) set by an explicit requirement
The intersection of the requirements is [0.0.3,0.3.0-).
None of the available versions can satisfy this requirement.
```

andyferris:
I guess adding similar upper bounds to dependencies of StaticArrays (like Rotations, CoordinateTransformations, Geodesy) to this PR might be necessary.

tkoolen:
> similar upper bounds to dependencies of StaticArrays (like Rotations [...])

Ah, sorry, I would have in https://github.com/JuliaLang/METADATA.jl/pull/8793 if I had known how the METADATA tests worked at the time; I didn't think it would be necessary to have an explicit Julia upper bound if that upper bound could just be inferred from the lack of compatible dependency versions.

carlobaldassi:
>I didn't think it would be necessary to have an explicit Julia upper bound if that upper bound could just be inferred from the lack of compatible dependency versions.

We could do that, but only for the relatively easy cases like these ones: in general inferring whether a version of a package can be installed at all is a hard problem and requires to run the full resolution machinery. So the current arrangement is to be explicit: each version which is registered in METADATA and is declared to be compatible with some julia version must be allowed to be installed somehow. Otherwise, METADATA is in an inconsistent state, in which a user may see that e.g. "Rotations" `v"0.3.0"` could be installed, but get an error when they try to install it (see above).

This also removes a burden for the package solver when it's used, since the incompatible packages are filtered out at the source, rather than requiring that the solver figures out this information every time it's called.

c42f:
@carlobaldassi thanks for the suggested fix.

I'm content to do that (it's easy in this case), but it seems like a nasty workaround when the breakage is not exactly `Rotations` fault, but the fault of a library it depends on.

In the abstract, consider three packages 

julia ← A ← B ← C

The author of `C` may not even know they're using `A` transitively via `B`; vice versa the author of some general package `A` shouldn't need to know that `C` is depending on it.

If the author of `A` now adds correct upper bounds, they might be able to fix `C` if it's in the public METADATA, but it's a chore.  If `C` were a private package, it won't even be detected, and the authors of `C` need to deal with the breakage.

This all just seems kind of unfortunate. If `A` automatically narrowed the permissable versions of `B` and julia for `C`, this would fix the Rotations case.

c42f:
@carlobaldassi Ah, looks like you've already answered my question while I was writing all that.  I think I can see why it's a hard problem.  There might be a nicer solution, but for now I'll just write it all out.

Thanks for explaining.

carlobaldassi:
For some reason, CI is not starting on the latest commit, I'd be curious to know if the EEG error goes away even without the solver fix.

Anyway, let me comment a little more on the issue of fixing the bounds of dependent packages: the scenario above with the julia ← A ← B ← C scheme doesn't convince me, for a number of reasons, and not only because it's a hard problem:

1. In general, the resolver can't "read your mind" about what you intend; i.e., I don't think there's a clear distinction between "implicit dependency" and "unintended consequences", see below. We may agree on some rules, but I think that would be complex and fragile.
2. This situation would only arise when fixing bounds of older versions of a package which make other packages' versions suddenly uninstallable, like in this PR or in analogous cases. I really don't think this is too much of a chore: first, this doesn't happen too often (only at julia updates, essentially); second, it only requires to do `PkgDev.check_metadata()`, which (assuming the resolver is doing its job proparly) doesn't take long and would point out all the problems; third, I see this as the responsibility of the PR author, which after all is the one that is trying to make METADATA more consistent, and shouldn't end up making it less consistent.
3. If the affected package that we're talking about (`C` in the example) is not registered in METADATA, then it wouldn't have any versions, only master; I don't see any scenario where setting a correct bound for package `A` would make things worse for package `C`: before the fix, it would fail, after the fix it would complain that there are no working versions of `A` available. The second case is actually a clearer indication to the developers/users of `C` of the origin of the problem (in this example, it may be required to fix `C` to make it work with newer versions of `B`, or that `B` is fixed to work with newer versions of `A`, for example, this will depend on the details). Assuming an "implicit dependency" on the Julia version would not fix anything in this case, it would just make things even more confusing, possibly.
4. What if the fix to `A` made *all* versions of `C` registered in METADATA not available. If the dependency were implicit, then a user would see a package `C` and not be able to install it at all. In this way at least the METADATA information ends up being correct, and the developer of `C` could be notified that their package needs updating to work with the latest Julia.
5. It is still possible that the requirements on `A` are too strict, or botched for some reasons; observing the most immediate consequences at PR time, and perhaps notifying the maintainers of the other affected packages, may catch that and is preferable to having users suddenly being unable to install a package.

So my general sentiment is strongly in favor of keeping things as they are — again, provided that the solver actually does its job correctly. (Plus, as I mentioned above, it a "do it once and for all" solution rather than "compute it every time `Pkg.update()/add()/remove()` is invoked" solution.)

c42f:
> In general, the resolver can't "read your mind" about what you intend; i.e., I don't think there's a clear distinction between "implicit dependency" and "unintended consequences"

Well put!  Particularly in regard to your last point.

> If the affected package that we're talking about (C in the example) is not registered in METADATA, then it wouldn't have any versions, only master

This doesn't seem quite true for people who maintain their own private METADATA. The reason I bring up unregistered packages is that we happen to have several at work containing internal code we don't want to release. Nevertheless, it should be possible for us to maintain our own internal METADATA of registered private versions (I understand that in Pkg3, this will be a lot easier).


Anyway, just to check I understand, is the rule something like:
When C declares an explicit dependency on A [v1,v2), and C also depends on B which depends on A [v3,v4), we require that [v1,v2) ⊆ [v3,v4)?

tkelman:
I'll cross-reference #8817 where this issue also came up, but the author of LightGraphs gave up on coming up with a consistent set of bounds since a downstream package had a somewhat complicated set of constraints on LightGraphs. I think there would be a few different solutions to that issue, but I'm not sure what strategy would be the most effective to pursue.

c42f:
Thanks for the xref @tkelman.  It's true that I'd feel unhappy to be modifying the upper bounds of `Rotations` if I wasn't one of the maintainers of that package.

carlobaldassi:
It seems Travis won't run on this? In the queue, I can only find the first failed attempt (https://travis-ci.org/JuliaLang/METADATA.jl/builds/225963034, testing https://github.com/JuliaLang/METADATA.jl/commit/2a567166a5e254e8099386fb7427023aff82a05e). @c42f: may you could try to force it? E.g. with
```
git commit --amend --no-edit
git push origin +StaticArrays-upper-bounds
```
(assuming `origin` is your repo, of course.)


> This doesn't seem quite true for people who maintain their own private METADATA. The reason I bring up unregistered packages is that we happen to have several at work containing internal code we don't want to release. Nevertheless, it should be possible for us to maintain our own internal METADATA of registered private versions (I understand that in Pkg3, this will be a lot easier).

Ah, I see what you meant now. Yes, that is a tricky situation. My solution, if I was maintaining such an alternative private METADATA, would probably be to use some script instead of pulling directly from origin, fetch the changes and and check whether there are only new entries or if something changed in the existing `requires` files (which doesn't happen often anyway), in which case then an extra `check_metadata()` run should be able to tell if there are any problems, and how to fix them before the pull. Another easier solution would be to disable the check_metadata entirely in the private version, I guess, and leave the discovery of inconsistencies up to the solver, whenever it's invoked and fails.

That is assuming an alternative version that pulls regularly from the official METADATA. Deciding how to proceed when merging different METADATAs together and one introduces inconsistencies in another would be more complicated, but it seems to me that basically the two options are still to either enforce consistency or ignore the errors until they come up.

As for #8817: is my reading correct that TrafficAnalysis is basically declaring the wrong dependencies? The correct way to handle that would be setting the correct julia bounds on TrafficAnalysis versions (after all, if they require a version of LightGraphs that doesn't work on 0.X, then they won't work on 0.X, right?). Also, the artificial dependency to avoid the downgrade should als be avoided, it would be best again to specify an upper bound on julia (or, if the previous versions that didn't depend on LightGraphs would have actually worked on recent versions of Julia, then the downgrade of the package would have been the correct choice - the only possible choice that makes everything work, until a newer version). Users that want to get the latest version and are willing get their hands dirty trying to make things work or that can deal with the failures have always the option to checkout.

c42f:
I just force pushed a trivially amended commit.  Hopefully Travis will pick it up properly this time.

c42f:
Oh good, looks like Travis has passed now.

> Another easier solution would be to disable the check_metadata entirely in the private version, I guess, and leave the discovery of inconsistencies up to the solver, whenever it's invoked and fails.

Yes, this is probably what we'll end up doing fairly soon.  We don't have a lot of private packages so we've been managing them in an even more basic way (via DeclarativePackages and basic git tags) so far, but things are getting messy at this stage and we'll be needing to maintain a proper private METADATA soon.

tkelman:
> or, if the previous versions that didn't depend on LightGraphs would have actually worked on recent versions of Julia, then the downgrade of the package would have been the correct choice - the only possible choice that makes everything work, until a newer version

Not so sure about that - if the old version is very old, would work on recent julia but has far fewer features, I think holding LightGraphs back to avoid the downgrade is the lesser evil.

aviks:
Hi @c42f is this PR still relevant?

c42f:
@aviks I think we can still merge these upper bounds - they are still valid, as far as I know, and it did take quite a while to come up with them.  Comments further up were addressed, AFAICT by looking at the thread and trying to remember back to what I was doing.

Conversely, I haven't seen people running into problems with this recently.

tkelman:
I'm not sure it's necessary at the moment, unless there have been situations where old versions are getting installed, or other bounds issues that this would help with (maybe at some future point there will be and we can revisit)

c42f:
Yeah, this is pretty stale and the need may have passed. I could go either way with it now (close or merge).

PR #9019, opened 2017-04-26
Tag Measurements v0.4.0 [https://github.com/giordano/Measurements.jl]

giordano:
From `NEWS.md`:

> ### Breaking Changes
> 
> * Support for Julia 0.4 and 0.5 was dropped.
> * `value` and `uncertainty` functions are no more exported.
> * `Measurements.gradient` function has been removed.  Use the vectorized
>   function `Measurements.derivative.(x, array)` instead.
> 
> ### Improvements
> 
> * There were several performance improvements, both in speed and memory usage.
>   These improvements are more evident for simple operations like addition and
>   subtraction (now these operations are 2-3 times faster than in v0.3.0).  In
>   addition, the overhead of operations with arrays of `Measurement`s and complex
>   `Measurement`s was greatly reduced by avoiding useless conversions (now
>   summing an array is 100 times faster than in v0.3.0).
> 
> ### New Features
> 
> * `quadgk` function from [`QuadGK.jl`](https://github.com/JuliaMath/QuadGK.jl)
>   package is extended to support `Measurement` objects as endpoints of
>   integration ([#8](https://github.com/giordano/Measurements.jl/pull/8)).  Note
>   that only the case of two real endpoints is supported.
> * Real `Measurement` objects can be printed with `"text/x-tex"` and `"text/x-latex"`
>   MIME types.  `\pm` TeX macro is used to render the `±` sign.
> * A new `parse` method is provided, to parse a string to
>   `Measurement{Float64}`.  This is the same as calling `measurement` with a
>   string argument.


Measurements/versions/0.4.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.6
```
tkelman: This won't be installable yet until 0.6.0 final gets released. You could set `0.6-` to allow all prereleases, or set some known-working more specific version or tag as the floor for it to work now.

giordano: Used a generic `0.6-`, thanks.

tkelman:
This caused a failure in your LombScargle package https://github.com/JuliaCI/pkg.julialang.org/blob/44084b06c55f5578762784ed716e56ee151f2eea/logs/LombScargle_0.6.log

giordano:
Thanks!  The development version of LombScargle (which I plan to release within the next few days) should be fine, though ;-)

PR #9025, opened 2017-04-26
Pull request/85a4894

emoszkowski:
Register StateSpaceRoutines.jl package


tkelman:
Seems like you could move HDF5 to a test-only dependency here

PR #9032, opened 2017-04-27
RecursiveArrayTools bounds

ChrisRackauckas:
For the type piracy elimination

### DIFF
--- a/DiffEqBase/versions/0.0.1/requires
+++ b/DiffEqBase/versions/0.0.1/requires
@@ -1,2 +1,3 @@
 julia 0.5
 RecipesBase 0.1.0
+RecursiveArrayTools 0.0.0 0.4.0
--- a/DiffEqBase/versions/0.0.2/requires
+++ b/DiffEqBase/versions/0.0.2/requires
@@ -1,3 +1,4 @@
 julia 0.5
 RecipesBase 0.1.0
 Ranges 0.0.1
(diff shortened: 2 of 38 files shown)


tkelman:
Almost like you've done this before. 
```
$ git grep -l RecursiveArrayTools | xargs dirname | xargs dirname | xargs dirname | sort -u
.
DelayDiffEq
DiffEqBase
DiffEqDevTools
DiffEqJump
DiffEqParamEstim
MultiScaleArrays
OrdinaryDiffEq
StochasticDiffEq
```

So will DelayDiffEq, DiffEqJump, DiffEqParamEstim, MultiScaleArrays, Ordinary- and StochasticDiffEq notice this? (That doesn't count test-dependencies)

ChrisRackauckas:
The only usage of the extended mean is inDiffEqBase and DiffEqDevTools. The other stuff in the updated RecursiveArrayTools is just new.

ChrisRackauckas:
>Almost like you've done this before.

Not on RecursiveArrayTools. But yes, these last few weeks have been tidying up the Base libraries for a 2.0.0 release, and that means cleaning up a lot of mistakes made really early on that are used in all of the testing and basic type constructions. Hopefully this is the last of it, which is why it's the 2.0.0 release.

tkelman:
Was mostly in response to the artificial dependency on the 0.0.1 tags, which I know we've gone through elsewhere.

PR #9038, opened 2017-04-28
Tag Hecke.jl v0.2.0

attobot:
Repository: [thofma/Hecke.jl](https://github.com/thofma/Hecke.jl)
Release: [v0.2.0](https://github.com/thofma/Hecke.jl/releases/tag/v0.2.0)
Diff: [vs v0.1.5](https://github.com/thofma/Hecke.jl/compare/f5f5b5607ba7e8b028c721d65cc178a4093f2a89...65bfc3b979d8600cca00e2ea472d5e135bb33119)
`requires` vs v0.1.5: 
```diff
--- 0.1.5/requires
+++ 0.2.0/requires
@@ -1,2 +1,2 @@
-julia 0.4
-Nemo 0.5.1
+julia 0.5
+Nemo 0.6
```
cc: @thofma


tkelman:
Shouldn't be importing `Base.isprime` then defining `isprime(a::Integer)` on Base types, that can change the behavior of other code. Better to just not import it from base since it's deprecated there, and have it be `Hecke.isprime`.

thofma:
I do the same as Primes.jl. There they also hide the deprecation warning by extending `isprime` and `factor` for Base types.

tkelman:
So if a package wants to import both Primes.jl and Hecke.jl, how do they get a predictable type back from `factor` ?

thofma:
I don't know why we are discussing this. It is done in Primes.jl. Why aren't we allowed to do it in Hecke.jl?

tkelman:
Primes.jl was taking the same code that used to be in base, and making it easy to transition by being compatible. Your code returns an entirely different factorization object type, so `factor(::Integer)` changes behavior dramatically if Hecke is imported.

thofma:
No, there is no changing bevahior, because julia 0.5 does not come with `factor(::Integer)`. So importing Hecke only provides functionality and does not override anything.

I have never read that all packages must be compatible with each other. Primes.jl and Hecke.jl don't play nice together. I don't see the problem.

tkelman:
julia 0.4 did have a `factor(::Integer)` though, which is why Primes.jl does so on julia 0.5. Packages working together isn't a requirement, but when it's simple to achieve just by modifying what you import and how you namespace things, what's the reason not to do so? Modifying the way other unrelated code behaves, like monkey-patching in many other languages, is possible but frowned upon. Especially when the methods and types in question belong to Base, or in a transitional formerly-part-of-Base package that temporarily still acts like the original Base code did.

thofma:
So if I wait for julia 0.6 there is no problem, right? Then there is no `factor(::Integer)` in Base and I can just define and export it in Hecke.jl.

tkelman:
That should work too, since Primes only imports these from Base on 0.5.

thofma:
Removed the monkey patching.

PR #9041, opened 2017-04-28
RFC: Add empty requires file to all past tags that didn't have one

tkelman:
This isn't strictly required but can help for uniformity's sake with some scripts

Just for some numbers, we currently have 9973 requires files on metadata-v2, of which roughly 45 of the existing ones are empty. This PR adds 406 more. Any objections to doing this?


timholy:
Will this add to the number of files certain operations have to open, check, and close? If so, there's some risk it could slow the performance of Pkg in ways we might notice.

tkelman:
By about 4%, or equivalent to the amount of real files that get added to metadata in roughly 16 days.

tkelman:
This could maybe allow us to skip a "does the file exist" check in a few places, since all recent tags have been enforcing that. Though Pkg3 will be restructuring this all, eventually.

tkelman:
If the file system has been warmed up, at least on Linux I don't see any noticeable difference in the timing of `Pkg.cd(Pkg.Entry.available)` on this branch vs metadata-v2.

tkelman:
20 of them need to be added anyway to accomplish #9048

tkelman:
This would allow us to remove the `isfile` check here https://github.com/JuliaLang/julia/blob/8eca02717bcea524877fad01aacedc807898bcd1/base/pkg/reqs.jl#L73, but in brief testing on Linux with a warmed-up file system that didn't seem to make much difference. I should do some testing on Windows and see if it helps there.

PR #9093, opened 2017-05-02
Tag JSON.jl v0.10.0

attobot:
Release: [v0.10.0](https://github.com/JuliaIO/JSON.jl/releases/tag/v0.10.0)
Diff: [vs v0.9.1](https://github.com/JuliaIO/JSON.jl/compare/cbd9169ab3aef3e91b3ec60cfbb78c0156351841...c3e5515019e76e0ff000b26050ed5b116aa7f3cd)
`requires` vs v0.9.1: no changes
cc: @TotalVerb


tkelman:
Would you mind if we retroactively made this the first 0.5-only version? The higher version number here resulted in the resolver doing some funny things on 0.4 which led to CoordinateTransformations and EEG getting downgraded to an old broken version when they ran on PkgEval, and I suspect something similar might happen to users.

TotalVerb:
No objection to that, makes sense.

PR #9099, opened 2017-05-02
Upper bound IntervalArithmetic and ValidatedNumerics' dependencies on StaticArrays to < 0.5

tkelman:
ref https://github.com/JuliaLang/METADATA.jl/pull/8999#issuecomment-297960883

since prior versions of ValidatedNumerics did not depend on StaticArrays,
this could result in downgrading ValidatedNumerics to 0.7.0 in order to
upgrade StaticArrays to 0.5

that would be fixed by releasing a new StaticArrays-0.5-compatible version
cc @dpsanders

### DIFF
--- a/IntervalArithmetic/versions/0.9.0/requires
+++ b/IntervalArithmetic/versions/0.9.0/requires
@@ -1,5 +1,5 @@
 julia 0.5
 CRlibm 0.5
-StaticArrays 0.3
+StaticArrays 0.3 0.5
 ForwardDiff 0.2
 RecipesBase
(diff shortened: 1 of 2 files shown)


dpsanders:
Thanks. Do you mean that the new version should have a minimum allowed version 0.5 of StaticArrays.jl?

tkelman:
The minimum version should be the lowest version with which it works correctly. If, after updating it to be compatible with StaticArrays 0.5, it no longer works with StaticArrays 0.3.x or 0.4.x, then yes the appropriate minimum version would be 0.5.

PR #9108, opened 2017-05-03
Make JSON 0.10 require Julia 0.5

tkelman:
ref https://github.com/JuliaLang/METADATA.jl/pull/9093#issuecomment-298819420


tkelman:
(merging a bit overly quickly since tonight's pkgeval run is starting right now)

PR #9083, opened 2017-05-01
Tag StatsBase.jl v0.14.0

attobot:
Repository: [JuliaStats/StatsBase.jl](https://github.com/JuliaStats/StatsBase.jl)
Release: [v0.14.0](https://github.com/JuliaStats/StatsBase.jl/releases/tag/v0.14.0)
Diff: [vs v0.13.1](https://github.com/JuliaStats/StatsBase.jl/compare/cb2668ba75d896c774762a323c1ffef0216c9b9f...38c62a15e2e3b24c732f9b90b7ded1abc83ce6ad)
`requires` vs v0.13.1: no changes
cc: @ararslan

PR #9115, opened 2017-05-04
Tag StatsBase.jl v0.14.1

attobot:
Repository: [JuliaStats/StatsBase.jl](https://github.com/JuliaStats/StatsBase.jl)
Release: [v0.14.1](https://github.com/JuliaStats/StatsBase.jl/releases/tag/v0.14.1)
Diff: [vs v0.14.0](https://github.com/JuliaStats/StatsBase.jl/compare/38c62a15e2e3b24c732f9b90b7ded1abc83ce6ad...e200a7299a40e1f9d719179b98968542edc19f3c)
`requires` vs v0.14.0: no changes
cc: @ararslan


tkelman:
will this fix DataArrays, ImageMetadata, and/or ImageTransformations? https://github.com/JuliaLang/METADATA.jl/pull/9083#issuecomment-298817852

ararslan:
Probably not, but it fixes LightGraphs

tkelman:
K. Any idea what's wrong with DataArrays then?

The Image* are showing that StatsBase has some ambiguities. Should probably enable ambiguity testing within StatsBase to avoid that going forward

ararslan:
> Any idea what's wrong with DataArrays then?

Not yet. :/

> Should probably enable ambiguity testing within StatsBase to avoid that going forward

I'm actually not sure how to do that, but it sounds like a good idea.

tkelman:
something like https://github.com/JuliaImages/ImageTransformations.jl/blob/ee9d5267309b2bce35be6c592ee376188120d82d/test/runtests.jl#L4-L7, there are a few variations out there

@timholy may have advice on the latest best-practice version of that

ararslan:
Turns out DataArrays had a weird case that we were testing but it doesn't actually make sense and was broken by a refactor of the weighting code in StatsBase. Our solution is to remove the test case in DataArrays.

PR #9129, opened 2017-05-04
Update urls for moved packages

tkelman:
Update urls for @rofinn (formerly @Rory-Finnegan) and @furface (formerly @brandonkmiller)'s packages. Github redirects should handle these, but if you plan on keeping these new names permanently, best to have the urls be accurate.


rofinn:
Thanks @tkelman, I've been meaning to update those links.

PR #9136, opened 2017-05-04
Register new package OhMyREPL.jl v0.1.0

attobot:
Repository: [KristofferC/OhMyREPL.jl](https://github.com/KristofferC/OhMyREPL.jl)
Release: [v0.1.0](https://github.com/KristofferC/OhMyREPL.jl/releases/tag/v0.1.0)
cc: @KristofferC


tkelman:
Does the change in error printing here interfere with any other packages' tests?

KristofferC:
I don't know. `Pkg.test` runs in its own process so probably not.

KristofferC:
Bump

tkelman:
Looks like Pkg.test doesn't set `--startup-file` though, so if it's put there as you recommend it could. Maybe worth restricting the output overrides to when `isinteractive()` is true

KristofferC:
`isinteractive()` is false when OhMyREPL.jl gets loaded (REPL hasn't started yet).

tkelman:
but it should be true by the time the functions execute?

KristofferC:
OhMyREPL loads a REPL by itself. Maybe this approach would work: https://github.com/KristofferC/OhMyREPL.jl/issues/68

PR #9147, opened 2017-05-04
Add Nemo 0.6 upper bound to Hecke 0.1.3 - 0.1.5

tkelman:
ref http://pkg.julialang.org/detail/Hecke.html, Hecke has been failing tests on julia 0.5 since April 14, which is when Nemo 0.6.0 was tagged

cc @thofma @wbhart does this look right to you? https://github.com/JuliaCI/pkg.julialang.org/blob/5b594c5d55cc0ddfd201f21a615dff02ff87de0c/logs/Hecke_0.5.log is showing `UndefVarError: is_negative not defined` at `Hecke/src/Map/NumberField.jl, in expression starting on line 289` for Hecke version 0.1.5. https://github.com/JuliaLang/METADATA.jl/pull/9038 will give a Nemo-0.6 compatible version once https://github.com/thofma/Hecke.jl/pull/16 is worked out, but in the meantime I believe these upper bounds on the past versions are still correct.


thofma:
Looks good.

PR #9189, opened 2017-05-08
Register new package PiCraft.jl v0.1.0

attobot:
Repository: [JuliaBerry/PiCraft.jl](https://github.com/JuliaBerry/PiCraft.jl)
Release: [v0.1.0](https://github.com/JuliaBerry/PiCraft.jl/releases/tag/v0.1.0)
cc: @aviks


PiCraft/versions/0.1.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.4
```
tkelman: are you still testing this on 0.4?

aviks: I did, actually. This code is unchanged from some time ago. However, given that we want all users on the Pi to use 0.5, I've updated require.

PR #9190, opened 2017-05-08
Tag IterativeSolvers v0.3.0 [https://github.com/JuliaMath/IterativeSo…

lopezm94:
…lvers.jl]


IterativeSolvers/versions/0.3.0/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.5
+UnicodePlots
+RecipesBase
+SugarBLAS
+LinearMaps
+Compat 0.17.0
```
tkelman: doesn't look like you have any `@compat abstract type` anywhere, and `@compat` isn't needed for `const foo = bar` unless `foo` has a type parameter

tkelman: ah missed that `UnrestartedHistory` and `RestartedHistory` do have a type parameter, in which case they need Compat 0.18

lopezm94:
@tkelman Could you follow [these](https://juliadocs.github.io/Documenter.jl/latest/man/hosting.html#SSH-Deploy-Keys-1) instructions to allow auto deployment of docs? I would do it but I don't have access to IterativeSolvers.jl setting. It will allow the deployment of new documentation every time a tag is made

tkelman:
Okay done. Looks like you deleted the existing git tag?

lopezm94:
Thanks. Yes, I'll re-tag so the new commits are included.

lopezm94:
The docs didn't auto deploy unfortunately. I copied the full hash of the new tag.

tkelman:
Looks like this broke
- GeostatInversion cc @montyvesselinov http://pkg.julialang.org/logs/GeostatInversion_0.5.log
- PyAMG cc @cortner http://pkg.julialang.org/logs/PyAMG_0.5.log
- RobustLeastSquares cc @andyferris http://pkg.julialang.org/logs/RobustLeastSquares_0.5.log

@nkottary why didn't reverse dependency CI run here?

tkelman:
PyAMG cannot be fixed by adjusting METADATA bounds since IterativeSolvers is a test-only dependency there. Would need to make a new tag, either supporting IterativeSolvers 0.3 or adding an upper bound in test/REQUIRE

nkottary:
@tkelman We expect the julia version bounds to be mentioned as the first line of the requires file. This was not the case for IterativeSolvers.jl:v0.2.2. So the test errored out.

tkelman:
that's not mandatory, should use the Pkg.Reqs functions to parse those files

montyvesselinov:
Are there plans v0.3.0 to be fixed? Now we are ignoring it.

-monty

tkelman:
a more isolated test case in https://github.com/JuliaMath/IterativeSolvers.jl/issues/116 would be helpful

the upper bounds here were at least easier than the Mads-Gadfly problem since GeostatInversion hasn't had as many past tags

cortner:
sorry for the slow reaction, I'll make PyAMG compatible with the latest version of IterativeSolvers within the next few days.

PR #9194, opened 2017-05-09
Register new package Avro.jl v0.0.2

Hydrotoast:
**Repository**: [Hydrotoast/Avro.jl](https://github.com/Hydrotoast/Avro.jl)


Avro/versions/0.0.2/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5 0.6-
```
tkelman: upper bounds will not be effective unless applied to all past tags

Hydrotoast: Thanks, I just decided to fix the tests that broke on v0.6. Cutting a new patch.

PR #9198, opened 2017-05-09
Register new package Poltergeist.jl v0.0.1

wormell:
Repo: https://github.com/wormell/Poltergeist.jl
Tagged 0.0.1


Poltergeist/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.4
+ApproxFun 0.6.0
+BandedMatrices 0.2
+Compat 0.24.0
```
tkelman: unless you're using a feature that was first released in 0.24, this is too tight of a minimum version requirement

wormell: OK, it's now set at 0.16.

tkelman: what's the newest feature you're relying on?

wormell: Broadcast dot syntax.

tkelman: unless you're using an updating operator (https://github.com/JuliaLang/Compat.jl/commit/6d60600e9003bf23a4b42792b7ee645682fdcdbf or https://github.com/JuliaLang/Compat.jl/commit/63beaae5cb2aa417d169baec9a17ac9cc239c6aa), that doesn't actually need Compat

wormell: Ah, looks like I don't need Compat at all any more. Thank goodness

wormell:
The tags are now pushed

Poltergeist/versions/0.0.1/requires
tkelman: ah right, do you want to publish the 0.4-supporting version at all? might be cleaner to only have the latest tag get published here

wormell: Possibly, but it would be an earlier version than this one. I'll have a think about it and start over.

wormell: Thanks for your patience!

tkelman: see also https://github.com/attobot/attobot which you may find useful for automating some of the publishing process

PR #9199, opened 2017-05-09
Tag OpenEphysLoader v0.1.3 [https://github.com/galenlynch/OpenEphysLo…

galenlynch:
…ader.jl]

Adding compatibility for julia 0.6

### DIFF
--- a/OpenEphysLoader/versions/0.1.3/requires
+++ b/OpenEphysLoader/versions/0.1.3/requires
@@ -1,0 +1,3 @@
+julia 0.5
+LightXML
+Compat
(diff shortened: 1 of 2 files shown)


OpenEphysLoader/versions/0.1.3/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+LightXML
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

galenlynch: I've added minimum version dependencies to my master branch, should I tag a new version or is it possible to change this tag?

tkelman: both - can click the edit button on the METADATA copy of this file for the 0.1.3 tag to modify the existing tag's minimum version requirement in place (only change requires, not the sha1), but for any code changes you'll need a new tag as well

PR #9200, opened 2017-05-09
Register PyCallJLD.jl [https://github.com/JuliaPy/PyCallJLD.git]

cstjean:
I called `PkgDev.register("PyCallJLD.jl")`, but it seems to have latched onto `PyCallJLD.git` even though I renamed that repo to `PyCallJLD.jl`. Is that an issue?

Also, I would have loved to use attobot for this, but I don't seem to have the proper JuliaPy permissions?

Hopefully this module can be merged into PyCall once we have conditional modules.

### DIFF
--- a/PyCallJLD.jl/url
+++ b/PyCallJLD.jl/url
@@ -1,0 +1,1 @@
+https://github.com/JuliaPy/PyCallJLD.git


tkelman:
JuliaPy settings seem to indicate that Attobot is enabled on all repos in that org? The missing .jl and lack of a tag here would need to be fixed.

PR #9205, opened 2017-05-09
Tag DSP v0.2.0 [https://github.com/JuliaDSP/DSP.jl]

gummif:


DSP/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+Polynomials v0.1.0 v0.1.2+
```
tkelman: it would be preferable if the package could be made compatible with latest Polynomials versions, rather than having to hold it back

gummif: Ok I'll take a look at that

DSP/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
...
+Reexport
+SpecialFunctions
+Compat 0.17.0
```
tkelman: this needs to be 0.18 since several of the typealias const replacements are parametric

gummif:
Changes have been made to allow more versions of Polynomials.

tkelman:
great! this should be squash merged if travis is green

PR #9216, opened 2017-05-10
Tag RandomNumbers v0.0.4 [https://github.com/sunoru/RandomNumbers.jl]

sunoru:
Updated for Julia 0.6.


RandomNumbers/versions/0.0.4/requires
```diff
@@ -0,0 +1 @@
+julia 0.6
```
tkelman: this won't be installable until 0.6.0 final is out

sunoru: Oh you're right.. Let me modify it and reopen a PR.

PR #9238, opened 2017-05-11
Register Reduce: v0.0.1

chakravala:
[https://github.com/chakravala/Reduce.jl.git]

Similar to [Maxima.jl](https://github.com/nsmith5/Maxima.jl), his package requires [Reduce PSL](http://www.reduce-algebra.com/index.htm) to be installed, it needs `redpsl` in the path.

[https://github.com/chakravala/Reduce.jl](https://github.com/chakravala/Reduce.jl)

As of now, I don't know yet how to automate the build with Reduce, but the packge is fully functional if you do have Reduce PSL installed.


MikeInnes:
Congrats on the release!

tkelman:
Please leave new package registrations open for at least 24 hours Mike.

MikeInnes:
Is it recorded somewhere that this is a generally agreed-upon rule?

tkelman:
This was discussed on julia-committers some time ago, yes.

tkelman:
Fatou defines several Base methods on Base types (aka "type piracy"), which is explicitly recommended against in thr style guide and changes the behavior of unrelated code when the package is imported.

MikeInnes:
Decisions like that should be documented (and discussed) publicly – this is a community project, after all.

chakravala:
Is there a way to define this Base method without it being a problem?

tkelman:
Sure we'll write it down somewhere - the readme is unfortunately hidden, but it has been discussed in the pr's and issues here on past occasions.

Rather than defining multiplication between Function objects, there is a function composition operator \circ  in Julia 0.6 and the Compat package

chakravala:
These are the 2 locations where the issue is. It's used to generate the function composition for a LaTeX expression. To do that, I need to be able to specify `j` function compositions.

```Julia
# define multiplication operator on functions to evaluate as function composition
import Base: *; *(f::Function,g::Function) = x -> f(g(x))

# we can convert the j-th function composition into a latex expresion
nL(f::Function,m,j) = SymPy.latex(j==1 ?
  newton(f,m)(Sym("z")) : (newton(f,m)^j)(Sym("z")))
```

tkelman:
that could probably be done with a package-local `repeated_composition` method? otherwise this could conceal bugs elsewhere, since exponentiation of functions is typically not defined

chakravala:
Okay, will commit this instead:
```Julia
# define multiplication operator on functions to evaluate as function composition
recomp(f::Function,x::Number,j::Int) = j > 1 ? f(recomp(f,x,j-1)) : f(x)

# we can convert the j-th function composition into a latex expresion
nL(f::Function,m,j) = SymPy.latex(recomp(newton(f,m),Sym("z"),j))
```

tkelman:
Reduce is also a very generic name, it might have helped to disambiguate a bit to make its purpose clearer

chakravala:
That is true. The first time I heard of Reduce, I thought it was too generic of a term as well. Google searching it is rather difficult due to the ambiguity.

However, you have to also appreciate the fact that Reduce is the [FIRST](http://www.reduce-algebra.com/reduce40.pdf) computer algebra software ever written, it is now over 40 years old! Technically and historically, this program has precedence over any other program named Reduce. When Reduce was first created, this was long before search engines and the internet were a thing.

I have thought about what to name this package and settled on `Reduce` because in the end I thought it was the most elegant way to name it and convenient. Other options could be

- `ReducePSL` less comfortable to type due to caps at end, but more specific about which lisp version of the program it is meant for
- `Redpsl` not much clearer to other people, but the name of the binary
- `ReduceAlgebra` similar to the website name of the software, but it is less convenient to constantly type a longer name, and technically this is not the name of the software
- `ReduceCAS` could emphasize that Reduce is a CAS (computer algebra software), but has the same issue of being a bit annoying to type compared to plain `Reduce`, also technically not the name of the software

If you think about it, this name is actually not much more generic than `Maxima`, which is kind of the opposite of `Reduce` in terms of the meaning of the actual word. It's just that less people have heard of it, so more people will recognize `Maxima` than `Reduce`. So in comparison to the `Maxima` package, it is actually named to fairly similar conventions. Then I would have to similarly say, how should anyone know from the title `Maxima` that it is a computer algebra software? Either you know it or you don't. Most people never heard of Reduce because it is obscure and used by very specialized people.

Also, the name of the REPL prompt in Julia is `reduce> `, so the package name `Reduce` has a one-to-one correspondence with this feature / name.

Let me know what your thoughts are based on that information.

MikeInnes:
Seems legit to me. There's really no one thing that "Reduce" could refer to, so it makes sense to just give it to the first taker.

PR #9243, opened 2017-05-12
Register new package Poltergeist.jl v0.0.4

attobot:
Repository: [wormell/Poltergeist.jl](https://github.com/wormell/Poltergeist.jl)
Release: [v0.0.4](https://github.com/wormell/Poltergeist.jl/releases/tag/v0.0.4)
cc: @wormell


dpsanders:
This looks like a very nice package! Any particular reason you chose that license? As far as I understand, it's not compatible with MIT, for example (only the 3-clause BSD is?)

dpsanders:
(And very clever, although obscure, package name!)

ararslan:
The BSD 2-clause license is quite similar to MIT, actually. And I don't think any of the BSD licenses are incompatible with MIT (which is itself a BSD-style license). [This](https://opensource.stackexchange.com/a/582) is a good explanation of BSD vs. MIT.

dpsanders:
Ah OK, thanks for the clarification - my mistake.

wormell:
Thanks @dpsanders, glad you like it! Two-clause seems like the default BSD licence in PkgDev for whatever (nefarious?) reason so I went with that

ararslan:
Is that the default? It might just be the first sorted alphanumerically. It's a good, solid license nonetheless though. 🙂

dpsanders:
I don't think there is a default: the license has to specified explicitly?

PR #9245, opened 2017-05-12
Tag LowRankApprox v0.0.2 [https://github.com/klho/LowRankApprox.jl]

klho:
With fixes by @MikaelSlevinsky for v0.6.


LowRankApprox/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Compat 0.19
```
tkelman: only need 0.18 since you aren't using IndexStyle

MikaelSlevinsky: Oops, I'm sorry this was my fault! What needs to be done? Deleting the current tag, committing the version change, re-tagging and re-submitting a PR to metadata?

tkelman: could be fixed in-place here (only editing the requires file, not changing the sha1) as long as the corresponding change gets made on master of the package before the next tag

Issue #6761, opened 2016-10-14
Establishing expectations for registered packages


ararslan:
There appears to be some dissatisfaction with the speed—be it too fast or too slow—of the METADATA workflow. I think this may stem from a lack of a common understanding of what's expected for registered packages. To that end, I think it may be a good idea to have a discussion about what exactly is expected of registered packages, including what is required for registration, so that everyone is on the same page. I think having a standard pull request template, as suggested in #5307, may help with this, but it would be nice to have consensus on expectations in the interim.

Note that this is **not** intended to be a debate on the merits of METADATA or registration, nor is it a suggestion that we should disallow certain packages or kinds of packages. I'm merely looking for input on what is/should be expected and how everyone can come to a mutual understanding to avoid future drama and/or confusion.

tkelman:
Standard things I check for when reviewing a new package:

- the guidelines listed here https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/README.md and here http://docs.julialang.org/en/release-0.4/manual/packages/#guidelines-for-naming-a-package
- do the git tags match the metadata tags (the tests added in https://github.com/JuliaLang/METADATA.jl/pull/6485)
- if the package can easily be made to pass its tests on Travis and/or AppVeyor, are they enabled and passing for all supported Julia versions?
- is any code clearly derived from or copied from some other project without appropriate license attribution?
- is the package doing anything obviously malicious, especially in a `deps/build.jl` script
- is every dependency that is used via `using`, `import`, or `importall` listed in `REQUIRE` or `test/REQUIRE`, preferably with minimum version requirements?
- is the package changing the behavior of other unrelated code via "type piracy" - aka extending methods from Base or some other package on type signatures that are entirely from Base or some other package. If either the method or at least one of the input types is owned by the same package, then there's no problem.
- is the package doing any fragile simulation of conditional dependencies that will interact poorly with precompilation - if so, it should explicitly mark itself as `__precompile__(false)`

Other far less important nitpicky things that I'll comment on but can easily be fixed in future tags:
- using `Pkg.dir` to refer to a package's own location instead of `dirname(@__FILE__)`
- not being specific enough about `VERSION` conditions
- `Pkg.test` making the package dirty or not cleaning up temporary test files

These were off the top of my head, I'll add more if I can think of them. If I've nitpicked about your package in a way that didn't fit in one of the above categories, feel free to refresh my memory.

[...]

PR #9256, opened 2017-05-12
Tag Twiddle.jl v0.3.0

attobot:
Repository: [Ward9250/Twiddle.jl](https://github.com/Ward9250/Twiddle.jl)
Release: [v0.3.0](https://github.com/Ward9250/Twiddle.jl/releases/tag/v0.3.0)
Diff: [vs v0.2.0](https://github.com/Ward9250/Twiddle.jl/compare/8502a7f191f7e19f11e2af9d1b11ea58e8e0f9a6...4318593032bb8777f46be2756d06936bb178f73d)
`requires` vs v0.2.0: no changes
cc: @Ward9250


tkelman:
This is failing its tests, looks like you're missing an export or import?

tkelman:
Mike, are you even looking at these? Please slow down a bit.

MikeInnes:
The CI is passing – anything we require for a release should be automated.

Ward9250:
Sorry, I forgot to qualify the new method names. Twiddle currently doesn't export, you explicitly import or qualify the name of the bit twiddling trick(s) you want. Should I leave this and just open a 0.3.1 hotfix version?

tkelman:
It's currently not, and can we have a discussion and implement some of these checks before insta merging everything without looking at it? Things merged to METADATA get downloaded to users almost instantly if they Pkg.update, and things will break a lot less often with a bit of care.

tkelman:
@Ward9250 yes a patch release will do

Ward9250:
Thanks!

MikeInnes:
We've already discussed it. Metadata isn't supposed to be a way for you to force review on packages. Make it work on Travis or open a PR.

Making sure that packages work is a responsibility for package maintainers to look after, not ours.

tkelman:
What does merging package tags without even letting the package's own CI finish running accomplish?

Incorrect requirements would need to be fixed in place in metadata, and isn't entirely automatable. Can you just slow down?

tkelman:
We now have reverse dependency CI running here which is very useful, but not if you don't even let it run.

ViralBShah:
Yeah, should at least wait for the package CI to go through. Metadata obviously is not a place for package code review which should be done in its own repo.

Hopefully the metadata CI should be ready soon and we will be one step closer to insta merge of PRs.

MikeInnes:
Luckily, incorrect requirements are easy to fix. I'd rather have a few incorrect requirements than let PRs languish for months.

I'll wait for package CI to go through.

tkelman:
With attobot it's now cleaner and less duplication to fix requirements before merging rather than after, otherwise you have to fix them in two places. It's also dangerous to merge a PR if anyone is in the process of revising it, since they may have deleted the git tag since CI ran.

tkelman:
> We've already discussed it

Yes, at length, in https://github.com/JuliaLang/METADATA.jl/issues/6761#issuecomment-253917642, and before that in #2777 - I would greatly appreciate you participating in that discussion before taking matters into your own hands somewhat recklessly and not respecting the existing standards, instead rushing through everything that comes across. Quoting @tbreloff 

> Having a curated source for package acquisition is key to the future of Julia.
> Pkg/METADATA should stay exactly as is, and remain the "super-robust and curated" source for packages. Tony: keep up the good work. Your "standard" and "nitpicky" lists are exactly what we need.

I'd like for there to be an anything-goes complete free for all option that people can opt into for new and experimental work, and Pkg3 should allow for that to be an option with multiple registries, but I'm not hearing from many people that the right level of curation in the current METADATA is zero, Travis-only. At least being careful about dependency requirements and quickly checking for malicious code, along with waiting to let the new reverse CI testing do its job, will result in fewer cases of "`Pkg.update()` broke my code."

MikeInnes:
That's out of context – I think there's widespread agreement that namespaces (or another solution for early stage packages) are needed before a fully-curated metadata makes sense, and we don't have that yet. Once we do there's no problem with having a core "commercially friendly" repo with high standards.

A finite set of clear requirements – like correct version bounds – would be totally fine by me. In practice that's not what happens; we get code review or demands for changes well outside of what's agreed upon, and good PRs get left for weeks. More automation may not be perfect, but it's less bad, and we can improve it as we go along (including taking any suggestions you have, like waiting for CI). Every other package manager I can think of manages just fine with far less than we have so I'm not sure why we would be a special case.

You really haven't noticed any frustrations with the existing system?

tkelman:
Frustration sure, but anything-goes without any spot checks at all is overreacting too far in the other direction. I'll be better about separating metadata-relevant feedback from package-code feedback, criticism received there.

My finite set of requirements is basically as laid out here https://github.com/JuliaLang/METADATA.jl/issues/6761#issuecomment-253893450, and for new tags of existing packages, I do a spot check that those items aren't getting violated in the diff. I'll try to automate a bit more of it, and the reverse-dependency testing is a good step in that direction.

A lot of things stall out when there's no response to feedback from the submitter. Once in a while it's my fault for losing track of something, but I try to make those rare - do point them out when they happen.

What's worse than the current system is having the standards be totally inconsistent depending on who happens to see and merge a PR. Can we collaborate and compromise here and come up with a mutually agreeable, uniformly enforced process?

tkelman:
We should put the standards (and this discussion, for that matter) somewhere visible and clear - I'm at a loss though for where that should be, since without restructuring the METADATA repo the readme here isn't visible. A PR template would do the job, but the attobot workflow makes that a lot less feasible now. An auto-comment bot maybe?

MikeInnes:
Good idea, perhaps attobot itself could post some information when making a PR? Communication is hard, especially as scale goes up, but improving it would help the situation a lot. Registrations have been particularly tricky in the past and I think it would ease things if it was totally clear that e.g. renaming feedback is optional (and even make people more receptive to it).

I can compromise on insta-tagging as long as the manual review has clear scope; e.g. sticking to the require file that's being put into metadata and generic info like CI status. It should be easy for a maintainer to mentally check a couple of boxes and know they won't face pushback.

I will put up a PR for attobot and we can carry on discussing there, or in better channels.

tkelman:
> Communication is hard, especially as scale goes up, but improving it would help the situation a lot.

Yes, very much so. With one or two exceptions you haven't recently merged anything too egregious here that I wouldn't have merged myself (possibly after some optional feedback and giving the submitter a chance to reply), but the lack of communication before this sudden spree of rushed merges concerns me a lot about how long it may take before you do so. If we could try to treat each other, and each other's viewpoints, with a bit more respect and work together here the outcome will be a lot better than if we don't.

Ward9250:
Not wanting to get involved too much in this, as I don't maintain any julia infrastructure of help with metadata currently, but I have to say for all the packages BioJulia releases, we've never had a problem with the current system, and any requests for changes or comments have always been reasonable, I would definitely not say they've ever constituted a code-review (they've been basic sanity checks), and no PR has ever gone stale for us, just our experience for what it is worth.

PR #9272, opened 2017-05-13
Register new package PGFPlotsX.jl v0.1.1

attobot:
Repository: [KristofferC/PGFPlotsX.jl](https://github.com/KristofferC/PGFPlotsX.jl)
Release: [v0.1.1](https://github.com/KristofferC/PGFPlotsX.jl/releases/tag/v0.1.1)
cc: @KristofferC


KristofferC:
Well.. can I get some details about the failed tests..?

tkelman:
click on the red x on the details page - ```could not spawn `pdflatex texfile`: no such file or directory (ENOENT)```

if the package doesn't install its own dependencies in build.jl, then JuliaCIBot currently won't handle it very well. we can probably try reading from .travis.yml and pick some things up that way

KristofferC:
Aha, yep. That makes sense.

tkelman:
@tawheeler any thoughts on there being a second related and very closely named pgf plotting package?

tawheeler:
@tkelman Yes, this reflects https://github.com/sisl/PGFPlots.jl/issues/69, in which @KristofferC is developing a package with lower-level LaTeX syntax support that PGFPlots.jl will sit on top of.

tkelman:
Ah cool. Was the X suggestion in that discussion meant literally, or as a standin for something more differentiating?

KristofferC:
It was me wanting to start coding and not thinking about names.

tkelman:
If the intent is for PGFPlots to later be rearchitected to build on top of this, then maybe PGFPlotsBase ?

KristofferC:
To me, this is a standalone package in the sense that it is written with the intention to be used directly and not only as a "Base"-dependency for another package. I have of course tried to make it flexible enough so that a more high level interfaces can be built on top of it (like potentially PGFPlots), but I don't really want that to influence the name.

KristofferC:
I think this is an OK name, you can talk about it like if you are using the X version or not lol.

PR #9301, opened 2017-05-15
Tag TrafficAssignment.jl v0.4.1

attobot:
Repository: [chkwon/TrafficAssignment.jl](https://github.com/chkwon/TrafficAssignment.jl)
Release: [v0.4.1](https://github.com/chkwon/TrafficAssignment.jl/releases/tag/v0.4.1)
Travis: [Travis Build Status](https://travis-ci.org/chkwon/TrafficAssignment.jl/builds/232406722)
Diff: [vs v0.4.0](https://github.com/chkwon/TrafficAssignment.jl/compare/7aa2e386c398f04f83b8d3a093e0d3fd23a5abb2...f58385b5c53cd3e52c6617c0cdcf4c1f254809e5)
cc: @chkwon

Please make sure that:
- [ ] CI passes for supported Julia versions (if applicable).
- [ ] Version bounds are up to date.


MikeInnes:
Thanks for the tag! A maintainer will check for issues and then merge the PR. Alternatively you can reply with `+ready` to do so yourself.

tkelman:
> Alternatively you can reply with +ready to do so yourself.

What are you talking about, how is this supposed to work?

tkelman:
I get the distinct impression you're ignoring that question. Can we discuss things and work together on this instead of unilaterally changing the way things work here?

MikeInnes:
If people want their prs merged I'll merge them; simple as that.

tkelman:
Even if the tag has requirements that are completely broken and would cause global havok for everyone who does a Pkg.update as soon as you do so?

MikeInnes:
If it's so easy to do that by accident, then we're in a perilous situation indeed; our entire package ecosystem has a bus factor of one, and will likely collapse as soon as you so much as go on holiday.

If this is the case – and I'm somewhat sceptical that it's likely in practice, especially now we have the reverse CI – an alternative approach is to give package authors enough guidance that they are confident get it right, and then trust them to do so. I'm very much open to your suggestions on how we can do that, but if it *must* involve distribution of workload.

tkelman:
The ease with which that can happen depends on what kind of sanity check the requirements and bounds get here before merging. A widely used package could add a handful of dependencies that would transitively pull in a few dozen or hundred other packages without looking that problematic at first glance, and travis and reverse dependency ci might still pass and give it a green checkbox.

I should indeed be better about writing these things down so other people can have the same intuition and apply checks uniformly. Asking how to help with that aspect hasn't been the approach you've taken here.

tkelman:
It's been a long time since we gave out commit access to metadata freely - it is fragile and easy to let through wrong or harmful things if we're not sanity checking them. Can you please stop with the auto comments? It's clearly automated, but infrastructure ​and tools like that should really be done publicly and in the open, with discussion and some consensus. METADATA does global resolution, upgrading everything is the default behavior, and updated packages can execute arbitrary code. This makes things pretty fragile. I'd be fine with adding the first part of what you've been commenting to the attobot message, but the second is a major change in policy that the current METADATA is not suited for.

MikeInnes:
Discussion of these exact issues has gone on for a long time with no progress; it's very unproductive when every suggestion is met with a worst case scenario. If you have alternative ideas for how to preserve quality while lowering the bus factor, I'm all ears, but if it's just picking holes then I'm out.

Allowing package maintainers to, you know, be responsible for their packages and users is the precedent in every other language and package manager I can think of, so it seems perfectly sensible to follow that precedent; the fact that we don't is, AFAICT, a historical accident. If you have a cogent reason why Julia is special, or a community consensus that we shouldn't do that, or any kind of evidence that this very small change is leading to anarchy and chaos, I'll happily bow to that.

tkelman:
CRAN, Homebrew, Debian, etc all have curated central repositories that moderate the release process for the sake of stability and testing. npm, pypi and rubygems (I think) are unmoderated. The Julia community is pretty split about whether curation is good or bad and how much is appropriate. Due to the risks of causing problems with the current metadata global resolution and auto update default Pkg behavior, I think the manual floor needs to be nonzero here until we have more flexibility with Pkg3, or split branches into a staging and release setup.

I have ideas and I would like to work with you and the rest of the community, on both sides of the curation fence, to implement them and automate what we can, but jumping straight to auto merging is too much, too fast. This dramatic change has come very suddenly from you on your own with no warning or public plan, and it's unsettling.

PR #9302, opened 2017-05-15
Update DoubleDouble URL to JuliaMath org

dpsanders:
DoubleDouble moved to JuliaMath.

cc @simonbyrne

### DIFF
--- a/DoubleDouble/url
+++ b/DoubleDouble/url
@@ -1,1 +1,1 @@
-git://github.com/simonbyrne/DoubleDouble.jl.git
+git://github.com/JuliaMath/DoubleDouble.jl.git


simonbyrne:
I believe @tkelman has a script to handle this?

simonbyrne:
Ah, it's so you can use AttoBot (https://github.com/JuliaMath/DoubleDouble.jl/issues/24).

tkelman:
https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/update-redirects.jl but doesn't hurt to do manually

lgtm

tkelman:
(as long as it's done via a repo transfer, so github redirects work on existing remotes, which this was)

be sure to double check that CI services are still running okay with the new repo location

PR #9305, opened 2017-05-15
Bump Stan.jl to 1.1.0

goedman:
Consolidate current state of Stan.jl master before releasing Stan.jl 2.0.0


Stan/versions/1.1.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5.0
+Mamba 0.10
+Gadfly 0.5.2
+BinDeps 0.4.5
```
tkelman: there are no new features between 0.4.3 and 0.4.5, and the bug fix there isn't something that would matter for your uses, so no need to raise the minimum version

goedman: Thanks again for helping out! I usually set the lower bound to the versions that are installed on my system when testing. But I see your point.

goedman: I'm currently testing with your other suggested change (using the upcoming Stan.jl 2.0.0 release) and also if the docs stuff works.

tkelman: I'll be writing up a longer explanation of this soon, but that's not what lower bounds are for. They should be set to the lowest version that would be expected to work. Pkg will usually upgrade things to the latest versions unless told otherwise, but if someone has BinDeps pinned to something less than 0.4.5, this too-high lower bound would mean they would not be allowed to upgrade to this latest version of Stan even though it would actually work fine.

goedman:
Thanks Tony, next up v"2.0.0" for Julia 0.6-RC1. Tonight or early tomorrow!

PR #9328, opened 2017-05-16
Bump Stan.jl to v"2.0.0"

goedman:
Compatible with Julia 0.5 and 0.6-RC1, added Github docs, a couple of breaking changes, hence a major version tag update.


Stan/versions/1.1.0/requires
```diff
@@ -1,5 +1,5 @@
 julia 0.5.0
 Mamba 0.10
 Gadfly 0.5.2
-BinDeps 0.4.3
+BinDeps 0.4.5
```
tkelman: isn't this reverting the change we made?

goedman: I think you are looking at the original 1.1.0 update, which we subsequently changed (just below it)

goedman: I actually was a bit surprised it showed yesterday's pull requests, maybe I should have closed it first. Or forked a brand-new METADATA.

tkelman: I'm looking at the overall diff https://github.com/JuliaLang/METADATA.jl/pull/9328/files

have you tried https://github.com/attobot/attobot ?

goedman: Ok, just set it up for future releases.

tkelman:
Sorry for the delay - you're missing `Compat 0.17` from your require file

goedman:
Thanks Tony, no problem for the delay. Since downgrading the BinDeps version online yesterday, Pkg.update() on Julia v0.5 complains about not being able to fast forward merge (it doesn't say what merge, but I'm assuming it is METADATA.jl) which I think is the root cause why it tried to bump up the BinDeps version level again when I created the PR for Stan.jl v2.0.0. So I'm flushing .julia/v0.5.

Then I'll see if I can somehow fix this missing Compat 0.17 issue. Would missing Compat 0.17 have been caught by autobot? Would it be acceptable if I do a patch release (Stan.jl 2.0.1)?

Thinking a bit more about your explanation for choosing the lowest possible lower bound. Either way I believe there is a testing 'hole'. Shouldn't Travis use that lower bound? I don't think that could easily be integrated into runtests.jl on a local system.

tkelman:
You could modify the requires file in this branch of your metadata fork in place without changing the sha, as long as you make the corresponding change on master before the next tag. Or do a separate patch release in a different pr on its own, and have 2.0.0 be a git tag but not published as a metadata tag. Having old tags with looser or missing requirements isn't good because Pkg could choose to downgrade to them given complicated constraints.

Most, but not all, users will generally be using latest versions of dependencies. So travis testing against latest is probably most representative. We could look into a testing mode that pins package versions to their lower bounds for testing purposes, but that could get a bit complicated. It's a combinatorial space of possible version combinations, and it doesn't need to be absolutely perfectly correct, but a reasonably educated guess for the lowest version that could work is usually closer to right than what you happened to test on.

PR #9338, opened 2017-05-16
Tag DeepDiffs v1.0.0 [https://github.com/ssfrr/DeepDiffs.jl]

ssfrr:


ssfrr:
whoops, forgot to push the tag. it's pushed now.

DeepDiffs/versions/0.1.0/requires
```diff
@@ -1 +1 @@
-julia 0.5
+julia 0.5 0.6-
```
tkelman: is it broken on every single prerelease? maybe better to set a ceiling where it was known to be broken, like alpha, beta, or an rc

ssfrr: It's not known to be broken on every prerelease, but it's not tested on them.

My plan for the half-dozen or so packages I maintain is to set `0.6-` upper bounds on all the old versions and release new versions that support `0.6-`. I guess I could instead put a `0.6` upper bound on the old and `0.6-` lower bound on the new, so that everything's installable on the 0.6 prereleases.

tkelman: the only changes to src here look like deprecation fixes, so I don't think the upper bound is really necessary - depwarns aren't really enough of a reason to rule out installation of this version if someone really needs to pin to it or upper bound it for whatever reason

ssfrr: fair enough - reverted.

ssfrr:
Hmm, JuliaCIBot is complaining that it doesn't support 0.5, but given that it doesn't claim to in `REQUIRE` that should be OK, right?

PR #9344, opened 2017-05-17
Register new package DataValueOperations.jl v0.0.1

attobot:
Repository: [davidanthoff/DataValueOperations.jl](https://github.com/davidanthoff/DataValueOperations.jl)
Release: [v0.0.1](https://github.com/davidanthoff/DataValueOperations.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/DataValueOperations.jl/builds/233088754)
cc: @davidanthoff

Please make sure that:


tkelman:
Is there a discussion somewhere that describes why these are defined as a separate package rather than in DataValues? This would be globally changing behavior depending on whether or not this package is imported, without the dependency being as explicit as if a function or type from a specific package is getting used.

davidanthoff:
I've added a section to the README of both packages that explains the rational for this.

tkelman:
Does it need to be registered if the second package isn't going to exist long term?

davidanthoff:
Realistically, I think this will be around for at least a year. I would much prefer if we could register it, I want to make it as easy as possible for the widest audience to try both approaches. If we make it clone only, I fear I won't get any feedback from folks that couldn't figure that step out...

tkelman:
The issue with doing it this way is if someone wants to limit just to the methods defined in DataValues, that's impossible to enforce if anything else they might load ever decides to depend on something (that depends on something that depends on something...) that depends on DataValuesOperations. There's only ever one DataValues module and DataValue type, and its method tables are global.

davidanthoff:
I guess another approach would be to have a function `load_whitelist()` in DataValues.jl that includes a file with all the stuff that is currently in DataValueOperations.jl. That function would essentially opt the current Julia process into the white list approach. Only problem with that might be precompile, right?

davidanthoff:
Ok, I've found a better way to do this, so this PR can be closed. Someone else will have to do that, I don't have the required access rights to close it.

PR #9347, opened 2017-05-17
adds upper bounds on JuliaAudio packages confirmed not to work on 0.6

ssfrr:


tkelman:
PortAudio looks like it's only broken because Devectorize is broken, right? I guess the chances of there ever being another tag of Devectorize are quite low though.

ssfrr:
possible it could be made to work, though I suspect that the reliance on a pure-julia callback called from another thread and depwarns are a segfaulty combo. I'd rather just clean break and get the next PortAudio version out quickly.

PR #9366, opened 2017-05-17
Use cached urls for travis julia downloads

tkelman:
@staticfloat should we be doing this here, on the website, PkgDev.generate etc too?


staticfloat:
Yeah, looks good to me, although I wonder why we have the `.tar.gz` part not as part of the `$url` ;)

tkelman:
Thanks. Just the website and pkgdev left then, I think

PR #9384, opened 2017-05-18
Tag OIFITS v0.3.1 [https://github.com/emmt/OIFITS.jl]

emmt:
Diff vs v0.3.0: https://github.com/emmt/OIFITS.jl/compare/45193a008004b4129db1f5745c7f9396240f5f0a...acff4b6b24bc241ca52982349b1478028443791f


OIFITS/versions/0.3.1/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+FITSIO 0.8
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

tkelman:
Can Travis CI be enabled here? Looks like this would break the package on 0.4, which is still supported according to the require file

OIFITS/versions/0.3.1/requires
emmt: I guess you mean for Compat? Is there some guidelines to determine the minimal needed version (I was not able to figure out by reading the doc. of Compat).

emmt:
OK I can enable Travis CI, but I do want to make it consistently, where can I find some info. about that?

tkelman:
https://docs.travis-ci.com/user/languages/julia/

OIFITS/versions/0.3.1/requires
tkelman: blame on the readme and checking when the features you're using were introduced is the best way at the moment, or I can help - looks like you're using String and Symbol, anything else? String needs to add a `using Compat: String` otherwise it had a different meaning in 0.4

emmt: Thanks for the hints! I have compiled julia-0.4 and julia-0.6 so that I can check on versions 0.4, 0.5, and 0.6 without waiting for Travis to complete.

emmt:
OK that's done Travis is running by now. Thanks a lot for your help. A last question: when all tests will be successfull, shall I issue a new pull request or is there something smarter to do?

tkelman:
you could modify this pr in place by pushing to this same branch of your metadata fork. or you could try out https://github.com/attobot/attobot which automates much of the process.

emmt:
Hi again,
In principle I have fixed everythings, following your good advices and discovered a bunch of nice tools and facilities ;-) I hope I will not have to bother you next time.
Thanks !

tkelman:
#9413 replaces this as far as I can tell?

emmt:
Yes it fixes the Compat version number I forgot to set

PR #9386, opened 2017-05-18
Pyamg test fix

cortner:


tkelman:
It looks like the changes here work with IterativeSolvers 0.3 but not 0.2.x, and since IterativeSolvers 0.3 is only available for julia 0.5, then this PyAMG change fails tests on julia 0.4. Probably best to make this julia 0.5 only (and set an IterativeSolvers 0.3 minimum version in test/REQUIRE)

cortner:
first time I've done it this way - I hope this is as intended?

tkelman:
mostly, though it'll need to use a new package minor version, 0.1.0

cortner:
Can you explain?

tkelman:
dropping support for a julia version requires using at least a new package minor version

cortner:
ok now?

tkelman:
the IterativeSolvers requirement in test/REQUIRE should have an 0.3 lower bound, but otherwise yes

cortner:
done. hopefully at some point I will understand this system . . .

tkelman:
forgot to push the 0.1.1 git tag? next time around, take a look at https://github.com/attobot/attobot

cortner:
ok, will set this up now

PR #9411, opened 2017-05-20
Tag FilePaths v0.1.0 [https://github.com/rofinn/FilePaths.jl]

rofinn:
Diff vs v0.0.1: https://github.com/rofinn/FilePaths.jl/compare/fe6ef4dd6072b4647b1f6e32608b8d4112988728...b5c8d72a084195af8140452b0d816b3184c2e50b

### DIFF
--- a/FilePaths/versions/0.1.0/requires
+++ b/FilePaths/versions/0.1.0/requires
@@ -1,0 +1,5 @@
+julia 0.5
+Compat 0.17.0
+URIParser 0.1.8
+Glob 1.1.1
+Humanize 0.4.1
--- a/FilePaths/versions/0.1.0/sha1
+++ b/FilePaths/versions/0.1.0/sha1
@@ -1,0 +1,1 @@
+b3c8be670e78dfaf593bba8579e9ed9ebf9be7d0


FilePaths/versions/0.1.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+Compat 0.24.0
```
tkelman: you aren't using anything newer than abstract type, are you?

rofinn: Yeah, I wasn't sure when that was added so I figured setting the minimum version number to what I currently have running seemed like reasonable choice.

rofinn:
FWIW, the JuliaCIBot is failing because the test cases assume that they're being run from an enviroment with the "USER" variable set. I suppose I could change this to call `whoami`, but I'm not sure if that's more likely to be available on all systems.

FilePaths/versions/0.1.0/requires
tkelman: that would prevent installation of this version if anyone has Compat pinned to something earlier than 0.24, unnecessarily since it would work fine with earlier - 0.17 on, if abstract type is the newest thing you're using

rofinn: I'm not sure why someone would have Compat pinned, but I'll change the requirement to 0.17.  Do you have any suggestions about the JuliaCIBot failure given the comment below?

tkelman: could fall back to something if the USER env var isn't set?

rofinn: Yeah, I could also just skip that particular test when USER isn't set for now I guess. I'm already testing the behaviour on travis and I'm not using the USER env var in the actual code.

tkelman: Newer versions of Compat can change behavior, just like any other package. Lower bounds (for all dependencies, not just Compat) should be the lowest version that could work correctly, not the latest thing you happened to test against. When requirements are too loose, it possibly allows installation of version combinations that might not work together. When requirements are too tight, it prevents installation of version combinations that would.

rofinn:
Updated compat min version and decided to skip the test depending on "USER" variable if it isn't set.

tkelman:
It's likely the minimum versions on your other dependencies are too tight as well, since none of those packages have had much more than deprecation fixes lately iirc

rofinn:
Thanks @tkelman! It's possible that the `URIParser` requirement is too high, but I know I needed the updated versions of Glob and Humanize to avoid errors when testing on 0.6. For future reference, is the preference to only use a high enough version to avoid errors, but allow versions of packages that would throw deprecation warnings?

tkelman:
yes, especially since version requirements are across all supported julia versions, shouldn't rule out installation of a working version combination on 0.5 to force upgrades on 0.6 that Pkg will default to anyway.

rofinn:
Alright, thanks for clarifying.

PR #9415, opened 2017-05-20
Tag FastCombinations v0.1.0 [https://github.com/anthonyclays/FastComb…

anthonyclays:
…inations.jl]

Diff vs v0.0.1: https://github.com/anthonyclays/FastCombinations.jl/compare/e58cf68517347a77e26c6846ef1306df1d5cdc6a...a1ea2f471045fd0ed93bb8b0bbf8a4451baf5c9c


tkelman:
This change breaks the package on julia 0.4. To drop support for 0.4, you should change the minimum version requirement in the REQUIRE file.

anthonyclays:
Ah yes, I only tested locally against 0.5 and 0.6. I changed the version requirement to julia 0.5, it should be fixed now

PR #9418, opened 2017-05-20
Tag ODE.jl v0.6.0

attobot:
Repository: [JuliaDiffEq/ODE.jl](https://github.com/JuliaDiffEq/ODE.jl)
Release: [v0.6.0](https://github.com/JuliaDiffEq/ODE.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/ODE.jl/builds/234379050)
Diff: [vs v0.5.0](https://github.com/JuliaDiffEq/ODE.jl/compare/bd3b236acc36c24e53731796d6e644e89c04385e...37e5d7a1fe6c468aa1a12e26b46578e4ddf9d599)
`requires` vs v0.5.0: 
```diff
--- 0.5.0/requires
+++ 0.6.0/requires
@@ -1,4 +1,4 @@
 julia 0.5
 Polynomials
-Compat 0.9
-DiffEqBase 1.0.0
+Compat 0.17.0
+DiffEqBase 1.5.1
```
cc: @ChrisRackauckas

Please make sure that:


tkelman:
I think something here caused a problem with QuantumOptics.jl, it's now freezing on 0.6 - check the most recent commit on https://github.com/JuliaCI/pkg.julialang.org/blob/gh-pages/logs/QuantumOptics_0.6.log

cc @bastikr

ChrisRackauckas:
I just re-ran ODE.jl's tests and did not get these precompilation errors. I have no idea what they are from.

tkelman:
The error in that log is just that something in QuantumOptics is trying to send a RowVector to a function that doesn't know how to handle them, things weren't 0.6-ready. The more problematic issue is that it hasn't finished running that package's tests since May 20th, when it should be running nightly.

ChrisRackauckas:
There was this issue with RowVector in the previous release of ODE.jl

```julia
INFO: Installing ODE v0.5.0
```

Looks like that's it. ODE.jl v0.6.0 fixes that, and was released 16 days ago.

tkelman:
And QuantumOptic's tests have been freezing ever since.

ChrisRackauckas:
The test you pointed to uses v0.5.0 of ODE.jl. I wouldn't've expected that to work on Julia v0.6 at all, and would be surprised if there was a Julia v0.6 test that worked with ODE v0.5. If it's failing on Julia v0.6 with ODE.jl v0.6, I'd need to see the error message because it would not be the same as the one you linked above.

tkelman:
The newer log files are not getting uploaded because they aren't finishing. Almost every other package gets new log files uploaded every day. The fact that the most recent commit for that log file was 16 days ago means there has been a problem ever since.

ChrisRackauckas:
Well the RowVector problem was fixed by https://github.com/JuliaDiffEq/ODE.jl/commit/ff1b4581622103c86808420bd236c90099ace8ec, so I have nothing to go off of without another log file. QuantumOptics does not test on v0.6 while ODE tests pass on v0.6, which indicates to me it must just be something due to that package.

tkelman:
`Pkg.add("QuantumOptics"); Pkg.test("QuantumOptics")`, it gets part way through then freezes

tkelman:
A successful run of that packages' tests looks like https://github.com/JuliaCI/pkg.julialang.org/blob/cfad0830fc32634fa83e6c587ff9d79f0eba4ce8/logs/QuantumOptics_0.5.log

On Julia 0.6 with ODE.jl 0.6.0, it gets to its spectralanalysis testset then freezes. The next testset should have been odedopri.

ChrisRackauckas:
That's not ODE.jl's problem. ODE.jl was preventing this error with its precompilation issue. This is stalling due to an issue in QuantumOptics' internal `odedopri` module. It's not immediately apparent to me why v0.6 breaks their code, but that's not something ODE.jl can provide anyways. That's an internal solver like `ode45` which allows for event handling, which ODE.jl doesn't offer. But DifferentialEquations.jl offers this through OrdinaryDiffEq.jl: the `DP5()` method with its event handling works on v0.6 and could be a replacement to this (and it has stabilized stepsizes among other differences). So I would recommend that change, but really this has nothing to do with ODE.jl.

bastikr:
Yes that's an issue in our ode solver. The problem was that the result from `max(Inf, NaN)` changed from `Inf` to `NaN` in julia 0.6. A pull request that fixes this and makes QuantumOptics.jl completely 0.6 ready is open and will be merged today or tomorrow. A new release is also not too far away, like in a week or so.

We plan to switch to DifferentialEquations.jl soon but we first wanted to create an extensive benchmark suite to make sure that there are no speed regressions. This work is now completed so this will be one of the first things I will look into after this release.

PR #9425, opened 2017-05-21
Deprecate PValueAdjust.jl

dirkschumacher:
2 years ago I wrote a small package PValueAdjust.jl that implemented some simple p-value adjustment methods for multiple testing. The recent MultipleTesting.jl package implements all the methods that are in PValueAdjust.jl and many more (and is actively maintained). In addition, PValueAdjust.jl now also seems to cause problems with Julia 0.5 and some functions break.

I would like to deprecate PValueAdjust.jl in favor of MultipleTesting.jl and will point users to that new package. I have already updated the Github README.


tkelman:
Sure. Though the first two versions will need to have a requires file created that says `julia 0.0- 0.5`

dirkschumacher:
Added the two files

PR #9416, opened 2017-05-20
Tag ReverseDiff.jl v0.1.3

attobot:
Repository: [JuliaDiff/ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl)
Release: [v0.1.3](https://github.com/JuliaDiff/ReverseDiff.jl/releases/tag/v0.1.3)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiff/ReverseDiff.jl/builds/234317209)
Diff: [vs v0.1.2](https://github.com/JuliaDiff/ReverseDiff.jl/compare/b1496d0e09e711e5b7b651b53c48e1f98dee43f3...9bc8b33ecf337f027670f0b34105001d0e9d7991)
`requires` vs v0.1.2: 
```diff
--- 0.1.2/requires
+++ 0.1.3/requires
@@ -1,5 +1,4 @@
 julia 0.5
-DiffBase 0.0.3
-ForwardDiff 0.3.4
-Calculus 0.2.0
+DiffBase 0.0.3 0.3.0
+ForwardDiff 0.3.4 0.5.0
 Compat 0.19.0
```
cc: @jrevels

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #9407, opened 2017-05-19
Tag ForwardDiff.jl v0.5.0

attobot:
Repository: [JuliaDiff/ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
Release: [v0.5.0](https://github.com/JuliaDiff/ForwardDiff.jl/releases/tag/v0.5.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiff/ForwardDiff.jl/builds/234196743)
Diff: [vs v0.4.2](https://github.com/JuliaDiff/ForwardDiff.jl/compare/82dcd98972f8af437d9b59d7083b0a22db9fe7bd...55222a0fe5509c9dc1ecca88c314f0b57f529cd6)
`requires` vs v0.4.2: 
```diff
--- 0.4.2/requires
+++ 0.5.0/requires
@@ -1,6 +1,7 @@
-julia 0.5
-DiffBase 0.0.3
-Compat 0.19.0
-Calculus 0.2.0
+julia 0.6-
+StaticArrays 0.5.0
+DiffBase 0.2.0 0.3.0
 NaNMath 0.2.2
 SpecialFunctions 0.1.0
+RealInterface 0.0.2
+CommonSubexpressions 0.0.1
```
cc: @jrevels

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #9437, opened 2017-05-22
upper bound ForwardDiff at 0.5.0 for downstream packages

jrevels:
I've placed upper bounds on ForwardDiff for the packages [marked by JuliaCIBot](http://juliarun-ci.s3.amazonaws.com/3a284f317fad06108bbef2fd7e5475743ce4a6f0/ForwardDiff_reverse_dependency_tests.html) in #9407.

ref #9416 

cc @ChrisRackauckas

### DIFF
--- a/FiniteElementDiffEq/versions/0.0.1/requires
+++ b/FiniteElementDiffEq/versions/0.0.1/requires
@@ -5,7 +5,7 @@ Parameters 0.5.0
 Plots 0.9.2
 RecipesBase 0.1.0
 GenericSVD 0.0.2
-ForwardDiff 0.2.4
+ForwardDiff 0.2.4 0.5.0
 NLsolve 0.7.3
 ChunkedArrays 0.1.0
 VectorizedRoutines 0.0.2
--- a/FiniteElementDiffEq/versions/0.0.2/requires
+++ b/FiniteElementDiffEq/versions/0.0.2/requires
@@ -5,7 +5,7 @@ Parameters 0.5.0
 Plots 0.9.2
 RecipesBase 0.1.0
 GenericSVD 0.0.2
-ForwardDiff 0.2.4
+ForwardDiff 0.2.4 0.5.0
 NLsolve 0.7.3
 ChunkedArrays 0.1.0
 VectorizedRoutines 0.0.2
--- a/FiniteElementDiffEq/versions/0.0.3/requires
+++ b/FiniteElementDiffEq/versions/0.0.3/requires
@@ -5,7 +5,7 @@ Parameters 0.5.0
 Plots 0.9.2
 RecipesBase 0.1.0
 GenericSVD 0.0.2
-ForwardDiff 0.2.4
+ForwardDiff 0.2.4 0.5.0
 NLsolve 0.7.3
 ChunkedArrays 0.1.0
 VectorizedRoutines 0.0.2
--- a/FiniteElementDiffEq/versions/0.0.4/requires
+++ b/FiniteElementDiffEq/versions/0.0.4/requires
@@ -4,7 +4,7 @@ IterativeSolvers 0.2.2
 Parameters 0.5.0
 RecipesBase 0.1.0
 GenericSVD 0.0.2
-ForwardDiff 0.2.4
+ForwardDiff 0.2.4 0.5.0
 NLsolve 0.7.3
 ChunkedArrays 0.1.0
 VectorizedRoutines 0.0.2
--- a/FiniteElementDiffEq/versions/0.0.5/requires
+++ b/FiniteElementDiffEq/versions/0.0.5/requires
@@ -4,7 +4,7 @@ IterativeSolvers 0.2.2
 Parameters 0.5.0
 RecipesBase 0.1.0
 GenericSVD 0.0.2
-ForwardDiff 0.2.4
+ForwardDiff 0.2.4 0.5.0
 NLsolve 0.7.3
 ChunkedArrays 0.1.0
 VectorizedRoutines 0.0.2
--- a/FiniteElementDiffEq/versions/0.1.0/requires
+++ b/FiniteElementDiffEq/versions/0.1.0/requires
@@ -4,7 +4,7 @@ IterativeSolvers 0.2.2
 Parameters 0.5.0
 RecipesBase 0.1.0
 GenericSVD 0.0.2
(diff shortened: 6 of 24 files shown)


ChrisRackauckas:
What exactly changed? I would be heavily surprised if this affected StochasticDiffEq.jl without affecting OrdinaryDiffEq.jl. They use ForwardDiff in their utils in exactly the same way, so much so that this is actually being refactored to a common library soon:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/OrdinaryDiffEq.jl#L22
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/misc_utils.jl#L14
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/StochasticDiffEq.jl#L17
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/misc_utils.jl#L14

Is there a reason why OrdinaryDiffEq.jl wasn't tested here?

This code uses the `JacobianConfig` in the same way too:

https://github.com/JuliaDiffEq/DiffEqCallbacks.jl/blob/master/src/manifold.jl#L18

Anyways, I would like to know what exactly the change is. Upper bounds are nice, but I'll just make sure a compatibility update is setup to be released ASAP. From what you mentioned in the other PRs, it seems the `AbstractConfig` is what changed?

http://www.juliadiff.org/ForwardDiff.jl/latest/user/api.html#Preallocating/Configuring-Work-Buffers-1

So is it just that the chunksize needs to be specified differently?

jrevels:
> Is there a reason why OrdinaryDiffEq.jl wasn't tested here?

Hmmm I'm not sure which packages JuliaCIBot is testing...

> So is it just that the chunksize needs to be specified differently?

There's that, and you should also now pass the target function you're using into the relevant constructor. You can also pass in `nothing`, which allows you to use it for any target function, but partially opts out of the tagging system. At this point, it would probably make sense to change the title of [this section in the version upgrade guide](http://www.juliadiff.org/ForwardDiff.jl/latest/user/upgrade.html#Setting-Chunk-Size-1) from "Setting Chunk Size" to "Using the AbstractConfig API".


There's a deprecation layer for the API changes, though, so it shouldn't actually cause breakage unless I messed something up (totally possible). Theoretically, there should only be breakage if packages touch ForwardDiff internals, e.g. handling `Dual`s directly. I expect handling `Dual`s directly is not too uncommon, since it's useful to do so when preallocating storage.

tkelman:
I think it bases what to test off of Pkg.dependents, but may be choosing some subset - I'll have to check the code to find what criterion it uses to pick. There are also 3 packages where the latest version doesn't depend on ForwardDiff any more but an earlier version did, and likely several that use ForwardDiff as a test-only dependency. We could alternately do things the older way where we wait for PkgEval to identify problems, but that would expose users to them too, so finding them out in advance where possible is useful.

jrevels:
I could just do a search-and-replace to apply an upper bound to *every* ForwardDiff requirement in METADATA, but that would probably be severe overkill. My hope is that most packages will only be using the documented API, and thus hit the deprecation layer rather than actually break. 

However, ForwardDiff only guarantees that deprecation layer for the previous major release (well, until `v1.0.0` is tagged, the previous minor release), so I suppose older versions of packages could break if they relied on older versions of ForwardDiff and never upper-bounded or upgraded. But I guess those packages would likely have already broken by now anyway. 

As for what to do here, I'll defer to your wisdom @tkelman. I'm assuming ForwardDiff isn't the first package to hit this specific brand of bounding issue? It'd be cool if Pkg3 somehow made this easier. Of course, this seems like a pretty tricky problem...

tkelman:
This isn't the first package to cause breakage downstream, not at all. Though this is one of the first (and definitely the first complicated case) since JuliaCIBot started identifying them ahead of time rather than post-merge PkgEval or a good prediction by the upstream package author identifying the issue. Want to figure out how to make the process as useful as possible, automate some more things and find a good mix between minimizing preventable breakage without making life too difficult for package authors.

ChrisRackauckas:
It was the dual constructors.

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/misc_utils.jl#L7
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/misc_utils.jl#L7

What's the `partials` argument in the `Dual` constructor?

jrevels:
> What's the partials argument in the Dual constructor?

All the `Dual` constructors are [here](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/dual.jl#L10). The `partials` are the partial derivatives stored with the `Dual` instance. They can also be thought of as the coefficients of orthogonal perturbations, i.e. non-interacting epsilon components.

Regarding the code you linked: If you have access to a relevant `AbstractConfig` instance at that point in the code, you can call `eltype` on it to get the `Dual` type rather than hardcoding the type. It might not matter for your case anyway, but you'd then also get the benefit of being protected from perturbation confusion by the tagging system. 

Also, it seems like [this](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/misc_utils.jl#L33):

```julia
jac_cfg = ForwardDiff.JacobianConfig(:DiffEqNLSolve, initial_x, ForwardDiff.Chunk{CS}())
```

...should probably be this (see [docs here](http://www.juliadiff.org/ForwardDiff.jl/latest/user/api.html#ForwardDiff.JacobianConfig)):

```julia
jac_cfg = ForwardDiff.JacobianConfig(permf!, fx2, initial_x, ForwardDiff.Chunk{CS}())
```

We should probably move any further JuliaDiffEq-specific discussion to the relevant repo, feel free to ping me on any issues/PRs there.

tkelman:
I suspect there will be a few more of these, I'll do some tests locally to find out faster than tonight's pkgeval results.

tkelman:
PkgEval beat me to it, looks like there was only one new visible regression here in Tensors.jl https://github.com/JuliaCI/pkg.julialang.org/blob/7602f878130b1f0132a32520e5e92b482d239b69/logs/Tensors_0.6.log

cc @KristofferC @fredrikekre

tkelman:
The existing upper bound on ForwardDiff that Optim is imposing is probably why this wasn't more visible, anything using both Optim and ForwardDiff is held back to 0.4.2 at the moment.

PR #9441, opened 2017-05-22
Pull request/babc8834

sadit:
I added two packages NNS (for near neighbor search) and TextModel (modeling text mostly based on bag of words and several tokenizers)


tkelman:
Hi @sadit. Note the package naming guidelines https://docs.julialang.org/en/release-0.5/manual/packages/#guidelines-for-naming-a-package, a more verbose name would usually be clearer and easier for potential users to guess what the package is for.

It's recommended to enable Travis CI, it looks like you're using quite a bit of Julia 0.6-only syntax which wouldn't work right away on Julia 0.5. And TextModel will need a tag.

sadit:
Hi Tony,

I am new on the package publishing. Is the pull-request discarded? In that case, I need to follow these suggestions and try again, right?

Regards,
Eric

tkelman:
You can modify the existing pull request in place. Though you may find the process easier if you use https://github.com/attobot/attobot instead of PkgDev.

PR #9455, opened 2017-05-23
Pull request/14c25d4a

jgorham:
Hi! I'd like to submit a first version of my package for public use.


tkelman:
Looks like your BinDeps setup isn't working correctly on Linux?

jgorham:
The problem occurs when not using a recent version of gcc. If gcc v5.3.0 or greater is used, I believe everything is fine. Is there a way to require this?

tkelman:
Oh right, you asked that in https://github.com/JuliaLang/BinDeps.jl/issues/299. I'll answer with some ideas there. On Travis you could try something like
```
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
    packages:
    - gcc-5
    - g++-5
```

PR #9457, opened 2017-05-23
Register `xtensor-julia`

SylvainCorlay:


SylvainCorlay:
It seems that CI requires the git repository URL to end with `.jl` which is weird since it is not a requirement for `Pkg.clone`.
Also it does not seem to be specified in the documentation that registered git repos should follow this pattern.

SylvainCorlay:
@tkelman is there any blocker to get this in?

SylvainCorlay:
:sparkles:

tkelman:
whoa, new package registrations should remain open for 24 hours

KristofferC:
Where is this documented?

tkelman:
It's now included in the attobot text for new package registrations that use attobot, but it's been the policy applied for initial feedback in multiple past discussions. I'd add it to the readme if that were visible.

Xtensor/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps 0.4.2
+Compat 0.21
```
tkelman: you don't appear to be actually using Compat anywhere

SylvainCorlay: Thanks, I did at some point and did not remove the line.

PR #9461, opened 2017-05-23
Register QuTiP [https://github.com/goropikari/QuTiP.jl.git]: v0.0.1

goropikari:
QuTiP.jl is a wrapper of QuTiP which is for simulating the dynamics of open quantum systems.


QuTiP/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+PyCall
+PyPlot
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

tkelman:
closing in favor of #9475

PR #9472, opened 2017-05-23
Register new package BioCore.jl v1.0.0

attobot:
Repository: [BioJulia/BioCore.jl](https://github.com/BioJulia/BioCore.jl)
Release: [v1.0.0](https://github.com/BioJulia/BioCore.jl/releases/tag/v1.0.0)
Travis: [Travis Build Status](https://travis-ci.org/BioJulia/BioCore.jl/builds/235434569)
cc: @Ward9250

Please make sure that:

@Ward9250 This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


tkelman:
Is there nothing to test here?

Ward9250:
It's possible this package may get some tests, but perhaps not for the first release(s) - so long as it builds: The primary reason for this package is to introduce function names and abstract types, common to BioJulia packages. Things which, on their own, are difficult to test because they're not really called or run, as it's down to the other packages to extend the definitions and add concrete instances. What functions can be run in this package are rather complicated Ragel and Automa helpers which generate code, and any parsers built using those functions are tested in their respective packages.

PR #9475, opened 2017-05-24
Register new package QuTiP.jl v0.0.1

attobot:
Repository: [goropikari/QuTiP.jl](https://github.com/goropikari/QuTiP.jl)
Release: [v0.0.1](https://github.com/goropikari/QuTiP.jl/releases/tag/v0.0.1)
cc: @goropikari

Please make sure that:

@goropikari This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


QuTiP/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+PyCall 1.12.0
+PyPlot 2.3.2
+Compat 0.25.2
```
tkelman: unless you're using a feature that wasn't in any prior version, this is too tight of a lower bound

goropikari: I'm new to julia package development, so I don't know how to find lower bound. Are there any techniques to check version dependency?

tkelman: Looking at https://github.com/JuliaLang/Compat.jl/blame/master/README.md is the currently recommended way. What features are you using from Compat?

goropikari: Because I didn't know how to show help documents of python functions, I quoted a part of PyPlot package code. But it seems my package doesn't require Compat. I am sorry for confusing you.

tkelman:
It's not a good idea to extend methods like
```
+(a::Number, b::PyCall.PyObject) = b + a
-(a::Number, b::PyCall.PyObject) = b - a
*(a::Number, b::PyCall.PyObject) = b * a
/(a::Number, b::PyCall.PyObject) = b / a
```
on types that aren't from your package - that can globally change the behavior of other packages that use PyCall when yours is imported. That kind of method should be defined in PyCall if anywhere, though the subtraction and division look backwards.

PR #9484, opened 2017-05-24
backport upper bounds on DiffBase for old Celeste, ForwardDiff, and ReverseDiff versions

jrevels:
ref https://github.com/JuliaLang/METADATA.jl/pull/9416

### DIFF
--- a/Celeste/versions/0.1.0/requires
+++ b/Celeste/versions/0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.6
 Compat
 DataFrames
 Distributions
--- a/Celeste/versions/0.2.0/requires
+++ b/Celeste/versions/0.2.0/requires
@@ -1,4 +1,4 @@
-julia 0.5-
+julia 0.5- 0.6
 DataFrames
 Distributions
 DocOpt
--- a/Celeste/versions/0.2.1/requires
+++ b/Celeste/versions/0.2.1/requires
@@ -1,4 +1,4 @@
-julia 0.5-
+julia 0.5- 0.6
 DataFrames
 Distributions
 FITSIO 0.8.4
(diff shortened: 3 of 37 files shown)


tkelman:
Yeah so this is most of the idea, but something that's counterintuitive about this is since not every version of Celeste or ForwardDiff depended on DiffBase, what could happen is Pkg could potentially choose to downgrade Celeste or ForwardDiff to an old version that did not depend on DiffBase, in order to upgrade DiffBase to the latest version. There are a few things we could do about that. What versions of Julia will DiffBase 0.3.0 support?

jrevels:
Ohhh that's tricky.

> What versions of Julia will DiffBase 0.3.0 support?

The plan is for DiffBase v0.3.0's REQUIRE to be the same as [current DiffBase master's `REQUIRE`, which specifies `julia 0.6`](https://github.com/JuliaDiff/DiffBase.jl/blob/master/REQUIRE).

tkelman:
Okay cool, that's what I was hoping you'd say, in that case we might have a reasonably nice solution. If none of those old versions (where old == did not depend on DiffBase) of ForwardDiff or Celeste can load properly on Julia 0.6, we can add a Julia 0.6 upper bound to them. That way we couldn't get into the downgrade-ForwardDiff-to-upgrade-DiffBase situation.

jrevels:
> If none of those old versions (where old == did not depend on DiffBase) of ForwardDiff or Celeste can load properly on Julia 0.6, we can add a Julia 0.6 upper bound to them.

Just applied this change (I didn't check whether each version could load properly on v0.6, though).

tkelman:
Thanks. I'll take a look at those unsatisfiable messages on travis, see what needs propagating.

tkelman:
I pushed an alternate fix. At least for those ForwardDiff versions that had minimum julia versions equal to 0.4.0 release or newer, we can add a fake dependency on DiffBase as another way of achieving the upper bound. Should be pretty harmless as DiffBase doesn't have many dependencies of its own.

PR #9485, opened 2017-05-24
Register Fire [https://github.com/ylxdzsw/Fire.jl.git]: v0.0.2

ylxdzsw:


ylxdzsw:
It should work on 0.5.x. The JuliaCIBot error should be test-only because it is from the `@test_nowarn`, which is not available before 0.6.

PR #9491, opened 2017-05-24
Add upper bound for a Query requirement

davidanthoff:


tkelman:
Since earlier versions of Query didn't depend on IterableTables, what could happen here is Pkg could choose to downgrade Query to an older version in order to get the latest version of IterableTables.

davidanthoff:
Things should be safe in combination with #9493, right? Ideally all of these would be tagged more or less simultaneously.

PR #9498, opened 2017-05-25
Tag Silo v0.2.0 [https://github.com/jgoldfar/Silo.jl]

jgoldfar:
Diff vs v0.1.0: https://github.com/jgoldfar/Silo.jl/compare/15d70dda19d3d9a9a44b5430c51a8b2e3035dc9e...bca41a2d246129042321ddab431188187824898e


Silo/versions/0.2.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.4
+BinDeps
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

jgoldfar: Got it; will look into that. Is there an easy way to fix that on this PR without having to delete the tag on the git side, then have PkgDev re-tag and publish the repo?

tkelman:
really shouldn't be extending `Base.names(file_name::String)` here, that can change the behavior of unrelated code if your package is loaded

jgoldfar:
The file in which `Base.names` is extended is code from the previous version of the package which won't be loaded at all if `using Silo` is called (it has been moved out of any file included in the module); the high-level interface has been slightly reduced in this version to remove this signature.

Silo/versions/0.2.0/requires
tkelman: if the sha doesn't need changing, could modify the requires file in place in this branch of your metadata fork, as long as you make the corresponding change on master of the package too before the next tag

PR #9512, opened 2017-05-25
Tag NLsolve.jl v0.10.0

attobot:
Repository: [JuliaNLSolvers/NLsolve.jl](https://github.com/JuliaNLSolvers/NLsolve.jl)
Release: [v0.10.0](https://github.com/JuliaNLSolvers/NLsolve.jl/releases/tag/v0.10.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/NLsolve.jl/builds/236026924)
Diff: [vs v0.9.1](https://github.com/JuliaNLSolvers/NLsolve.jl/compare/ced92f85612b4c62fda71e87c6e60eaa99a2acd3...cb009c319ec2398ce15ce29057bba14112013020)
`requires` vs v0.9.1: 
```diff
--- 0.9.1/requires
+++ 0.10.0/requires
@@ -1,8 +1,7 @@
-julia 0.4
+julia 0.6-
 Optim
 Calculus
 Distances
-ForwardDiff 0.3.0
-Compat 0.8.4
-LineSearches 0.1.2 1.0.0-
+ForwardDiff 0.5.0
+LineSearches
 DiffBase
```
cc: @KristofferC

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #9518, opened 2017-05-25
Upper bound early JuMP versions to 0.6.0-pre

tkelman:
these all used lower-case `symbol()` which no longer exists

ref https://github.com/JuliaLang/METADATA.jl/pull/9512#issuecomment-304057632, cc @mlubin @joehuchette, the existence of ForwardDiff 0.5.0 is causing issues since the only JuMP versions that don't have ForwardDiff upper bounds are the old ones that don't depend on ForwardDiff at all.


mlubin:
Ok then

PR #9523, opened 2017-05-25
Pull request/c06a15cb

lanaperisa:


tkelman:
Since you renamed the package to `TensorToolbox`, should change the `using Tensors` to `using TensorToolbox` in `test/runtests.jl`

lanaperisa:
Trying to do all this while super busy was not a good idea, sorry about all the mistakes. So the idea is to register my package under TensorToolbox name to show clear analogy to Matlab Tensor Toolbox, by first copying Tensors to TensorsToolbox and then setting Tensors to deprecated. 

I've went through the documentation properly now, fixed the changes in the TensorToolbox package, but I can't figure out how to commit changes now, or how to push to the pull-request/c06a15cb branch on lanaperisa/METADATA.jl. When running 
`git checkout pull-request/c06a15cb`
from local .julia/v0.5/METADATA, I get
error: pathspec 'pull-request/c06a15cb' did not match any file(s) known to git

Kind of same problem as [here](https://discourse.julialang.org/t/how-to-update-pr-to-metadata-jl/1550/5). 

Or do you want me to start over using AttoBot? 

Thanks.

tkelman:
There's now a different separate package registered under the Tensors.jl name (which probably should have been more specific to reduce the chance of conflicts like this), so that should be left alone in metadata. The order in which you have merge commits here makes it okay for now, but this branch is a bit messy. Up to you whether to try to clean it up, or start from scratch with attobot.

You're probably missing a `git remote add -f lanaperisa https://github.com/lanaperisa/METADATA.jl` before you can checkout from or push modifications to your fork.

lanaperisa:
OK, starting from scratch.

PR #9524, opened 2017-05-25
Tag CUSPARSE.jl v0.6.0

attobot:
Repository: [JuliaGPU/CUSPARSE.jl](https://github.com/JuliaGPU/CUSPARSE.jl)
Release: [v0.6.0](https://github.com/JuliaGPU/CUSPARSE.jl/releases/tag/v0.6.0)
Diff: [vs v0.5.0](https://github.com/JuliaGPU/CUSPARSE.jl/compare/0f31aa131ee5a4f988dbd17fe38db84a2ebdd489...66b699cff194975fcfe916e728fa54873e3ea6b7)
`requires` vs v0.5.0: 
```diff
--- 0.5.0/requires
+++ 0.6.0/requires
@@ -1,2 +1,3 @@
 julia 0.5
 CUDArt
+Compat
```
cc: @kshyatt

Please make sure that:


CUSPARSE/versions/0.6.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+CUDArt
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

tkelman: looks like 0.18 for `@compat const AbstractCudaSparseVector{Tv}` is the newest thing you're using

(and any of the `const foo = bar` where there's no type parameter on the LHS don't need `@compat` at all)

kshyatt:
I've tested this on 0.5 and 0.6 locally and it works. The CI build shown is on an old version, hopefully we can run it again later today?

PR #9531, opened 2017-05-26
Tag Optim.jl v0.7.10

attobot:
Release: [v0.7.10](https://github.com/JuliaNLSolvers/Optim.jl/releases/tag/v0.7.10)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/Optim.jl/builds/236272962)
Diff: [vs v0.7.8](https://github.com/JuliaNLSolvers/Optim.jl/compare/250b50ab5f9bb87576c623fa20fdafb981ce40a1...6542d33fe16cbb1c989f51faad530e9fb249d893)
`requires` vs v0.7.8: 
```diff
--- 0.7.8/requires
+++ 0.7.10/requires
@@ -1,6 +1,8 @@
-julia 0.4
-Calculus
-ForwardDiff 0.3.0 0.5.0
+julia 0.5
 PositiveFactorizations
 Compat 0.18.0
-LineSearches 0.1.2 1.0.0-
+LineSearches 2.1.0
+Calculus
+NLSolversBase
+ForwardDiff 0.3.0 0.5.0
+ReverseDiff 0.0.3
```
cc: @pkofod

Please make sure that:


tkelman:
if there's no code change and only requirements, we could loosen the bound in place by modifying the existing metadata copy of the requires file in place

pkofod:
> if there's no code change and only requirements, we could loosen the bound in place by modifying the existing metadata copy of the requires file in place

sure, that would work. This was simply to fix the downgrades of Optim you described. I'll get to it a little later today.

tkelman:
looks like the idea is that v0.7.8 does work fine with ForwardDiff 0.5, and doesn't need an upper bound after all?

pkofod:
> looks like the idea is that v0.7.8 does work fine with ForwardDiff 0.5, and doesn't need an upper bound after all?

exactly; I'm not sure why that bound was placed there

pkofod:
though https://travis-ci.org/JuliaNLSolvers/Optim.jl/jobs/236274927#L398

pkofod:
Looking at http://juliarun-ci.s3.amazonaws.com/347c49c08a6aeeff165b9ffebe71a00002c1ad3c/Optim_reverse_dependency_tests.html could it be that it's needed for v0.6 to work?

tkelman:
Oh it is actually breaking. But ForwardDiff 0.5.0 is julia 0.6-only

pkofod:
Will your ForwardDiff patch fix this, or is there still something we need to do?

tkelman:
my commits have been very minor changes, https://github.com/JuliaDiff/ForwardDiff.jl/commit/0e0a5e1e72fe213b0c8298dcb6c95057ee09b2e1 might make a difference though. Test Optim against ForwardDiff master on rc2 and see if there's any difference?

pkofod:
Okay, so I just have to understand something. Is this downgrade thing only on v0.6 ? @KristofferC you said you had observed it.

KristofferC:
For example https://travis-ci.org/JuliaNLSolvers/NLsolve.jl/jobs/236214083#L190

tkelman:
Optim 0.6.x can use ForwardDiff 0.2.x on Julia 0.4, Optim 0.7.x can use ForwardDiff 0.3.x on Julia 0.4 or 0.5, and Optim 0.7.8 can use ForwardDiff 0.4.x on Julia 0.5. So only on Julia 0.6 or 0.7, which are the only places where ForwardDiff 0.5.0 is available, would you hit this downgrading to Optim 0.5.0 in order to upgrade ForwardDiff to 0.5.0.

pkofod:
Okay. I propose we close this then. I know v0.6 is around the corner, but we're tagging a new version soon. That one will have some breaking changes that we will try to smooth out using some helper functions that checks if people used the old syntax. We can then tag a new version after that, that only works on v0.6 and uses the new ForwardDiff. That should resolve all conflict possibilities  (bold claim, I know...)

PR #9532, opened 2017-05-26
Pull request/23ec4e65 OOP.jl

ipod825:
This package aims to provide macros that let programmers write in object-oriented style.


tkelman:
Hi @ipod825. Note the naming guidelines at https://docs.julialang.org/en/release-0.5/manual/packages/#guidelines-for-naming-a-package, it's recommended to use a bit more descriptive names rather than acronyms where possible. Something like ObjectOrientedMacros or OOPMacros might make it more obvious what the package is for.

You should add a top-level `test/runtests.jl` file that includes the rest of your test files, so `Pkg.test` works.

PR #9562, opened 2017-05-28
Tag Etcd v0.1.0 [https://github.com/rofinn/Etcd.jl]

rofinn:
Diff vs v0.0.1: https://github.com/rofinn/Etcd.jl/compare/a8829eac75d9ad501d118d2fe0d03a034fe24fca...620e09eaada17aa6548c9f8ac30848dd364239bd

NOTE: Since the original author appears to be unresponsive this PR changes the pkg url.


tkelman:
Because this isn't a fork, anyone who already has the package installed will probably not be able to get this update. It's a long-standing bug in Pkg that changing the repo url doesn't update existing clones' remotes, so it really only works for repo transfers where github puts redirects in place. It may be better to make this a new package.

rofinn:
In this case that behaviour would still work because the versions of julia supported by each version are mutually exclusive.

v0.0.1: `julia 0.2- 0.4`
v0.1.0: `julia 0.5`

As a result, there shouldn't ever be a situation where someone has already installed `Etcd` and can install the updated version.

tkelman:
Ah yes you're right, hadn't checked that the package has been deprecated in metadata for some time.

tkelman:
HttpCommon should be listed explicitly in REQUIRE though since it's imported, it's not guaranteed that every future version of Requests will always depend on HttpCommon

rofinn:
Good point, I've updated the REQUIRE file and the sha1 is pointing to the updated tag revision.

PR #9567, opened 2017-05-28
Add upper bounds on NamedTuples in Query and IterableTables

davidanthoff:
I mark all the versions that don't work with julia 0.6- as such.

I also upper bound NamedTuples so that we can phase in the next breaking change in NamedTuples.

### DIFF
--- a/IterableTables/versions/0.0.1/requires
+++ b/IterableTables/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.5
-NamedTuples 2.0.0
+NamedTuples 2.0.0 3.0.0
 Requires 0.3.0
--- a/IterableTables/versions/0.0.2/requires
+++ b/IterableTables/versions/0.0.2/requires
@@ -1,3 +1,3 @@
 julia 0.5
-NamedTuples 2.0.0
+NamedTuples 2.0.0 3.0.0
 Requires 0.3.0
--- a/IterableTables/versions/0.1.0/requires
+++ b/IterableTables/versions/0.1.0/requires
@@ -1,4 +1,4 @@
 julia 0.6.0-rc1
-NamedTuples 3.0.2
+NamedTuples 3.0.2 4.0.0
 Requires 0.4.3
 DataValues 0.0.3
--- a/Query/versions/0.4.0/requires
+++ b/Query/versions/0.4.0/requires
@@ -1,5 +1,5 @@
 julia 0.5
-NamedTuples 2.0.0
+NamedTuples 2.0.0 3.0.0
 DataStructures 0.4.5
 Requires 0.3.0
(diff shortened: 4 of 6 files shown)


tkelman:
`0.6-` isn't good to use as an upper bound, it's exclusive so should use a known-broken upper bound rather than the very first thing after 0.5 branched

davidanthoff:
I'm not going to look through early builds of 0.6 to see when it broke these packages. I do know that things don't work on 0.6 alpha, that is when I started testing. So 0.6- seems like a pragmatic choice to me. The only negatively affected would be people running early 0.6 builds, which is probably no one?

As a matter of policy, I think it should be left to the package author to decide about restrictive bounds like this. If I don't want to support early builds of the 0.6 branch then that should be my choice.

tkelman:
Why do you want to prevent installation of these package versions on Julia versions where they would work fine? Bounds that are too restrictive cause a resolver failure or inability to install the package at all, which is worse than allowing installation of a maybe-broken version of the package.

tkelman:
This would make `Pkg.add("Query")` or `Pkg.add("IterableTables")` error on any pre-rc1 version of 0.6, which would make bisecting any base issues triggered in either of these packages more trouble than it needs to be.

davidanthoff:
Should I make the bound 0.6.pre? Or 0.6.0.pre-alpha? I know for sure that things don't work on alpha and newer.

tkelman:
I'd recommend setting the upper bound for these early versions to be the same as the lower bound for the most recent version, at least if you're going to apply the upper bound to every single past version. The julia upper bounds aren't usually strictly necessary unless there's some other situation you're trying to prevent the resolver from doing. If there's a newer version available that supports julia 0.6 then that's what will be used in most cases.

davidanthoff:
Alright, I just have the bounds on NamedTuples in this PR now, which solves a concrete problem we have with the breaking change in NamedTuples.

tkelman:
The upper bound should apply to the past versions of Query and IterableTables too, right? Query 0.4.0 and 0.4.1, and IterableTables 0.0.1 and 0.0.2 don't currently upper-bound their NamedTuples dependency

davidanthoff:
None of those versions work on julia 0.6, and NamedTuples went julia 0.6 only starting with v3.0.0, so this shouldn't matter. Can we merge this here now? If someone else wants to sort things out for other versions, great, but I don't think its worth the effort and time (and so I won't do it).

tkelman:
Applying an upper bound only to the most recent version accomplishes very little. Pkg is allowed to downgrade Query or IterableTables in order to upgrade NamedTuples.

davidanthoff:
Ok, now with upper bounds on NamedTuples everywhere.

PR #9570, opened 2017-05-28
Register new package ZipCode.jl

equinetic:
Repository: [equinetic/ZipCode.jl](https://github.com/equinetic/ZipCode.jl)


quinnj:
Great documentation! Looks like a great new package.

PR #9575, opened 2017-05-29
Register new package PtFEM.jl v0.0.3

attobot:
Repository: [PtFEM/PtFEM.jl](https://github.com/PtFEM/PtFEM.jl)
Release: [v0.0.3](https://github.com/PtFEM/PtFEM.jl/releases/tag/v0.0.3)
Travis: [Travis Build Status](https://travis-ci.org/PtFEM/PtFEM.jl/builds/237009055)
cc: @goedman

Please make sure that:

@goedman This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


PtFEM/versions/0.0.3/requires
```diff
@@ -0,0 +1,12 @@
+julia 0.5
+Documenter 0.10.0
+LightXML 0.4.0
+#Plots 0.10.3
+#GR 0.19.0
+Codecs 0.2.0
+Interpolations 0.3.6
+DataTables 0.0.2
+WriteVTK 0.5.1
```
tkelman: you're relying on some feature that wouldn't have worked with any previous version of the package?

goedman: Tony, you mean I shouldn't be using lower bounds at all unless I know for a fact I need at least that lower bound version or higher for a package my package depends on? Amazing I got away with my total misconception how to use this feature for this long, even after your feedback for Stan.jl. I'll fix it. Thanks again for your help!

tkelman: Getting the exactly accurate lower bound can be more effort than it's usually worth, but a few minutes looking through blame on when functions you're using were introduced, what changed between different versions of the dependencies, etc to get a rough guess is usually better and closer to right than either no bound at all or requiring the most recent release.

Lower bounds are saying that no earlier versions will work at all, and Pkg should not allow simultaneous installation of a lower version of the dependency at the same time as this version of this package. If you leave off the lower bound, it's saying that any released version of the dependency should work. If that's not quite right, the consequences are that someone who has pinned the dependency to an older version, or has it held back due to *upper* bounds somewhere, may be allowed to install version combinations that won't work together.

If your bounds are too tight on the other hand, then what will happen is anyone who has the dependency package pinned or bounded to something earlier than your minimum version either won't be allowed to install your package at all, or would stay held back to some older version of your package that didn't have as tight of a lower bound. So bounds too tight would prevent installation of version combinations that would actually work together.

There are different consequences to having bounds too loose vs too tight - too loose could result in errors from the package, too tight could result in errors or unintuitive behavior (not using the newest versions of things) from Pkg version resolution. I'll be writing all this up in a doc section soon, it's not very well explained.

goedman: In the past I tried to minimize possible causes of problems for users of my package by making the bounds (too) tight. By making them possibly too wide, I expect users will create issues if they run into a problem. That's ok and I think is part of the responsibility of any package creator/maintainer. If I can't work around that issue I will introduce/update a lower bound, (initially in the master branch) until the next release. With attobot more frequent releases are possible although it unfortunately means more work for you and a few others to review them.

The other aspect I have clearly struggled with is dependencies on packages like Plots, Mamba, GR, Gadfly which prevented me from testing on upcoming Julia releases. So I am now taking the approach not including these in REQUIRE and not using these in runtests.jl. That of course lowers coverage etc. I'm not sure if you want to touch on that as well in your write-up but just thought to mention it.

goedman: Editing the release did not update the REQUIRE file properly. I have delete the release and created a new one with the same version (v0.0.3).

goedman:
Thanks Tony!

PR #9592, opened 2017-05-29
remove unnecessary FiniteElementDiffEq upper bound

ChrisRackauckas:


tkelman:
Thanks. Same for Stochastic?

PR #9594, opened 2017-05-29
Tag GLAbstraction.jl v0.3.3

attobot:
Repository: [JuliaGL/GLAbstraction.jl](https://github.com/JuliaGL/GLAbstraction.jl)
Release: [v0.3.3](https://github.com/JuliaGL/GLAbstraction.jl/releases/tag/v0.3.3)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGL/GLAbstraction.jl/builds/237281719)
Diff: [vs v0.3.2](https://github.com/JuliaGL/GLAbstraction.jl/compare/fb559abd9ddaafa1bbdd8ce6910029c116fb5470...013542a54acfc4d7400620053c5ca50c504cf32a)
`requires` vs v0.3.2: 
```diff
--- 0.3.2/requires
+++ 0.3.3/requires
@@ -1,12 +1,15 @@
-julia 0.5
+julia 0.5 0.6.0
 
-Quaternions
+
+GeometryTypes 0.2 0.3
+Reactive 0.3.0 0.3.7 0.5-
+
 FixedSizeArrays
-GeometryTypes
 ModernGL
-Reactive
+
 FixedPointNumbers
 ColorTypes
 Compat 0.8.0
 FileIO
 GLFW
+Quaternions
```
cc: @SimonDanisch

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #9595, opened 2017-05-29
Tag GLWindow.jl v0.3.4

attobot:
Repository: [JuliaGL/GLWindow.jl](https://github.com/JuliaGL/GLWindow.jl)
Release: [v0.3.4](https://github.com/JuliaGL/GLWindow.jl/releases/tag/v0.3.4)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGL/GLWindow.jl/builds/237282586)
Diff: [vs v0.3.3](https://github.com/JuliaGL/GLWindow.jl/compare/5be338cc71c39c3a5a9f202321c2fa9fbcc1767c...080f476c7b3897d6ec356c321a16d088a090016c)
`requires` vs v0.3.3: 
```diff
--- 0.3.3/requires
+++ 0.3.4/requires
@@ -1,10 +1,12 @@
-julia 0.5
+julia 0.5 0.6.0
 
+Reactive 0.3.0 0.3.7 0.5-
+GLAbstraction 0.3.1 0.4
+GeometryTypes 0.2 0.3
+
+GLFW
 ModernGL
-GLAbstraction 0.3.1
-GLFW
-Reactive
-GeometryTypes
+
 Compat 0.7.15
 ColorTypes
 FixedPointNumbers
```
cc: @SimonDanisch

Please make sure that:


SimonDanisch:
battle plan:
merge this and #9594, and then backport the require for GLVisualize/GLWindow/GLAbstraction...
After that I should be able to release a version that works on 0.6 only

GLWindow/versions/0.3.4/requires
```diff
@@ -0,0 +1,14 @@
+julia 0.5 0.6.0
```
tkelman: same comment as in #9594

GLWindow/versions/0.3.4/requires
```diff
@@ -0,0 +1,14 @@
+julia 0.5
+
+Reactive 0.5
+GLAbstraction 0.3.1 0.5
+GeometryTypes 0.2 0.4
```
tkelman: will GeometryTypes 0.4 also break the earlier versions of this package?

SimonDanisch: GeometryTypes 0.4 is 0.6 only and this release is 0.5 only

tkelman: there's no julia upper bound here, as discussed earlier

SimonDanisch: Yes? Didn't you convince me, that the bound isn't a good idea?

tkelman: I did. But that means "this release is 0.5 only" is not correct

SimonDanisch: So what do you want me to do? It is 0.5 only and you didn't want me to put in an upper bound...

tkelman: what breaks first if you try to install this version of the package on 0.6? will it work at all? would it be better to use GeometryTypes 0.3.x? if you have this upper bound on GeometryTypes but only on this latest package version, then Pkg can downgrade GLWindow to 0.3.3 in order to upgrade GeometryTypes to 0.4.

SimonDanisch: Yeah well, I guess I won't ever get around this problem without backporting the bounds.
Also, on 0.6 it should hopefully just install the newest version, which should be 0.6 compatible.

tkelman: backporting the bounds would solve the problem. is there an 0.6-compatible version yet? it won't install the newest version if bounds don't allow it to.

SimonDanisch: There is, in branches... And I want to get this tagged to finally merge them and tag a 0.6 compatible version

tkelman: If GeometryTypes 0.4 will break all of the past tags (in what way?), then it would be best to add the bounds to the past tags' requires files.

SimonDanisch: GeometryTypes 0.4 is 0.6 only, so on the 0.5 only version of GLWindow it will obviously be broken ;)

tkelman: If there's going to be an 0.6 version of GLWindow soon where it's not broken, then applying an upper bound only to a single 0.5 version of GLWindow doesn't serve much purpose. Applying the bounds to all of the version would serve a purpose though since there aren't julia upper bounds on GLWindow, do these existing GLWindow versions work at all on julia 0.6?

PR #9597, opened 2017-05-29
Register Eirene [https://github.com/Eetion/Eirene.jl.git]

Eetion:
…l.git]

### DIFF
--- a/Eirene/url
+++ b/Eirene/url
@@ -1,0 +1,1 @@
+https://github.com/greg.henselman@gmail.com/Eirene.jl.git


tkelman:
the email in the url should be replaced with a github username. is this a publicly visible repo?

Eetion:
Thank you!  I think the repo is publicly visible now.

Eirene/url
```diff
@@ -0,0 +1 @@
+https://github.com/greg.henselman@gmail.com/Eirene.jl.git
```
tkelman: you'll need to update the url here, and tag a version

tkelman:
See https://docs.julialang.org/en/release-0.5/manual/packages/#generating-the-package, or look at the structure of existing packages, for how a package should be structured. You should declare your dependencies with a REQUIRE file, put your julia code in a file named `src/Eirene.jl` that defines a `module Eirene`, and add some unit tests in a file named `test/runtests.jl`.

jingpengwu:
@Eetion I suggest using attobot, it is really easy to release code!

Eetion:
@tkelman and @jingpengwu, thank you!  I've been a bit swamped so progress is slow, but will follow up with attobot.  best.

MikeInnes:

Eetion:
Thanks for the message - I've been tied up the past few months but do plan to continue, hopefully starting mid November.

PR #9607, opened 2017-05-30
Register new package NodeJS.jl v0.0.1

attobot:
Repository: [davidanthoff/NodeJS.jl](https://github.com/davidanthoff/NodeJS.jl)
Release: [v0.0.1](https://github.com/davidanthoff/NodeJS.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/NodeJS.jl/builds/237687571)
cc: @davidanthoff

Please make sure that:


davidanthoff:
Hm, not sure what is going wrong here, the travis build works... It also seems that the [first test](https://github.com/davidanthoff/NodeJS.jl/blob/master/test/runtests.jl#L4) seems to pass and only the [second](https://github.com/davidanthoff/NodeJS.jl/blob/master/test/runtests.jl#L5) one fails?

I wonder whether that is somehow related to [this](https://github.com/davidanthoff/NodeJS.jl/blob/master/deps/build.jl#L35) symlink...

@tkelman is there some way to run these tests in the same environment as the JuliaCIBot, but locally? Debugging this by triggering the JuliaCIBot seems quite painful.

tkelman:
Try using this in an environment that doesn't have a different version of node already installed, npm assumes it's on the path.

davidanthoff:
Ah, good catch, tahnks, it was actually never using the private node in any of the tests I ran! I now changed it and tested on Windows and Ubuntu machines without node or npm and it seems to work. Lets see what JuliaCIBot says.

davidanthoff:
I think this can be merged now, I believe I addressed all suggestions and all tests seem to pass.

PR #9608, opened 2017-05-30
Tag GeometryTypes.jl v0.3.0

attobot:
Repository: [JuliaGeometry/GeometryTypes.jl](https://github.com/JuliaGeometry/GeometryTypes.jl)
Release: [v0.3.0](https://github.com/JuliaGeometry/GeometryTypes.jl/releases/tag/v0.3.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGeometry/GeometryTypes.jl/builds/237695276)
Diff: [vs v0.2.2](https://github.com/JuliaGeometry/GeometryTypes.jl/compare/2df57243de3c8b702d97ec8ee6c062a3a2a76b8c...60a9a5ebfb8ce530d520c7565ce6117ef5ddf050)
`requires` vs v0.2.2: 
```diff
--- 0.2.2/requires
+++ 0.3.0/requires
@@ -1,5 +1,5 @@
-julia 0.4
-FixedSizeArrays
+julia 0.6-
+StaticArrays
 ColorTypes
-Compat 0.7.15
-Iterators
+Compat 0.18
+FixedPointNumbers
```
cc: @SimonDanisch

Please make sure that:


GeometryTypes/versions/0.3.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.6-
```
tkelman: should use 0.6.0-pre, since where syntax wouldn't work on early dev versions, better to keep using the 0.5-compatible versions of the package there​

since both 0.4 and 0.5 are being dropped, it would be worth incrementing the minor version by 2, just in case it becomes necessary to make 0.5-only bugfix releases at some point

andyferris: > since both 0.4 and 0.5 are being dropped, it would be worth incrementing the minor version by 2, just in case it becomes necessary to make 0.5-only bugfix releases at some point

Yes, I can say from experience that this is a good idea.

SimonDanisch: makes sense, thanks!

SimonDanisch:
can be closed: https://github.com/JuliaLang/METADATA.jl/pull/9624

PR #9636, opened 2017-06-02
Register new package StructJuMP.jl v0.0.1

blegat:
Repository: [StructJuMP/StructJuMP.jl](https://github.com/StructJuMP/StructJuMP.jl)
Release: [v0.0.1](https://github.com/StructJuMP/StructJuMP.jl/releases/tag/v0.0.1)
Travis: [Build Status](https://travis-ci.org/StructJuMP/StructJuMP.jl)
cc: @joehuchette @blegat @ktarplee @kibaekkim @cnpetra @michel2323 @fqiang @emreyamangil

### DIFF
--- a/StructJuMP/url
+++ b/StructJuMP/url
@@ -1,0 +1,1 @@
+git://github.com/StructJuMP/StructJuMP.jl.git
--- a/StructJuMP/versions/0.0.1/requires
+++ b/StructJuMP/versions/0.0.1/requires
@@ -1,0 +1,4 @@
+julia 0.5
+JuMP 0.17
+MathProgBase
+ReverseDiffSparse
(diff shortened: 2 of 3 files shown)


tkelman:
Since MathProgBase and ReverseDiffSparse are both imported directly in `src/StructJuMP.jl`, they should be listed explicitly in REQUIRE

tkelman:
Why is this changing the behavior of `JuMP.parseNLExpr_runtime`? That's type piracy, and could change the results of unrelated code that used JuMP if it happened to call through that method signature.

joehuchette:
The piracy is intentional and necessary, barring some nontrivial changes to JuMP nonlinear parsing code. It won't change behavior for any valid JuMP models, only ones containing variables in expressions that are "owned" by other models, which would result in errors otherwise.

PR #9637, opened 2017-06-02
New URL and release of ROOT.jl

oschulz:
* Changes the URL of ROOT.jl - ROOT.jl has moved to the JuliaHEP organization. 
* New release, Cxx.jl-based, ROOT-6 and Julia v0.6 compatible.

Both changes are in agreement with the original author, Joosep Pata (who's a co-owner of JuliaHEP).

### DIFF
--- a/ROOT/url
+++ b/ROOT/url
@@ -1,1 +1,1 @@
-git://github.com/jpata/ROOT.jl.git
+git://github.com/JuliaHEP/ROOT.jl.git


tkelman:
This won't work for anyone who has previously installed the package - repo moves generally need to be done as repository transfers so github will put redirects in place. Otherwise any existing clones will still have their remotes pointing to the old repository, and won't be able to access this update.

oschulz:
> This won't work for anyone who has previously installed the package.

I understand - would it be possible to make an exception in this case? The previous state or ROOT.jl was more of a proof-of-concept, and it's unlikely that there are active users.

tkelman:
If @jpata is an owner on the organization, then he should be able to do a repo transfer. You could rename your new version to JuliaHEP/ROOT.jl-new, transfer the existing jpata/ROOT.jl over to JuliaHEP/ROOT.jl, then merge all the new content from JuliaHEP/ROOT.jl-new into JuliaHEP/ROOT.jl. It's not like you have any issues or anything significant other than the git history in the new repo, so you could even delete the JuliaHEP/ROOT.jl-new repo once all its commits are merged in.

oschulz:
Ok, I've renamed it to JuliaHEP/ROOT.jl-new. @jpata, would it be Ok if we do a move, like @tkelman suggests? I you move ROOT.jl to JuliaHEP, I can take care of the rest, afterwards I'll remove ROOT.jl-new. The whole history will be preserved in any case.

jpata:
done: https://github.com/JuliaHEP/ROOT.jl

oschulz:
Thanks, @jpata!!

@tkelman: I pushed the release back into the move repo. Hope all is fine, now?

oschulz:
Fixed travis-ci failure.

tkelman:
If `using Cxx` is in `src/ROOT.jl` then it needs to be listed in REQUIRE

oschulz:
> If using Cxx is in src/ROOT.jl then it needs to be listed in REQUIRE

@tkelman, I understand the reasons, but is there any possibility of waiving this requirement until we have Pkg3? Cxx is kinda unique, because it takes an extremely long time to build (since it has to build LLVM and Clang), and after build it's almost 1GB. So it's an ideal candidate for system-wide installation (I use "share/julia/site"), especially in a multi-user context.

I have a bunch of users here, all with NFS-based home directories, and if there we to all install Cxx in their home ... not good. We actually use [Singularity](http://singularity.lbl.gov/) containers which have ROOT, Anaconda, Julia, Cxx.jl and a lot of other stuff our people need baked in, and serve them via CephFS. This keeps all metadata operations on the clients, which is very beneficial. Unfortunately, while Julia will happily use the baked in Cxx.jl (as it's on the default `LOAD_PATH`) the current package manager will still insist on installing another Cxx in `Pkg.dir()` as soon as a package requires Cxx.

This is why it's like to keep Cxx out of REQUIRE, if at all possible. Or is there another solution, until Pkg3? I know it's kind of a hack not to REQUIRE Cxx, I'd love to have a cleaner solution, but I can't really wait for Pkg3.

tkelman:
If you want to put packages outside of Pkg.dir and on LOAD_PATH, do it for ROOT.jl too.

oschulz:
> If you want to put packages outside of Pkg.dir and on LOAD_PATH, do it for ROOT.jl too.

I could, but then what about the packages we use/develop that depend on ROOT.jl? People work on those, so we can't install them system-wide too, resp. bake them into a read-only container?

oschulz:
Any idea when Pkg3 will land? It should solve these kind of problems, right?

tkelman:
You can add multiple locations to your LOAD_PATH, some system-wide and some user-writable

tkelman:
If you don't list dependencies, then `Pkg.add("ROOT"); using ROOT` will fail.

oschulz:
> You can add multiple locations to your LOAD_PATH, some system-wide and some user-writable

I know, but people have to manually `git clone` the packages into those locations then. An things like `Pkg.test` don't work with packages on LOAD_PATH either.

tkelman:
Or have your users make symlinks from the system-wide copies into their Pkg.dir - as long as they don't try to run an operation that needs to modify anything I think that could work

oschulz:
> If you don't list dependencies, then Pkg.add("ROOT"); using ROOT will fail.

Oh, darn - I was not aware. I use `Pkg.clone` for ROOT.jl at the moment, obviously. It's different with `Pkg.add`?

oschulz:
> Or have your users make symlinks from the system-wide copies into their Pkg.dir

Yes, that's a workaround I considered and tried - but things go very bad at the first `Pkg.update()` the user does ...

tkelman:
> It's different with `Pkg.add`?

Not in a way that matters on this point. What's different is you happen to have Cxx available somewhere, either Pkg.dir or LOAD_PATH. That's not guaranteed to be the case if you don't mark the dependency.

tkelman:
If you remove the `.git` or artificially make the system-wide copy dirty in the git sense, then I think Pkg should skip trying to modify it.

oschulz:
Oh, right, of course - the user would have to install Cxx.jl themselves, obviously.

Like I said, I hoped we could make an exception with Cxx, but I get that you have to protect the integrity of the Julia ecosystem as a whole, Tony, and that compromises of this kind kinda ... well, compromise it a bit. :-)

oschulz:
> If you remove the .git or artificially make the system-wide copy dirty in the git sense, then I think Pkg should skip trying to modify it.

Thanks for the tip, I'll try it! Kinda dirty, but at least the dirt will be local, on our side. ;-)

Just for the future - Pkg3 will be able to handle things like this (first-class system-wide packages), right?

tkelman:
Yes, this is certainly in the design scope of what Pkg3 is trying to address. No idea on an ETA. There's a JuliaCon talk on the schedule about it, but the code is still in a private repo that only Stefan has seen AIUI.

oschulz:
Yes, I saw that talk on the schedule, I'm eager to learn more about it. Until then, I'll just have to trust Stefan to solve this, eventually, and find a workaround on our end. :-)

KristofferC:
cc @StefanKarpinski

oschulz:
@tkelman, I'll add Cxx to the requirements as requested and find some way to deal with it on our side. However, ROOT currently need the Cxx master branch, as it needs the new RTTI feature. So I guess I should shelve this release of the renewed ROOT.jl until the next Cxx.jl release.

@Keno: Could you do a new (minor version?) release of Cxx.jl (so that I can add that version to the requirements of ROOT.jl and release it)?

Keno:
Yes, I'll do that tomorrow

oschulz:
> Yes, I'll do that tomorrow

Thanks a bunch!

oschulz:
@tkelman; I modified this PR to just change the URL of ROOT.jl, and will make a separate ROOT.jl release PR after Keno's new Cxx.jl release tomorrow, Ok?

tkelman:
sounds like a plan

oschulz:
Thanks for your explanations and your patience, Tony!

PR #9642, opened 2017-06-02
Tag DSGE v0.2.1 [https://github.com/FRBNY-DSGE/DSGE.jl]

emoszkowski:
Diff vs v0.2.0: https://github.com/FRBNY-DSGE/DSGE.jl/compare/e3347771c5bf9bc52ffee9ae886aff1edcdaa40d...f42a9bc2035878a3d4a0f7e33a77df118020b6b0


DSGE/versions/0.2.1/requires
```diff
@@ -0,0 +1,12 @@
+julia 0.5.0+
```
tkelman: the trailing + probably doesn't mean what you intend - it means the package is installable on 0.5.1 but not 0.5.0

tkelman:
If this is the first julia-0.5-only release, it should use a new package minor version, so 0.3.0 of the package

DSGE/versions/0.2.1/requires
emoszkowski: I just haven't tried it - certainly could/should do! I'll let you know once we've checked it out.

tkelman: should be pretty easy on travis, can add another entry to the matrix with the full version number 0.5.0

PR #9648, opened 2017-06-03
Deprecate `DynamicalBilliardsPlotting`

Datseris:
(plotting moved back to core package under on-demand availability)


DynamicalBilliardsPlotting/versions/0.1.0/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.5
+julia 0.0- 0.6
```
tkelman: unless the package now works on julia 0.4 or earlier, leave the lower bound

Datseris: Yes my mistake, clearly.

PR #9658, opened 2017-06-04
Tag Hinton v0.1.2 [https://github.com/ninjin/Hinton.jl]

ghost:


Hinton/versions/0.1.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Colors
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

tkelman: you can probably actually drop the Compat dependency, doesn't look like you're relying on anything that isn't in 0.5

when dropping support for old julia versions, should use a new package minor version, so at least 0.2.0

aviks:
Hi @ninjin The METADATA tests failed with `ERROR: LoadError: New tag 0.1.2 of package Hinton requires julia 0.5.0, but version 0.1.0 of Hinton requires julia 0.3.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.2.0`. Could you update this please?

andreasnoack:

PR #9687, opened 2017-06-06
Tag TikzGraphs v0.4.1 [https://github.com/sisl/TikzGraphs.jl]

tawheeler:
Diff vs v0.4.0: https://github.com/sisl/TikzGraphs.jl/compare/b0e79a7be974f15cd9e5c8e8bce2d7bea0b3cb9a...4ddc1aba773432dcefbad54f1d7f79c1bf2d2d7e

Some changes to support 0.6

### DIFF
--- a/TikzGraphs/versions/0.4.1/requires
+++ b/TikzGraphs/versions/0.4.1/requires
@@ -1,0 +1,4 @@
+julia 0.4
+TikzPictures 0.3.5
+LightGraphs 0.6.0
+Compat 0.8.0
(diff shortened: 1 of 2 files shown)


TikzGraphs/versions/0.4.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.4
+TikzPictures 0.3.5
+LightGraphs 0.6.0
+Compat 0.8.0
```
tkelman: needs 0.17 for abstract type, and I don't think `@compat struct` actually parses correctly pre 0.6

tawheeler: Thanks for mentioning that! Travis was marking the tests as passed but apparently there is some issue running Julia. (https://github.com/sisl/TikzGraphs.jl/issues/18)

PR #9707, opened 2017-06-07
Tag ProximalBase.jl v0.0.3

attobot:
Repository: [mlakolar/ProximalBase.jl](https://github.com/mlakolar/ProximalBase.jl)
Release: [v0.0.3](https://github.com/mlakolar/ProximalBase.jl/releases/tag/v0.0.3)
Travis: [Travis Build Status](https://travis-ci.org/mlakolar/ProximalBase.jl/builds/240554949)
Diff: [vs v0.0.2](https://github.com/mlakolar/ProximalBase.jl/compare/2d73fae36e53ebb5b6e02ef0c1dc2ad4af482c26...3c2488fa5ba14b16a167ee1d00c57860da8c6102)
`requires` vs v0.0.2: no changes
cc: @mlakolar

Please make sure that:


tkelman:
Your CoordinateDescent package was using the `AProxL1` type, may want to deprecate it if possible?

mlakolar:
Is there a convenient way to simultaneously update both packages?
I have changed the master branch of the CoordinateDescent package after the change here, so that package should work with the master branch here.

tkelman:
You can retroactively add upper bounds to the existing tags of CoordinateDescent here in metadata, only changing the requires files of the metadata copies (leave the sha content alone). That would say the existing versions of CoordinateDescent cannot be simultaneously installed with this new breaking version of ProximalBase, but once a new version of CoordinateDescent is released without an upper bound, they can both be upgraded if both installed. Though in this case since both packages are new, it may not be worth the effort.

mlakolar:
Thank you for the instructions.
I've put an upper bound in CoordinateDescent retrospectively in #9725

PR #9725, opened 2017-06-08
Upper bound on ProximalBase

mlakolar:
In #9707 with the version of ProximalBase 0.0.3 there are breaking changes.

### DIFF
--- a/CoordinateDescent/versions/0.0.1/requires
+++ b/CoordinateDescent/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.6.0-pre
-ProximalBase
+ProximalBase 0.0.2-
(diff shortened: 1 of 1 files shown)


CoordinateDescent/versions/0.0.1/requires
```diff
@@ -1,3 +1,3 @@
 julia 0.6.0-pre
-ProximalBase
+ProximalBase 0.0.2-
```
tkelman: if you only list one number then it's a lower bound (inclusive), a second number is an upper bound (exclusive) so I think you want `ProximalBase 0.0.0 0.0.3` to say any version strictly less than 0.0.3 is okay

PR #9733, opened 2017-06-09
Register new package StatFiles.jl v0.0.1

attobot:
Repository: [davidanthoff/StatFiles.jl](https://github.com/davidanthoff/StatFiles.jl)
Release: [v0.0.1](https://github.com/davidanthoff/StatFiles.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/StatFiles.jl/builds/241065338)
cc: @davidanthoff

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

@davidanthoff This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.

PR #9734, opened 2017-06-09
Register new package CSVFiles.jl v0.0.1

attobot:
Repository: [davidanthoff/CSVFiles.jl](https://github.com/davidanthoff/CSVFiles.jl)
Release: [v0.0.1](https://github.com/davidanthoff/CSVFiles.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/CSVFiles.jl/builds/241065280)
cc: @davidanthoff

Please make sure that:


davidanthoff:

CSVFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6-
```
tkelman: same comment as https://github.com/JuliaLang/METADATA.jl/pull/9733/files#r121231480

davidanthoff:

davidanthoff:
Could this please be merged? It was open more than 24 hours and I addressed all feedback.

ViralBShah:
Are we waiting on JuliaCIBot?

ViralBShah:
Merging for now.

CSVFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6-pre
+TextParse 0.1.6
```
tkelman: why does this need the absolute latest TextParse? lower bounds should be the earliest version that would work

ViralBShah: TextParse is fast moving. Best to not use older versions.

tkelman: setting bounds too tight causes pkg resolver errors and inability to install versions that would work together

davidanthoff: I don't understand how package authors are realistically supposed to figure out a lower bound. TextParse doesn't have a NEWS, so all I can look at are the releases. v0.1.6 says "use _pointer in place of pointer", which suggests to me that previous versions had an error. But I have no clue, and I certainly don't have the time to test my package with every old version of every package I depend on to figure out what the lowest possible bound is. If there was some tool that did this automatically, great, but asking package devs to do this manually seems like an enormous misallocations of their time, quite frankly, and I want to use the time I have for julia to improve my packages.

ViralBShah: I know, but all these are new packages and with quick development. Best to let things move fast and in a month or two, the bounds will get older and things will stabilize. Technically you are right, but it seems that in early stages, rules could perhaps be a bit more lax.

tkelman: a few minutes looking at blame and /compare can usually get much closer to accurate than forcing latest

PR #9735, opened 2017-06-09
Register new package FeatherFiles.jl v0.0.1

attobot:
Repository: [davidanthoff/FeatherFiles.jl](https://github.com/davidanthoff/FeatherFiles.jl)
Release: [v0.0.1](https://github.com/davidanthoff/FeatherFiles.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/FeatherFiles.jl/builds/241065324)
cc: @davidanthoff

Please make sure that:


davidanthoff:

FeatherFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6-
+Feather 0.2.5
```
tkelman: same comment as https://github.com/JuliaLang/METADATA.jl/pull/9733/files#r121231480 on the julia requirement

why does this need the absolute latest Feather.jl version, seems like you're not using any methods that haven't been around for longer

davidanthoff:

FeatherFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6-pre
+Feather 0.2.5
```
tkelman: why does this need the absolute latest Feather.jl version? seems like you're not using any methods that haven't been around for longer

davidanthoff: Actually, we should wait with merging this PR here. Feather is broken on julia 0.6, so that should probably be fixed first and then the minimum on Feather here should be bumped to the first version of Feather that actually works on 0.6.

ViralBShah:
I for one am looking forward to Anthoffverse!

davidanthoff:
This can be merged now! Sooner rather than later would be great, I'm trying to get everything shipshape before my juliacon talk.

FeatherFiles/versions/0.0.1/requires
tkelman: there's no difference between the last several feather versions that your package is relying on - lower bounds are not foravoiding depwarns, they're for features you're using

tkelman:
Feather lower bound is still too restrictive

davidanthoff:
> Feather lower bound is still too restrictive

Done.

davidanthoff:
Thanks!

PR #9736, opened 2017-06-09
Register new package ExcelFiles.jl v0.0.1

attobot:
Repository: [davidanthoff/ExcelFiles.jl](https://github.com/davidanthoff/ExcelFiles.jl)
Release: [v0.0.1](https://github.com/davidanthoff/ExcelFiles.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/davidanthoff/ExcelFiles.jl/builds/241065309)
cc: @davidanthoff

Please make sure that:


davidanthoff:

ExcelFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6-
```
tkelman: same comment as https://github.com/JuliaLang/METADATA.jl/pull/9733/files#r121231480

davidanthoff:

ExcelFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6-pre
+ExcelReaders 0.8.0
```
tkelman: doesn't look like there are any new features in 0.8.0 that weren't in 0.7.0, why are you depending on that? lower bounds should reflect the lowest version that could possibly work, otherwise you're restricting Pkg from installing version combinations that would work together fine

davidanthoff: Done.

davidanthoff:
And the same here as well, could this please be merged? It was open more than 24 hours and I addressed all feedback.

ExcelFiles/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6-pre
+ExcelReaders 0.7.0
```
tkelman: 0.7 also doesn't look like it introduced any features that you're relying on

ViralBShah: I think package authors should be free to use a minimum version of a package they are comfortable with. I personally would prefer to use the latest versions as I start off something new, especially where the dependent packages do not necessarily have a large audience.

davidanthoff: It was the first release to support julia 0.6.

tkelman: forcing latest versions of dependencies is unnecessarily restrictive if there aren't actual dependencies on features that were first added in that version

PR #9738, opened 2017-06-09
Update QuantumOptics.jl to version v0.4.0 and change github url

bastikr:


tkelman:
You need to have `Compat 0.17` in your require file to use `@compat abstract type`

bastikr:
Thank you! I changed the pull request accordingly.

tkelman:
be sure to also make the corresponding change in https://github.com/qojulia/QuantumOptics.jl/blob/master/REQUIRE

bastikr:
I made a pull request, thanks again!

bastikr:
Since there was this copy bug you reported, it's probably not a good idea to use this release. Is it okay to move the 0.4.0 tag to a commit that fixes this problem or is it cleaner to just release 0.4.1?

tkelman:
you can do either, whichever you prefer

bastikr:
I moved the v0.4.0 tag to the latest commit and changed the sha1 key in METADATA accordingly.

PR #9742, opened 2017-06-09
Tag Plots.jl v0.12.0

attobot:
Repository: [JuliaPlots/Plots.jl](https://github.com/JuliaPlots/Plots.jl)
Release: [v0.12.0](https://github.com/JuliaPlots/Plots.jl/releases/tag/v0.12.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaPlots/Plots.jl/builds/241203759)
Diff: [vs v0.11.3](https://github.com/JuliaPlots/Plots.jl/compare/df729cd0f6dd115e1e509372ae691efbedb0784a...3e720f8baec9563385272f6dfc4f5383e251f4c0)
`requires` vs v0.11.3: 
```diff
--- 0.11.3/requires
+++ 0.12.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6-pre
 
 RecipesBase
 PlotUtils 0.4.1
```
cc: @mkborregaard

Please make sure that:


mkborregaard:
Hold this for a few days - if we end up merging Chris' PR on RecipesBase (https://github.com/JuliaPlots/RecipesBase.jl/pull/16) I'll incorporate the required changes immediately here for the first 0.6 release - to simplify versioning.

Plots/versions/0.12.0/requires
```diff
@@ -0,0 +1,13 @@
+julia 0.6-pre
+
+RecipesBase
```
tkelman: this needs a minimum version for the new functions, right?

mkborregaard: No, we agreed on instead of using the julia-version number to align. That's why we are releasing our first 0.6-only releases at the same time across the board.

mkborregaard:
I've merged now, so as far as I can see this should be good to go

Plots/versions/0.12.0/requires
tkelman: That isn't correct - the fact that there aren't Julia upper bounds on the earlier versions means you can pin to them, or Pkg can downgrade to them if upper bounds tell it to, on Julia 0.6.

tkelman: If you choose to run `Pkg.pin("RecipesBase", v"0.1.0")`, that would break this version of Plots. If the dependency were indicated correctly with lower bounds on Plots 0.12.0's dependency on functions that were first introduced in RecipesBase 0.2.0, then running `Pkg.pin("RecipesBase", v"0.1.0")` would result in Pkg downgrading Plots to the most recent version that is compatible with RecipesBase 0.1.x. While this wouldn't necessarily work perfectly on 0.6, it's a lot better than being unable to import Plots at all.

mkborregaard: Ah - that's unfortunate. This was based on the advice here: https://github.com/JuliaPlots/RecipesBase.jl/pull/16#issuecomment-299846416 . That in fact was the rationale behind this whole exercise.
What do you suggest then?

tkelman: That advice still stands, but has more to do with when you choose to introduce or remove deprecations that will be loudly visible to all downstream users. I suggest indicating the fact that Plots 0.12.0 relies on importing functions that were first introduced in RecipesBase 0.2.0 (if I'm correct about this part) by indicating an 0.2.0 lower bound on Plots 0.12.0's RecipesBase dependency.

mkborregaard: The only thing is that not having a need for explicit version bounds adds to the no-conflict-potential that is so important for RecipesBase. We believe strongly that having a no-conflict base package that can be required by deep library code without problems is the best way to ensure basic plotting functionality to the julia ecosystem.
Would it help to revert the RecipesBase release, make a new release from a branch prior to Chris' commit with upper version bounds on Julia, then redo this release? And would it be sufficient to do it for RecipesBase?

tkelman: Upper bounds would have to be added retroactively on all past versions, and I'm not sure they're usually a good idea unless the package fails to load completely.

Adding a lower bound on Plots' dependency on RecipesBase that reflects the functions it's relying on doesn't make RecipesBase itself any more likely to cause conflicts.

mkborregaard: Fair enough. I've added the version bound as requested.

mkborregaard:
and this

PR #9754, opened 2017-06-10
Tag ROOT v0.1.0 [https://github.com/JuliaHEP/ROOT.jl]

oschulz:
New release, Cxx.jl-based, ROOT-6 and Julia v0.5 and v0.6 compatible.


ROOT/versions/0.1.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Cxx
```
tkelman: does this need 0.1.1 to work correctly?

oschulz: Yes, of course. Sloppy me - sorry, and thanks for catching this, @tkelman!

PR #9768, opened 2017-06-11
Rename Dendriform.jl from GroveAlg.jl

chakravala:
Package is being renamed to Dendriform from GroveAlg. This name is more appropriate for the future generalizations I am planning, as the program currently does dendriform dialgebra and I intend to extend this to dendriform tiralgebra as well. Therefore, Dendriform is a more appropriate name for this package.

…v0.0.1, v0.0.2, v0.0.3

The new version v0.0.3 has these changes [diff](https://github.com/chakravala/Dendriform.jl/commit/85cd17d312f6d5159554a10609332387cfc91e27), mainly renaming things and 4 extra function definitions to help with multiple dispatch.

Wanted to rename it now, since the next version is going to have dendriform trialgebra features as well

The original GroveAlg.jl package can be completely removed from the Julia package manager, as far as I know there shouldn't be any user base for it at this time. What is the deprecation procedure?


tkelman:
So since the first 2 tags didn't have a `src/Dendriform.jl` file, they won't work. This new-name package should be registered with just the one latest tag that will work under the new package name. Should also either do a repository rename, or make this a new separate repo, since the url should match the package name. Once the new name is all registered, the deprecation procedure is to add a julia upper bound to the metadata requires files for all tags of the old name.

chakravala:
Alright, deleted the first 2 tags from this commit and fixed the URL. There is already an upper bound of 0.5 for the old name on the requires files. Sounds like we are set then?

tkelman:
0.0.1 and 0.0.2 only have a julia lower bound, an upper bound is a second number if present

chakravala:
Alright, I added `0.5-` to the requires

tkelman:
That's an empty set. The upper bound is exclusive. Probably best to list `0.6` as the upper bound, that way it won't be installable once julia 0.6 final comes out.

chakravala:
changed to 0.6

PR #9780, opened 2017-06-12
Tag Phylo.jl v0.1.1

attobot:
Repository: [richardreeve/Phylo.jl](https://github.com/richardreeve/Phylo.jl)
Release: [v0.1.1](https://github.com/richardreeve/Phylo.jl/releases/tag/v0.1.1)
Travis: [Travis Build Status](https://travis-ci.org/richardreeve/Phylo.jl/builds/242178753)
Diff: [vs v0.1.0](https://github.com/richardreeve/Phylo.jl/compare/ccd6d208ddb09c58cbc23164e0a9f81d0518d85d...ca9e3aed3aad4c1c3e8b2db4efa12742c41637d5)
`requires` vs v0.1.0: 
```diff
--- 0.1.0/requires
+++ 0.1.1/requires
@@ -1,3 +1,4 @@
 julia 0.5
 Compat 0.18.0
 Distributions 0.12.2
+DataStructures 0.5.3
```
cc: @richardreeve

Please make sure that:


Phylo/versions/0.1.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+Compat 0.18.0
+Distributions 0.12.2
+DataStructures 0.5.3
```
tkelman: do you really require the absolute latest version? looks like what you're using should work okay with earlier versions,so this too tight a lower bound

richardreeve: @tkelman this was to remove the 0.6 deprecation warnings that were only fixed in 0.5.3

tkelman: Right but that's not required on Julia 0.5 which this package still supports, and lower bounds aren't for warnings anyway. Pkg will upgrade by default so most users will be on the latest version of things, but for those that aren't, you shouldn't prevent them from installing this version of Phylo if they happen to be using an older version of DataStructures.

richardreeve: Okay. I'll switch to 0.4.3 which git blame suggests covers most of the code for OrderedDict. Once I start ignoring commits it becomes impossible to use git blame to determine the correct release to use though afaik.

tkelman: close is good enough

richardreeve:
I just updated the release with some minor tidying to trigger JuliaCIBot to actually run (since it seemed to be permanently pending).

PR #9788, opened 2017-06-13
Pull request/56f4b00f

pazzo83:
Tagged version v0.0.1 for NOAAData

### DIFF
--- a/NOAAData/versions/0.0.1/requires
+++ b/NOAAData/versions/0.0.1/requires
@@ -1,0 +1,4 @@
+julia 0.6-
+Requests
+DataTables
+IndexedTables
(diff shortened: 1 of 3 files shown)


NOAAData/versions/0.0.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.6-
```
tkelman: Since you're using new syntax that wouldn't work on early 0.6-dev versions, it would be best not to claim to support them. A lower bound like 0.6.0-pre or one of the rc's would be closer.

tkelman:
closing in favor of https://github.com/JuliaLang/METADATA.jl/pull/9803

PR #9811, opened 2017-06-15
Tag Atom.jl v0.6.0

attobot:
Repository: [JunoLab/Atom.jl](https://github.com/JunoLab/Atom.jl)
Release: [v0.6.0](https://github.com/JunoLab/Atom.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JunoLab/Atom.jl/builds/243263071)
Diff: [vs v0.5.10](https://github.com/JunoLab/Atom.jl/compare/f13924faddfff4efd25fe60b1188e2c45ae6e76a...7a55deda388658fdafa875746cce3dbe71089054)
`requires` vs v0.5.10: 
```diff
--- 0.5.10/requires
+++ 0.6.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6-
 Juno 0.2.6
 Lazy 0.11.3
 LNR
```
cc: @MikeInnes

Please make sure that:


Atom/versions/0.6.0/requires
```diff
@@ -0,0 +1,13 @@
+julia 0.6-
```
tkelman: should use 0.6-pre instead of claiming to support early dev versions that where syntax wouldn't work on

ararslan:
Hey @MikeInnes, looks like @tkelman's comment wasn't addressed. Versioning can be annoying and it may seem like a minor detail, but getting the bound right can be pretty important to ensure people on certain versions of Julia don't end up with broken packages. I'm not really sure what the protocol is for adjusting this after the tag has been registered, but if you could make the adjustment on the repo so that it's all good for the next tag, that would be 💯.

MikeInnes:
As a package maintainer I have to make decisions about what users and use cases to support. People using source builds of Julia from months ago are very low on that list; as far as I'm concerned, they should not expect packages to keep things correct for them and can do a `git pull`.

Of course, as is typical of open source, if someone else wants to put in that effort and make the changes themselves I'm all for it.

ararslan:
> As a package maintainer I have to make decisions about what users and use cases to support.

Totally. I don't mean to suggest you have to care about people using old, non-release versions, just that it's helpful to use more specific cutoffs as Tony suggested.

tkelman:
It's a handful of characters more to be much more accurate, and the fact that there are previous versions of the package that worked on older Julia versions means it would be much better to have Pkg keep using those older package versions on that range of Julia versions where they're known not to work. And it's blatantly disrespectful to not even respond to valid review and merge anyway. You have made a habit of that for a very long time and I'd really appreciate if it could stop.

https://github.com/JuliaLang/METADATA.jl/pull/9824
https://github.com/JunoLab/Media.jl/pull/14
https://github.com/JunoLab/Atom.jl/pull/88
https://github.com/MikeInnes/Lazy.jl/pull/63

tkelman:
You should also really pay attention to your package's CI. Many of these have pitifully few tests, but even a basic verification that the package loads is better than nothing. JuliaCIBot will test your releases but not your master commits or PR's.

PR #9820, opened 2017-06-15
Fix Julia limits on older versions of QuartzImageIO

rsrock:


tkelman:
Upper bounds are exclusive, I wouldn't have the trailing dash there since that excludes the very first commit after branching. Are you sure these don't load at all at the upper bound?

MikeInnes:

PR #9825, opened 2017-06-15
Register new package IterTools.jl v0.1.0

attobot:
Repository: [JuliaCollections/IterTools.jl](https://github.com/JuliaCollections/IterTools.jl)
Release: [v0.1.0](https://github.com/JuliaCollections/IterTools.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaCollections/IterTools.jl/builds/243397370)
cc: @iamed2

Please make sure that:

@iamed2 This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


iamed2:
This is the new home for Iterators.jl. Users will be encouraged to move their dependency on Iterators.jl to this package.

ararslan:
Perhaps we should make some PRs to dependent packages to ensure everyone gets properly updated. We could even add
```julia
function __init__()
    warn("The Iterators.jl package is deprecated in favor of IterTools.jl and Base.Iterators")
end
```
to Iterators, though I don't know how legit it is to do something like that.

iamed2:
@ararslan I was thinking that would be a good idea. Is there a particular reason why that might not be legit?

ararslan:
¯\\\_(ツ)\_/¯ I just don't know

iamed2:
Well, assuming this is accepted, I will do that and start PRs (we'll have to work our way down the dependency graph)

ararslan:
Okay, I'd be happy to help divide and conquer.

tkelman:
You decided against messing with the history to make renamed versions of all the old tags then?

ararslan:
Random idea: Could we upper bound Iterators to 0.6.0, effectively deprecating it in METADATA since it won't be installable on 0.6.0, then people will have to use IterTools instead?

tkelman:
That would make everything that's being slow about switching to it also not installable on 0.6, so it's pretty drastic and would be better done after everyone has had a chance to switch.

ararslan:
Ah yeah, good call.

tkelman:
I'd also avoid making a new 0.4-compatible tag with an init-time warning, since almost all packages have stopped making new versions supporting that. Maybe even make the init-time warning an 0.6-only thing, if some of the packages that depend on Iterators have gone 0.6-only already?

iamed2:
> You decided against messing with the history to make renamed versions of all the old tags then?

Yep, since you pointed out that only one package upper-bounds Iterators.

PR #9831, opened 2017-06-15
Register new package Dolang.jl v0.1.0

attobot:
Repository: [EconForge/Dolang.jl](https://github.com/EconForge/Dolang.jl)
Release: [v0.1.0](https://github.com/EconForge/Dolang.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/EconForge/Dolang.jl/builds/243459624)
cc: @sglyon

Please make sure that:

@sglyon This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


tkelman:
Importing an optional dependency in a try-catch is not compatible with precompilation, since the state of whether the dependency is available won't be tracked correctly.

sglyon:
Is the only issue that users who install the optional dependency after this package has been precompiled won't see the optional dependency until after another precompile happens?

If so, I know about that issue and am totally happy to live with it.

tkelman:
and the reverse, if someone removes the dependency after your package has been precompiled. Having to manually manage the cache defeats the purpose of automatic precompilation.

sglyon:
Last I checked the optional dependency (SymEngine) is difficult to install behind a corporate firewall. That's the only reason we have it as conditional. 

Is there any other way I can do this that would work with Precompilation

If not, unfortunately, I don't see a way around what I'm doing.

tkelman:
Disable precompilation if you do what you're doing.

sglyon:
OK, I could do that as a last resort.

Do you know of any other alternatives. I'd love to keep pre-compilation _and_ the optional dependency

tkelman:
Without https://github.com/JuliaLang/julia/pull/21743, nothing nice and fully precompilation-compatible. Could make a separate package that depends on both. Or maybe refactor so all references to the optional package happen only at runtime, assuming it's been loaded into Main.

sglyon:
That's helpful, thanks.

I think for now I'll just do `__precompile__(false)`

PR #9840, opened 2017-06-16
Register new package Microeconometrics.jl v0.0.1

lbittarello:
Repository: https://github.com/lbittarello/Microeconometrics.jl


lbittarello:
I'd forgot to end the name of the repository with ".jl"... Fixed!

Microeconometrics/versions/0.0.1/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6-
```
tkelman: this includes the very first dev version after 0.5 branched, so `mutable struct` syntax wouldn't work there - better to use `0.6.0-pre` as the minimum

lbittarello: Fixed!

tkelman: you'll need to update this pr. consider using https://github.com/attobot/attobot

PR #9861, opened 2017-06-18
Switched to new GR run-time v0.25.0

jheinen:


jheinen:
May be, in the future, the GR.jh wrapper (0.22.0l) should have the same version number as the gr repo (0.25.0).

PR #9884, opened 2017-06-20
Register UniqueVectors, which was formerly known as IndexedArrays

garrison:
This pull request also makes a final release of IndexedArrays to deprecate its usage.

Ref #2236.


IndexedArrays/versions/0.3.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5
+Compat 0.24.0
```
tkelman: only need 0.17 for abstract type

garrison:
OK, both `requires` files have been updated now.

PR #9929, opened 2017-06-22
Tag EasyPkg v0.3.0

oschulz:
Diff vs v0.2.0: https://github.com/oschulz/EasyPkg.jl/compare/2c03a8fcaf859d956d25139962debcbf117d7267...f5362cadd3362cbcb6e0e592feb5bd00729ccd8c


EasyPkg/versions/0.3.0/requires
```diff
@@ -0,0 +1 @@
+julia 0.5- 0.6-
```
tkelman: adding an upper bound actually doesn't help directly unless you also add it to all past versions - just the metadata copies of the require files

oschulz: Thanks, @tkelman! Better now?

tkelman: I'd avoid using the trailing dash on the upper bound, since upper bounds are exclusive and the trailing dash means the first commit after the previous release branched

oschulz: Thanks, Tony - it's my first upper bound so far. :-)

Ok now?

tkelman:
Missing the git tag for 0.3.0

oschulz:
> Missing the git tag for 0.3.0

Sorry - don't know how that happened. Fixed now.

EasyPkg/versions/0.0.1/requires
```diff
@@ -1 +1 @@
-julia 0.4
+julia 0.4 0.5
```
tkelman: pinning to this version, seems to pass tests fine on 0.5?

oschulz: Hm, strange - v0.2.0 should be the first that's 0.5-compatible.

tkelman: 0.0.1 looks simple / small enough that there doesn't seem to be anything there that wouldn't work on 0.5

oschulz: So be it. :-)

oschulz:
Thanks for the help, @tkelman!

PR #9950, opened 2017-06-24
StatprofilerHTML: bump julia version

tkluck:
Tested to work with julia 0.6-rc2; no reason to suspect it doesn't work in every current 0.5+.

### DIFF
--- a/StatProfilerHTML/versions/0.1.1/requires
+++ b/StatProfilerHTML/versions/0.1.1/requires
@@ -1,1 +1,1 @@
-julia 0.5
+julia 0.5+


tkluck:
FWIW, let me know if this change should be accompanied by a version bump of the package. I tested that the code works without modification, though.

tkelman:
That's not what + means here. See https://docs.julialang.org/en/stable/manual/strings/#man-version-number-literals-1. When one version number is listed in REQUIRE, it's an inclusive lower bound https://docs.julialang.org/en/stable/manual/packages/#Requirements-Specification-1.

tkluck:
@tkelman nice, thanks for checking! I'll close this pull-request then, to decrease clutter for you all.

Do I then understand correctly it *should* install correctly as-is? If so, I'll have to look into my config and see why, exactly, that failed earlier for me.

tkelman:
Looks like it's working fine? https://pkg.julialang.org/detail/StatProfilerHTML

Keep in mind there are separate package directories for different Julia minor versions, `~/.julia/v0.5` vs `~/.julia/v0.6`

tkluck:
It must have been something related to that. Thanks for showing me all those enlightening links and sorry for the noise!

PR #10031, opened 2017-06-29
Update URL after move GenericSVD

simonbyrne:
The repository was moved to JuliaMath. This needs to be updated to trigger AttoBot.


simonbyrne:
cc @andreasnoack

tkelman:
double-check that various CI services are re-enabled

PR #10051, opened 2017-06-30
Release Xtensor.jl 0.0.2

SylvainCorlay:


Xtensor/versions/0.0.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps 0.4.2
+Compat 0.21
```
tkelman: you're still not using Compat anywhere

SylvainCorlay: Fixed. Thanks for the review.

tkelman: be sure to make the corresponding change at https://github.com/QuantStack/Xtensor.jl/blob/master/REQUIRE before the next tag

SylvainCorlay: Done

PR #10053, opened 2017-07-01
Tag JWAS v0.1.2 [https://github.com/reworkhow/JWAS.jl]

reworkhow:
last version for Julia0.5


tkelman:
Looks like this causes a `UndefVarError: ST not defined` in XSim. If a new version of XSim isn't ready yet, the problem can be avoided by adding an upper bound on the dependencies of the past XSim versions on JWAS. That can be done in-place on the METADATA requires files, without needing to modify the package repo, tags or the sha files.

tkelman:
Since the first two versions of XSim didn't depend on JWAS, the bound can do something counterintuitive - Pkg could be allowed to downgrade XSim in order to upgrade JWAS. To prevent that, could add an "artificial" dependency `JWAS 0.0 0.2` to the files `XSim/versions/0.0.2/requires` and `XSim/versions/0.0.3/requires`

XSim/versions/0.0.2/requires
```diff
@@ -1,2 +1,3 @@
-julia 0.3-
+julia 0.3 0.6-
```
tkelman: Upper bounds are exclusive, and `0.6-` means the very first commit after 0.5 branched. If you want to make the package not even installable on 0.6 release, better to use `0.6` with no trailing `-` as the upper bound

XSim/versions/0.1.0/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.6
 DataFrames
 Distributions
-JWAS 0.1-
+JWAS 0.1.1 0.2-
```
tkelman: the lower bound part of this seems to have broken things on pkgeval https://github.com/JuliaCI/pkg.julialang.org/blob/345bf6547d534e4238f05e3aeacda45bb5db5379/logs/XSim_0.4.log

PR #10057, opened 2017-07-01
Modof

aritrasep:
Multiobjective Discrete Optimization Framework in Julia


Modof/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.5.2
```
tkelman: what is required that was in 0.5.2 but not 0.5.0?

Modof/versions/0.0.1/requires
```diff
@@ -0,0 +1,7 @@
...
+Match 0.3.0
+DataFrames 0.9.0 0.9.1 0.10.0
+DataFramesMeta
+PyCall 1.13.0
```
tkelman: this is the absolute latest version, are you sure you require a feature that didn't exist in any earlier versions?

tkelman:
Registered packages must use an OSI approved license: https://opensource.org/licenses
Academic-use-only licenses are too restrictive for registration on METADATA.

PR #10059, opened 2017-07-02
Tag GtkBuilderAid v0.1.0

Matt5sean3:
Repository: [Matt5sean3/GtkBuilderAid.jl](https://github.com/Matt5sean3/GtkBuilderAid.jl)
Release: v0.1.0
Travis: [Build Status](https://travis-ci.org/Matt5sean3/GtkBuilderAid.jl)
Diff vs v0.0.7: https://github.com/Matt5sean3/GtkBuilderAid.jl/compare/5be3b115dce821e2f5c4ccdb8b146631ef6a42d4...2e05d9d3ac3643eed870b16f3ed59e88bb20f946

This version updates the package for julia 0.6, removes usage of deprecated pieces of Gtk.jl, and makes some major internal simplifications. Mostly, this just makes sure this package isn't completely dead and unusable. I removed some minor functionality that was much more trouble than it was worth leaving things closer to how they behave using Gtk.jl's `signal_connect` function directly.


GtkBuilderAid/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.5
+Cairo 0.2.24
+Gtk 0.8.4
```
tkelman: were all the pieces of Gtk you're using available in this version?

Matt5sean3: I haven't changed the functionalities used from Gtk since I last checked the necessary versions, but Gtk 0.8.4 won't load in julia 0.5 or julia 0.6 due to outmoded syntax and the switch from creating widgets in Gtk.jl using macros to using ordinary functions has broken compatibility with earlier versions.

The earliest version that will load in julia 0.5 successfully is Gtk 0.10.2. The earliest version that will load in julia 0.6 successfully is Gtk 0.12.0. Tests are successful in julia 0.6 with Gtk 0.12.0, but Gtk 0.10.2 on julia 0.5 has issues due to leaving macro syntax, but tests are successful with Gtk 0.11.0.

Cairo 0.2.24 won't load in julia 0.5 or julia 0.6 as the `Color` package does not have a supported version and type renaming. The earliest version that loads in julia 0.5 is Cairo 0.2.31 and GtkBuilderAid tests with it successfully. The earliest version that loads in julia 0.6 is Cairo 0.2.32 and GtkBuilderAid tests with it successfully.

This seems to put the dependencies as Cairo 0.2.31 and Gtk 0.11.0, so I'll open a new pull request reflecting as such.

Matt5sean3:
Need to try yet again with updates to supported dependency versions *sigh*

PR #10066, opened 2017-07-02
Pull request/2ee95e0d

montyvesselinov:
v0.6 fixes


Mads/versions/0.4.16/requires
```diff
@@ -0,0 +1,34 @@
...
+PyPlot 2.2
+Gadfly 0.6.2
+Compose 0.5.2
+Measures
+Images
+YAML
+JSON
+JLD
+ProgressMeter
+Distributions 0.10.0 0.14.0
```
tkelman: does Distributions 0.14 cause problems with all past Mads versions? If so, applying the bound only on this one version of Mads won't accomplish much, Pkg is allowed to downgrade Mads in order to upgrade Distributions.

montyvesselinov:
You are correct; Distributions constraint is not needed; there are
still Mads issues with v0.6; Distributions 0.14 was causing problems
in v0.6 but we are not ready for 0.6 anyway. I will remove it.




-monty

PR #10074, opened 2017-07-03
diffeq upper bounds

ChrisRackauckas:
For incoming DiffEqBase update


StochasticDiffEq/versions/2.7.0/requires
```diff
@@ -1,6 +1,6 @@
 julia 0.6.0
 Parameters 0.5.0
-DiffEqBase 1.13.0
+DiffEqBase 1.13.0 1.15.0
```
tkelman: what about 2.7.1?

ChrisRackauckas: fixed

PR #10083, opened 2017-07-03
Tag Lumberjack v2.1.1 [https://github.com/WestleyArgentum/Lumberjack.jl]

WestleyArgentum:


tkelman:
need to list `Compat 0.17` explicitly in REQUIRE and do `using Compat` to use `@compat abstract type`

WestleyArgentum:
Lol, I was wondering why it worked for me without `using`. Still, good call on the REQUIRE, and I think using is better this way anyway.

PR #10091, opened 2017-07-04
Tag Actors v0.1.0 [https://github.com/daqcore/Actors.jl]

oschulz:


Actors/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5-
+Compat 0.20.0
```
tkelman: abstract type only requires 0.17

Actors/versions/0.0.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5-
```
tkelman: drop the trailing dash if you don't need to support prereleases of 0.5

oschulz:
Thanks for the hints, Tony.

oschulz:
> missing an important line from https://github.com/daqcore/Actors.jl/pull/2/files

Sorry, @tkelman , no idea how that got lost in the merge. I'll fix it.

PR #10102, opened 2017-07-05
Tag Proj4.jl v0.1.0

attobot:
Repository: [JuliaGeo/Proj4.jl](https://github.com/JuliaGeo/Proj4.jl)
Release: [v0.1.0](https://github.com/JuliaGeo/Proj4.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGeo/Proj4.jl/builds/250175887)
Diff: [vs v0.0.1](https://github.com/JuliaGeo/Proj4.jl/compare/c7bac29f3aa79f071417347dc609edb14626d829...cdb45538b0d1244bee643de6a0df5a7219af6957)
`requires` vs v0.0.1: 
```diff
--- 0.0.1/requires
+++ 0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.5
 BinDeps
-Compat 0.8
+Compat 0.26.0
 @osx Homebrew
```
cc: @c42f

Please make sure that:


Proj4/versions/0.1.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps
+Compat 0.26.0
```
tkelman: only need 0.17 for abstract type

c42f: Thanks Tony.

Is it important to fix this?  I'd have thought Compat was the one package that other packages should avoid have an upper bound on, so being a bit too restrictive on the lower bound shouldn't be a big deal?

c42f: I ask this as a general question about Compat - of course it's not a big deal on a case by case basis and we can fix it here - but getting the correct lower bound for Compat seems to be a constant source of errors.

tkelman: Upper bounds from other packages might be rare, but pinning everything for reproducibility is fairly common. Ruling out version combinations that would work fine together would hold your package back unnecessarily.

c42f: Pinning for production deployment is definitely a good thing. We do it ourselves, if we didn't I think we'd go crazy :-)  However, we do this across the board for a given deployed ie, pinning both Proj4 and Compat.

I think a reasonable answer might be that "in principle this shouldn't matter, but fix it anyway because everyone currently invents their own solution for pinning in production"?

tkelman: I've seen it matter. Say you have a set of pinned packages installed then want to add something new. Having too-tight or too-loose dependency bounds on the pinned package versions can cause things to go very wrong in terms of what version of the newly-added package you get (if any).

Really people should be thinking about what features they're using from all of their dependencies, Compat is a case where this is especially well-defined to figure out.

mauro3: Oh, sorry, I didn't see this conversation yet. I just pushed a fix.

mauro3: Concerning finding the right version of Compat.jl: it would be nice if it was a bit easier to find which one is the right one.  Maybe the readme should add which compat tag each entry corresponds to?

tkelman: right now blame on the readme is fairly accurate, but yes going through and annotating each would be good too

c42f: See also

https://github.com/JuliaLang/Compat.jl/pull/331

which unfortunately hasn't been merged due to issues discussed on the PR, but would probably be better than the current situation.

c42f: Thanks @mauro3 for fixing the tag!

c42f: > I've seen it matter

Ok, good to know.

PR #10129, opened 2017-07-06
Tag Mongo v0.3.0 [https://github.com/pzion/Mongo.jl]

pzion:
Diff vs v0.2.3: https://github.com/pzion/Mongo.jl/compare/ea46eab8e3e1c6fcaddc178b96c000e138f57573...e85a9a4c8e78f9ed11740a8abc5decd4d615514b

Note that this drops support for Julia v0.4

### DIFF
--- a/Mongo/versions/0.3.0/requires
+++ b/Mongo/versions/0.3.0/requires
@@ -1,0 +1,5 @@
+julia 0.5
+LibBSON 0.3
+BinDeps 0.4
+Compat 0.9
+@osx Homebrew
--- a/Mongo/versions/0.3.0/sha1
+++ b/Mongo/versions/0.3.0/sha1
@@ -1,0 +1,1 @@
+e85a9a4c8e78f9ed11740a8abc5decd4d615514b


pzion:
Is there some way to see the failing output of JuliaCIBot?

tkelman:
click on details

Mongo/versions/0.3.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+LibBSON 0.3
```
tkelman: were there API changes at all?

pzion: Nope

pzion:
screen shot 2017-07-06 at 4 38 35 pm

This is all I get for the details, nothing is clickable past that as far as I can tell.

tkelman:
click the red x'es

pzion:
Ah thanks, I'll take a look and see if something stands out

Mongo/versions/0.3.0/requires
tkelman: then doesn't seem 0.3 is strictly required unless you're relying on new functionality? the older tags might still work on julia 0.5 with this new Mongo release?

pzion: I thought the same for LibBSON, but then the Travis build failed telling me that you can't change the julia required version without bumping the minor version.

pzion:
OK, I think what is happening is that in the .travis.yml file we have the following:
```
services:
    - mongodb
```
This starts a local mongodb service to test against.  I'm guessing that the JuliaCIBot doesn't respect this.  I'll see if I can find docs on doing the same in JuliaCIBot.

Mongo/versions/0.3.0/requires
tkelman: minor version of the package itself, not its dependency requirements if it would still work against older releases

aviks:
>I'll see if I can find docs on doing the same in JuliaCIBot

Unfortunately, that is not possible yet. Will get done at some point.

Mongo/versions/0.3.0/requires
pzion: right.  I'll fix things up then

PR #10165, opened 2017-07-08
Tag DataStructures.jl v0.6.0

attobot:
Repository: [JuliaCollections/DataStructures.jl](https://github.com/JuliaCollections/DataStructures.jl)
Release: [v0.6.0](https://github.com/JuliaCollections/DataStructures.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaCollections/DataStructures.jl/builds/251406378)
Diff: [vs v0.5.3](https://github.com/JuliaCollections/DataStructures.jl/compare/63038135479f953ade6571662f4638d9953ebb82...8a555b60adb901776444f33fccf1153a4d7733d9)
`requires` vs v0.5.3: no changes
cc: @kmsquire

Please make sure that:


tkelman:
With this many changes to such a widely-used package, it's probably better to leave julia 0.4 alone and make this 0.5-plus.

kmsquire:
Ok

kmsquire:
The last push drops v0.4 support and should be tagged correctly.

DataStructures/versions/0.6.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.4
```
tkelman: still says it supports 0.4?

kmsquire:
@attobot, why didn't you update?  :-/

(Possibly because I mistagged as `0.6.0` at one point...?)

tkelman:
It might need you to delete both the github release and the git tag before retagging/re-releasing.

kmsquire:
I thought I did that, but I'll try that again

kmsquire:
Okay, that did it.

tkelman:
Looks like this caused a failure in PolynomialMatrices.jl, but only on 0.6? https://github.com/JuliaCI/pkg.julialang.org/blob/fbded8901df6963f13676d491cbc5d2b9898a782/logs/PolynomialMatrices_0.6.log cc @neveritt

kmsquire:
Was PolynomialMatrices recently updated?  I have to run and can't check
right now, but that test was run agains PolynomialMatrices v0.1.1, and the
current version is v0.2.0 (which passes fine locally).

tkelman:
The tag in https://github.com/JuliaLang/METADATA.jl/pull/10245 fixed this already, so I think you're okay, but you may want to pin it to 0.1.1 and see whether the usage it used to be doing is something you'd like to preserve.

kmsquire:
I checked this.  I don't think it's worth preserving.  The functionality that was removed (as you saw) was allowing

```
SortedDict{Int,Float64,Base.Order.ForwardOrdering}()
```

The latest tag broke that, with your fix being to remove the `ForwardOrdering` type parameter

```
SortedDict{Int,Float64}()
```

The functionality could be added back to try to construct the ordering object if it was specified, but for the `Orderings` defined in `Base`, that would currently only work for `ForwardOrdering`s--all other orderings currently require an argument, which is cumbersome (or impossible) to infer from the type parameter.

`ReverseOrdering` could actually be updated (in `Base`) to have a useful no-argument constructor, but it's likely that the order `Order`  will be entirely removed from `Base` before v1.0 now that we can specialize on function arguments.

PR #10168, opened 2017-07-08
Tag Measurements v0.5.0 and update URL

giordano:
From `NEWS.md`:

> ### Breaking Changes
> 
> * `isinteger` and `iszero` now check also that the uncertainty is zero.
> * Comparison between `Measurement` and `Real` with `==` now compares also the
>   uncertainty, so that the `Measurement` must have uncertainty equal to zero to
>   be equal to a real number.
> * Comparison between `Measurement` and `Irrational` with `==` now gives always
>   `false`, consistently with the rest of `Real` types.
> 
> ### New Features
> 
> * New mathematical operations supported: `sinpi`, `cospi`, `sinc`, `cosc`,
>   `asec`, `acsc`, `acot`, `asech`, `acsch`, `acoth`, `sincos` (the last one only
>   on Julia 0.7).
> * It is now possible to parse a string as a `Measurement{T}` with any
>   `T<:AbstractFloat` (not only `Measurement{Float64}`), as long as the parsing
>   method is able to digest the string.  Tested with `T` equal to `Float16`,
>   `Float32`, `Float64`, and `BigFloat`.

Diff `v0.4.0...v0.5.0`: https://github.com/JuliaPhysics/Measurements.jl/compare/v0.4.0...v0.5.0


tkelman:
Be sure to re-enable CI services after the repo move.

giordano:
Didn't I do so?  Both [Travis](https://travis-ci.org/JuliaPhysics/Measurements.jl) and [AppVeyor](https://ci.appveyor.com/project/giordano/measurements-jl) seems to be fine.

Measurements/versions/0.5.0/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.6.0-pre
```
tkelman: which was the first rc where float16 parsing worked? maybe drop the `-pre` when you don't need to support prerereleases any more

giordano: I didn't even know parsing of `Float16` was introduced in Julia 0.6 (couldn't even find it by quickly skimming NEWS.md).  Anyway I dropped the `-pre`, no need to support those versions.  Thanks!

PR #10174, opened 2017-07-08
Register new package SimpleWeightedGraphs.jl v0.0.1

attobot:
Repository: [JuliaGraphs/SimpleWeightedGraphs.jl](https://github.com/JuliaGraphs/SimpleWeightedGraphs.jl)
Release: [v0.0.1](https://github.com/JuliaGraphs/SimpleWeightedGraphs.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGraphs/SimpleWeightedGraphs.jl/builds/251593956)
cc: @sbromberger

Please make sure that:

@sbromberger This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


SimpleWeightedGraphs/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.6
+LightGraphs 0.9.1
```
tkelman: this wouldn't work with 0.8 or 0.9.0 ?

sbromberger: It might but I don't want to support them with this.

tkelman: why not? what if someone has one of those versions pinned to reproduce a result?

sbromberger: Then they won't have a need for this new package, I guess. This is a standalone addition to LightGraphs. I will be supporting the latest release of LightGraphs with each new version.

tkelman: Please don't tighten bounds any more than required for features that are being uses or bug fixes that you're relying on. Bounds aren't how to support what you want to support should be expressed, they're how absolute minimum dependency requirements for things to work should be expressed.

sbromberger: How does one indicate that one doesn't want to support less than the latest release of a dependency?

tkelman: A readme note, or response to any issues that may be opened.

sbromberger: I respectfully disagree that that is an effective method of ensuring the correct level of dependencies. If this were integrated into LightGraphs.jl directly, then it would not be available to anyone who was not at the (newly-tagged) latest version. In my opinion this is the equivalent method of ensuring a consistent, supportable ecosystem when the packages are separated.

tkelman: Otherwise you're forcing Pkg into preventing installation of a range of version combinations that could otherwise work fine together, which can end up with resolver errors, or this package being held back unnecessarily. The tested configuration is worth documenting, but not enforcing since it's overly restrictive.

sbromberger:
> that could otherwise work fine together

As the sole supporter of this package (and a core contributor to LightGraphs), I don't *want* people who are not at the latest version of LG using this package. Even if it works with v0.8.0, I don't have time to test it and subsequent versions, and I don't have time to support it on that old version. If someone finds that it does work and wants to support a fork that works on v0.8.0, they are free to do so.

It should be the package maintainer's decision what versions of dependencies s/he wishes to support, regardless of the technical compatibility of the packages themselves.

SimpleWeightedGraphs/versions/0.0.1/requires
sbromberger: > that could otherwise work fine together

As the sole supporter of this package (and a core contributor to LightGraphs), I don't *want* people who are not at the latest version of LG using this package. Even if it works with v0.8.0, I don't have time to test it and subsequent versions, and I don't have time to support it on that old version. If someone finds that it does work and wants to support a fork that works on v0.8.0, they are free to do so.

It should be the package maintainer's decision what versions of dependencies s/he wishes to support, regardless of the technical compatibility of the packages themselves.

tkelman: The point is that if you don't absolutely require the latest version of your dependency, then a restrictive requirement isn't the "correct level." If they should always move in exact lockstep, why are they separate packages? If they would work against different version combinations, then don't rule those out since there may be valid reasons to need to do things that way.

tkelman:
That's fine in terms of what you want to support. But enforcing that via bounds is going to result in Pkg errors or unintuitive resolver behavior - Pkg bounds are not the right method to express maintainer preferences, only minimum technical compatibility.

SimpleWeightedGraphs/versions/0.0.1/requires
sbromberger: They are separate packages because this is first in a series of what will hopefully be specialized graph implementations, some of which may have other external dependencies that do not apply generally to LightGraphs or other graph implementations. An example will be a graph implementation that uses JuliaDB for storage. Until there are conditional dependencies, there is no great way to do this without splitting into separate packages.

sbromberger:
@tkelman I understand and appreciate your position, though I disagree with it. It speaks to some larger issues, though (but perhaps I'm conflating things):

1) An insistence on technical compatibility being the driver for version restrictions would lead to some pretty perverse incentives to create versions with deliberate incompatibilities.

2) There needs to be a better way of handling the "we only support version X of package A with version > Y of its dependent package B" without creating the risk of resolver errors. (This could take the form of bundled dependencies, I guess, but I'm certainly not an expert in this area.)

3) Conditional dependencies might fix a bunch of this headache. Then we could put things that are "optional" into the main package and have them work *if* some prerequisite dependencies are already installed (and perhaps warn on use if they're not.)

In any case, thank you for this merge and for the discussion.

tkelman:
Pkg3 should probably have a way of distinguishing "recommended" vs "required" dependencies (and bounds), and a way of marking package versions as maintained vs unmaintained, in ways that influence the resolver score to favor more recent things but not absolutely rule out older combinations with errors or infeasibility when they would work and might still be required by someone doing something out of the ordinary.

PR #10181, opened 2017-07-10
Gauss quadrature

billmclean:
Patch release for Julia 0.6 to fix deprecation warnings.


tkelman:
You're using syntax here that doesn't work in julia 0.4 without Compat. Better to drop julia 0
r support by changing the minimum in REQUIRE to 0.5 and re-tagging with a new package minor version

tkelman:
replaced by https://github.com/JuliaLang/METADATA.jl/pull/10183

PR #10182, opened 2017-07-10
Tag InspectDR v0.2.2 [https://github.com/ma-laforge/InspectDR.jl]

ma-laforge:
Fix notebooks to work with Julia v0.6 now that `Interact.jl` works again.

Diff vs v0.2.1: https://github.com/ma-laforge/InspectDR.jl/compare/31bc0b1389a3e3cccf98fc6d4031e1ff6ad31acd...6205596359ca3f10d2431578179d8729364a17a5


InspectDR/versions/0.2.2/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6.0-pre
```
tkelman: drop the trailing -pre when you no longer need to support prereleases

ma-laforge:
@tkelman :

As far as I know, my new code still works with -pre "releases"....

However: I am no longer really testing for -pre "releases" now that v0.6.0 is out.

Question:
Should we be dropping the -pre from REQUIRES if we are not testing on pre-releases? - or do we leave it in just in case someone wants to try it?

tkelman:
It is possible to manually check things out to specific versions even if not technically supported in the REQUIRE file, but Pkg might fight you by undoing those changes next time you add or update anything. If you happen to be using a prerelease of julia for whatever reason (the main valid reason for needing to do that is tracking down the cause of a bug in Base that may be triggered by package code) then running `Pkg.add` would pick the most recent version that claims to support prereleases. But you could equally get a newer version by running `Pkg.add` with release julia then not touching Pkg when testing something with prerelease. If you're sure you're not relying on anything that was fixed or added between the alpha tag and the final release then it doesn't hurt to leave the `-pre` but it may be good to clean it up eventually - and be aware that your dependencies might not have working tags that claim to support `-pre`.

PR #10208, opened 2017-07-11
Pull request/8722a7a5

trappmartin:
Changed version to 0.1.0 as this package has been re-factored to use julia 0.6.


tkelman:
Should probably only register the package with a single initial tagged version. Note that you can modify the branch of your metadata fork in-place instead of opening separate PR's.

trappmartin:
Thx, I'll have a look.

PR #10214, opened 2017-07-11
Tag Gadfly.jl v0.6.3

bjarthur:
1. wide-form data
2. IterTools
3. fix Geom.beeswarm
4. test on 0.6


aviks:
Thanks @bjarthur . I would suggest, for next time, the get the Gadfly and Compose version numbers back in sync. Makes it easier to track things.

PR #10220, opened 2017-07-11
Pull request/22412f72

yakir12:


tkelman:
Would be good to fill in the package description to something other than "No description, website, or topics provided" since that will get used on pkg.julialang.org

yakir12:
Done!

PR #10245, opened 2017-07-13
Tag PolynomialMatrices v0.1.2 [https://github.com/neveritt/Polynomial…

neveritt:
…Matrices.jl]

Diff vs v0.1.1: https://github.com/neveritt/PolynomialMatrices.jl/compare/13ad6b6bf3ae305ab11382fe1b79869c4983a62c...a9f317f8f5c98f004929519412d4e4674f7c5e36

Hopefully, this should fix #10165.


tkelman:
Thanks! Should use a new package minor version since support for julia 0.4 is being dropped

PR #10251, opened 2017-07-13
Tag Images.jl v0.11.0

attobot:
Repository: [JuliaImages/Images.jl](https://github.com/JuliaImages/Images.jl)
Release: [v0.11.0](https://github.com/JuliaImages/Images.jl/releases/tag/v0.11.0)
Diff: [vs v0.10.1](https://github.com/JuliaImages/Images.jl/compare/aa04e63487d46f0e41041e4ca65365ea3aa240ba...2f78e6b9674b39bcad16d7711ac3e3071fe0d755)
`requires` vs v0.10.1: 
```diff
--- 0.10.1/requires
+++ 0.11.0/requires
@@ -16,3 +16,4 @@
 FileIO
 Compat 0.19
 StatsBase   # for histrange
+Requires
```
cc: @timholy

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #10261, opened 2017-07-14
Place upper bound on Images dependency of Pyramids

timholy:
Ref #10251


timholy:
See also https://github.com/loganwilliams/Pyramids.jl/pull/3

PR #10269, opened 2017-07-14
Tag JDBC.jl v0.2.0

attobot:
Repository: [JuliaDB/JDBC.jl](https://github.com/JuliaDB/JDBC.jl)
Release: [v0.2.0](https://github.com/JuliaDB/JDBC.jl/releases/tag/v0.2.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDB/JDBC.jl/builds/253797169)
Diff: [vs v0.1.0](https://github.com/JuliaDB/JDBC.jl/compare/f41a3b3729c1d7bdc5d8cccbd95aaf3a6d03e50b...491d726a3019b73899376c9771ccbe8a31fa9099)
`requires` vs v0.1.0: 
```diff
--- 0.1.0/requires
+++ 0.2.0/requires
@@ -1,5 +1,5 @@
-julia 0.4
-JavaCall 0.4.1
+julia 0.5
+JavaCall 0.5.0
 Requires
-Compat
+Compat 0.17
 DBAPI
```
cc: @aviks

Please make sure that:


JDBC/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+JavaCall 0.5.0
```
tkelman: this does not need to be tightened? slow down here

aviks: Nah, this is good. there were only deprecation fixes between this and 0.5.1. 0.5.0 is good for julia 0.6. So its fine.

tkelman: this is too tight for julia 0.5 though

tkelman: the changes in JDBC aren't related to its uses of JavaCall, and https://github.com/JuliaInterop/JavaCall.jl/compare/v0.4.1...v0.5.0 was otherwise warning fixes

aviks: But I don't wan't people to use lower versions on julia 0.5. I still actively support JavaCall on julia 0.5, and so don't want to support lower javacall versions.

aviks: So its tight due to (my) policy rather than technical reasons, FWIW

tkelman: that's fine, you can recommend, and most people will be on, newer javacall versions. but you're not relying on anything that didn't work on the old version, so you shouldn't rule it out. if someone has javacall 0.4.x pinned or bounded, they shouldn't have jdbc held back for no reason

tkelman: having your bounds too tight doesn't force anyone to upgrade javacall, it just prevents them from upgrading jdbc, unnecessarily ruling out version combinations that would work fine together

tkelman: bounds are for expressing technical requirements, not support preferences. too tight bounds means resolver errors or unintuitive behavior holding versions back more often by reducing the space of allowed combinations.

aviks: No, bounds are also for expressing support preferences. I DO absolutely want to reduce the space of allowed combinations. That is exactly what I want to do here.

tkelman: If someone asks for JavaCall 0.4.1 and JDBC 0.2.0, why should that be an error? It would work correctly, it should be allowed.

aviks: Because I don't have the time or inclination to worry about if JavaCall 0.4.1 works with JDBC 0.2.0. In this case you have seen that it will, but i can't be bothered to check this across all the packages that I maintain. JavaCall 0.4.1 was tagged months ago, I have no clue what was in it (rhetorically). It is impractical (and unfair) to insist that package maintainers keep all of this in their heads when tagging.

tkelman: We should add separate requirements vs recommendations for Pkg3 so the resolver favors more recent supported versions of things. Maybe even warn for non-recommended combinations. But given requirements end up visible to users as Pkg errors or an inability to upgrade bounded packages, it's a heavy handed way to express "I don't think you should do that" when there isn't an actual problem or technical issue that warrants a tighter bound.

I'm asking that bounds only be tightened to address known incompatibilities or real requirements.

PR #10271, opened 2017-07-15
Tag GraphIO.jl v0.1.0

attobot:
Repository: [JuliaGraphs/GraphIO.jl](https://github.com/JuliaGraphs/GraphIO.jl)
Release: [v0.1.0](https://github.com/JuliaGraphs/GraphIO.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaGraphs/GraphIO.jl/builds/253843256)
Diff: [vs v0.0.1](https://github.com/JuliaGraphs/GraphIO.jl/compare/fcceba1f8db57f5ab6ac2135069de3e53358d8c7...89831d40841fdc0053db0124ecc0d071cceb4232)
`requires` vs v0.0.1: 
```diff
--- 0.0.1/requires
+++ 0.1.0/requires
@@ -1,6 +1,7 @@
-julia 0.6.0-pre
-LightGraphs 0.8.0
-GZip 0.2.20
-EzXML 0.3.0
+julia 0.6-
+LightGraphs 0.9.2
+GZip 0.3.0
+EzXML 0.4.4
 ParserCombinator 1.7.11
-JLD 0.6.3
+JLD 0.6.11
+SimpleTraits 0.5.0
```
cc: @sbromberger

Please make sure that:


GraphIO/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6-
```
tkelman: you're using struct syntax, this should be at least 0.6.0-pre https://github.com/JuliaGraphs/GraphIO.jl/pull/3 or release now

GraphIO/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
...
+LightGraphs 0.9.2
+GZip 0.3.0
```
tkelman: there are only warning fixes in https://github.com/JuliaIO/GZip.jl/compare/v0.2.20...v0.3.0, please only raise lower bounds for a hard requirement or to avoid a known problem

GraphIO/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
...
+EzXML 0.4.4
```
tkelman: similarly you aren't relying on any of the new features between https://github.com/bicycle1885/EzXML.jl/compare/v0.3.0...v0.4.4

GraphIO/versions/0.1.0/requires
sbromberger: I don't / can't know this to be true. I don't have a way to do matrix testing against multiple versions of each package - what's listed is what I know works, because it's what I use in my environment.

GraphIO/versions/0.1.0/requires
sbromberger: See above. I am only willing to support this package with these versions of dependencies because that's what I'm using locally. I don't know whether there are performance regressions or other issues with different combinations of dependencies.

tkelman: I locally tested against the lower bounds you had at the previous tag, worked fine. Enforcing bounds that are too tight is problematic, please only raise your lower bounds to express a known requirement of a bug fix or feature.

GraphIO/versions/0.1.0/requires
sbromberger: In fact, after a quick look, I don't believe that's true.
https://github.com/JuliaIO/GZip.jl/compare/JuliaIO:69b6765...JuliaIO:d42aca3#diff-67162e70d7b3c9f54f2ecd99d469a98bL30 shows that `readall` was replaced by `readstring`. `readall` now throws an error in 0.6.

GraphIO/versions/0.1.0/requires
sbromberger: I appreciate that you've tested this, but I haven't and am not comfortable supporting versions that I won't be able to troubleshoot should people have problems.

GraphIO/versions/0.1.0/requires
tkelman: it gives a warning on import and defines a package-local version

tkelman: doesn't look like GraphIO calls GZip's readstring

GraphIO/versions/0.1.0/requires
tkelman: That's fine - very few people will ever need to, Pkg will auto-update most users. But for anyone who needs to pin to something older, you're causing Pkg resolver errors or an inability to upgrade by bounding things too tightly.

sbromberger: Can't they just use the previous version of GraphIO then? I mean, if they're unable or unwilling to upgrade some of their packages, then they should be comfortable with the fact that they won't get the latest versions of other packages that use those as dependencies.

jpfairbanks: Is there a good way to do the matrix testing? 

Or how would one handle a case where we don't bump the dependency requirements and then there is a problem and we need to bump it later?

tkelman: Pkg.pin and Pkg.test in a loop works well. If you find a known issue later, you can retroactively tighten the bounds in-place in the metadata requires file. It's better to have things too loose until there's a known problem that can be addressed separately. Or spend a minute or two with blame to find when the features or bugfixes you know you're relying on were originally introduced.

tkelman: Yes they can keep using the previous version of GraphIO, but it also means that if someone needs to downgrade or pin to a tiny bit older version of one of your dependencies than the bound you set here, Pkg will downgrade GraphIO which you probably don't want.

sbromberger:
@tkelman – as this is the second time you and I have run into this issue, and because I'm apparently still not understanding it adequately from your perspective, perhaps we could get some core committers to weigh in on the discussion and help me understand why it's not the package maintainer's prerogative to specify the versions of dependent packages s/he wishes to support? I'm happy to do it offline, but I really don't think it's a long-term correct policy to have to do matrix testing on all dependencies, and this is putting a serious barrier in front of releasing new versions of packages that I can support.

GraphIO/versions/0.1.0/requires
jpfairbanks: "Tighten the bounds in-place" does that mean "change the REQUIRE file associated with that version without creating a new version"?

jpfairbanks:
I think part of the miscommunication comes from the fact that @sbromberger wants to certify, "I know this configuration works" and only be held responsible for supporting users using that configuration. But the METADATA system want to certify "if you use versions outside these bounds it will definitely not work" and those are two different sets of configurations.

sbromberger:
> the METADATA system want to certify "if you use versions outside these bounds it will definitely not work" 

I think you've got it. The problem is, I have *no way of practically testing* this condition.

jpfairbanks:
We only need two real checks "everything at the lower bound you want to impose" and "everything at the upper bound you want to impose" and if those two pass it should be fine because everyone is using semantic versioning

sbromberger:
> We only need two real checks "everything at the lower bound you want to impose" and "everything at the upper bound you want to impose" 

We have 7 dependencies in GraphIO.

> everyone is using semantic versioning

Semver [is largely undefined for < 1.0.0](http://semver.org/#spec-item-4).

tkelman:
I'd like to help you understand the global resolution consequences of having bounds too tight. I frequently see Pkg end up using very old versions of packages due to some upper bound that is put in place to avoid a known problem. Resolution constraints can have very unintuitive consequences. METADATA lower bounds should be absolute minimum requirements for things to work, not a documentation of known-to-work and tested versions. I'm hoping in Pkg3 we can more clearly separate those pieces of information, between mandatory requirements and recommended versions.

I'm not asking that you test against your lower bound versions, I agree that's a combinatorial problem and impractical. But I do hope that you recognize what you're using and require from each of your dependencies, and given everything is on github and fairly simple to track changes, a few minutes should be enough to get a reasonable guess on an approximate minimum. Better to be too loose and allow more version combinations that might be too wide (for the rare individual who might need to do so and would appreciate not getting a Pkg error from that), than be too tight and prevent simultaneous installation of version combinations that would actually work. An error from mismatched dependencies can be addressed with after-the-fact tightening (changing just the metadata copy of requires bounds later, not touching the package since it's a past published version), but Pkg errors or unexpectedly using old versions of some packages are messier.

I also need to write up a thorough piece of documentation explaining all this with some example scenarios.

GraphIO/versions/0.1.0/requires
tkelman: Change the metadata requires file of an existing tag. The git sha won't change, just metadata's copy of the requirements data. It starts as a copy of the package REQUIRE file at tagging time, but doesn't have to remain the same.

sbromberger:
Crazy idea: what if we just drop the version bounds entirely? If it works, great. If it doesn't, we say, "upgrade to these versions if you want support".

tkelman:
Not that crazy. Though if you're using some functions or relying on some bug fix that didn't exist in every single tagged version ever of your dependencies, you can set a known floor below which it's obvious that things won't work.

jpfairbanks:
We can probably get to a good place by having the loosest bounds like @tkelman has described and then a note in the README with "known to work" combinations that have been tested. We could put `print(Pkg.status())` in the `.travis.yml` file so that travis will print out a set of versions that are known to work.

sbromberger:
I don't know what the lowest bounds are and don't have time to test, so I'm in favor of just removing them.

jpfairbanks:
The old lower bounds should roll over to the new version unless there is a known patch that we needed to bump

sbromberger:
But where did the lower bounds themselves come from? I've been doing this forever.

jpfairbanks:
They should have started at 0.0.0 and got bumped only when necessary

sbromberger:
Right. So we start over. "0.0.0" is the same as "no lower bound".

sbromberger:
@tkelman I'm even more convinced now that packages should be able to include their own versions of dependencies and not rely on a "global install" of a specific version.

GraphIO/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6
+LightGraphs
```
sbromberger: But here's the thing: I don't want to have to manually manage each dependency. That way lies madness.

sbromberger: Since I'm responsible for LightGraphs, I guess I can do it there.

tkelman:
Environments will still have to be consistent, unfortunately there's no way to have two different versions of the same module loaded in memory at the same time in Julia at the moment. But having fewer packages at a time, only on a project basis rather than everything that you ever felt like installing will mean fewer interdependencies to satisfy.

jpfairbanks:
I think @sbromberger just came around to the *vendor all the deps* point of view

sbromberger:
> I think @sbromberger just came around to the vendor all the deps point of view

I've always been a strong supporter of this. It's why OSX did things right with bundling libs into apps, as opposed to how many Linux systems did it relying on shared libraries. Disk space is cheap enough nowadays. Packages could be named internally *with their version* to provide uniqueness, so that package A uses `LightGraphs@0.8.2` and package B uses `LightGraphs@0.9.0`, and each bundles its own version.

GraphIO/versions/0.1.0/requires
tkelman: What do you mean manage? List what you require? The rest probably have had much more stable API's over time at least. This becomes easier if it's tracked incrementally. Maybe a big parsed corpus of all exports and function names of all package tags could automatically determine what's being used.

sbromberger: It means that in order to put lower bounds on everything the way you suggest, I have to track the changes to these dependent packages and only change things when I know the old version will break the code. I don't have time to do that for a package with 7 dependencies where all the dependents have active development communities and tend to release frequently, and I can't do my own testing because I make sure that I'm running the latest and greatest of ALL packages on my development system (which happens also to be my work system).

tkelman:
Right but then if package A and B have to talk to each other, what happens if A tries to send a type from a one version of LightGraphs to a function from B that's expecting a different version of that type?

sbromberger:
> Right but then if package A and B have to talk to each other, what happens if A tries to send a type from a one version of LightGraphs to a function from B that's expecting a different version of that type?

The same thing that happens now when package A requires a version of package B that's not compatible: the code breaks.

sbromberger:
Also, attobot's message might be confusing:
> Please make sure that:

> CI passes for supported Julia versions (if applicable).
> Version bounds are up to date.

My version bounds *were* up to date. That's what caused all this kerfuffle in the first place.

PR #10279, opened 2017-07-16
Tag Clang v0.2.0 and v0.3.0

ihnorton:


Clang/versions/0.2.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.6-
```
tkelman: shouldn't claim to support early dev versions where `abstract type` wouldn't work

ihnorton: fixed. thanks.

Clang/versions/0.1.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5- 0.6
```
tkelman: none of the earlier versions have upper bounds, is this needed if 0.2.0 is available?

tkelman:
It's unlikely that you'll ever make another 0.4 release, but this should probably be 0.2.0 and 0.3.0 rather than 0.1.1 and 0.2.0

PR #10285, opened 2017-07-16
Tag ColorSchemes v1.4.0

cormullion:
I moved this repo to JuliaGraphics and attempted to tag a new release without the help of Attobot. I have a suspicion that I've done something wrong, but haven't yet worked out what it might be...


ColorSchemes/versions/1.4.0/requires
```diff
@@ -0,0 +1,8 @@
+julia 0.6-
+Images 0.10.0
```
tkelman: does this need to be tightened? the `reshape` with colon wouldn't work on early 0.6-dev versions so better to use 0.6 release as the minimum julia version

cormullion: Good thinking. Will change it.

tkelman: and what about Images, are you relying on a feature or bugfix that was in 0.10 but not 0.6?

cormullion: Oh, I'll have to research that. I just got it working with the current version of Images and updated the versions. Images moves fast....

tkelman:
replaced by https://github.com/JuliaLang/METADATA.jl/pull/10325 ?

cormullion:
I suppose I could delete this one?

tkelman:
if redundant, yeah

cormullion:
Moving repos is a bit tricky, IMHO.

PR #10288, opened 2017-07-16
Tag BinDeps.jl v0.7.0

attobot:
Repository: [JuliaLang/BinDeps.jl](https://github.com/JuliaLang/BinDeps.jl)
Release: [v0.7.0](https://github.com/JuliaLang/BinDeps.jl/releases/tag/v0.7.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaLang/BinDeps.jl/builds/254215255)
Diff: [vs v0.6.0](https://github.com/JuliaLang/BinDeps.jl/compare/264e086062f3c39adba7997f187bd54bc0e6b218...e2553e57468d5692f7e13f45ddc888670bcbbfda)
`requires` vs v0.6.0: no changes
cc: @vtjnash

Please make sure that:


tkelman:
It would be much, much less risky if https://github.com/JuliaLang/BinDeps.jl/pull/277 only used the BinDeps-local version of this code on nightly Julia.

vtjnash:
This is moving code the code from C (where we know it occasionally fails silently, because it used to crash CI until we added code to make it fail silently), to Julia (where that failure should be much less likely, and also get reported correctly).

tkelman:
I've asked you three or four times now in https://github.com/JuliaLang/BinDeps.jl/pull/277 how many packages you tested this against, and not gotten a clear answer. Simple question, simple answer.

vtjnash:
Yes, I made sure it 
```
julia> BinDeps.debug("HDF5")
INFO: Reading build script...
 - Library "libhdf5"
    - Satisfied by:
      - BinDeps.AptGet package hdf5-tools at /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.10
    - Providers:
      - BinDeps.AptGet package hdf5-tools
      - BinDeps.Pacman package hdf5 (can't provide)
      - BinDeps.Yum package hdf5 (can't provide)
      - Autotools Build
```

tkelman:
In how many cases? Please answer the question.

vtjnash:
Several. Above I'm just showing the success of the latest test.

vtjnash:
CI seems to have gotten confused by the accidental close. Do I need to make a new PR?

tkelman:
The existing code isn't going anywhere on 0.5 or 0.6. Does this fix anything that didn't already work? It's rather risky that it could break in a situation you haven't tested, so I still think this change should be nightly only.

vtjnash:
> Does this fix anything that didn't already work? 

Yes. This handles error reporting correctly and reliably, and removes one minor source of unpredictably and undetectable failures from BinDeps.

>  it could break in a situation you haven't tested

The existing code breaks in situations that I _have_ encountered. As a bugfix, this improves the reliability of the ldconfig handling on all versions.

tkelman:
Can you give a concrete example of the old code doing the wrong thing? Should add tests for that if possible.

tkelman:
If this breaks anything it's getting reverted. And I'd still appreciate a specific example of the old code doing the wrong thing. "Unpredictabl[e] and undetectable failures" and "situations that I *have* encountered" don't help anyone else understand what the problem is.

PR #10290, opened 2017-07-16
Tag TextAnalysis v0.2.0 [https://github.com/johnmyleswhite/TextAnalys…

aviks:
…is.jl]

Diff vs v0.1.0: https://github.com/johnmyleswhite/TextAnalysis.jl/compare/0a1c2b1a5d62b715ae3e919f39a10b316fc8e631...9724c7e55f40bc450d3b60721d0df2adc1bfe0e8


TextAnalysis/versions/0.2.0/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.5
+Languages 0.1.0+
```
tkelman: the trailing rules out 0.1, is that what you meant here?

aviks:
yes

tkelman:
that's an odd way to express that and would allow an 0.1.1 if that happened to get released, wouldn't it be better to list the inclusive lower bound matching what you require?

aviks:
There was some combination of dependencies, local environment and travis that forced me do it this way. Can't be bothered to remember why exactly unless this breaks anything now .

tkelman:
It doesn't make a lot of sense to say "anything higher than 0.1.0, but not including it" is your minimum requirement, if your requirement is actually the first version after it that exists right now

aviks:
As i said, there was a reason why I had to do it this way, and at this point, it doesn't hurt much, and i have higher priority issues to fix.

tkelman:
There aren't even any code changes in https://github.com/johnmyleswhite/Languages.jl/compare/v0.1.0...v0.2.0. Seems totally unnecessary.

tkelman:
It does hurt, quite visibly. Pinning Languages.jl to 0.1.0 downgrades TextAnalysis, asking for Languages 0.1.0 and TextAnalysis 0.2.0 at the same time errors. Neither of those are helpful behaviors for any users to encounter.

PR #10304, opened 2017-07-17
Tag DiffBase.jl v0.3.0

attobot:
Repository: [JuliaDiff/DiffBase.jl](https://github.com/JuliaDiff/DiffBase.jl)
Release: [v0.3.0](https://github.com/JuliaDiff/DiffBase.jl/releases/tag/v0.3.0)
Diff: [vs v0.2.0](https://github.com/JuliaDiff/DiffBase.jl/compare/c987aed2639ae24f0a031d27d159baa62d06c6db...d070cffd49a9a90ae4a183e7124fab66dc17ea03)
`requires` vs v0.2.0: 
```diff
--- 0.2.0/requires
+++ 0.3.0/requires
@@ -1 +1,2 @@
 julia 0.6.0-pre
+StaticArrays 0.5.0
```
cc: @jrevels

Please make sure that:


jrevels:
This release has one API change; the mutating DiffResult functions have license to construct a new output rather than mutate the output given to them. This is because some new DiffResult types are fully immutable (see https://github.com/JuliaDiff/ForwardDiff.jl/pull/236). Thus, it's now consider "good style" when using DiffBase's mutating functions to re-alias the output. For exampe:

```julia
using ForwardDiff, DiffBase
f(x) = sum(sin, x) + prod(tan, x) * sum(sqrt, x)
x = rand(4)
result = DiffBase.HessianResult(x)

# old style
ForwardDiff.hessian!(result, f, x)

# new style
result = ForwardDiff.hessian!(result, f, x)
```

Theoretically, this shouldn't affect anybody at the moment, since ForwardDiff/ReverseDiff are currently non-inclusively upper-bounded at DiffBase v0.3.0.

tkelman:
Only the most recent versions of ForwardDiff and ReverseDiff apply upper bounds on DiffBase. In order for that bound to be effective (if this is breaking for those packages), should apply that upper bound to all the previous tags as well.

tkelman:
(which is #9484 - only thing that needs addressing there is doing something about the older versions of packages that initially didn't depend on DiffBase)

DiffBase/versions/0.3.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.6.0-pre
```
tkelman: to match https://github.com/JuliaLang/METADATA.jl/pull/9484 this should probably be 0.6 release

PR #10309, opened 2017-07-17
Tag POMDPs.jl v0.5.2

attobot:
Repository: [JuliaPOMDP/POMDPs.jl](https://github.com/JuliaPOMDP/POMDPs.jl)
Release: [v0.5.2](https://github.com/JuliaPOMDP/POMDPs.jl/releases/tag/v0.5.2)
Diff: [vs v0.5.1](https://github.com/JuliaPOMDP/POMDPs.jl/compare/5b41716b01985a6e6b83ba2b5a8c89febc4ea66b...e0f4195a21711f4f51a4f69ab6f08a0194f49e89)
`requires` vs v0.5.1: 
```diff
--- 0.5.1/requires
+++ 0.5.2/requires
@@ -1,2 +1,2 @@
-julia 0.5
+julia 0.5 0.6
 Distributions
```
cc: @MaximeBouton

Please make sure that:


POMDPs/versions/0.5.2/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.5 0.6
```
tkelman: upper bounds are not effective unless applies to all past versions. this just means on 0.6, people will be held back to POMDPs 0.5.1

MaximeBouton: Should we just leave julia 0.5 as the requirement?

tkelman: now that there's a newer version available, most people on 0.6 will be using that. anyone who isn't will have a good reason not to be.

MaximeBouton: You mean people on 0.6 will be using the 0.6 version of POMDPs.jl? The release that we are submitting here only supports up to julia 0.5 (included) but we submitted another PR for the version that supports 0.6.

tkelman: right but there were no code changes here, just the readme and bounds - the bounds aren't accomplishing much, and I don't think many people look at local copies of readmes (I could be wrong there)

zsunberg: Ok, yeah, we don't need this. Sorry for the clutter. I was thinking there was a case where the upper bound would help the package manager figure things out, but now I don't think that is true. We can abandon this release.

PR #10311, opened 2017-07-17
Tag POMDPs.jl v0.6.0

attobot:
Repository: [JuliaPOMDP/POMDPs.jl](https://github.com/JuliaPOMDP/POMDPs.jl)
Release: [v0.6.0](https://github.com/JuliaPOMDP/POMDPs.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaPOMDP/POMDPs.jl/builds/254650375)
Diff: [vs v0.5.1](https://github.com/JuliaPOMDP/POMDPs.jl/compare/5b41716b01985a6e6b83ba2b5a8c89febc4ea66b...ea497223ec3aa8c1b28c1fdd51a789d51f259a81)
`requires` vs v0.5.1: 
```diff
--- 0.5.1/requires
+++ 0.6.0/requires
@@ -1,2 +1,2 @@
-julia 0.5
+julia 0.6
 Distributions
```
cc: @MaximeBouton

Please make sure that:


tkelman:
Looks like ParticleFilters.jl won't work against this new version of POMDPs? We can add an upper bound to the POMDPs dependency in https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/ParticleFilters/versions/0.0.1/requires, https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/ParticleFilters/versions/0.0.2/requires, and https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/ParticleFilters/versions/0.0.3/requires to hold POMDPs back to 0.5.x if someone has ParticleFilters installed, until there's a version of ParticleFilters available that's compatible with POMDPs 0.6.

MaximeBouton:
ParticleFilters.jl will work, there is a branch 0_6 ready to merge. It is the same for all the package from JuliaPOMDP. They all rely on the three base modules that are POMDPs, POMDPToolbox and POMDPModels. Do you recommend we submit the PR for all the packages at the same time? We made the changes so that they all work with the 0.6 version of the three base modules.
Thanks for the help

tkelman:
The new version will work, but the existing version won't, right? The METADATA copy of the requires files can be retroactively edited, and upper bounds can be added for dependencies on old versions that won't work with the new version of the dependency.

MaximeBouton:
Is it best to just tag new compatible versions of all of our registered packages (POMDPs, POMDPToolbox, POMDPModels and ParticleFilters.jl) simultaneously?

MaximeBouton:
(I mean first merging versions of POMDPToolbox, POMDPModels, and ParticleFilters that are compatible with POMDPs 0.6 and julia 0.6 and then tagging all of them and submitting PRs to METADATA)

tkelman:
Perhaps, but if they require one another, attobot doesn't currently handle putting multiple packages in the same PR. If they're all ready, can open them now and once each is merged, we can reset CI on the others.

zsunberg:
Hi @tkelman , so to make sure I understand this correctly, the JuliaCIBot is reporting that after this tag, ParticleFilters won't work, which is what is holding this PR up?

Is there documentation for the JuliaCIBot somewhere? I am interpreting that the red x in the ParticleFilters-0.6.0-rc1-New cell of the JuliaCIBot output means that "the Travis CI tests for the ParticleFIlters packages will fail on julia 0.6 if this PR is merged." Am I correct, or is it some other tests that are failing.

Is a passing JuliaCIBot test required to pass for merging this PR into METADATA?

When you say "we can reset CI on the others", are you referring to JuliaCIBot or Travis CI for METADATA or Travis CI for the individual packages, or some combination?

zsunberg:
I think the easiest solution is to tag compatible versions of all four packages (POMDPs, POMDPToolbox, POMDPModels, and ParticleFilters). Once they are all tagged, everything should work, but I do not know all the versioning steps and modifications that would be needed to get JuliaCIBot to pass at each step of merging them in sequence.

Should we submit METADATA PRs for all four simultaneously or try to figure out a sequence of changes that will get everything to pass at every step?

tkelman:
> Is a passing JuliaCIBot test required to pass for merging this PR into METADATA?

Passing JuliaCIBot is not required, but it provides some useful pre-merge testing - we will be documenting exactly how it works, and probably opening up the repo, pretty soon. If you click on the X or checkbox it'll take you to a test log. What it does is test the reverse-dependencies (everything that depends on this package) to see if it was working against the previous version, does this new version break it. In this case it identified that yes, ParticleFilters would start failing its tests on Julia 0.6 after a Pkg.update() if this gets merged without doing anything about requirement bounds.

> I am interpreting that the red x in the ParticleFilters-0.6.0-rc1-New cell of the JuliaCIBot output means that "the Travis CI tests for the ParticleFIlters packages will fail on julia 0.6 if this PR is merged." Am I correct, or is it some other tests that are failing?

Correct.

> When you say "we can reset CI on the others", are you referring to JuliaCIBot or Travis CI for METADATA or Travis CI for the individual packages, or some combination?

I'm referring to Travis CI for METADATA, if the tags were individually opened via attobot, one package at a time. Since presumably the new version of ParticleFilters (and other downstream packages) will depend on this new version of POMDPs, if you open a separate ParticleFilters tag PR right now, Travis CI for METADATA would complain that it depends on a version of POMDPs that might not be published yet (if Travis for that PR runs before this PR gets merged).

I think JuliaCIBot doesn't handle PR's that tag multiple packages at the same time correctly, but that's okay, we can let PkgEval or post-merge Travis CI on the individual packages identify problems if needed.

There are two or three options that would minimize the amount of time Pkg.add would give you a broken combination of packages.

1. Upper bound the existing ParticleFilters versions' dependencies on POMDPs to < POMDPs v0.6. Part of the change in https://github.com/JuliaLang/METADATA.jl/issues/10321 etc, but better to remove the Julia upper bound that's there right now. What this would do is hold POMDPs back to 0.5.x whenever you have or ask to install one of these existing ParticleFilters versions. We can merge this safely, POMDPs by itself will work, ParticleFilters would use an old POMDPs version until you release the new ParticleFilters version which could be done on its own later.

2. Prepare individual tags for each package without touching the bounds. Would be good to get these all ready so they can be merged in quick succession when they all look okay.

2.a) use attobot and do them in separate one-package-at-a-time PR's - the downstream packages would fail METADATA travis if they depend on POMDPs >= v0.6 and that's not merged yet, but I can restart METADATA travis immediately after merging the POMDPs tag to fix that.

2.b) put all the related tags in the same PR - have to use PkgDev for this, and JuliaCIBot might not run the right thing, but it would also work.

zsunberg:
Ok, I think @MaximeBouton and I have it figured out. Will submit the correct PRs soon.

zsunberg:
Ok, so once #10331 gets merged, the JuliaCIBot should pass this PR, then we can merge this, and then after that do ParticleFilters, POMDPModels, and POMDPToolbox

tkelman:
JuliaCIBot needs to be tweaked a bit to actually obey bounds - currently it checks things out in a funny way that would ignore bounds, so it's not quite representative of what happens when people to Pkg.add or Pkg.update. But we should be safe now. If you do `Pkg.add("POMDPs")` on Julia 0.6 and don't have ParticleFilters installed, you'll get POMDPs 0.6.0. If you do have ParticleFilters installed, POMDPs will stay at 0.5.1 for now, until the new version of ParticleFilters gets published.

zsunberg:
great, thanks!

PR #10320, opened 2017-07-18
Tag SwitchTimeOpt v0.2.0

bstellato:
New patch release SwitchTimeOpt.jl

* Support for DifferentialEquations.jl
* Removed deprecations and warnings in Julia 0.6


SwitchTimeOpt/versions/0.1.1/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5-
```
tkelman: if you don't need to support prereleases of julia 0.5, better to drop the trailing dash

tkelman: if you use `abstract type` syntax, then you don't support 0.5 at all so the minimum julia version should be changed to 0.6 and this should be tagged as a new package minor version

PR #10321, opened 2017-07-18
Add upper bound in require

MaximeBouton:


ParticleFilters/versions/0.0.1/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.5
+julia 0.5 0.6
```
tkelman: Does it load at all on 0.6? upper bounding means the package will not be installable at all on julia 0.6. Probably better to leave it without an upper bound, and once an 0.6-compatible version is released most people will upgrade to it with Pkg.update.

MaximeBouton: This release is made to run with POMDPs.jl v0.5.2 and won't run on 0.6.

tkelman: I get some deprecation warnings, but ParticleFilters 0.0.3 and POMDPs 0.5.1 pass tests for me on Julia 0.6.0.

tkelman: Having Julia upper bounds on all the versions would make `Pkg.add("ParticleFilters")` an error, which seems extreme to me and isn't needed to avoid the incompatibility noted in https://github.com/JuliaLang/METADATA.jl/pull/10311

PR #10332, opened 2017-07-18
Tag POMDPModels v0.2.0, POMDPToolbox v0.2.0, ParticleFilters v0.1.0

zsunberg:
These are all dependent on each other for their Travis tests to pass, but we have verified locally that they will pass once this is merged. Is this the best way to do it, or should we request these changes some other way?


ParticleFilters/versions/0.1.0/requires
```diff
@@ -0,0 +1,3 @@
+julia 0.6
+StatsBase
+POMDPs 0.4
```
tkelman: this requires 0.6 minimum, right?

MaximeBouton: julia 0.6 yes; it also needs POMDPs 0.6 ; I will make the change and submit a new PR. Do you think it would be better to submit a new one for the three pkgs at the same time? (same as what @zsunberg did with the fix on the requirements)

tkelman: You actually don't need a new PR, you can just edit this branch of your metadata fork in place. As long as you make the corresponding change to the REQUIRE file on master of ParticleFilters it's fine

MaximeBouton: I don't have access to the metadata fork unfortunately @zsunberg does but will not be available, I will fork and do another PR if you don't mind

tkelman: If he checked the "allow maintainers to edit" box then I could make changes here, if that would be easier

MaximeBouton: I think that would help thanks, the change is just POMDPs 0.4 -> POMDPs 0.6

tkelman: and looks like you moved the tag sha to https://github.com/JuliaPOMDP/ParticleFilters.jl/commit/f526b040b2a0a6d206bb64c6824dbb104999442b

PR #10339, opened 2017-07-19
PDFIO Registration v0.0.1

sambitdash:
PDFIO is a PDF parser written in pure Julia. 

v0.0.1 requested for registration.


PDFIO/versions/0.0.1/requires
```diff
@@ -0,0 +1,5 @@
+julia 0.6
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need. Though since you've made the package 0.6-only, do you need Compat at all right now?

sambitdash: My original plan was to implement for 0.5 and 0.6. Half way through the development 0.6 was released. So I removed the support for 0.5. 

The current intent is to be able to support 0.7 and 0.6. While I have not embraced the 0.7 syntax yet any library developed currently should keep that in mind. Hence, this Compat lib was added. It does not actually use much as such as I have not used any 0.7 syntax explicitly yet.

PR #10346, opened 2017-07-19
Tag BlackBoxOptim v0.2.2 [https://github.com/robertfeldt/BlackBoxOpti…

robertfeldt:
…m.jl]

Many changes since v0.2.1 since we had tagged but not published them in between. Latest drops julia 0.4 support and fixes many deprecation warnings on 0.6 (though inner ctor warnings still there since we will need to drop 0.5 support to fix this so holding off some more time until most users have moved to 0.6).

Diff vs v0.2.1: https://github.com/robertfeldt/BlackBoxOptim.jl/compare/4fc185d8009f9f7d6c3e14219fe7742f072842ed...5862706fb4725ee6584edf9ef316c7388155f944


tkelman:
If you're dropping 0.4 support you'll need to use a new package minor version

robertfeldt:
Yes, you're right. Will tag as v0.3.0 and try again...

PR #10351, opened 2017-07-20
new package: HCubature

stevengj:
This package provides a pure-Julia alternative to the Cubature.jl and Cuba.jl packages (both of which rely on C libraries, and e.g. are not type-generic), complementing the pure-Julia QuadGK.jl (1d integrals).

(Closes stevengj/HCubature.jl#1)

cc @ararslan

### DIFF
--- a/HCubature/versions/1.0.0/requires
+++ b/HCubature/versions/1.0.0/requires
@@ -1,0 +1,6 @@
+julia 0.6
+Compat
+StaticArrays
+Combinatorics
+DataStructures
+QuadGK
--- a/HCubature/url
+++ b/HCubature/url
@@ -1,0 +1,1 @@
+https://github.com/stevengj/HCubature.jl.git
--- a/HCubature/versions/1.0.0/sha1
+++ b/HCubature/versions/1.0.0/sha1
@@ -1,0 +1,1 @@
+d5bef9c5bd6055985b1efcb2f1cbfa4d41baa777


HCubature/versions/1.0.0/requires
```diff
@@ -0,0 +1,6 @@
+julia 0.6
+Compat
```
tkelman: Are you relying on anything post-0.6 from this? If so, best to declare a minimum version dependency that implements the features you need.

tkelman:
Also should avoid the intermittent download flakiness that the Cubature package has been hitting https://pkg.julialang.org/detail/Cubature.html

HCubature/versions/1.0.0/requires
stevengj: So far, I'm not actually using anything from Compat as far as I know.  I put it in more as a precaution to help keep things from breaking in 0.7.

stevengj:
Yeah, I need to move those Cubature binaries to bintray.

stevengj:
Performance is pretty good (a bit better than Cubature.jl), but it is not as good as I would like, I think primarily because `struct Box` isn't stack-allocated.  I think we are all hoping that a future Julia version will do less allocation for transient structs.

stevengj:
Okay to merge?

tkelman:
Prefer keeping new package registrations open for at least 24 hours of initial feedback, but looks okay to me

PR #10352, opened 2017-07-20
Tag Chemfiles v0.7.4 [https://github.com/chemfiles/Chemfiles.jl]

Luthaf:
Diff vs v0.6.0: https://github.com/chemfiles/Chemfiles.jl/compare/980d92c5f67ea6ccf2c745ee49ce5fd30edb86be...7165227d23dad7702dbe8807c4ff4b0b1ca93f50


Chemfiles/versions/0.7.4/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+BinDeps
+Compat
```
tkelman: It's best to declare a minimum version dependency that implements the features you need.

Luthaf: Ok I'll add them. 

Should `PkgDev.tag` send a warning about this?

tkelman: we should add some checking that looks at your package and suggests bounds automatically

is there something in BinDeps 0.4 that wasn't in earlier versions you're relying on?

Luthaf: > is there something in BinDeps 0.4 that wasn't in earlier versions you're relying on?

Not really, but that was the version I had on my machine and I ran the tests with. Considering that most Julia users are always on the last version of every package (because `Pkg.update()` updates everything to the latest version), I think this should be fine.

tkelman: it can be harmful to set bounds overly tight though, it can cause downgrades or held back packages or Pkg resolver errors if anyone needs to bound or pin to a dependency older than what you tested with, even if it might work fine

Luthaf: Is there any way for me to automatically pin the packages to the version in `REQUIRE` when testing on Travis ? Else it will be pretty hard for me to guarantee that the code is working with a specific version of the dependencies.

I've heard about a Pkg3 proposal for Julia, will it still work with a global shared package library, or will it be more akin virtualenv, where the resolver does not need to satisfy the constrains of all packages, even unrelated ones?

tkelman: you can call Pkg.pin but it's probably not worth the hassle most of the time, just set bounds if there's a known issue or something obviously won't work from looking at what features the code needs to use.

yes pkg3 is a prototype but being worked on, there's a julep and there was a juliacon talk on it. it will have environments for doing project specific resolution

aviks:
@Luthaf is this ready to go?

Luthaf:
Yes, fine for me ! 👍

PR #10355, opened 2017-07-20
PDFIO registration

sambitdash:
1. Fixed the temp file clean up.
2. Fixed the tag issues reported. 
3. Compat has been left intact to accommodate for 0.7 Julia development.


tkelman:
0.0.2, 0.0.3, and 0.0.4 all look like the same sha?

you may want to check out https://github.com/attobot/attobot which automates some of the tagging

sambitdash:
I was using PkgDev.tag to tag. It just kept incrementing the tags. Now when I published all the 3 versions got added to a pull request. I do not think I can manually do any changes to tags as that may lead to more confusions.

tkelman:
You can manually modify this branch of your metadata fork if needed. Not strictly necessary but could be confusing to have duplicated tags.

sambitdash:
I am not very conversant with git as such. So will not like to add further confusion to my understanding at this point in time. I will fix this in a subsequent time when my knowledge improves in that area. If it's not mandatory please accept the pull request and move forward.

tkelman:
Sure. I'd recommend using attobot next time around, especially if you're not very familiar with git.

PR #10399, opened 2017-07-24
Register new package FirstOrderSolvers.jl v0.1.0

attobot:
Repository: [mfalt/FirstOrderSolvers.jl](https://github.com/mfalt/FirstOrderSolvers.jl)
Release: [v0.1.0](https://github.com/mfalt/FirstOrderSolvers.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/mfalt/FirstOrderSolvers.jl/builds/256833653)
cc: @mfalt

Please make sure that:

@mfalt This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


mfalt:
JuliaCIBot fails on 0.6 because the tests require a newer version of Convex.jl than what currently exists in METADATA. I intend to update the test/REQUIRE file to point to it once Convex.jl tags a new version.
It works on my travis since I pull the latest version. Can this still be accepted or do I have to wait until Convex tags a new version?

FirstOrderSolvers/versions/0.1.0/requires
```diff
@@ -0,0 +1,7 @@
+julia 0.6-
```
tkelman: do you need to support prereleases? if not, drop the trailing dash

mfalt: All right, removed the tag, and added it again to the new commit, it seems like attobot was able handle it properly.
Is there a reason why its not good to support pre-releases?

tkelman: 0.6- means the very first commit after 0.5 branched, so to support that you'd basically need to support 0.5

tkelman:
You should be listing Convex in test/REQUIRE if it's a test-only requirement. Your tests may fail for the majority of users who are on release versions, but that's not the end of the world.

mfalt:
All right, fixed.

PR #10413, opened 2017-07-25
Tag ODEInterface v0.1.4 [https://github.com/luchr/ODEInterface.jl]

luchr:
Added interface to 

* [Colsys/Colnew](https://people.sc.fsu.edu/~jburkardt/f77_src/colnew/colnew.html), a multi-point boundary value problem solver. 

It is licensed under GNU LGPL. `ODEInterface` compiles `colnew` and creates a shared library, which is "only" used (=dynamically loaded) by `ODEInterface` and no parts of the `colnew` code are included in `ODEInterface`. Hence this is IMHO an use-case covered by the LGPL license.

Is it possible to support different julia versions in different branches? Like:
```
   v0.1.4          v0.2.0 (for julia ≥ 0.7)
─────┴──┬────────────┴───────── branch: master
        │
        └────┬───────────────── branch: julia_v0.4-v0.6
           v0.1.5  ... v0.1.*  (for 0.4 ≤ julia ≤ 0.6)
```
Because of [Julia Issue 17146](https://github.com/JuliaLang/julia/issues/17146) I don't dare to tag `ODEInterface` commits that can only be found in the branch `julia_v0.4-v0.6` and then use the `master` branch for julia v0.7/nightly only.


tkelman:
> Is it possible to support different julia versions in different branches? Like:

Yes, though I'd recommend enabling branch protection on any branch you plan on making releases from just so you don't accidentally delete it. If it works any better for you, you could choose to have your 0.7-only branch the one that isn't master for a while.

luchr:
A new Pull-Request for the v0.7-only version (which I've already prepared) will follow soon.

Thank you for your support.

PR #10415, opened 2017-07-25
Tag ODEInterface v0.2.0 [https://github.com/luchr/ODEInterface.jl]

luchr:
Attemt:
Splitting `ODEInterface` in two different branches, depending on the julia version:
* Branch `julia_v0.4-v0.6`: `ODEInterface` for 0.4 ≤ julia ≤ 0.6 with future `ODEInterface`-versions: v0.1.5, v0.1.6, etc.
* Branch `master`: `ODEInterface` for julia ≥ 0.7 with `ODEInterface`-versions: v0.2.0 and fugure versions.

So Version v0.2.0 has the same features as v0.1.4, but written in julia-v0.7 style (`struct`, `mutable struct`, spaces around binary operators like `≥`, compatibility code/macros deleted, etc.).

Picture:
```
   v0.1.4          v0.2.0 (for julia ≥ 0.7)
─────┴──┬────────────┴───────── branch: master
        │
        └────┬───────────────── branch: julia_v0.4-v0.6
           v0.1.5  ... v0.1.*  (for 0.4 ≤ julia ≤ 0.6)
```

I hope the contents of `REQUIRE` and `requires` is right for the current v0.7/nightly build.

### DIFF
--- a/ODEInterface/versions/0.2.0/requires
+++ b/ODEInterface/versions/0.2.0/requires
@@ -1,0 +1,1 @@
+julia 0.7-DEV
--- a/ODEInterface/versions/0.2.0/sha1
+++ b/ODEInterface/versions/0.2.0/sha1
@@ -1,0 +1,1 @@
+e2bf5c3f013b2759d7c289bab2cebd5373fcf54c


tkelman:
This looks okay, though it might be a good idea to leave yourself more versioning room since you're dropping support for 0.4, 0.5, *and* 0.6 all at the same time here. If you ever need to make another bugfix release that supports say 0.5 and 0.6 but not 0.4 any more, may want to leave a separate space for it.

luchr:
Thank you very very much for this "versioning"-advice. I'll do this and open a separate pull request (today or tomorrow). Please don't merge this one.

luchr:
Sorry, me again with one question about "more versioning room".

Here:

https://github.com/luchr/ODEInterface.jl/blob/master/VERSIONS.md

is one approach. But that doesn't work because `PkgDev` doesn't allow the creation of v0.4.0 (in time) before v0.2.0.
Are the some "best practice" numbering schemes?

tkelman:
That's an annoying limitation of PkgDev. You're probably better off using https://github.com/attobot/attobot

http://semver.org is the best practice for versioning, though we usually recommend a slightly stronger version that says "0.1.x" versions *can* have breaking changes or features but it's a good idea to use new minor versions when those changes are large, even if the major version is still 0. And if you drop support for a specific minor version of Julia, that's considered a form of breaking change even if the package code doesn't actually change significantly, so worth a new package minor version.

Having version numbers that increase quickly or leave gaps isn't a problem (though I didn't realize PkgDev would complain), and if you end up needing to go back and make backport bugfix releases in the future it's useful to have space to put them in case they need to be distinguished.

PR #10467, opened 2017-07-28
upper bound Tokenize for OhMyREPL 0.5

KristofferC:
Fixes https://github.com/KristofferC/OhMyREPL.jl/issues/93

Is this correct @tkelman?

### DIFF
--- a/OhMyREPL/versions/0.1.0/requires
+++ b/OhMyREPL/versions/0.1.0/requires
@@ -1,3 +1,3 @@
 julia 0.5
-Tokenize 0.0.1 0.1.8
+Tokenize 0.0.1 0.1.9
 Compat 0.9.5
--- a/OhMyREPL/versions/0.1.0/requires
+++ b/OhMyREPL/versions/0.1.0/requires
@@ -1,3 +1,3 @@
 julia 0.5
-Tokenize
+Tokenize v0.0.1 v0.1.8
 Compat 0.9.5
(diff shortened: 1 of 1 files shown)


tkelman:
I would leave off the v prefixes. Can this be fixed in Tokenize? If not, may want to start using a new minor there to reflect the API break.

KristofferC:
Yeah, I didn't realize that would be an API break. If this fixes it, I would like to go with this simple fix since it is technically for a version of OMR that is not supported anymore.

tkelman:
Right but I'd request that you start using 0.2.x numbering in Tokenize ASAP, maybe even making a duplicate of your latest tag as 0.2.0. People are still using OhMyREPL on 0.5 and it would be good to leave versioning room for possible backport bug fixes just in case.

tkelman:
actually which version caused the problem, 0.1.8 or 0.1.9? upper bounds are exclusive

KristofferC:
The 0.1.9 version has problems. I tagged a 2.0 Tokenize

tkelman:
so if 0.1.8 worked fine, the upper bound should be 0.1.9

KristofferC:
Fixed

KristofferC:
Seems like attobot didnt pick up the new tag, perhaps because it is on the same commit as the previous one.

tkelman:
Maybe it assumes that was unintentional? Hm, could manually propose copying the 0.1.10 metadata files to a new 0.2.0 then

PR #10498, opened 2017-07-31
Register package AbaqusReader.jl

ahojukka5:
A package to parse ABAQUS .inp files.


tkelman:
Needs a minimum Julia version requirement

PR #10529, opened 2017-08-01
Tag DiffEqBase.jl v1.16.0

attobot:
Repository: [JuliaDiffEq/DiffEqBase.jl](https://github.com/JuliaDiffEq/DiffEqBase.jl)
Release: [v1.16.0](https://github.com/JuliaDiffEq/DiffEqBase.jl/releases/tag/v1.16.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqBase.jl/builds/259788644)
Diff: [vs v1.15.1](https://github.com/JuliaDiffEq/DiffEqBase.jl/compare/f7603656bf59128b3a3146d04920c70e841cf726...80df0f6fe831a819fa6f549d72a3f251ba96ac32)
`requires` vs v1.15.1: no changes
cc: @ChrisRackauckas

Please make sure that:


ChrisRackauckas:
Tests are expected to fail because they use OrdinaryDiffEq.jl and StochasticDiffEq.jl which on the latest release are bound to not use this version because of the removal of the `@muladd` macro. That was just removed and made its own package, MuladdMacro.jl, which is now used on master, but that cannot be tagged easily either because of a naming dispute since the old DiffEqBase exports the old `@muladd`. Since it's heavily used throughout, we hope that we can just get this to pass through without tests passing, which would then fix everything. Thanks.

PR #10530, opened 2017-08-01
Tag NIDAQ.jl v0.3.0

bjarthur:
support julia 0.6 and NI-DAQmx 17.1.0


tkelman:
Compat needs to be listed in REQUIRE if it's used in your source. Though other than the `using Compat` doesn't look like you're using it for anything?

bjarthur:
oi.  my bad.  fixed now.  probably not kosher though to try and move the git tag to point to this new commit.  will i just have to tag and publish 0.3.1?

tkelman:
Either way, since it hasn't been published to metadata yet it's fine to modify if you want to change it.

bjarthur:
fixed here https://github.com/JuliaLang/METADATA.jl/pull/10556

PR #10544, opened 2017-08-01
Tag FFTW.jl v0.0.4

attobot:
Repository: [JuliaMath/FFTW.jl](https://github.com/JuliaMath/FFTW.jl)
Release: [v0.0.4](https://github.com/JuliaMath/FFTW.jl/releases/tag/v0.0.4)
Travis: [Travis Build Status](https://travis-ci.org/JuliaMath/FFTW.jl/builds/259957626)
Diff: [vs v0.0.3](https://github.com/JuliaMath/FFTW.jl/compare/070cd891fda64f956aab5c0f5062948ade060a16...9d57158dc646852d430a06a11584b35643b52048)
`requires` vs v0.0.3: no changes
cc: @ararslan

Please make sure that:


ararslan:
I still have no idea what the JuliaCIBot failure is supposed to convey.
screen shot 2017-08-01 at 3 03 36 pm
There is literally no information in this.

KristofferC:
```
LoadError: LoadError: failed process: Process(`apt-get install libfftw3-double3`, ProcessExited(100)) [100]
while loading string, in expression starting on line 87
while loading /mnt/pkgevalempty/34ac051cf321dd2e9f9116dd4c041bc9d126c805/DSP/.julia/v0.7/FFTW/deps/build.jl, in expression starting on line 3

================================================================================
WARNING: Compat.ASCIIString is deprecated, use String instead.
  likely near string:87
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

StefanKarpinski:
It tells you exactly what it tested! :trollface:

ararslan:
Ah, had to allow loading unsafe scripts, then the links show up. Yes, looks like it's failing on 0.7 because it can't install things using the BinDeps APT interface.

aviks:
@ararslan That probably explains why some people have been complaining about not being able to see the logs. However, I'm surprised that this is an issue, since the links are plain href's, and there should be no javascript involved. There is a CSS, that is used to display the icons, which in turn would load a font. So we'd appreciate some help debugging this. 

* What browser?
* Where are you setting your "allow unsafe script" preference? In the base browser settings, or do you have some plugin installed?

ararslan:
> What browser?

Firefox, though I get the same thing in Chrome

> Where are you setting your "allow unsafe script" preference? In the base browser settings, or do you have some plugin installed?

In Firefox it shows up as a yellow triangle over a lock. Clicking that gives you the ability to load unsafe scripts.
screen shot 2017-08-01 at 3 40 26 pm

In Chrome it shows up as a shield with a red line through it. Similarly, clicking that allows you to select loading unsafe scripts.

I don't think I have Chrome plugins installed but on Firefox I have HTTPS Everywhere, so simply going to the HTTP version of the site doesn't do it for me. The inability to load unsafe scripts seems to be the default in both browsers and I don't see a way to change that (at least in Firefox).

aviks:
Ah, OK, gottit. You must have some equivalent of HTTPS-Everywhere in chrome as well, since the JuliaBot CI "Details" link on this page is `http`. The `<link>` element  specifying the CSS on that page is also specified as `http://`, so when you land on the page with `https://` you get an _unsafe script_ warning.  I can replicate this now. 

So the solution is to specify the link to the CSS with relative protocol.  We'll get that done. Thanks !

tkelman:
We should also add a plaintext link next to the icons, since even when they load correctly it's not obvious to everyone that clicking on the icon links to the logs

ararslan:
Maybe the words "Passed" and "Failed"? Sascha Mann was saying that his ad blocker blocks custom fonts, so he wasn't even seeing the icons, just □.

PR #10545, opened 2017-08-01
Upper bound older versions of AbstractTrees

Keno:
These versions of AbstractTrees don't work with julia 0.6, don't let the user install them


tkelman:
What doesn't work about them? They all seem to pass their tests on Julia 0.6

Keno:
There's invalid subtyping syntax in the package prior to 0.1.0, I'm surprised that tests would pass.

tkelman:
I spoke a little too soon, 0.0.4 is indeed broken on julia 0.6, but 0.0.1 through 0.0.3 don't seem to be

aviks:
Tests fail with `ERROR: ERROR: LoadError: LoadError: packages with unsatisfiable requirements found:`

Seems risky to merge as is.

MikeInnes:

PR #10548, opened 2017-08-01
Register and tag FEMBasis.jl

ahojukka5:
Package allows user to define new interpolation basis functions used in FEM calculations easily by defining vertices of interpolation domain + polynomial used to interpolate


tkelman:
looks like TimerOutputs is needed in `test/REQUIRE`

ahojukka5:
Can I have different REQUIRE file for tests?

tkelman:
yes, a file called `test/REQUIRE` is for test-only dependencies

PR #10584, opened 2017-08-04
Tag mPulseAPI v1.0.6

bluesmoon:
1.0.6 handles some new bugs introduced in the upstream API.


tkelman:
You need to push the git tag. Are you still testing against Julia 0.4?

bluesmoon:
Ok, I've set DataFrames version to 0.7.3 and updated 1.0.5's requires as well.  Older versions don't care about the DataFrames version.

PR #10600, opened 2017-08-04
ParallelAccelerator and CompilerTools for 0.6.

DrTodd13:


tkelman:
ParallelAccelerator is missing a dependency on SpecialFunctions

tkelman:
I pushed a commit to your branch here that removes the duplicate tags (CompilerTools 0.2.2 and ParallelAccelerator 0.2.3), and the tags of ParallelAccelerator that were missing a dependency so would not work (0.3.0 and 0.3.1). We can add them back if there's a need to do so, but I don't think there's any reason for that.

PR #10606, opened 2017-08-05
Tag UniqueVectors v1.0.0 [https://github.com/garrison/UniqueVectors.jl]

garrison:
Diff vs v0.4.0: https://github.com/garrison/UniqueVectors.jl/compare/a7b83e82bc49763d31612fcf6595a60230d35e25...a7b83e82bc49763d31612fcf6595a60230d35e25


UniqueVectors/versions/1.0.0/sha1
```diff
@@ -0,0 +1 @@
+a7b83e82bc49763d31612fcf6595a60230d35e25
```
tkelman: this is the same commit as 0.4.0? https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/UniqueVectors/versions/0.4.0/sha1

garrison: Sorry, and thanks for catching this.  I was confused why the diff above listed nothing and should have figured this out myself.  [Turns out my last action before running PkgDev was to "test" on julia master, and while I thought I had set `~/.julia/v0.7/UniqueVectors` to be a symlink to my development directory, it was not, and therefore `PkgDev.tag` and `PkgDev.publish` instead used the latest commit there, the same as `v0.4.0`.]

Since I already pushed the faulty `v1.0.0` tag, I fear that I may break `Pkg.update()` for people who already received the tag if I force push another, different `v1.0.0`.  So my current plan is to call this release `v1.0.1` instead and delete the `v1.0.0` tag.  Do you have a better idea/suggestion?

tkelman: that's fine, but since this pr wasn't merged, the tag isn't in anyone's metadata. the only people who would see a difference if you replaced 1.0.0 in place would be someone who fetched and checked out 1.0.0 locally. anyone else will get the tag replaced next time they fetch

garrison:
OK, fixed, and here is the correct diff: https://github.com/garrison/UniqueVectors.jl/compare/a7b83e82bc49763d31612fcf6595a60230d35e25...538a5f89cbf63a1e89019dc0361c9767331354ba

Now just waiting on CI

garrison:
Actually, I am going to abort this release due to a small backwards-incompatible change I'd like to make to the package before declaring 1.0.  Will open a new pull request when it's ready.

PR #10616, opened 2017-08-05
Register and Tag ModelReduction.jl

ahojukka5:
ModelReduction.jl uses various of techniques to decrease the dimensionality of discretized FE problem to make it easier to solve in long transient simulations. Main algorithms are Guyan reduction and Craig-Bampton methods. As a result one obtain a reduced system, so called "super element" what can be used in multibody dynamic softwares for rapid time simulations. Currently, Guyan reduction is implemented.

CC: @maikkirapo


tkelman:
missing minimum julia version requirement

tkelman:
And are these fairly application-specific methods of model reduction? That's a pretty generic name that could apply to a lot of fields, not necessarily strictly finite element methods.

ahojukka5:
Good question. For example, Guyan reduction is basically eliminating dofs from linear system of equations to make model smaller. It includes things like calculating Schur complement. So it's purely mathematical operation where unknown are divided into two sets, the ones to keep and other ones to eliminate.

I'm not so familiar with Craig-Bampton but I think it's also quite general way to reduce model described by differential equations. So we have (discretized) system `Mu'' + Cu' + Ku = f` and we want to reduce degrees of freedom from that. The starting point for the algorithms are M, C, K given without taking assumptions how they are actually calculated.

I would not say that these can be used *only* in FEM calculations. In the other hand, I cannot figure out any other ways to achieve this kind of big models with a clear explanation of unknown so that we know what we actually want to eliminate from system. Maybe these techniques could be used in FDM, FVM of CDF also.

Any comments, @maikkirapo, @TeroFrondelius, @jvaara ?

ahojukka5:
The book we are following here is [Model Order Reduction Techniques with Applications in Finite Element Analysis](http://www.springer.com/gp/book/9781852338077). From mathematical point of view the same methods are applicable  to not only FEM but with other kind of discretization schemes also, although the author is focusing on FEM. Based on the name of book I would conclude that model order reduction techniques are quite general and book just introduces applications in FEM world. Do you have some suggestion how the name of package should be improved?

PR #10635, opened 2017-08-07
Register and Tag Mortar3D.jl

ahojukka5:
Mortar3D.jl is a Julia package to calculate discrete projections between non-conforming finite element mesheds. The resulting "mortar matrices" can be used to tie non-conforming finite element meshes together which are meshed separately to construct bigger models.


ahojukka5:
What means?

tkelman:
think that was a bug, restarted it

Mortar3D/versions/0.0.1/sha1
```diff
@@ -0,0 +1 @@
+a692a300de6e3657b2a0bf57d0df100a0119f2c9
```
tkelman: package is kinda empty at that commit, but okay

PR #10644, opened 2017-08-08
Tag AWS* and dependancies

samoconnor:
Various 0.6 compatibility tweaks.
Various bug fixes and feature enhancements per user requests.
Support in AWSCore for 102 new service packages: https://github.com/JuliaAWS (not yet registered).
Lots of new documentation. Deployed here: https://juliacloud.github.io/AWSCore.jl/build/index.html


tkelman:
FNVHash should use a new package minor version since julia 0.4 support is being dropped

samoconnor:
Thanks for picking that up @tkelman, do you have any other concerns before I update the PR?

tkelman:
Looks like any of the downstream packages that are using new modules in AWSCore should declare the corresponding minimum version requirement

samoconnor:
I agree with "should". However, I feel that the benefit to end-users is marginal given that they typically do `Pkg.update()` frequently anyway and rarely update just `AWSCore. I feel like the time it would take me to re-jigger the tags and fight with git to fix up the PR would be better spent getting the stuff I have just written for `AWSCore v0.2.1` released.
Are your "should" comments hard blocking requirements for merging the PR>

tkelman:
If you create a v0.1.0 git tag for FNVHash with the same commit, everything else can be fixed with metadata-only edits to this branch, as long as the corresponding changes get made on master before the next tag. None of the downstream packages look like they'll load at all for AWSCore prior to 0.2.0, either due to referencing `AWSCore.services` or `AWSCore.escape_path`. If anyone has local modifications or has pinned AWSCore, they'd be allowed to upgrade the other packages and they wouldn't load unless we add the bounds that tell Pkg not to do that.

samoconnor:
I've update all the REQURESs like this https://github.com/samoconnor/AWSS3.jl/commit/aa11bbeeffbd891f344af3d25f5ce3aaea236cd8

samoconnor:
FNVHash is re-tagged https://github.com/samoconnor/FNVHash.jl/releases/tag/v0.1.0

tkelman:
Thanks, check that e00e34ccbfa6ac577e8c19cb1c6e30780e3f300e looks okay to you - should pass travis now (after an unfortunately long queue right now, sorry bout that)

PR #10646, opened 2017-08-08
Tag Turing.jl v0.3.0

attobot:
Repository: [yebai/Turing.jl](https://github.com/yebai/Turing.jl)
Release: [v0.3.0](https://github.com/yebai/Turing.jl/releases/tag/v0.3.0)
Travis: [Travis Build Status](https://travis-ci.org/yebai/Turing.jl/builds/262298353)
Diff: [vs v0.2.0](https://github.com/yebai/Turing.jl/compare/2165bd8ba0f9608c00b6e01b025f49240702b59f...f2e4437d8740c960048f1bb0d302579ca4fefde8)
`requires` vs v0.2.0: 
```diff
--- 0.2.0/requires
+++ 0.3.0/requires
@@ -1,8 +1,6 @@
 julia 0.5
 
 Distributions 0.11.0
-PDMats 0.5.1
 ForwardDiff
 Mamba
-Stats
 ProgressMeter
```
cc: @yebai

Please make sure that:


ararslan:
It looks like Turing is only tested on Julia 0.5. Is it only known to work there? It appears you're defining `@~`, which no longer parses as of Julia 0.6 (or maybe early 0.7-dev), so I imagine the tests would likely fail at least on Julia nightly. It might be worth adding upper bounds to the required Julia version if newer versions are known not to work, otherwise it would be good to add more recent Julia to the testing matrix.

tkelman:
> It might be worth adding upper bounds to the required Julia version if newer versions are known not to work

I rarely recommend this actually, all it would mean is older versions of Turing would get used instead. Unless you were to upper bound every past tag of Turing, in which case the package would effectively be deprecated and not installable at all on post-bound Julia versions.

yebai:
We'd like to start support Julia 0.6 in Turing 0.3.* releases. However as @ararslan mentioned, the `@~` is dropped from Julia 0.6. This causes issues on the Turing side. The `@~` macro is very helpful for defining concise statistical notations, we hope that we can continue to write code like `x ~ Normal(0,1)`. We are currently investigating into how to support this feature for Julia 0.6, and it might take some time to find a solution. On the other side, it is slightly surprising that `@~` syntax is dropped, since not much work is really required to keep supporting it from the Julia code base.

https://github.com/yebai/Turing.jl/issues/173

ararslan:
FWIW DataFrames (and the forthcoming StatsModels) uses `@formula(y ~ x)` as a replacement. You can take a look at the docs and implementation to see whether something similar would suit your needs. `~` still parses as an infix operator with assignment precedence, so parsing what you need inside of formulas is quite straightforward.

> it is slightly surprising that `@~` syntax is dropped, since not much work is really required to keep supporting it from the Julia code base.

`@~` was the sole infix macro, so it had to be special-cased in the parser; removing that exception simplifies the parser and makes things more consistent. It really only existed as an R-ism that can be accomplished in a more Julian manner with regular macros. But anyway, that's an aside.

aviks:
@yebai should this still be merged?

yebai:
@aviks Yes, please. We'll soon make a new release that fixes compatibility with Julia 0.6 and fixes other known issues.

cc. @xukai92

PR #10662, opened 2017-08-09
Upper bounds for Hecke

thofma:


Hecke/versions/0.2.1/requires
```diff
@@ -1,2 +1,2 @@
 julia 0.5
-Nemo 0.6.2
+Nemo 0.6.2 0.6.2
```
tkelman: upper bounds are exclusive, so should be 0.6.3

thofma: Oops, fixed.

PR #10689, opened 2017-08-11
update TranscodingStreams.jl and friends

bicycle1885:
This updates the protocol used in TranscodingStreams.jl and its friends, which must be merged at the same time.


tkelman:
I see you caught and fixed the Pkg.checkout that TranscodingStreams shouldn't be doing in its tests on a released version, thanks for that.

It looks like the existing versions of Codec* would be broken if TranscodingStreams got updated but a Codec* package was still using 0.2.x or earlier (from pinning or local modifications or an upper bound), would it be worth adding a TranscodingStreams 0.3 upper bound on the previous tags of Codec* ?

bicycle1885:
I think nobody depends on specific versions of Codec* packages and there is no reason to pin a version or set an upper bound of a package. Fortunately (or unfortunately?), not so many people use these packages at the moment, so I think it is mostly safe to merge this without setting the upper bounds.

tkelman:
FYI, this started failing on packageevaluator: https://github.com/JuliaCI/pkg.julialang.org/blob/30dd9a418cf3b4076a21cd0ec26b22cb8de47a97/logs/TranscodingStreams_0.6.log

bicycle1885:
Thank you. There may be a rare bug because I cannot reproduce it locally and on [TravisCI](https://travis-ci.org/bicycle1885/CodecXz.jl).

PR #10697, opened 2017-08-12
Tag ApproxFun v0.7.0 [https://github.com/JuliaApproximation/ApproxFun…

dlfivefifty:
….jl]

Diff vs v0.6.1: https://github.com/JuliaApproximation/ApproxFun.jl/compare/7a20a13d029560ad7dc95ec897fcf12ed1d0ca3a...7cb661a27331e74191d6a6fe1e85db16b3652ad9

This version drops support for Julia 0.5


ararslan:
The custom `RowVector` implementation that shadows the one in Base isn't exported, right? Perhaps better to use a separate name for your implementation?

dlfivefifty:
There arequite a few Base shadowing functions that are not exported, for example many of the infinite iterators have analogues that support getindex.

Since they are only used internally it doesn't affect other packages.

ararslan:
Okay cool, just checking.

dlfivefifty:
I'm going to retag and submit a new pull request using the suggestions from @tkelman

PR #10702, opened 2017-08-13
Register LabelNumerals [https://github.com/sambitdash/LabelNumerals.j…

sambitdash:
…l.git]: v0.0.1

Numeric quantities are used sometimes for pure representational purposes without any true a numeric significance. For example, the page numbering is carried out using simple arabic numerals, roman numerals, alphabets. These numbers have additive properties but may not have any multiplicative significance. Such numbers can also have prefix notations as well. In some cases, the representation can be in upper case or lower case as well. This package implements such a numerals. Such numeric schemes are used as page numbers in PDF file specification. However, the need may be felt else where as well, which prompted the author to implement it as an independent package. The interface has been also influenced significantly by the RomanNumerals package.


sambitdash:
Tags have been pushed. Please rerun the build.

sambitdash:
closing and reopening fixed the issue.

sambitdash:
Please hold this for now. I will need to add a new revision (0.0.2) which will have additional capabilities.

sambitdash:
0.0.2 already added. Please review for merging.

PR #10713, opened 2017-08-14
Tag OAuth and LogParser

randyzwitch:


tkelman:
OAuth still has a use of Compat in its tests that should be removed if it's no longer in REQUIRE

randyzwitch:
Sorry about that. Removed Compat import from tests, deleted/re-tagged OAuth v0.5, changed sha1 commit here

ararslan:
Looks like the check here is failing because it's expecting a leading `v` in the tag name, i.e. v0.5.0 rather than just 0.5.0.

randyzwitch:
@ararslan I updated it on GitHub, but since it's not a push, is there a way for you to re-run the Travis?

Needing a v in front seems excessively specific ;)

PR #10716, opened 2017-08-14
mPulseAPI does not currently work with julia 0.7

bluesmoon:
Updating requires files for all versions since our current tests show that we don't work with julia 0.7.

### DIFF
--- a/mPulseAPI/versions/1.0.0/requires
+++ b/mPulseAPI/versions/1.0.0/requires
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.7-
 Requests
 LightXML
 HttpCommon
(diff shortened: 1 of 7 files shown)


tkelman:
Are you sure you want to make Pkg.add an error? This is effectively deprecating the package.

bluesmoon:
How would you suggest I specify that the package only works with 0.4, 0.5 and 0.6?  At the moment I can't even test on 0.7 because it complains about my documentation being in the wrong format.

tkelman:
It's fine for the package not to work there yet (a note in the readme maybe), the question is whether you want to disallow installing it there altogether.

bluesmoon:
nvm, I've made it work with 0.7.

PR #10722, opened 2017-08-14
Tag CSTParser.jl v0.1.3

attobot:
Repository: [ZacLN/CSTParser.jl](https://github.com/ZacLN/CSTParser.jl)
Release: [v0.1.3](https://github.com/ZacLN/CSTParser.jl/releases/tag/v0.1.3)
Travis: [Travis Build Status](https://travis-ci.org/ZacLN/CSTParser.jl/builds/264465082)
Diff: [vs v0.1.1](https://github.com/ZacLN/CSTParser.jl/compare/276263ebf1ca9c4cffae08c9749daa03b9a52ace...6ce7bbb8bd4f96fd657ca4a064667490a7713d65)
`requires` vs v0.1.1: 
```diff
--- 0.1.1/requires
+++ 0.1.3/requires
@@ -1,2 +1,3 @@
 julia 0.6.0-pre.beta
 Tokenize 0.1.10
+AbstractTrees
```
cc: @ZacLN

Please make sure that:


ZacLN:
How can I prevent it testing against 0.7?

ararslan:
Prevent what, your package? It already looks like you aren't testing against 0.7 on Travis.

ZacLN:
Oh I'm getting confused by the JuliaCIBot

tkelman:
The deletion of `Variable` appears to be breaking for LanguageServer, so that should have upper-bounds added

ZacLN:
I've added it to the waiting PR

tkelman:
Which? The bounds need to be on the existing versions in METADATA

ZacLN:
Do you need us to publish a new release (of LanguageServer) with the upper bounds limit before a release using CST 0.1.3?

tkelman:
No, we should amend the requires files of the existing LanguageServer tags, only within METADATA, to indicate that this new release of CSTParser will break them. Anyone who has LanguageServer installed should therefore have their CSTParser held back to a version that would work correctly, until a new LanguageServer is available that doesn't have an upper bound (so is compatible with the new CSTParser).

ZacLN:
Ah I see, I think @davidanthoff  could do that (?)

davidanthoff:
The proper thing to do would be to bump the version of CSTParser.jl to v0.2.0 because it breaks things, and then add upper limits to LanguageServer.jl, right? We should do that in any case, just in case we have to publish a bug fix release only on the CSTParser.jl 0.1.x branch.

I'll open a PR with upper bounds for LanguageServer.jl in a minute.

davidanthoff:
#10759 adds an upper bound of v0.2.0 on CSTParser.jl in all the registered LanguageServer.jl versions. Next step would be for @ZacLN to remove the v0.1.3 tag from CSTParser.jl and replace it with a v0.2.0 tag.

ZacLN:
Alright, will do

PR #10728, opened 2017-08-15
Add TotalVariation Package

fundamental:
Adds a package implementing total variation and group sparse total variation denoising


fundamental:
Per the 0.7-dev failure it seems to be an issue with loading and/or precompiling DSP.jl's FFTW.jl dependency (i.e. it's DSP.jl's bug and they appear to be aware of it). It takes an extra load and close locally, but on my machine with julia built with today's git head DSP.jl loads and TotalVariation.jl runs through its tests without problems.

PR #10759, opened 2017-08-16
Add bounds to LanguageServer.jl requirements

davidanthoff:
I also had to add lower bounds, and I just picked the first version of CSTParser.jl that was registered in METADATA.jl. I'm almost certain that those lower bounds are not right, but the main purpose of this PR is to get the upper bounds in place.

### DIFF
--- a/LanguageServer/versions/0.1.0/requires
+++ b/LanguageServer/versions/0.1.0/requires
@@ -1,5 +1,5 @@
 julia 0.6.0-rc1
 JSON 0.8.1
 URIParser 0.1.8
-CSTParser
+CSTParser 0.0.2 0.2.0
 Tokenize
--- a/LanguageServer/versions/0.1.1/requires
+++ b/LanguageServer/versions/0.1.1/requires
@@ -1,5 +1,5 @@
 julia 0.6.0-rc1
 JSON 0.8.1
 URIParser 0.1.8
-CSTParser
+CSTParser 0.0.2 0.2.0
 Tokenize
--- a/LanguageServer/versions/0.1.2/requires
+++ b/LanguageServer/versions/0.1.2/requires
@@ -1,5 +1,5 @@
 julia 0.6.0-rc1
 JSON 0.8.1
 URIParser 0.1.8
-CSTParser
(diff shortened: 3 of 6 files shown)


tkelman:
Lower bounds matching the first registered CSTParser version are fine, you could also even set them to 0.0.

One somewhat counterintuitive issue here is since the very first version of LanguageServer didn't depend on CSTParser at all, Pkg could potentially choose to downgrade LanguageServer to 0.0.1 in order to upgrade CSTParser to 0.2.0. Two ways of potentially avoiding that would be adding a "fake" dependency on CSTParser with an upper bound of 0.2.0 to LanguageServer 0.0.1, or upper-bounding LanguageServer 0.0.1 to julia < 0.6 since all CSTParser versions are 0.6-only. Neither of those are ideal though since it does look like LanguageServer 0.0.1 would actually work fine on Julia 0.6, so maybe the best option is not to worry about that possibility and just get a CSTParser-0.2-compatible version of LanguageServer released soon.

davidanthoff:
I added the fake dependency on CSTParser for now.

tkelman:
Sorry I should have been a little clearer, that actually would make LanguageServer 0.0.1 not installable at all on Julia 0.5, so that' actually not a usable solution here

tkelman:
We could make an empty dummy CSTParser 0.0.1 that is installable on Julia 0.5, kind of a silly thing to do but it would be an option

tkelman:
"not worry about it for now" is simpler since as I said LanguageServer 0.0.1 does actually work on Julia 0.6. If we see it downgrade and it becomes a real issue then we can address it, but whether it will happen depends a bit on the weighting in the objective function for the Pkg dependency resolver.

davidanthoff:
Can we merge this for now? There is absolutely no need for LS 0.0.1 to work anywhere, and this PR here seems to solve the main issue, namely the upper bounds.

tkelman:
The first commit I'd merge, the second I'd rather not unless there's evidence of the downgrading LanguageServer actually happening to anyone

davidanthoff:
Alright, now with only the first commit.

PR #10764, opened 2017-08-16
Register LCA.jl Package

fundamental:
LCA.jl provides an implementation of the Locally Competitive Algorithm, a sparse approximation algorithm originally designed for analog implementations.


ararslan:
Hey @fundamental, looks good overall. I put in a PR to move stuff around a little bit. In particular, the code that uses PyPlot should be separated from the package code, otherwise PyPlot should be declared a dependency of the package.

fundamental:
@ararslan Thanks again for taking the time to make a PR. Let me know if you think I should update the tagged version, though it shouldn't be an issue as the "using PyPlot" in the v0.0.1 tag was within an "if(false)" block.

ararslan:
Yeah, the current tagged version should be fine. I guess the only thing at this point is that we typically discourage acronyms in package names unless they're quite well-known. Is that the case for LCA? I've not heard of it, so I just wanted to check.

fundamental:
Within literature/in-person-interactions discussing/referencing LCA is typically referred to as 'LCA' and not the full 'Locally Competitive Algorithm'. So, at least within the signal processing community LCA should only refer to this algorithm. 

If relevant, (for context) I have worked indirectly with several individuals publishing work specifically on LCA (including a dissertation and one of the original algorithm creators) and several individuals working in the sparse signal processing subfield. While working with said individuals I do not recall any of them mentioning any conflicts with the 'LCA' abbreviation.

ararslan:
Good enough for me!

PR #10799, opened 2017-08-18
Tag Flux.jl v0.2.2

attobot:
Repository: [MikeInnes/Flux.jl](https://github.com/MikeInnes/Flux.jl)
Release: [v0.2.2](https://github.com/MikeInnes/Flux.jl/releases/tag/v0.2.2)
Travis: [Travis Build Status](https://travis-ci.org/MikeInnes/Flux.jl/builds/265894544)
Diff: [vs v0.2.1](https://github.com/MikeInnes/Flux.jl/compare/7a85eff370b7c68d587b49699fa3f71e44993397...b17eb78e0a7131de18501b621c594a9c4d459372)
`requires` vs v0.2.1: 
```diff
--- 0.2.1/requires
+++ 0.2.2/requires
@@ -1,5 +1,4 @@
-julia 0.6.0-pre
-Iterators
-DataFlow
+julia 0.6.0
+DataFlow 0.2.1 0.3-
 Juno
 MacroTools 0.3.3
```
cc: @MikeInnes

Please make sure that:


Flux/versions/0.2.2/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.6.0
+DataFlow 0.2.1 0.3-
```
tkelman: note that this upper bound will not accomplish anything unless it's applied to the past Flux versions as well - otherwise when DataFlow 0.3 comes out (assuming you anticipate that will break Flux), Pkg could choose to downgrade Flux to an earlier version that doesn't have the upper bound in order to upgrade DataFlow

MikeInnes: Is there a script for this or something?

tkelman: Some work in progress at https://github.com/JuliaLang/PkgDev.jl/pull/94, but I tend to do `$EDITOR ~/.julia/v0.6/METADATA/Flux/versions/*/requires` and change the bounds manually to make a PR - can put 0.0 as a lower bound, 0.2.1 as an upper bound since it looks like the old versions of Flux don't work with DataFlow 0.2.1?

MikeInnes: yeah, that would be correct

PR #10800, opened 2017-08-18
Tag OnlineStatsBase.jl v0.0.7

attobot:
Repository: [joshday/OnlineStatsBase.jl](https://github.com/joshday/OnlineStatsBase.jl)
Release: [v0.0.7](https://github.com/joshday/OnlineStatsBase.jl/releases/tag/v0.0.7)
Travis: [Travis Build Status](https://travis-ci.org/joshday/OnlineStatsBase.jl/builds/265949222)
Diff: [vs v0.0.6](https://github.com/joshday/OnlineStatsBase.jl/compare/cab12070db57571d194137cc4355db5a10fccc6e...3aa989ae5170b7d7aeeefe2788ace4d27ef5650b)
`requires` vs v0.0.6: no changes
cc: @joshday

Please make sure that:


tkelman:
This looks like it causes a failure downstream in OnlineStats?

joshday:
I do that every time...The next OnlineStats release (in the next couple minutes) will have an upper bound for OnlineStatsBase.

tkelman:
the upper bound will need to be applied to all the past releases, otherwise Pkg could downgrade OnlineStats to a version without a bound in order to upgrade OnlineStatsBase

joshday:
Oh, yikes.  Is it better to leave off the upper bound then?

tkelman:
Either put it on all versions, or don't have it on any and deal with a large possible number of version combinations that don't work together. Another issue is OnlineStatsBase is 0.6-only, but OnlineStats has been around for a while and has versions for julia 0.3, 0.4 and 0.5 (and even a few without minimum julia versions at all) that didn't depend on OnlineStatsBase. So even if all of the versions of OnlineStats that do depend on OnlineStatsBase had an upper bound added, Pkg could downgrade to an earlier version that didn't depend on OnlineStatsBase. If those early versions of OnlineStats don't load at all on julia 0.6, then we could potentially add a julia upper bound to them to avoid the possibility of downgrading.

joshday:
I decided to go without the upper bound.  That sounds like the simpler solution and I think it's pretty unlikely that someone would have one of them pinned.  

The failing tests in OnlineStats is because I removed a type in OnlineStatsBase.  The PR for the OnlineStats release removes the problem tests.

aviks:
@joshday Is this OK to merge now?

joshday:
Yes, good to go

PR #10819, opened 2017-08-20
Tag PowerModelsAnnex v0.0.4 [https://github.com/lanl-ansi/PowerModels…

ccoffrin:
…Annex.jl]

Diff vs v0.0.3: https://github.com/lanl-ansi/PowerModelsAnnex.jl/compare/13fdf1f9c523f0300cbd7190c9f36c83fa0aca75...b236f37aa3c838a65f2dae8b4974ae7714cf9326

Adding a few from-scratch power system models.


ccoffrin:
@tkelman a little help getting this merged in?

PowerModelsAnnex/versions/0.0.4/requires
```diff
@@ -0,0 +1,4 @@
+julia 0.5
+JuMP 0.17 0.18-
```
ararslan: Why such specific version bounds? Does the package not work at all with JuMP versions other than 0.17? Likewise for PowerModels.

ccoffrin: Both PowerModels and JuMP have a version numbering convention where breaking changes occur in the second version number.  Non-breaking releases occur in the third version number.  Hence, it's is not reasonable to assume that a code that works on version 0.x will still work in version 0.x+1.  This is the reason for the tight version bounds.

PR #10817, opened 2017-08-20
Tag DiffEqNoiseProcess.jl v0.5.0

attobot:
Repository: [JuliaDiffEq/DiffEqNoiseProcess.jl](https://github.com/JuliaDiffEq/DiffEqNoiseProcess.jl)
Release: [v0.5.0](https://github.com/JuliaDiffEq/DiffEqNoiseProcess.jl/releases/tag/v0.5.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaDiffEq/DiffEqNoiseProcess.jl/builds/266409332)
Diff: [vs v0.4.0](https://github.com/JuliaDiffEq/DiffEqNoiseProcess.jl/compare/42db3d5cf9a33cf5cd6139fb91bb1b3f0afd54a1...cb9d1e7e6e14ae45a53566464170c4b068c7ad9b)
`requires` vs v0.4.0: 
```diff
--- 0.4.0/requires
+++ 0.5.0/requires
@@ -1,5 +1,5 @@
 julia 0.6
-ResettableStacks
+ResettableStacks 0.2.0
 DataStructures
 DiffEqBase 1.13.0
 RecipesBase
```
cc: @ChrisRackauckas

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #10828, opened 2017-08-21
bound StochasticDiffEq

ChrisRackauckas:
Fixes CIBot failures of https://github.com/JuliaLang/METADATA.jl/pull/10817


ChrisRackauckas:
What exactly is the issue here?

PR #10840, opened 2017-08-22
Add bounds to BedgraphsFiles package

davidanthoff:
Missed this one when I did #10832.

@CiaranOMara Cool package! I'm in the process of moving the trait definition from IterableTables to TableTraits to get rid of some annoying Requires.jl dependency. This here will prevent things from breaking, and I'll open PRs against BedgraphFiles.jl to make this a smooth transition.


CiaranOMara:
@davidanthoff, that'd be grand. I like your work.

PR #10860, opened 2017-08-23
Add bounds to a LazyQuery dependency

davidanthoff:
Current LazyQuery.jl versions don't work with Query.jl v0.7.0, so this adds the appropriate bounds.

CC @bramtayl

### DIFF
--- a/LazyQuery/versions/0.0.1/requires
+++ b/LazyQuery/versions/0.0.1/requires
@@ -2,3 +2,4 @@ julia 0.5
 DataFrames
 MacroTools
 LazyContext 0.0.1 0.1.0
+Query 0.0.0 0.7.0
--- a/LazyQuery/versions/0.1.0/requires
+++ b/LazyQuery/versions/0.1.0/requires
@@ -2,6 +2,6 @@ julia 0.6
 LazyContext 0.1.1
 DataFrames
 MacroTools 0.3.3
-Query
+Query 0.0.1 0.7.0
 NamedTuples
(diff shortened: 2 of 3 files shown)


tkelman:
This will likely downgrade LazyQuery to 0.0.1 in order to upgrade Query.

davidanthoff:
I believe @gramtayl is working on a LazyQuery.jl version that works with Query.jl v0.7.0, so I assume this is just going to be a temporary problem. Still better to put this bound in, right?

tkelman:
Since there are versions of Query available for all the julia versions supported by LazyQuery 0.0.1, safest to add an artificial `Query 0.0 0.7` to it

bramtayl:
I'm having a world age problem. The anonymous functions I feed into Query are based on eval, which worked in the previous version but not this one. @davidanthoff did Query change such that it is working in the current world age? If so I'll have to go through and add a whole bunch of evals.

davidanthoff:
> safest to add an artificial Query 0.0 0.7 to it

Done! That seems the right thing to do, thanks for pointing it out.

> I'm having a world age problem. The anonymous functions I feed into Query are based on eval, which used to work. @davidanthoff did Query change such that it stopped supporting this functionality? If so I'll have to go through and add a whole bunch of evals.

Of the top of my head I can't think of any change that would trigger this... In my mind the v0.7.0 update was breaking because I moved some functions from IterableTables.jl to TableTraits.jl, but everything else should be pretty harmless...

PR #10882, opened 2017-08-24
Upper limit on NLSolversBase in packages that depend on it.

pkofod:


tkelman:
This is going to be problematic, since not all versions of these packages depend on NLSolversBase. Pkg can choose to downgrade them to an old version that doesn't depend on NLSolversBase in order to upgrade NLSolversBase to the latest version. You probably don't want it to do this.

pkofod:
Hm, okay. Then how can I do this? The failing tests in the other or are not important, so we could just let it fail and merge the tag anyway, and then fix in a patch? I don't know what you'd be comfortable doing here.

aviks:
I'll close this (so as not clutter the queue). Please reopen when ready.

pkofod:
Couldn't you just cancel the job? :) Anyway, I'll reopen when I find the solution

pkofod:
@tkelman I'm not really sure how to fix this. Could we use the fact that the two packages have a `pre` tag on v0.6? I mean, would it work if I only did it for the latest version of the two packages and simultaneously changed the lower bound to v0.6 proper on the latest tags, and adding the `v0.6-pre` tag to all previous versions?

tkelman:
There are a few possible solutions. Either make old versions depend on NLSolversBase even though they don't technically need to, or add julia upper bounds to those old package versions (preferably only if they can't actually load on recent julia), or some combination. I should be able to take a closer look and suggest some things within the next few days

pkofod:
> Either make old versions depend on NLSolversBase even though they don't technically need to, or add julia upper bounds to those old package versions (preferably only if they can't actually load on recent julia), or some combination. I should be able to take a closer look and suggest some things within the next few days

The Julia version bounds should work. Only the most recent tags actually work on v0.6 I believe.

anriseth:
Hi @tkelman, I'm gonna try to give this versioning issue a shot. Let me know if the following will work:
Add `require NLSolversBase 1.0 3.0` to all versions of LineSearches, NLSolve and Optim, if:
- They require Julia v0.5
- They don't already have a NLSolversBase dependency

PS: All NLSolversBase versions require v0.5, and the first version was NLSolversBase v1.0.0

tkelman:
That should get you part of the way there. Easiest to start from a piece of it and carefully go from there until it's a complete solution.

PR #10880, opened 2017-08-24
Tag WeakRefStrings.jl v0.3.0

attobot:
Repository: [quinnj/WeakRefStrings.jl](https://github.com/quinnj/WeakRefStrings.jl)
Release: [v0.3.0](https://github.com/quinnj/WeakRefStrings.jl/releases/tag/v0.3.0)
Travis: [Travis Build Status](https://travis-ci.org/quinnj/WeakRefStrings.jl/builds/267842913)
Diff: [vs v0.2.0](https://github.com/quinnj/WeakRefStrings.jl/compare/34d5207e0c3e0c4b133ccc8d226a9e509e530df1...152cb5e6698fb183429e9d9cc57c37917b208e5a)
`requires` vs v0.2.0: 
```diff
--- 0.2.0/requires
+++ 0.3.0/requires
@@ -1 +1,2 @@
-julia 0.5-
+julia 0.6
+Nulls
```
cc: @quinnj

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #10915, opened 2017-08-25
DataStreams v0.2.0

quinnj:


nalimilan:
Don't DataFrames and DataTables also need an upper bound?

quinnj:
Good catch! DataFrames has actually never had a dependency.

nalimilan:
Thanks. Can you make a PR to add the bounds to DataTables's REQUIRES? Looks like we also need an upper bound for WeakRefStrings (cf. https://github.com/JuliaLang/METADATA.jl/pull/10880#issuecomment-324927946).

tkelman:
See https://github.com/JuliaLang/METADATA.jl/pull/10882#issuecomment-324638656, same problem here.

nalimilan:
Yes, that's an issue. Maybe we should just add a fake dependency on DataStreams to old releases to ensure no downgrading happens automatically.

quinnj:
Yeah, I went thru all the packages and I don't think we'll run into any serious problems. I added a `julia 0.4 0.5` upperbounds to a past ODBC version which didnt' have any bounds on packages anyway, and added a fake dependency for past DataTables, but I'm not too worried about DataTables since it's been an advertised experiment and hopefully will be archived soon.

davidanthoff:
Is there some documentation somewhere what changed in this release and what packages that depend on it need to update? The whole iterable tables integration will have to follow this update. It doesn't show up in terms in REQUIRE files because the integration is behind a Requires.jl ``@require`` clause, but this will presumably break the existing integration.

quinnj:
This is a preliminary release so that final testing/travis runs can be made on downstream packages, hence the upper bounds everywhere (meaning, no one should be using this release w/ this tag). Once downstream packages have been thoroughly tested/upgraded, I'll tag another DataStreams release w/ any final updates + docs and then can tag downstream packages. 

Changes for this release are pretty well captured in this PR: https://github.com/JuliaData/DataStreams.jl/pull/36/. Does Requires.jl allow you to require specific package versions? It should just be a simple matter of putting an upper bounds on DataStreams for now. I'll try to notify everyone once the real public release is ready for downstream upgrades.

nalimilan:
I'm not too worried about DataTables either, we should just ensure that it keeps working as it does right now. A fake dependency on older DataStreams releases (or on Julia itself) is enough IMHO.

davidanthoff:
> Does Requires.jl allow you to require specific package versions?

No, once this is tagged in METADATA.jl, I'll have to deal with it in the iterable tables universe, there is no way to hold things back with upper bounds in when Requires.jl is used. Having said that, in practice it might be less of a problem because the only way that folks are getting DataStreams.jl is probably via one of the dependent packages, and those should have the bounds in place and that should do the trick for now, I guess.

tkelman:
you only changed one past odbc version, that won't be good enough to avoid downgrading to an even older version

quinnj:
Good catch, there were a bunch of `julia 0.2` old requirements; that would have been bad :)

tkelman:
and with early CSV versions that didn't depend on WeakRefStrings

quinnj:
kk, just pushed another one; hopefully this is the last whack-a-mole.

quinnj:
BioMedQuery failures look unrelated; probably got in that state before we did more rigorous testing.

nalimilan:
There a

nalimilan:
There are suspicious errors related to BinDeps, but this one sounds related, isn't it?
```
     BioMedQuery v0.0.1 – no valid versions exist for package WeakRefStrings
```

CSV/versions/0.0.11/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5
```
tkelman: the julia upper bounds shouldn't be needed on CSV versions that have existing DataStreams and WeakRefStrings dependencies?

quinnj: I think they're bounded by the DataStreams upper bounds to 0.5 anyway, so I figured might as well put it in CSV as well.

tkelman: doesn't look like it? https://github.com/JuliaLang/METADATA.jl/blob/2688df9c8e59be6a5fdf166e61a7981fe6df61d3/DataStreams/versions/0.0.1/requires

DataTables/versions/0.0.2/requires
```diff
@@ -7,3 +7,4 @@ SortingAlgorithms
 Reexport
 Compat 0.19.0
 FileIO 0.1.2
+DataStreams 0.1.0 0.2.0
```
tkelman: if these are fake, why not allow earlier?

Feather/versions/0.1.5/requires
```diff
@@ -3,5 +3,5 @@ FlatBuffers 0.1.1 0.2.0
 NullableArrays 0.0.8
 CategoricalArrays 0.0.4 0.0.5
 DataFrames
-DataStreams 0.0.13
-WeakRefStrings 0.1.3
+DataStreams 0.0.13 0.2.0
+WeakRefStrings 0.1.3 0.3.0
```
tkelman: do the earlier Feather versions also need WeakRefStrings upper bounds?

SQLite/versions/0.3.7/requires
```diff
@@ -2,7 +2,7 @@ julia 0.4.1
 BinDeps
 @osx Homebrew
 @windows WinRPM
-DataStreams 0.0.13
+DataStreams 0.0.13 0.2.0
 DataFrames
-WeakRefStrings 0.1.3
+WeakRefStrings 0.1.3 0.3.0
```
tkelman: earlier SQLite versions will need WeakRefStrings upper bounds as well, right?

quinnj:
Ok, replaced the mess here w/ just the original DataStreams tag.

PR #10921, opened 2017-08-25
Tag Fatou v0.1.2

chakravala:
This patch for `Fatou` solves the 0.5 to 0.6 world counter compatibility issue ~

https://github.com/chakravala/Fatou.jl/commit/caf0537be36e21d90009f39dfa15e1db00215484


tkelman:
Old tags should not be deleted.

chakravala:
Alright, got it. 

Yea, I had deleted 0.0.1 and 0.0.2 a while back ago from the repo since the 0.0.3 release is more official and that was pre-release, so they didn't exist anymore and didn't point to anything anymore, but I won't do that again then.

chakravala:
will this be able to merge though?

not sure why the JuliaCIBot tests failed but the package tests fine according to the tests linked on the badge [Build Status](https://travis-ci.org/chakravala/Fatou.jl) [Build status](https://ci.appveyor.com/project/chakravala/fatou-jl)

tkelman:
Please add the git tags back for all tags that are registered in metadata.

Fatou/versions/0.1.1/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5 0.6-
```
tkelman: does the package load at all? this makes these versions not installable even on the very first prerelease after release-0.5 branched. if there's a newer working version available, the upper bounds aren't strictly necessary and aren't solving a specific problem here

chakravala: Julia 0.5 should work, but julia 0.6 does not work before this patch.

tkelman: upper bounds are exclusive though, and `0.6-` means the very first prerelease commit of 0.6 (which code-wise is essentially identical to 0.5)

chakravala: these require version bounds are confusing me to be honest. so the package is not functional on julia 0.6 before the 0.1.2 Fatou patch, but it does work for julia 0.5 and 0.5.2. How can I best update these `requires` files then? as I understand it `0.6-` means up to (but not including) 0.6

chakravala: so the `-` just needs to be removed?

chakravala:
alright, I think it should be all set now

chakravala:
are we all set now?

tkelman:
Looks like the commit sha listed doesn't match the git-tagged commit in the repo? There should be a test that those match but for some reason that isn't flagging the mismatch here.

Do any of the previous versions even partially work at all on Julia 0.6? If there's some very limited functionality that can still work, it might not be worth disallowing installation of those versions with the upper bounds. A newer version being available should be sufficient for most people to upgrade to.

chakravala:
Ah yes, I overwrote a commit to fix a spelling mistake. I updated the sha in this pull request but did not update the actual tag in the repo. Now that should be fixed, I think

None of the previous versions worked with Julia 0.6 because of the World Counter issue, so only the newest tag can work with 0.6

chakravala:
It seems the checks are passing now, I have fixed the sha

PR #10933, opened 2017-08-26
Register new package InferenceUtilities.jl

afniedermayer:
See discussion at https://discourse.julialang.org/t/more-informative-inferred-especially-for-unit-testing/5481/12


tkelman:
You'll need to tag a version here. You may find it easiest to use https://github.com/attobot/attobot

MikeInnes:
Closing as it has been a while without an update. @afniedermayer the package is of course still welcome if you can tag a version.

PR #10943, opened 2017-08-27
Correct changes to DynamicalSystems.jl

Datseris:
1. Changed the REQUIRE to be correct
2. Change the git tag at the sha file to be the same as the release git tag


DynamicalSystems/versions/0.1.0/sha1
```diff
@@ -1 +1 @@
-839f85ab012297aa241bf7f4d58becfb110a6fd8
+36fb8da89417c7d31c5a3ead5289e472ebed4242
```
tkelman: this part should be left alone, I meant the other way, put the git tag back to the merged sha

Datseris:
i removed the sha change...

Datseris:
Hi, is this fine now? If not please let me know what more is there to be done.

(I would like to tag a new version soon and this should happen *after* this PR has been resolved)

tkelman:
The git tag at https://github.com/Datseris/DynamicalSystems.jl/releases/tag/v0.1.0 still doesn't match the originally tagged sha value at https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/DynamicalSystems/versions/0.1.0/sha1 - that should be causing a travis failure

Datseris:
Hm, please bear with me because I really don't understand.

I thought that this was the change I did with the previous commit, where you commented that this part should be left alone. Where I changed the contents of versions/0.1.0/sha1 with the git tag I find at my  package, at https://github.com/Datseris/DynamicalSystems.jl/releases/tag/v0.1.0 ...?

Can you explain to me what I should do, because I am confused? Sorry for not getting it :D
Is there a way I can change a git tag at the tagged version, or...? Because when I got to my tagged release and press "edit tag" it lets me change the name but I don't see any option for something called "sha".

tkelman:
There should be an option to delete the git tag as well (if you can't find it in the web ui, we can look up the command-line invocation for it), then you could replace it with a "new" v0.1.0 at the originally tagged commit sha of 839f85ab012297aa241bf7f4d58becfb110a6fd8

Datseris:
I could not find a way to do this. The only way I could find was to completely delete the tag and then create a new one with the same name. In fact, this [stack question](https://stackoverflow.com/questions/8044583/how-can-i-move-a-tag-on-a-git-branch-to-a-different-commit) also suggests the same thing, but via git instead of github.

There are 3 things that I have to say/ask:

1. The original sha for the tag corresponds to a point in the package where the require file was NOT correct. This was in fact the reason I made a new tag, trying to correct the old, but the old was already merged in metadata.
2. Why can't we simply change the sha that is written in the METADATA file, like I tried to do here? I am asking so I can learn.
3. It is not possible to remove the tag from metadata and replace it with a new one completely? Like If I now create a version v0.2.0 and make this as the first release in METADATA?

tkelman:
The METADATA requires file is allowed to be different than the package REQUIRE at the tagged commit. For tagged versions of registered packages, the METADATA copy takes priority.

Datseris:
Okay! I think I did it! 
I have made a tag and a new release at the github page at the commit that you mentioned! I am super happy that everything went smoothly! Thankfully GitKraken made everything very easy!

This PR has the correct changes to the require, so everything will be ok when its merged!

tkelman:
I've loosened the lower bound versions slightly for consistency with master of the package. Agree it should be good now if Travis passes.

re: your second and third questions, the meaning in terms of code content for a given version number of a Julia package shouldn't change over time. That's why the sha shouldn't change once registered, and existing tags shouldn't be removed. But requirement bounds of past versions can be modified in METADATA, in case they are found to be incorrect.

PR #10969, opened 2017-08-28
Add appropriate upper bounds for DataStreams and WeakRefStrings packages

quinnj:
Note: no new tags are introduced in this PR, only appropriate DataStreams & WeakRefStrings upper bounds.

### DIFF
--- a/CSV/versions/0.0.12/requires
+++ b/CSV/versions/0.0.12/requires
@@ -1,4 +1,4 @@
-julia 0.4.1 0.6
+julia 0.4.1
 DataStreams 0.0.10 0.0.13
 DataFrames
-WeakRefStrings 0.1.3
+WeakRefStrings 0.1.3 0.3.0
--- a/CSV/versions/0.0.12/requires
+++ b/CSV/versions/0.0.12/requires
@@ -1,4 +1,4 @@
-julia 0.4.1 0.5
+julia 0.4.1 0.6
 DataStreams 0.0.10 0.0.13
 DataFrames
 WeakRefStrings 0.1.3
--- a/CSV/versions/0.0.12/requires
+++ b/CSV/versions/0.0.12/requires
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5
 DataStreams 0.0.10 0.0.13
 DataFrames
 WeakRefStrings 0.1.3
--- a/CSV/versions/0.0.5/requires
+++ b/CSV/versions/0.0.5/requires
@@ -1,4 +1,4 @@
 julia 0.4.1 0.5
 Compat
 NullableArrays
-DataStreams 0.0.1 0.0.13
+DataStreams 0.0.1 0.0.5
--- a/CSV/versions/0.0.5/requires
+++ b/CSV/versions/0.0.5/requires
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5
 Compat
 NullableArrays
-DataStreams 0.0.1 0.0.5
+DataStreams 0.0.1 0.0.13
--- a/CSV/versions/0.1.0/requires
+++ b/CSV/versions/0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.5 0.7
+julia 0.5
 DataStreams 0.1.0 0.2.0
 DataFrames
 WeakRefStrings 0.1.3 0.3.0
--- a/CSV/versions/0.1.0/requires
+++ b/CSV/versions/0.1.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
-DataStreams 0.1.0
+julia 0.5 0.7
+DataStreams 0.1.0 0.2.0
 DataFrames
-WeakRefStrings 0.1.3
+WeakRefStrings 0.1.3 0.3.0
--- a/DataStreams/versions/0.0.1/requires
+++ b/DataStreams/versions/0.0.1/requires
@@ -1,4 +1,4 @@
-julia 0.4 0.5
+julia 0.4 0.6
 NullableArrays
 CategoricalArrays 0.0.1 0.0.5
 WeakRefStrings 0.1.2 0.3.0
--- a/DataStreams/versions/0.0.1/requires
+++ b/DataStreams/versions/0.0.1/requires
@@ -1,4 +1,3 @@
 julia 0.4 0.6
 NullableArrays
 CategoricalArrays 0.0.1 0.0.5
-WeakRefStrings 0.1.2 0.3.0
--- a/DataStreams/versions/0.0.1/requires
+++ b/DataStreams/versions/0.0.1/requires
@@ -1,3 +1,4 @@
-julia 0.4
+julia 0.4 0.5
 NullableArrays
 CategoricalArrays 0.0.1 0.0.5
+WeakRefStrings 0.1.2 0.3.0
--- a/DataStreamsIntegrationTests/versions/0.0.1/requires
+++ b/DataStreamsIntegrationTests/versions/0.0.1/requires
@@ -1,3 +1,3 @@
-julia 0.5
-DataStreams
+julia 0.5 0.6
+DataStreams 0.1.0 0.2.0
 NullableArrays
--- a/DataStreamsIntegrationTests/versions/0.0.1/requires
+++ b/DataStreamsIntegrationTests/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.5 0.7
-DataStreams 0.1.0 0.2.0
+DataStreams 0.0.0 0.2.0
 NullableArrays
--- a/DataStreamsIntegrationTests/versions/0.0.1/requires
+++ b/DataStreamsIntegrationTests/versions/0.0.1/requires
@@ -1,3 +1,3 @@
-julia 0.5 0.6
+julia 0.5 0.7
 DataStreams 0.1.0 0.2.0
 NullableArrays
--- a/DataTables/versions/0.0.1/requires
+++ b/DataTables/versions/0.0.1/requires
@@ -8,3 +8,4 @@ Reexport
 Compat 0.19.0
 FileIO 0.1.2
 Juno 0.2.4
+DataStreams 0.1.0 0.2.0
--- a/DataTables/versions/0.0.1/requires
+++ b/DataTables/versions/0.0.1/requires
@@ -8,4 +8,4 @@ Reexport
 Compat 0.19.0
 FileIO 0.1.2
 Juno 0.2.4
-DataStreams 0.1.0 0.2.0
+DataStreams 0.0.0 0.2.0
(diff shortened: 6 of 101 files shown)


tkelman:
please try to use your fork going forward for metadata PR's to avoid doubling the travis builds, and creating temporary branches that everyone who does Pkg.update has to fetch

quinnj:
Ah, noted.

CSV/versions/0.1.0/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.5
-DataStreams 0.1.0
+julia 0.5 0.7
```
tkelman: why the julia upper bound here?

quinnj: Pre-emptive upper bounds. I'll actually change this to `julia 0.5 0.6` once I have the next tag of CSV ready, which will be `julia 0.6`. It's helping me segment the releases when I think about them per julia release; i.e. for each julia release, what's the latest tag that is compatible.

CSV/versions/0.0.12/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5
```
tkelman: this, 0.0.13, and 0.0.14 all work fine on julia 0.5

CSV/versions/0.1.0/requires
tkelman: I don't think it's a good idea to rule out installation on a given Julia version if it works correctly.

quinnj: Well, it's the situation we live in, this is a way I can keep my sanity and actually understand what's happening with all these versions.

DataStreams/versions/0.0.1/requires
```diff
@@ -1,3 +1,4 @@
-julia 0.4
+julia 0.4 0.5
 NullableArrays
 CategoricalArrays 0.0.1 0.0.5
+WeakRefStrings 0.1.2 0.3.0
```
tkelman: if this is fake, shouldn't have a lower bound

DataStreams/versions/0.0.1/requires
```diff
@@ -1,3 +1,4 @@
-julia 0.4
+julia 0.4 0.5
```
tkelman: this version passes on 0.5, shouldn't need a julia upper bound if WeakRefStrings is bounded

DataStreams/versions/0.0.1/requires
quinnj: It's the lowest "reasonable" bound, there were some serious bug fixes and this is the lowest I'd want anyone to get onto.

DataStreams/versions/0.0.1/requires
quinnj: I found it easier to reason about putting upper bounds on julia itself. Sure it limits a few packages that *_may_* have been forward compatible, but they weren't necessarily meant to be. If someone is on 0.5, I'd rather they use the minimum version that I planned on vs. an earlier version that happened to work.

DataStreamsIntegrationTests/versions/0.0.1/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.5
-DataStreams
+julia 0.5 0.6
```
tkelman: this should be 0.7 at lowest, otherwise anyone who has this package currently installed will get an error

DataTables/versions/0.0.1/requires
```diff
@@ -8,3 +8,4 @@ Reexport
 Compat 0.19.0
 FileIO 0.1.2
 Juno 0.2.4
+DataStreams 0.1.0 0.2.0
```
tkelman: shouldn't need a nonzero lower bound if this is a fake dependency

DataStreamsIntegrationTests/versions/0.0.1/requires
quinnj: No one should really have this package installed and if it manages to load, it's certainly not expected to work. This is a good upper bound; I'll be archiving this package soon anyway.

DataStreams/versions/0.0.1/requires
tkelman: but this version of DataStreams didn't actually depend on WeakRefStrings, so it shouldn't matter except for the upper bound

DataStreamsIntegrationTests/versions/0.0.1/requires
tkelman: Pkg.update() will cause confusing errors. I'm happy to help you with any resolver issues, but I really don't want to be causing more restrictive resolution than necessary to avoid real problems.

DataStreams/versions/0.0.1/requires
tkelman: The vast majority of people on 0.5 will be using the latest version that's allowed. Disallowing version combinations that do work correctly makes Pkg go out of its way to cause errors or packages getting installed at unintuitive versions - it causes more problems than it solves. Version bounds should be as loose as would work correctly, any tighter is harmful.

CSV/versions/0.1.0/requires
tkelman: > what's the latest tag that is compatible.

That's quite easy to determine, and doesn't require adding upper bounds.

Feather/versions/0.1.4/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.5
```
tkelman: this passes fine on julia 0.5, please don't add the julia upper bound

quinnj:
K, pushed some updates based on your recommendations, but honestly, this process is super wearing me out. Is there seriously not a better way to handle this situation? Can we not leverage all the package version code to compute necessary upper bounds? For some reason this is really boiling my blood. Sorry if I'm coming off hot at times.

quinnj:
Hey look at that, finally got it to pass. any other tweaks needed, @tkelman ?

CSV/versions/0.0.5/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4.1
+julia 0.4.1 0.5
 Compat
 NullableArrays
-DataStreams 0.0.1 0.0.5
+DataStreams 0.0.1 0.0.13
```
tkelman: had this been too tight somehow?

quinnj: Yes, both versions before & after allowed `0.0.1 0.0.13`, so I think it was just a missed update at some point.

Metamath/versions/0.0.1/requires
```diff
@@ -1,3 +1,3 @@
-julia 0.4.1
+julia 0.4.1 0.5
```
tkelman: this could cause issues if anyone has the package installed on 0.5 or newer, it effectively deprecates the package cc @getzdan 

it had been passing on 0.5 until june 20 2016 - https://pkg.julialang.org/detail/Metamath.html

quinnj: There hasn't been any non-@tkelman activity in over a year, plus it's easy enough to clean a few things up and do a new release with new REQUIRES if we really want to be careful.

getzdan: Hey, guys, I'll clean this up RealSoonNow. Not much depends on this. But, true, it should be cleaned for 0.6 and beyond.

quinnj: I went ahead and did a PR: https://github.com/getzdan/Metamath.jl/pull/3. Turns out you don't even need CSV for what you're doing. I updated the REQUIRES to be `julia 0.5`, so that should allow the tag update I have here to be correct. Looks like linux travis & appveyor are happy w/ that PR if you'd like to merge (travis OSX has been pretty flaky/delayed lately) and then we can do a new tag for Metamath.

getzdan: Thanks @quinnj . Refreshing the package was in my TODO list. Trouble is, things get inserted on top and bottom of the list all the time. Mergin the PR, but tending the package is still on the list ;)

quinnj: I'd recommend enabling https://github.com/integration/attobot; with that enabled, you just go to https://github.com/getzdan/Metamath.jl/releases and create a new release and attobot will create the tag and METADATA PR for you.

quinnj: @getzdan, will you be able to do a new tag soon? That would help facilitate merging all these version revisions. If you go thru the steps I mentioned in my previous comment, it should take all of 2 minutes or so :)

quinnj: Thanks!

tkelman:
> Can we not leverage all the package version code to compute necessary upper bounds?

Not sure what you mean. Sorry this is fiddly and manual at the moment, especially when your reverse-dependencies have had complicated histories and didn't always depend on the same set of packages over time.

CSV/versions/0.0.5/requires
tkelman: There was something that had been removed from DataStreams at some point, which maybe only this version of CSV ever used?

quinnj:
Ok, Metamath is tagged now, good to go here?

tkelman:
Going over it one last time, hopefully will be but let me double-check.

CSV/versions/0.0.5/requires
tkelman: PointerString - this is the only version here that has the correct upper bound, the earlier bounds were too loose

Feather/versions/0.2.5/requires
```diff
@@ -3,6 +3,6 @@ FlatBuffers 0.2.0
 NullableArrays 0.0.8
 CategoricalArrays 0.0.6
 DataFrames
-DataStreams 0.1.0
+DataStreams 0.1.0 0.2.0
 WeakRefStrings 0.1.3
```
tkelman: judging by https://pkg.julialang.org/detail/Feather.html it looks like these will need upper bounds too, but I can go ahead and add them

nalimilan: Please do, anything which can help getting this PR merged is welcome! ;-)

nalimilan:
Looks like we're good to go?

tkelman:
Not quite, still verifying ODBC and SQLite

tkelman:
I'm fairly satisfied with this version, and I think it should pass the verification tests. I generally want to avoid upper-bounding Julia on anything unless it's absolutely known not to work at all. The problem that this is intended to address is that the new DataStreams and WeakRefStrings versions are breaking, so any past package that depended on them should upper-bound them. It's only when the package version is so old that it didn't depend on either that it makes more sense to bound julia. If someone had a package version pinned on a julia version that would have been ruled out with a tight upper bound here, they'd suddenly start getting errors on Pkg.update that aren't needed to solve the real problem here.

quinnj:
all green.....

tkelman:
If you're okay with the commits I added to this, we can squash merge.

quinnj:
Yep, I'm hugely appreciative of your help on this; I know it's not fun for anyone.

PR #10990, opened 2017-08-31
Register and Tag PySyntax [https://github.com/kdheepak/PySyntax.jl.git]

kdheepak:


Py/versions/0.0.1/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.6
+PyCall 1.14.0
```
tkelman: is this using a feature of 1.14 that doesn't work with any earlier versions?

tkelman:
Should fill in the package description to something other than "No description, website, or topics provided," and enable Travis CI.

A more descriptive name might be a good idea as well.

Py/versions/0.0.1/requires
kdheepak: No, I don't believe so. It should be backwards compatible. I've made the change in the repo to remove the require for the specific PyCall version. I've also added julia 0.5 and 0.6 as compatible versions.

kdheepak:
I've updated the package description. Do you think `PySyntax.jl` be a more descriptive name?

kdheepak:
I changed the package name to `PySyntax.jl` and added a description to the repo.

MikeInnes:
Thanks for the new package! You might consider announcing it on [Julia's forum](https://discourse.julialang.org/).

kdheepak:
Thanks for the suggestion! I posted it on the forum, this will be a great way to get feedback.

PR #11001, opened 2017-08-31
Register MultivariateAnomalies [https://github.com/milanflach/Multiva…

milanflach:
…riateAnomalies.jl.git]


tkelman:
You'll need to tag a version here. May want to look into https://github.com/attobot/attobot which makes the process a bit easier.

PR #11006, opened 2017-08-31
Tag RationalFunctions v0.1.1

aytekinar:
Tagging [RationalFunctions](https://github.com/aytekinar/RationalFunctions.jl).

Diff vs v0.1.0: https://github.com/aytekinar/RationalFunctions.jl/compare/68fd334ef44ed2a532ce01c6f266c04cde8b2666...b16ec1a072f359c81af38b1a1a981170324f19e3


aytekinar:
I don't know why JuliaCIBot is not passing, but you can check my [Travis](https://travis-ci.org/aytekinar/RationalFunctions.jl/builds/270364795) and [Appveyor](https://ci.appveyor.com/project/aytekinar/rationalfunctions-jl/build/1.0.36) logs, if you want.

Failures are not my package's problem, unfortunately :(

tkelman:
Looks like https://github.com/jheinen/GR.jl/issues/35

aytekinar:
What am I supposed to do? I mean, the same package **passes** Unix and OSX for all versions, but **fails** Win x86 *only* on `v0.6` (and the build stops due to fast fail option).

My package does not depend on `GR`, but in my tests I am requiring it to have the fancy plot examples uploaded to my [releases](https://github.com/aytekinar/RationalFunctions.jl/releases).

Do you know of any (light) backend for `Plots` that I can use and `savefig` for some `.png` output that can also work on all platforms? Or should I simply remove the plotting thing?

tkelman:
The win32 failure on 0.6 may be an underlying Julia bug, so might be worth looking into that in more detail.

aytekinar:
@aviks has already merged the PR, but do you want something from me, @tkelman, in terms of the above-mentioned changes? Shall I keep it as is, and simply wait for `GR` to switch to a proper configuration to support headless VMs. Or do you think I should abandon the code for plotting altogether?

tkelman:
Up to you. Would make your tests a bit easier to run a wider range of places if it didn't depend on plotting, or if the plotting were separated a bit. I think there may be backends that do work more places, but I don't use Plots all that much so could be wrong.

aytekinar:
In a way, I define plotting recipes in my source, and through `Plots`, I can have more coverage. Otherwise I only miss `print`ing and `plot`ing functionality in my tests.

Anyways, thank you for your time and suggestions. I will try to find a solution myself. Or else, I will generate the plot in Travis only on some systems/architectures.

PR #11014, opened 2017-08-31
Register LTISystems

aytekinar:
Registering [LTISystems][js-ltisystems]: `v0.1.0`.

[js-ltisystems]: https://github.com/JuliaSystems/LTISystems.jl.git


davidanthoff:
The name seems too generic for METADATA.jl.

aytekinar:
@davidanthoff any other suggestions?

It is, for now, a collection of different representations of linear, time-invariant system models (LTI system models) to be used for control analysis and design as well as system identification.

Previously, it was [`ControlCore`](https://github.com/KTH-AC/ControlCore.jl), but we thought of such a name as [`IdentificationToolbox`](https://github.com/JuliaSystems/IdentificationToolbox.jl) will also be using it.

davidanthoff:
I don't have a better name, but "Systems" means all sorts of things in all sorts of areas. My first association when I saw it was "Oh, this must be something about system level stuff in julia".

Even just adding the term "Models" in some form would clear things up, I think.

aytekinar:
Since we also have frequency- and time-responses of the corresponding system models, we thought it would be a good name. However, now, I see your point, even though I still think that any person searching packages through https://pkg.julialang.org would read the description "Core control systems functionality for analysis, design and identification tools."

davidanthoff:
Would probably be good if other folks also chimed in, I'm not sure what the official policy on this kind of stuff is right now, I really was just reporting my first reaction to the name :)

aytekinar:
Thank you for your feedback. As I said, we hadn't thought about it --- much appreciated. I still hope that this would be accepted 😄

aviks:
Yes, _Systems_ is a very overloaded name. I feel this would confuse everyone who sees it when scrolling through the list, and as we all know, no one reads extended documentation :)

aytekinar:
We have changed the package's name to be more specific. I hope this helps now.

Cc: @davidanthoff, @aviks and @tkelman

tkelman:
DiffEqBase and Polynomials are missing from REQUIRE here - if they're directly imported, they should be listed as direct dependencies

tkelman:
Anything based on SLICOT in here needs to maintain license attribution, unless they're all clean-room reimplementations based on non-copyrighted material.

aytekinar:
> DiffEqBase and Polynomials are missing from REQUIRE here - if they're directly imported, they should be listed as direct dependencies

Do you mean `using` when you say `... if they're directly imported, ...`, @tkelman? I couldn't see any `import` statements related to those packages. If so, we will make sure `DiffEqBase` and `Polynomials` are in `REQURE`. Thanks!

> Anything based on SLICOT in here needs to maintain license attribution, unless they're all clean-room reimplementations based on non-copyrighted material.

In this one, we might need some more help/elaboration. We have used in two places, *i.e.*, `luhessfact!` and `hesssolve!`, *translations* of SLICOT routines `MB02SZ` and `MB02RZ`. The version we have used contains GPL2 license. Citing the corresponding section of the license, we have:

> 2. You may modify your copy or copies of the Program or any portion of it, thus forming a work based on the Program, and copy and distribute such modifications or work under the terms of Section 1 above, provided that you also meet all of these conditions:
> a) You must cause the modified files to carry prominent notices stating that you changed the files and the date of any change.
> b) You must cause any work that you distribute or publish, that in whole or in part contains or is derived from the Program or any part thereof, to be licensed as a whole at no charge to all third parties under the terms of this License.
> c) If the modified program normally reads commands interactively when run, you must cause it, when started running for such interactive use in the most ordinary way, to print or display an announcement including an appropriate copyright notice and a notice that there is no warranty (or else, saying that you provide a warranty) and that users may redistribute the program under these conditions, and telling the user how to view a copy of this License. (Exception: if the Program itself is interactive but does not normally print such an announcement, your work based on the Program is not required to print an announcement.)
> These requirements apply to the modified work as a whole. If identifiable sections of that work are not derived from the Program, and can be reasonably considered independent and separate works in themselves, then this License, and its terms, do not apply to those sections when you distribute them as separate works. But when you distribute the same sections as part of a whole which is a work based on the Program, the distribution of the whole must be on the terms of this License, whose permissions for other licensees extend to the entire whole, and thus to each and every part regardless of who wrote it.

**2.a)** above is what you want us to do, @tkelman, and I think that putting a disclaimer in `?luhessfact!` and `?hesssolve!` which states that we have translated the portions to Julia (with a date) should suffice. **2.c)** does **not** apply to SLICOT, as it does not print any announcements. However, for **2.b)** above, we might need to ask you about the licensing. You already [told](https://github.com/JuliaLang/METADATA.jl/pull/6454#issuecomment-249127726) us what needs to be done when something is under GPL, but in this case I am not sure what *we* should do. I mean, **2.b)** clearly states that we need to switch to GPL2 (or, GPL3, which is the newer version), which is not preferrable in Julia community. Then,

> These requirements apply to the modified work as a whole. If identifiable sections of that work are not derived from the Program, and can be reasonably considered independent and separate works in themselves, then this License, and its terms, do not apply to those sections when you distribute them as separate works.

should make it fine for us to keep GPL2 for the corresponding two function names. However, I am not sure what

> But when you distribute the same sections as part of a whole which is a work based on the Program, the distribution of the whole must be on the terms of this License, whose permissions for other licensees extend to the entire whole, and thus to each and every part regardless of who wrote it.

translates to. Any opinion?

Cc: @CristianRojas, @neveritt.

tkelman:
Violations of copyright law are against the Julia community standards - https://julialang.org/community/standards/

Respect the copyright of SLICOT. That requires distributing derived works under GPL. Or removing all code derived from SLICOT and doing a clean room re-implementation by someone who has not looked at the SLICOT code.

tkelman:
> Do you mean using when you say ... if they're directly imported, ...

yes, either import or using. Note that the merged requirements are incorrect and should now be amended in place, otherwise Pkg will see this initial version as if it did not depend on Polynomials or DiffEqBase and be allowed to downgrade to it if it so chooses

PR #11015, opened 2017-08-31
Deprecate ControlCore

aytekinar:
Deprecated [`ControlCore`](https://github.com/KTH-AC/ControlCore.jl) in favour of [`SystemsBase`](https://github.com/JuliaSystems/SystemsBase.jl).

### DIFF
--- a/ControlCore/versions/0.0.1/requires
+++ b/ControlCore/versions/0.0.1/requires
@@ -1,2 +1,2 @@
-julia 0.4
+julia 0.4 0.7
 Polynomials


aytekinar:
I have checked the closed PR's to see how the deprecation process is handled. However, I couldn't understand if I should put an upper-bound such as `0.5`, or put a strict version such as `0.0`, so I simply kept this version.

tkelman:
It should be a release version that doesn't exist yet, so 0.7 right now. Otherwise anyone who has installed the package on 0.5 or 0.6 would start getting errors on Pkg.update.

aytekinar:
Aha.. Thanks! Will update now...

tkelman:
No more or less supported than it is right now - it would continue to be installable, Pkg will allow you do add it if you ask. `0.7-` with the trailing dash would mean not installable (and would error) for anyone on current nightlies of julia as well. `0.7` without the trailing dash will mean once people use 0.7.0 final release, then if they try to `Pkg.add("ControlCore")` they'll get an error.

aytekinar:
Last question then... Should I use `0.7` or `0.7-` in this case? If you think this would cause trouble (I am not sure if anyone ever uses this, though), then I will make a last change to 0.7.

PR #11022, opened 2017-09-01
Register Modof.jl and Modolib.jl

aritrasep:
Register 2 new packages for Multi-objective Discrete Optimization in Julia


tkelman:
Modolib shouldn't be cloning Modof in its deps/build.jl script. That would mean anyone who installs Modof will be forced to use master of Modof, possibly getting breaking changes unpredictably instead of letting Pkg manage release version compatibility as usual.

Can you separate the registrations, Modof first then we'll do Modolib separately after that goes through?

tkelman:
Modof should also have a dependency on Conda if it uses it in its deps/build.jl script

PR #11035, opened 2017-09-02
Register new package RiemannComplexNumbers.jl v0.0.1

attobot:
Repository: [scheinerman/RiemannComplexNumbers.jl](https://github.com/scheinerman/RiemannComplexNumbers.jl)
Release: [v0.0.1](https://github.com/scheinerman/RiemannComplexNumbers.jl/releases/tag/v0.0.1)
cc: @scheinerman

Please make sure that:

@scheinerman This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


tkelman:
It's not usually recommended to change the behavior of a base type like this, since that's global. It would be preferable to define a package-local type with whatever behavior you want, then importing your package can't possibly break any unrelated code.

scheinerman:
Yes, I understand about global changes. There's pros and cons though. On the up side, a simple addition of `using RiemannComplexNumbers` at the top of the code automatically changes everything without needed to define a separate `RiemannComplex` type and then wrap complex numbers is that type.
(And  I'm not likely to redo this package for the new type and take care of all the corner cases that might arise.)
This probably is not that important in calculations with complex numbers where one is not going to divide by 0 or get an infinite result.

tkelman:
The problem is that if any dependency does `using RiemannComplexNumbers`, you suddenly get a behavior change that you might not want. Defining new types and behaviors on them is idiomatic and done everywhere in the Julia ecosystem - if code is written with generic enough type signatures, then calling it on new types usually works fine. What corner cases? A straightforward loop over functions where you aren't changing behavior can delegate functionality to existing methods.

scheinerman:
Let me thing about it ... :-)

PR #11041, opened 2017-09-03
Register new package Counters.jl v0.0.1

attobot:
Repository: [scheinerman/Counters.jl](https://github.com/scheinerman/Counters.jl)
Release: [v0.0.1](https://github.com/scheinerman/Counters.jl/releases/tag/v0.0.1)
cc: @scheinerman

Please make sure that:

@scheinerman This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


kmsquire:
Hello @scheinerman, can you contrast this with the `Accumulator` type in `DataStructures.jl` (http://datastructuresjl.readthedocs.io/en/latest/accumulators.html)?  

If yours has some functionality not contained there, would you consider making a pull request there with the updates?  

There may still be reasons to have a separate package, of course.

scheinerman:
Thanks. It looks like the `Accumulator` type serves the same purpose as my `Counter` and so mine is redundant. If you'd rather not list my package in `METADATA`, that's fine.  There are minor differences. For example, I prefer `incr!(c,x)` or `c[x]+=1` for incrementing as opposed to `push!`. I also like using `+` for adding two counters, whereas the `Accumulator` package has `merge`. There are a few other minor differences such as, I can directly set the count of an object, I can decrease the value of a counter (but not below 0), and if I'm counting (say) `String` objects, asking for the count of something that isn't a `String` is an error.

Unfortunately, while I'm OK at writing some Julia code, the whole world of git/travis/METADATA has me pretty baffled, so I don't really how I would modify someone else's code and do a "pull request". 

Bottom line: I think the case for this as a separate package is modest at best, but it's not nothing :-).

quinnj:
I think it's fine to register; DataStructures has gotten quite large, and several of the individual datastructures should probably live in their own package anyway (so dependent projects don't have to include the kitchen sink). I think this is a very well-done package w/ docs, so I'm all for merging.

It _would_ be interesting to do some performance benchmarking on both implementations; I know that's the main reason I haven't used the DataStructures Accumulator before and rolled my own for a specific use-case. I don't remember exactly what was slow, but seems like the type should be optimized more than just using a regular dict.

PR #11068, opened 2017-09-05
NLSolvesBase bounds fix, take 2

anriseth:
Ref: https://github.com/JuliaLang/METADATA.jl/pull/10882/

@tkelman this goes a slightly different route than what I proposed originally. None of the affected packages (LineSearches, NLsolve, Optim) have `require julia 0.5`, so what I wanted to do will not work: They all use lower bound 0.4. I have added an upper bound v0.6 to these, to ensure that no packages are downgraded in order to upgrade NLSolversBase to version 3.0

cc @pkofod


tkelman:
Do any of these versions work correctly on julia 0.6? If so it would be unfortunate to rule out installation there. I wonder if we could release a dummy empty version of NLSolversBase just for julia 0.4?

anriseth:
> I wonder if we could release a dummy empty version of NLSolversBase just for julia 0.4?

So would that mean branching off the initial commit https://github.com/JuliaNLSolvers/NLSolversBase.jl/commit/4ad12ce387f3f9c9983e1a7d59593106701b2dcd, fixing the `runtests.jl` and `REQUIRES` files there, and then tag it as e.g. v0.0.1?

And would the best option be to set REQUIRES `julia 0.4 0.6`, to cover all relevant versions that NLSolversBase never supported?

I don't have tagging rights on NLSolversBase. @pkofod your call whether you want to do it yourself

pkofod:
What does empty mean in ths context @tkelman ?

tkelman:
a package with nothing in the module, that only exists for version resolution's sake

PR #11072, opened 2017-09-06
Register Q [https://github.com/enlnt/Q.jl.git]: v0.0.1

abalkin:


aviks:
We generally dislike single character package names, but in this case it might be the appropriate. I'll wait a bit to see if there are any further comments, but merge otherwise.

PR #11080, opened 2017-09-07
ImageSegmentation: set required version on Clustering

kmsquire:
Noted here: https://github.com/JuliaImages/ImageSegmentation.jl/issues/16


tkelman:
lgtm, should also be done on master of the package

PR #11083, opened 2017-09-07
Register JuliaKara v0.0.1

sebastianpech:
I know I'm violating the guidelines for naming julia packages.
However, the original author of Kara asked me to name the package in accordance to the other Kara ports e.g. JavaKara, PythonKara, RubyKara


sebastianpech:
JuliaCIBot tests fail due to Gtk.

quinnj:
I think the name is fine in this case; thanks for explaining.

PR #11095, opened 2017-09-08
Upper bound DelayDiffEq

devmotion:
This PR adds upper bounds to `DelayDiffEq` versions since there was a breaking change in `OrdinaryDiffEq`, see https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/36 for more information.


ChrisRackauckas:
These bounds are wrong. The upper bound is exclusive, so it should've been 2.19.0. This is causing JuliaCIBot failures.

PR #11098, opened 2017-09-08
Register new package SubsetSelection.jl v.0.0.1

jeanpauphilet:


MikeInnes:
Thanks for the registration! You might consider announcing it on [Julia's forum](https://discourse.julialang.org/).

ViralBShah:
This seems to have failed on all the releases. Perhaps should have been fixed before merging?

PR #11109, opened 2017-09-08
Tag DataCubes v0.2.0 [https://github.com/c-s/DataCubes.jl]

c-s:
Diff vs v0.1.0: https://github.com/c-s/DataCubes.jl/compare/ec27b6dafb700c16e0739c0e547d7b9528134384...5573e2d3bef571e35c51ad9ec95ca67ff1b6d60c

### DIFF
--- a/DataCubes/versions/0.2.0/requires
+++ b/DataCubes/versions/0.2.0/requires
@@ -1,0 +1,4 @@
+julia 0.6
+DataFrames
+Formatting
+DataStructures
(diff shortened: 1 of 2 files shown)


ararslan:
Are you actively testing against 0.7? Looks like you're using `tty_size`, which is causing an `UndefVarError`.

c-s:
I am about to update DataCubes so that it can run under Julia 0.7. Is there a way to run tests for both v0.6 and 0.7 whenever I push my changes, but the above "JulicaCIBot" shows success as long as v0.6 test is successful?

ararslan:
No, JuliaCIBot always reports failure if tests fail on 0.7. But if you're not actively targeting 0.7 and the tests pass on 0.6, that's fine.

PR #11110, opened 2017-09-08
Pull request/c2dfaaef

pjabardo:
Package to calculate thermodynamic properties of moist air


ararslan:
The failure on JuliaCIBot for Julia 0.7 appears to be unrelated to your package.

PR #11123, opened 2017-09-10
Tag NLSolversBase 3.0.0, 0.0.1 and requires changes

anriseth:
This introduces NLSolversBase v0.0.1, which is added for older NLSolversBase package dependants that didn't depend on NLSolversBase before. Hopefully this ensures that NLSolversBase 3.0 can be introduced without downgrading any other stuff.

@pkofod 

Ref #11068


pkofod:
Uhm OptimBase needs req changes

LineSearches/versions/0.0.1/requires
```diff
@@ -1 +1,2 @@
 julia 0.4
+NLSolversBase
```
tkelman: these should all be upper bounded

anriseth: By what version?
`NLSolversBase 0.0.1 1.0`, or `NLSolversBase 0.0.1 3.0` ?

NLsolve/versions/0.10.1/requires
```diff
@@ -4,4 +4,4 @@ Distances
 ForwardDiff 0.5.0
 LineSearches
 DiffBase
-NLSolversBase 2.1.0
+NLSolversBase 2.1.0 3.0
```
tkelman: 0.10.0 also needs an upper bound

anriseth: Ah, thank you - forgot that one in the process. I'm glad you catch all these issues

tkelman: They're going to need to be caught by other people going forward.

This looks good now.

anriseth:
Hmm. I don't understand what the problem is with JuliaCIBot and OptimBase:
- It complains that it doesn't know `NonDifferentiable`, which was introduced in `NLSolversBase 1.0`
- OptimBase previously had a `NLSolversBase` requirement, with no version bounds
- JuliaCIBot "current" passes and shows that it installs `NLSolversBase 2.1.3`
- This PR puts a lower bound of `NLSolversBase 2.1.3`, but it seems like OptimBase installs a version that does not support `NonDifferentiable`.

tkelman:
That last commit should be reverted if OptimBase works with any versions prior to 2.1.3. The problem is JuliaCIBot is trying to check out this newly registered empty version, which it shouldn't do.

anriseth:
> That last commit should be reverted if OptimBase works with any versions prior to 2.1.3

OptimBase should work on 2.0 as well. I've reverted it.

Then this PR is good to go from my side :)

pkofod:
Is this good, or is the cibot an issue?

KristofferC:
This sholud be good to go?

pkofod:
> This should be good to go?

If you say so, I'm not brave enough to say if it is or not :)

KristofferC:
Im gonna close and reopen to hopefully rerun CIBot.

KristofferC:
Actually, I think CI bot checks out the package so it doesn't apply upper bounds properly.

pkofod:
So, I'm just going to bump this one again. Maybe @KristofferC @aviks or @MikeInnes  (I've seen you merging stuff lately) can give a clue as to whether this can be merged now? I think everything is fine, the only problem is with JuliaCIBot itself. It seems to be checking out v3.0.0 of NLSolversBase manually which it shouldn't do.

KristofferC:
Yes, JuliaCIBot does not work well with upper bounds...

KristofferC:
One thing you can do is to start a new Pkg folder (with `JULIA_PKGDIR`), run `Pkg.init()`, check out this PR in METADATA. And then add some packages and see that things behave like you expect.

pkofod:
I cannot say so for all combinations of everything, but a fresh install that adds the three packages that are supposed to use this package seems to do the correct thing. It installs NLSolversBase 3.0, but then it downgrades it to 2.x.y if you try to add Optim, LsqFit and NLsolve, and all three packages have successful `Pkg.test()`.

https://gist.github.com/pkofod/b2a79be336619f30088737171682315f

KristofferC:
Seems good. Let's try it then.

pkofod:
Alright, it didn't break anything as such, but now all PRs for the affected packages refuse to run tests because of package resolution won't resolve on travis and AV. Anyway, I'll create package updates to fix this (fix the count tests and allow them to upgrade to v3.0.0 of NLSolversBase). I'm not sure this is the best setup for ease of maintenance. I like that NLsolve and LsqFit don't depend on Optim anymore, but I think that I may have to rethink the current setup a bit.

KristofferC:
> " refuse to run tests because of package resolution won't resolve on travis and AV."

Why would that happen though? It worked when you did it locally?

pkofod:
It still works but I didn't try to set up Travis and evaluate a pr locally :) the problem is that all tags have upper boundaries in metadata, so whenever option lsqfit or nlsolve wants to move to nlsolversbase v3.0.0 it's basically impossible before a tag is made. I'll fix it later. I just need to tag versions that work started by with line searches and optimbase and then lsqfit nlsolve and option

PR #11147, opened 2017-09-11
Tokamak

MikeInnes:


aviks:
`ERROR: LoadError: Repository name Tokamak does not end in .jl`

MikeInnes:
That's fine as a convention, but it seems strange and unnecessary that it's enforced.

PR #11158, opened 2017-09-12
Upper bound TypeSortedCollections for RigidBodyDynamics.

tkoolen:
Fixes https://github.com/tkoolen/RigidBodyDynamics.jl/issues/332.

### DIFF
--- a/RigidBodyDynamics/versions/0.3.0/requires
+++ b/RigidBodyDynamics/versions/0.3.0/requires
@@ -1,7 +1,7 @@
 julia 0.6-rc3
 StaticArrays 0.6.2
 Rotations 0.6.0
-TypeSortedCollections 0.1
+TypeSortedCollections 0.1 0.1.1
 LightXML 0.4.0
 DocStringExtensions 0.4.0
 Compat 0.27 # for Val constructor


tkelman:
Not necessarily. Pkg could choose to downgrade RigidBodyDynamics to a previous version that doesn't depend on TypeSortedCollections in order to upgrade to the latest TypeSortedCollections version.

tkoolen:
> Pkg could choose to downgrade RigidBodyDynamics to a previous version that doesn't depend on TypeSortedCollections in order to upgrade to the latest TypeSortedCollections version.

That would be better than the current situation. And if users want to avoid that situation, they can set a lower bound on RigidBodyDynamics in their REQUIRE file (either the one for a dependent package, or the global one).

tkoolen:
Any update on this?

tkoolen:
See also https://github.com/JuliaLang/METADATA.jl/pull/11281.

KristofferC:
Should this still be merged?

tkoolen:
In my opinion, yes. It's just a fact that version 0.3.0 of RigidBodyDynamics doesn't work with version 0.1.1 of TypeSortedCollections, so I think that should be modeled. I believe that the scenario from https://github.com/JuliaLang/METADATA.jl/pull/11158#issuecomment-328945323 can't happen anymore because of the new RigidBodyDynamics tag. And again, even that scenario would have been better than the situation before this PR and the release of RigidBodyDynamics 0.4.0, so I don't understand the holdup.

PR #11170, opened 2017-09-12
Pull request/720afe9c (first release of NCDatasets) 

Alexander-Barth:
NCDatasets is a module to read NetCDF files.


sbromberger:
@Alexander-Barth would you mind taking a look at [GraphIO.jl](https://github.com/JuliaGraphs/GraphIO.jl) and seeing if your code makes sense to adopt there? (I'm not a CDF expert.)

Alexander-Barth:
NetCDF is rather a format to store multidimensional arrays. As far as I can tell, it's use case is quite different from graph data.

ararslan:
Do you support Windows with your package? Looking at the build script, I don't see a fallback build or anything like that; will Windows use the Conda downloads? It might be good to add Windows testing using AppVeyor.

Aside from that, things look good here.

Alexander-Barth:
Indeed Windows users will install the NetCDF dependency using Conda. I will use  AppVeyor as you suggested to make sure that it works.

ararslan:
Sounds good. That's not blocking for a tag, so this is good to go.

ararslan:
Looks like you have your DataAssim registration commits in here but those can be squashed out in the merge. Might be good for next time to use Attobot (highly recommended, it's great) or sync your local METADATA branch before opening the PR.

PR #11174, opened 2017-09-13
Register new packages UnsafeAtomics.jl and StingerGraphs.jl

rohitvarkey:
UnsafeAtomics.jl allows for atomic operations to be applied on native Julia arrays rather than arrays of `Atomic` types. We find that this helps improve performance for applications that need a vector of atomics like the parents array of a parallel BFS.

StingerGraphs.jl is a Julia wrapper to the STINGER (http://www.stingergraph.com/) graph database.

These packages are to be presented as part of a paper presentation at HPEC 2017 on Sep 13 2017 around 2pm EDT. It would be awesome if we could have these registered by then so the audience could try it out with just a `Pkg.add`! 

cc: @jpfairbanks


jpfairbanks:
@tkelman Do we need to worry about the fact that JuliaCIBot is failed on 0.7? We aren't targeting 0.7 yet.

aviks:
No we don't need to worry about the JuliaCI result yet. New package registrations are usually kept open for 24 hours to allow feedback.

PR #11190, opened 2017-09-14
Register new packages Persa.jl, Surprise.jl and DatasetsCF.jl

filipebraida:
I have created some packages for the recommendation problem (specifically collaborative filtering).

Persa.jl is the core package. It's the kernel for others.

DatasetsCF.jl is the package for automating the download and loading of classic CFs datasets.

Surprise.jl is a wrapper for the python library and thus converting to use Persa.jl package.

I have more things implemented, but I would like to do it gradually and these are the main ones.


KristofferC:
You need to make sure the packages have tagged releases.

PR #11203, opened 2017-09-16
Add upper bounds to CSTParser's Tokenize requirement

ZacLN:


ZacLN:
@KristofferC , does this look right?

KristofferC:
No, the upper bound is exclusive.

ZacLN:
Got ya,changed it

PR #11293, opened 2017-09-20
Tag Digits.jl v0.2.0

greenflash1357:
retagging Digits v0.2.0


KristofferC:
```
Do not modify existing published tag sha1 files (policy 8).
```

Why not tag v0.2.1?

andreasnoack:

PR #11294, opened 2017-09-20
Tag LineSearches.jl v3.0.0

attobot:
Repository: [JuliaNLSolvers/LineSearches.jl](https://github.com/JuliaNLSolvers/LineSearches.jl)
Release: [v3.0.0](https://github.com/JuliaNLSolvers/LineSearches.jl/releases/tag/v3.0.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/LineSearches.jl/builds/277864473)
Diff: [vs v2.2.1](https://github.com/JuliaNLSolvers/LineSearches.jl/compare/f382f9c0f4ca18cbb205f61d41cc62342e784b1a...f382f9c0f4ca18cbb205f61d41cc62342e784b1a)
`requires` vs v2.2.1: 
```diff
--- 2.2.1/requires
+++ 3.0.0/requires
@@ -1,3 +1,3 @@
 julia 0.6-pre
-NLSolversBase 2.0 3.0
+NLSolversBase 2.0
 Parameters
```
cc: @pkofod

Please make sure that:


pkofod:
I guess I should check this to see if it still uses the 0.9.3 branch of optim and 2.2.1 branch of LineSearches, etc, right?

pkofod:
I checked locally, just as before, and this is fine. Want me to post the output again @KristofferC  ? Or do you trust me :angel: . The reason why CI fails (thanks @ChrisRackauckas for the explanation) is because of https://github.com/JuliaNLSolvers/LineSearches.jl/blob/f382f9c0f4ca18cbb205f61d41cc62342e784b1a/test/optim.jl .

ChrisRackauckas:
You're also using Optim in two other places:

https://github.com/JuliaNLSolvers/LineSearches.jl/blob/f382f9c0f4ca18cbb205f61d41cc62342e784b1a/test/backtracking.jl
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/f382f9c0f4ca18cbb205f61d41cc62342e784b1a/test/counter.jl

That will always cause this to downgrade on CI. Few ways to handle this. One is to just accept that CI won't do well with LineSearches.jl when NLSolversBase.jl changes. Another is to force Travis to grab Optim master if necessary (though @tkelman was against this before), and this can be done in the .yml. Or you can decouple the LineSearches.jl tests completely with Optim. You can add these integration tests in Optim.jl and then JuliaCIBot will find if they fail when tagging, or you can make an internal "small Optim" for testing, or you can make Travis have a different set of tests for running specifically these integration tests, but not have them part of the standard setup (by using an environment variable check).

pkofod:
Actually, come to think about it we did it like this exactly because JuliaCIBot was *not* a thing, so Optim could silently (wrt METADATA CI at least) fail. So I think we can safely remove them from the tests.

pkofod:
I guess we'll be removing these eventually, but for now I'd prefer to just leave the Optim tests in, and merge as is. Then when we tag Optim and NLsolve, I think JuliaCIBot will be so so happy.

PR #11307, opened 2017-09-21
Tag TextParse v0.3.0, and JuliaDB v0.3.0 upperbound old releases

shashi:
cc @davidanthoff this is the DataValues switch.

Also requires #11306


davidanthoff:
Great, thanks! I'll do a CSVFiles.jl release asap that uses the new TextParse.

davidanthoff:
Actually, wait one moment with this one. We might not even need bounds on CSVFiles, I think it might just work with the new TextParse out of the box. I'll double check now.

davidanthoff:
I don't think we need the bounds on CSVFiles.jl. I just tested things locally with ``master`` of TextParse.jl, and everything just works.

And it works because [here](https://github.com/davidanthoff/TableTraitsUtils.jl/blob/master/src/TableTraitsUtils.jl#L18) we have to special case any missing value representation that is not ``DataValue``, but now that TextParse.jl returns ``DataValueArray``, we don't need to convert anything and the default code path just works.

davidanthoff:
Wait, this PR has lots of upper bounds on TextParse in the CSVFiles REQUIRE file that shouldn't be there, as far as I can tell.

shashi:
@davidanthoff #11359

PR #11327, opened 2017-09-22
add upper bounds for sugar + transpiler

SimonDanisch:
I think this is how we can assure, that GPUArrays still works after tagging Sugar + Transpiler for the new changes!


tkelman:
No, this isn't correct. You have to add upper bounds to all past releases' requires files, otherwise Pkg can downgrade these packages in order to upgrade the bounded package.

SimonDanisch:
see https://github.com/JuliaLang/METADATA.jl/pull/11331

PR #11331, opened 2017-09-22
correct upper bounds

SimonDanisch:


Transpiler/versions/0.0.1/requires
```diff
@@ -1,6 +1,6 @@
 julia 0.5
 Compat 0.17.0
-Sugar
+Sugar 0.3 0.4
```
tkelman: don't tighten the lower bound, set it to 0

SimonDanisch:
okay, done :)

PR #11342, opened 2017-09-24
Tag Box0 v0.3.0 [https://gitlab.com/madresistor/Box0.jl]

kuldeepdhaka:


aviks:
Please consider adding a `test/runtests.jl` file, so that the package tests can be run with a single command via the REPL. I appreciate that the tests probably cannot be run without the hardware, but for users who have all the dependencies, making it easy to run the tests might be useful. Also, in the same vein, please consider adding a bindeps script to automatically download the C library dependency.

PR #11353, opened 2017-09-25
New versions of FaSTLMM package

sens:
These are updates to the FaSTLMM package.  Changes are minor.

Please note that I develop in Mercurial and so am opening this pull request manually after making changes to METADATA manually.  Please pardon mistakes.


andreasnoack:
It would be great if you could use https://github.com/attobot/attobot when tagging your package

sens:
Would you like me to redo the pull request?

andreasnoack:
No, I think it is fine but good idea to use Attobot in the future because it provides convenience to check that things are working as expected.

PR #11376, opened 2017-09-27
Register Clarus [https://github.com/ClarusFinancialTechnology/Clarus.… 

garykennedy:


garykennedy:
We can now ignore this PR.
We have found attobot and now use it.
attobot has registered Clarus with v0.0.4 in the Julia metadata

PR #11388, opened 2017-09-28
Registering Hop

mistguy:


mistguy:
I wonder why other packages are registered by attobot?

kmsquire:
> I wonder why other packages are registered by attobot?

@mistguy, can you clarify this comment/question?  It's unclear if

1. you're not sure what Attobot is
2. you you're feeling frustrated at opening three pull requests to get this right, and regretting not using Attobot
3. you genuinely wonder why Attobot is opening so many packages...
4. None of the above

(I admit that one or two of those seem somewhat  unlikely.). Anyway, just in case:

https://github.com/attobot

mistguy:
Oh, I don't know what attobot is and thought other people is just using `PkgDev.publish()` to ask attobot work for them. Now I understand it and thanks for the link.

PR #11391, opened 2017-09-28
Tag DSP.jl v0.3.4

attobot:
Repository: [JuliaDSP/DSP.jl](https://github.com/JuliaDSP/DSP.jl)
Release: [v0.3.4](https://github.com/JuliaDSP/DSP.jl/releases/tag/v0.3.4)
Diff: [vs v0.3.3](https://github.com/JuliaDSP/DSP.jl/compare/f2c056d9b367303d5e6e54f300056c0c693f52a1...69abc408a8d3eaeba8e3999b8c3d8f7a87150f24)
`requires` vs v0.3.3: 
```diff
--- 0.3.3/requires
+++ 0.3.4/requires
@@ -1,4 +1,5 @@
 julia 0.6-rc1
+Compat 0.32
 Polynomials 0.1.0
 Reexport
 SpecialFunctions
```
cc: @ararslan

Please make sure that:


martinholters:
Due to the additional dependency, this should probably become 0.4.0.

ararslan:
Packages can add and drop dependencies in point releases so long as the APIs don't change. That's the case with the addition of Compat; nothing about the addition is user facing.

martinholters:
Hm, but if package `A` depends on `Compat <= 0.10`, then `DSP` will introduce a clash in a point release, which _is_ user facing. (Dropping a dependency would be ok.)

martinholters:
Surely, there is precedent for this kind of question. Probably @tkelman knows whether we typically bump the minor version because of additional requirements?

ararslan:
I don't think there are any packages that place an upper bound on Compat, so this should be entirely safe.

martinholters:
True, that was purely theoretical. And with major version 0, we don't have to be too pedantic anyway. So I'm ok with going ahead here.

PR #11405, opened 2017-09-29
diffeq upper bounds to forwarddiff update

ChrisRackauckas:


tkelman:
This is incomplete, when you add upper bounds it needs to be on all the past versions. Otherwise Pkg can downgrade to earlier DiffEqCallbacks, OrdinaryDiffEq, or StochasticDiffEq versions that you missed here in order to upgrade ForwardDiff to 0.6.

PR #11449, opened 2017-10-03
Temporal Updates / New `Strategems.jl` Package

dysonance:
Temporal updates are mostly focused on better organizing the indexing logic.

Additionally, I would also like to add a new package I've been working on (`Strategems`). While still in early stages of development, the goal is to provide a straightforward and efficient systematic trading strategy backtesting engine. So far it is still somewhat immature, but has enough functionality built in to be able to pretty easily develop basic trading strategies.

Once the `Temporal` updates are integrated, tests should pass on `Strategems` that run through a very basic trading algorithm backtest. More tests still to come, but would like to get this package out to encourage more use and feedback about things to work on / improve.

Been a while since I've added a new package to the ecosystem, so let me know if I've there's anything I'm missing or if there's any questions.


aviks:
It'd be better to have a separate PR for each package.

PR #11497, opened 2017-10-08
Tag Dynare v0.0.2.

stepan-a:


KristofferC:
>ERROR: LoadError: New tag 0.0.2 of package Dynare requires julia 0.6.0, but version 0.0.1 of Dynare requires julia 0.3.0-. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.1.0 using `PkgDev.tag("Dynare", :minor)`.

PR #11559, opened 2017-10-13
Pull request/2e6aa657

wolfiex:
Added Reload.jl


KristofferC:
Have you seen https://github.com/timholy/Revise.jl? It seems you package does exactly what `Revise.track(file)` does?

wolfiex:
Thanks, I though there must be something similar!

 Having just tried it Revise does update variables correctly, however it does not tell you if you have introduced an error in your code (and thus fails to update the data when this happens). 

Additionally it does not seem to rerun print commands when needed.  The idea of this was to help with constantly having to click on the interpreter window to see a variable result, whilst still being able to play with the data when needed (you know RSI and all). 

I guess i should postpone the pull request and see if i can adapt the code form Revise to suit my needs.

KristofferC:
Note that including a file at arbitrary times is quite dangerous. What if the user is in process of writing `rm("somefile...")` and the include fires before the filename is complete? Then you could delete a file that was not intended to be deleted.

aviks:
Closing this PR (to prevent slipped fingers). Please let us know if you need this reopened.

PR #11560, opened 2017-10-14
Register TupleTools [https://github.com/jutho/TupleTools.jl.git]: v0.1.0

Jutho:
This name will probably trigger some debate? I am open to suggestions, though the current name does describe what this package does.


Jutho:
URL is not completely correct (my name should be capitalised). Fixing this now.

PR #11561, opened 2017-10-14
Register TupleTools [https://github.com/Jutho/TupleTools.jl.git]: v0.1.0

Jutho:
This name will probably trigger some debate? I am open to suggestions. This name does however capture the intent of this package, and it's just a bunch of functionality that otherwise I keep redefining in every package (public or private) that I create.


Jutho:
I once again forgot that I should be using `attobot`. Should I close this and use attobot instead?

PR #11594, opened 2017-10-16
PDSampler.jl

tlienart:


aviks:
Hi @tlienart 

* You seem to have changed the name of the package from PDMP to PDSampler. You may want to redo the package registration as well? 
* Do you need to register all these versions? 
* The requires for the later versions do not have a minimum julia version in the REQUIRES. that causes the tests to fail. 

In general, it may be easier to use attobot to register packages. See https://github.com/attobot/attobot#attobot

PR #11606, opened 2017-10-17
Register LempelZiv [https://github.com/Naereen/LempelZiv.jl.git]

Naereen:


Naereen:
The travis build seems to fail because I didn't push any tags, and even though I pushed a tag on my branch, it does not appear here, and I cannot ask a new pull request... :confused: https://github.com/JuliaLang/METADATA.jl/compare/metadata-v2...Naereen:metadata-v2

Naereen:
Thanks !

PR #11665, opened 2017-10-20
Add upper bounds of Celeste dependencies

Keno:
cc @jeff-regier. Without this, the tagged version of Celeste doesn't load on 0.6.


jeff-regier:
Looks good, thanks!

ChrisRackauckas:
You need to upper bound previous versions too or Pkg could lower the version of Celeste to go around the bounds.

Keno:
This is the first version of Celeste compatible with julia 0.6

ChrisRackauckas:
I see, and it lower bounds Julia so it won't be a problem.

PR #11824, opened 2017-11-01
Pull request/7361887f

montyvesselinov:
julia require fixed


KristofferC:
> New tag 0.1.1 of package FEHM requires julia 0.5.0, but version 0.1.0 of FEHM requires julia 0.0.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.2.0 using `PkgDev.tag("FEHM", :minor)`.

I think it is best to drop the v0.1.0 tag which has a REQUIRE without a `julia` entry.

montyvesselinov:
Sure; can you tell me how to do this? I can delete all the tags, retag and
resubmit if you want. Thank you very much!

-monty

martinholters:
Superseded by #11886.

PR #11839, opened 2017-11-02
Put Julia version upper bound on a bunch of DataStructures versions

martinholters:
These don't even load on Julia 0.4.


ChrisRackauckas:
What about earlier versions?

martinholters:
At least they load without error.

PR #11845, opened 2017-11-02
Register new package LocalFilters.jl v0.1.0

attobot:
Repository: [emmt/LocalFilters.jl](https://github.com/emmt/LocalFilters.jl)
Release: [v0.1.0](https://github.com/emmt/LocalFilters.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/emmt/LocalFilters.jl/builds/296283778)
cc: @emmt

Please make sure that:

@emmt This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


timholy:
One does wonder why we need multiple packages for convolution (see https://github.com/emmt/LocalFilters.jl/issues/1), but whatever. I'm slowly coming to realize that people seem to be allergic to "Images" in the name of packages that perform operations on arrays. In this case I'm surprised because those morphology operations originated in the world of image processing, so I wonder how ImageMorphology.jl wasn't discovered before writing LocalFilters.jl.

emmt:
I knew ImageMorphology.jl but I needed more flexible choices for the shape of the structuring elements (not just rectangular boxes).

timholy:
Great, thanks for explaining. But perhaps we should look into ways to avoid splitting the development community---either have ImageMorphology leverage (or be replaced by) LocalFilters or vice versa. I'm not sure it does a lot of good to have two almost-identical packages with very slightly different capabilities, because then there are two packages that need bugfixes, updates to new Julia versions, feature extensions, etc., and maintaining twice the code is twice the work.

I'm traveling now and short on time, but let's discuss in issues.

PR #11894, opened 2017-11-06
Add upper bound on a Feather dependency

davidanthoff:
This should be approved by someone from the JuliaData organization, ideally @quinnj.

It is a short-term fix for https://github.com/JuliaData/Feather.jl/issues/58, which is a real issue for all the downstream packages because it breaks all the tests.


quinnj:
Thanks @davidanthoff!

Issue #11200, opened 2017-09-16
Removing packages


rennis250:
I would like to remove my packages from METADATA.jl. I submitted them years ago and have barely worked on them since. I don't have the time or interest to keep up with them and I'm quite certain that barely anyone uses them. The following packages are what I would like to remove:

- Sparrow.jl
- Processing.jl
- SDL.jl
- Tealeaf.jl (may not actually be in METADATA, but I can't remember...)
- Willow.jl (may not actually be in METADATA, but I can't remember...)
- Arduino.jl (provides almost no functionality)
- GLUT.jl
- OpenGL.jl (not even my package, really. modified from something Jasper den Ouden had made)
- GetC.jl (may not actually be in METADATA, but I can't remember... not even my package, really. modified from something Jasper den Ouden had made)

tkelman:
Packages shouldn't be completely removed in case anyone was depending on them, but they can be deprecated. Arduino, GetC, GLUT, OpenGL, and SDL were already deprecated in https://github.com/JuliaLang/METADATA.jl/pull/9048 since none of them load on Julia 0.6.

Processing never had a release version tagged, so nothing to do there. Tealeaf and Willow were never registered. That leaves just Sparrow. But since the only release Sparrow depends on Color.jl, and Color.jl was deprecated in https://github.com/JuliaLang/METADATA.jl/commit/474b51a5d6fe63140fa759f54bddc7497d567bda, Sparrow is effectively already deprecated too. So I don't think you have to do anything here. If you want to completely delete your copies of the repositories, it would be best to first redirect the url files to point to JuliaPackageMirrors.

rennis250:
Thanks for the info. I opened a Pull Request in #11911 .

PallHaraldsson:
"were already deprecated in #9048 since none of them load on Julia 0.6."

Interesting you can do that. If anyone wants to take over development, I guess it's possible (only with permission of [here] @rennis250; or would you say all are reduntant, e.g. SDL with other (non-SDL-based) better packages? Or don't you just have time?).

After some time, can a name be reclaim for a new package?

rennis250:
@tkelman

PR #11911, opened 2017-11-07
Redirect deprecated package urls, according to discussion in #11200

rennis250:
I no longer maintain any of these packages and would like to remove them from my account. According to the discussion in #11200, it is acceptable if I redirect the urls to the JuliaPackageMirrors versions.

### DIFF
--- a/Arduino/url
+++ b/Arduino/url
@@ -1,1 +1,1 @@
-git://github.com/rennis250/Arduino.jl.git
+git://github.com/JuliaPackageMirrors/Arduino.jl.git
--- a/GLUT/url
+++ b/GLUT/url
@@ -1,1 +1,1 @@
-git://github.com/rennis250/GLUT.jl.git
+git://github.com/JuliaPackageMirrors/GLUT.jl.git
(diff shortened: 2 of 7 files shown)


rennis250:
Any chance that this is acceptable as is?

rennis250:
?

rennis250:
@tkelman

martinholters:
Bump. What's up with JuliaPackageMirrors? Are they going to stay? Is it safe to redirect there?

rennis250:
@tkelman @StefanKarpinski @KristofferC @JeffBezanson

StefanKarpinski:
Not sure – I believe that @tkelman set this up in the first place, but it hasn't been maintained for a long time. We should probably have something like it, but I'm not even part of the org.

rennis250:
Considering the amount of time and the lack of interest in my packages, I will make backups and be removing them from my account today. If you need me to put the backups in an accessible location, then just send me a message thanks.

andreasnoack:
The packages have all been deprecated so I'll just close this.

PR #11917, opened 2017-11-08
Register new package TupledNames.jl v0.0.1

attobot:
Repository: [JeffreySarnoff/TupledNames.jl](https://github.com/JeffreySarnoff/TupledNames.jl)
Release: [v0.0.1](https://github.com/JeffreySarnoff/TupledNames.jl/releases/tag/v0.0.1)
cc: @JeffreySarnoff

Please make sure that:

@JeffreySarnoff This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


KristofferC:
Since this is pure type piracy, wouldn't this make sense as a PR to Base instead of having it as a package?

JeffreySarnoff:
I  could give that a go -- but before the effort,
@JeffBezanson Is there some of this that would be appropriate for Base?  My impetus was coding smoothness.  At present the only occasions I have for stuffing the parametric part of a type/constructor is to multifurcate a state-of-affairs.  And it does not seem necessary to disallow an inlined rewriting step:
```julia

@inline function NamedTuple(names::NTuple{N, Symbol}, values::NTuple{N, T}) where T where N
    return NamedTuple{names}(values)
end
```
I know that value getting by string introduces delay.  My thought is that many uses of NamedTuples are going to be relatively perdurant captures of interest, and for those uses the availability of string-form fetching will be much more natural to many.  To me, this is not as compelling an addition as functional form rewrites and and the ability to instantiate from [these_values...] as well as (these_values...). 

I wanted your opinion first, before combing through code I have not yet seen.

JeffreySarnoff:
@KristofferC (i) If, without introducing a new type, I renamed the rewriting functions and kept their operational definitions is it still a form of type piracy?  (ii) What of introducing a new type that does nothing other than determine the renaming to use?

KristofferC:
If you define constructors for NamedTuples with Base arguments then that will be type piracy. 
If the functions are named to something else, or if the argument types are owned by the package, then it will not be type piracy

In my humble opinion, while these method might look convenient, they will be slow and prevent many of the reasons why NamedTuples was introduced. I would argue that the convenience here is  a performance trap and should perhaps be discouraged?

JeffreySarnoff:
I understand.  I will withdraw it.

PR #11937, opened 2017-11-10
Register PLS [https://github.com/lalvim/PLS.jl.git]

lalvim:


ChrisRackauckas:
I don't know about the name because the acronym is not well known.

lalvim:
Hi cris. There are packages in https://github.com/JuliaStats, such as NMF.jl, that used this naming convention. PLS is a well know naming for Partial Least Squares regressor. How can we proceed?

ChrisRackauckas:
This isn't my domain so I'll let someone else make the call.

aviks:
We very much prefer packages not to have acronyms. Would you be wiling to change it?

lalvim:
@aviks , maybe PartialFit or PLSFit. 
Both are ok?

aviks:
Either is good, I think.

PR #11976, opened 2017-11-13
Augmentor.jl: restrict upper version of ImageCore and StaticArrays

Evizero:
The tagged versions of Augmentor rely on a function in ImageCore that was removed with ImageCore 0.5


Evizero:
I also added upper bounds for StaticArrays because of breaking changes.

see https://github.com/FugroRoames/Rotations.jl/issues/49

PR #11985, opened 2017-11-13
RFC: Freeze METADATA for 0.4

ararslan:
After this, new tags and packages cannot be registered if they support Julia 0.4, as 0.4 is officially unmaintained as of the release of 0.6.


.test/METADATA.jl
```diff
@@ -1,17 +1,13 @@
...
 const gh_path_reg_git=r"^/(.*)?/(.*)?.git$"
 
-const releasejuliaver = v"0.4" # Current release version of Julia
-const minjuliaver = v"0.4.0" # Oldest Julia version allowed to be registered
+const releasejuliaver = v"0.6" # Current release version of Julia
+const minjuliaver = v"0.5.0" # Oldest Julia version allowed to be registered
 const minpkgver = v"0.0.1"   # Oldest package version allowed to be registered
 
-print_list_3582 = false # set this to true to generate the list of grandfathered
+print_list_3582 = true  # set this to true to generate the list of grandfathered
```
tkelman: don't commit this, just use it locally

.test/ci.sh
```diff
@@ -12,7 +12,7 @@ git checkout -b localbranch
 
 mkdir -p $CI_TMP_DIR
 cd $CI_TMP_DIR
-for ver in 0.4 0.5 0.6 nightly; do
+for ver in 0.5 0.6 nightly; do
```
tkelman: leave this alone for at least a little while, nothing should be allowed to break consistency on 0.4

ararslan:
Want to take another look, @tkelman?

PR #12008, opened 2017-11-14
Pull request/883c91c4

lalvim:
Improvement on package organization


aviks:
Hi @lalvim we generally discourage acronyms for package names, see https://docs.julialang.org/en/stable/manual/packages/#Guidelines-for-naming-a-package-1

PR #12026, opened 2017-11-16
update MathProgBase upper bounds

mlubin:
A "major" version bump for MathProgBase 0.7 was required as it dropped support for Julia 0.5, no changes otherwise.


mlubin:
Travis is reporting:
```
ERROR: LoadError: LoadError: Do not modify existing published tag sha1 files (policy 8).
```
Is it no longer appropriate to modify `requires` files like this?

KristofferC:
It is. The Travis test is just unreliable...

PR #12060, opened 2017-11-18
Register PList [https://github.com/ordovician/PLists.jl.git]: v0.0.1

ordovician:
This is an encoder and decoder for the old NeXTSTEP property list ASCII format (plist). macOS can still read these files. In the future I hope to also support the XML and binary version used on macs today.


ordovician:
The JuliaCIBot says the key` 370cdcb2ffc2be606bb5e7846b6cb8418f01f722/PList_julia_0_6.log` is missing, but I don't know what that means or what I can do about it.

PR #12078, opened 2017-11-19
Register new package Expokit.jl

acroy:
This package provides Julia implementations of some routines contained in EXPOKIT. Those routines allow an efficient calculation of the action of matrix exponentials on vectors for large sparse matrices.

cc @ChrisRackauckas


ChrisRackauckas:
I reviewed this earlier. This is good to go once the CI passes.

mforets:
> LoadError: METADATA/Expokit/versions/0.0.1/requires: oldest allowed julia version 0.4.0 too old (>= 0.5.0 needed)

😮

ChrisRackauckas:
Oh yes, v0.4 METADATA is frozen. You should increase the minimum version to v0.5.

ChrisRackauckas:
Instead of modifying

acroy:
Sorry for the mess. I retagged the package and now it should be fine.

PR #12106, opened 2017-11-20
Upper bound julia versions for old Gallium

Keno:
In preparation for tagging a new 0.6 Gallium that just re-exports ASTInterpreter2.


aviks:
I think this should be reverted. This breaks users who had a current Gallium installed.

PR #12132, opened 2017-11-21
 Tag PyTest.jl v0.0.3

pdobacz:
https://github.com/pdobacz/PyTest.jl


KristofferC:
> New tag 0.0.3 of package PyTest requires julia 0.6.0, but version 0.0.1 of PyTest requires julia 0.5.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.1.0 using `PkgDev.tag("PyTest", :minor)`.

PR #12156, opened 2017-11-22
Register FillArrays v0.0.1

dlfivefifty:
FillArrays is a package to represent lazy versions of `fill(x, n...)`, `zeros(T,n...)`, `ones(T,n...)`, and `eye(T,n,m)` as `Fill{T}(n...)`, `Zeros{T}(n...)`, `Ones{T}(n...)` and `Eye{T}(n,m)`. 

This is particularly useful for providing a standard syntax for creating matrices. For example, pull requests will be added to enable `BandedMatrix(Eye(n))`, `CLArray(Zeros(Float32,n,m))` (without any memory transfer), etc. This also mimics Julia v0.7's constructors `Matrix(I, n,m)`.


ChrisRackauckas:
This is a sweet idea! Should it have laziness in the name though?

dlfivefifty:
I'm open to a different name if you have a reasonable suggestion.  (I'm not that big a fan of "lazy" though since it's "at compile time", which seems the exact opposite of lazy...lazy always sounds like it's dynamic.)

PR #12138, opened 2017-11-22
Register new package Case.jl v0.0.1

djsegal:
Repository: [djsegal/Case.jl](https://github.com/djsegal/Case.jl)
Release: [v0.0.1](https://github.com/djsegal/Case.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/djsegal/Case.jl)
cc: @djsegal

PR #12158, opened 2017-11-23
Register new package Underscore.jl v0.0.1

djsegal:
Repository: [djsegal/Underscore.jl](https://github.com/djsegal/Underscore.jl)
Release: [v0.0.1](https://github.com/djsegal/Underscore.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/djsegal/Underscore.jl)
cc: @djsegal


aviks:
Is this really worth a registered packages? Looks like one function, can it go somewhere else?

KristofferC:
I also had the feeling that this is almost a specialization of `replace` to a few values. The use case also seems quite specialised so it makes sense to just bundle this function.

djsegal:
I'm all for putting it in another package, as long as it's not LegacyStrings.

Also, it's simple now. But that's because it has a simple test case.

// things get more complex over time.

KristofferC:
I think having all of this (Case and Underscore) in Fake makes the most sense. There is no need to extract every function into its own package.

djsegal:
see: https://github.com/JuliaLang/METADATA.jl/pull/12138#issuecomment-346997152

// not for lumping case (&underscore) into the faker gem. they are at different levels of abstraction

timholy:
That depends on your perspective about what you mean by "abstraction." I'd call this something like `NamingConventions.jl` and put them together.

djsegal:
Renaming Case.jl and packaging Underscore with it sounds like an acceptable solution. I'll try to get to it this week.

And what I meant is that a painter who makes their own brushes doesn't sell them with the portrait. They use them for future landscapes and abstracts too.

KristofferC:
And that is why you don't ship your editor together with the package.

The brushes can be thrown away when the painting is complete, the utility code that you factor out into its own package here, not so much.

djsegal:
I'm sorry dude. But string manipulation is a very common task. Just cause you don't use something, doesn't mean it's worthless...

Which is what you make clear in every post ever

KristofferC:
I was just disagreeing with the relevance of the metaphor.

djsegal:
closing in favor of:

+ [StringManipulations.jl](https://github.com/JuliaLang/METADATA.jl/pull/12226)

PR #12174, opened 2017-11-24
Added OSQP.jl package

bstellato:
Added OSQP.jl package and tagged version `0.1.0`.


aviks:
We generally dislike acronyms for our package names, but this seems to be a julia wrapper around a native library, so I support this is alright.

PR #12193, opened 2017-11-24
Update LRUCache URL for move to JuliaCollections

ararslan:
cc @jcrist


jcrist:
👍 Looks good to me. Thanks for taking care of this.

PR #12189, opened 2017-11-24
Tag DataArrays.jl v0.7.0

attobot:
Repository: [JuliaStats/DataArrays.jl](https://github.com/JuliaStats/DataArrays.jl)
Release: [v0.7.0](https://github.com/JuliaStats/DataArrays.jl/releases/tag/v0.7.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaStats/DataArrays.jl/builds/306869920)
Diff: [vs v0.6.2](https://github.com/JuliaStats/DataArrays.jl/compare/78327d1e6a8cd27c44d9afbaec419caf6a98893c...e8fe4f25727dad1f7155c784621bbc50737bb064)
`requires` vs v0.6.2: 
```diff
--- 0.6.2/requires
+++ 0.7.0/requires
@@ -1,4 +1,5 @@
 julia 0.6
+Missings 0.2.1
 StatsBase 0.15.0
 Reexport
 SpecialFunctions
```
cc: @nalimilan

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #12220, opened 2017-11-26
Fix upper bounds added for DataArrays 0.7.0

nalimilan:
@tkelman noted at https://github.com/JuliaLang/METADATA.jl/pull/12189 that I forgot to add upper bounds to older releases of DataArrays dependencies which did not depend on it yet, which means that the package manager may downgrade these packages. The check was quite tedious, but overall there aren't many problematic situations, since almost all packages either also depended on DataFrames (which itself depended on DataArrays, so adding the dependency is OK) or were very old (so that an upper bounds on Julia 0.6 is fine). The only case where I added an upper bound on Julia 0.6 to a relatively recent package is EasyPhys 0.1.3, but well...

For reference, the command I found to identify packages which depended on DataArrays in some but not all releases:
```sh
grep -Rl DataArrays | sed "s/\/.*$//g" | uniq | xargs grep -RL DataArrays | grep requires | sed "s/\/.*$//g" | uniq
```
(By removing the last two commands, you get the list of releases which do not depend on DataArrays. That's only a problem if there are more recent releases which do depend on it.)

This gives:
```
Benchmarks: added DataArrays bounds since it also depended on DataFrames
DataArrays: ...
DataFrames: added julia 0.6 upper bound since only old releases did not depend on DataArrays
Discretizers: only old versions depended on DataArrays, so OK
FactorModels: already has an upper bound on julia 0.6
Feather: added DataArrays bounds since it also depended on DataFrames
Gadfly: added DataArrays bounds since it also depended on DataFrames
GeoIP: added DataArrays bounds since it also depended on DataFrames
GoogleCharts: added DataArrays bounds since it also depended on DataFrames
MarketTechnicals: added DataArrays bounds since it also depended on DataFrames (version 0.0.0 only)
MixedModels:  added DataArrays bounds to  since it also depended on DataFrames (versions 0.3.0 to 0.5.1, since others are OK or have an upper bound on julia 0.6)
NormalizeQuantiles: only older releases depended on DataArrays
RCall: only older releases depended on DataArrays
RDatasets: added DataArrays bounds since it also depended on DataFrames
RobustStats: added DataArrays bounds since it also depended on DataFrames
SALSA: only older releases depended on DataArrays
SQLite: only older releases depended on DataArrays, except one release which has julia 0.4 upper bound
Taro: added DataArrays bounds since it also depended on DataFrames
TermWin: added julia 0.6 upper bound since only old releases did not depend on DataArrays
TimeData: added DataArrays bounds since it also depended on DataFrames (only old releases)
TimeSeries: added DataArrays bounds since it also depended on DataFrames (only old releases) 
EasyPhys: added julia 0.6 upper bound to version 0.1.3 since it was the only release not to depend on DataArrays
LazyQuery: added DataArrays bounds since it also depended on DataFrames
```

### DIFF
--- a/TermWin/versions/0.0.10/requires
+++ b/TermWin/versions/0.0.10/requires
@@ -1,3 +1,4 @@
 Lint
 julia 0.3-
 DataFrames 0.0.0 0.11.0
+DataArrays 0.0.0 0.7.0
--- a/Benchmark/versions/0.0.0/requires
+++ b/Benchmark/versions/0.0.0/requires
@@ -1,1 +1,2 @@
 DataFrames 0.0.0 0.11.0
+DataArrays 0.0.1 0.7.0
--- a/Benchmark/versions/0.0.0/requires
+++ b/Benchmark/versions/0.0.0/requires
@@ -1,2 +1,2 @@
 DataFrames 0.0.0 0.11.0
-DataArrays 0.0.1 0.7.0
+DataArrays 0.0.0 0.7.0
--- a/Benchmark/versions/0.0.1/requires
+++ b/Benchmark/versions/0.0.1/requires
@@ -1,3 +1,3 @@
 julia 0.2-
 DataFrames 0.0.0 0.11.0
-DataArrays 0.0.1 0.7.0
+DataArrays 0.0.0 0.7.0
--- a/Benchmark/versions/0.0.1/requires
+++ b/Benchmark/versions/0.0.1/requires
@@ -1,2 +1,3 @@
 julia 0.2-
 DataFrames 0.0.0 0.11.0
+DataArrays 0.0.1 0.7.0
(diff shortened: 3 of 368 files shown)


tkelman:
Are you sure none of the versions with julia upper bounds added load at all on that version of Julia?

nalimilan:
In this PR, the only one is BioMedQuery. Others are really too old to work on 0.6. I've added a commit to use a fake DataArrays dep.

TermWin/versions/0.0.10/requires
```diff
@@ -1,2 +1,2 @@
 Lint
-julia 0.3-
+julia 0.3- 0.6 0.6
```
tkelman: this is effectively not bounded above at all

nalimilan: Hmm, I probably made a mistake with `sed`.

nalimilan:
Merge?

tkelman:
DataArrays had an 0.0.0, so the lower bounds shouldn't be 0.0.1

tkelman:
#12189, #12188, and #12223 also did that incorrectly, the lower bounds shouldn't be excluding a valid version

nalimilan:
Fixed.

PR #12256, opened 2017-11-27
Tag MixedModels.jl v0.14.0

attobot:
Repository: [dmbates/MixedModels.jl](https://github.com/dmbates/MixedModels.jl)
Release: [v0.14.0](https://github.com/dmbates/MixedModels.jl/releases/tag/v0.14.0)
Diff: [vs v0.13.0](https://github.com/dmbates/MixedModels.jl/compare/6d99ac613fd84364fcd8f79a47fd611413e0ffa9...7d152ad8ac69338fd19c41f8fcf5a82a201dd0e8)
`requires` vs v0.13.0: 
```diff
--- 0.13.0/requires
+++ 0.14.0/requires
@@ -1,13 +1,14 @@
 julia 0.6
 ArgCheck
 BlockArrays
-DataArrays
-DataFrames 0.9 0.11.0
+CategoricalArrays
+DataFrames 0.11.0
 Distributions 0.11
 GLM 0.7
 NamedArrays
 NLopt 0.3
 Showoff
 StaticArrays 0.6
-StatsBase 0.11
-StatsFuns 0.3
+StatsBase 0.19
+StatsFuns 0.5
+StatsModels 0.1
```
cc: @dmbates

Please make sure that:


aviks:
Package Tests seem to fail on `0.6`. Is that again one of the `0.6.1` vs `0.6.0` differences?

dmbates:
The problem occurs when creating a vector of matrices that are the results of reshaping contiguous blocks of elements of a vector.  I may be getting too cute here but I would like this vector to be
```jl
StridedMatrix{T}[]
```
The elements of this vector are multiplied in place by triangular matrices and to use the BLAS they should be recognized as `StridedMatrix{T}`.

In 0.6.1
```
julia> isa(reshape(view(rand(25), 2:21), (4, 5)), StridedMatrix{Float64})
true
```

I'll see if the same is true in 0.6.0

dmbates:
I would suggest that I bump the version of julia in `REQUIRE` to 0.6.1.

dmbates:
I'm not quite sure how to do that.  I can make the change in my package repository then create a v0.14.1 release.  Does that seem to be the best approach?

dmbates:
The alternative is to edit `REQUIRE` in this PR but that would cause `REQUIRE` in `METADATA` to be different from `REQUIRE` in the package directory.  I guess that is okay because the files in `METADATA` are sometimes edited after the fact to include upper bounds on package versions.

aviks:
>make the change in my package repository then create a v0.14.1 release

That sounds like the best option. If you do that with attobot, it should update this PR.

PR #12264, opened 2017-11-28
Reverts https://github.com/JuliaLang/METADATA.jl/pull/12106/

bmharsha:
The user's who have already installed older versions of Gallium and pinned it's version, are now unable to perform `Pkg.add()` operations because of this change, reverting this PR will avoid user's from encountering this issue in future.

### DIFF
--- a/Gallium/versions/0.0.1/requires
+++ b/Gallium/versions/0.0.1/requires
@@ -1,4 +1,4 @@
-julia 0.5- 0.6
+julia 0.5-
 TerminalUI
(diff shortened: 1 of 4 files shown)


aviks:
Reverts #12106. cc: @Keno

Keno:
While I'm ok with this to avoid breaking users l, it will make it more likely that Gallium will be broken for users on 0.6. ideally we'd figure out how to teach the package manager to not get confused by this.

PR #12306, opened 2017-12-01
Tag JSON.jl v0.16.2

attobot:
Release: [v0.16.2](https://github.com/JuliaIO/JSON.jl/releases/tag/v0.16.2)
Travis: [Travis Build Status](https://travis-ci.org/JuliaIO/JSON.jl/builds/309854830)
Diff: [vs v0.16.1](https://github.com/JuliaIO/JSON.jl/compare/0b191d9403efeb5939af6a4b094b1e96ca97fd8e...600a40bb11f9c82d5a4590cb61de99460dd0f17b)
`requires` vs v0.16.1: 
```diff
--- 0.16.1/requires
+++ 0.16.2/requires
@@ -1,2 +1,3 @@
 julia 0.6
 Compat 0.37.0
+Nullables 0.0.1
```
cc: @ararslan

Please make sure that:


ararslan:
Needed for Nanosoldier

nalimilan:
The Travis failure persists after restarting. :-/

Looking at the code, that "ok" assertion corresponds to this (pkg/resolve.jl:126):
```julia
        ok, sol = greedysolver(interface)

        if !ok
            try
                graph = Graph(interface)
                msgs = Messages(interface, graph)
                sol = maxsum(graph, msgs)
                ok = verify_solution(sol, interface)
                @assert ok
            catch err
                isa(err, UnsatError) || rethrow(err)
                pp = red_pkgs[err.info]
                for vneq in eq_classes[p][vn]
                    push!(problematic, (p, vneq, pp))
                end
            end
        end
```

I imagine it could be related to the fact that PkgDev uses JSON, but Nullables.jl should merely reexport `Base.Nullable`, so nothing is supposed to break. PkgDev tests pass locally with JSON from git master...

TotalVerb:
I can reproduce the issue on this METADATA branch locally. It must be some artifact with how the solver sanity checks things.

Extremely strangely, the failure seems to be on

```
p = "DiffEqDevTools"
vn = v"0.7.1"
nvn = v"0.8.1"
```

TotalVerb:
It is, I believe, unrelated to the JSON change itself, as it persists even when reverting JSON to 0.16.1. The problem seems to be in the METADATA entry...

TotalVerb:
Someone who knows how the resolver works will need to take a look at this. Here is a patch that "fixes" the issue on 0.7, by (seemingly) making the problematic DiffEqDevTools versions unavailable.

```
diff --git a/DiffEqDevTools/versions/0.7.0/requires b/DiffEqDevTools/versions/0.7.0/requires
index b3a13ee..36b8372 100644
--- a/DiffEqDevTools/versions/0.7.0/requires
+++ b/DiffEqDevTools/versions/0.7.0/requires
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5 0.6
 RecipesBase 0.1.0
 DiffEqBase 0.15.0 1.0.0
 RecursiveArrayTools 0.1.2 0.4.0
diff --git a/DiffEqDevTools/versions/0.7.1/requires b/DiffEqDevTools/versions/0.7.1/requires
index b3a13ee..36b8372 100644
--- a/DiffEqDevTools/versions/0.7.1/requires
+++ b/DiffEqDevTools/versions/0.7.1/requires
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5 0.6
 RecipesBase 0.1.0
 DiffEqBase 0.15.0 1.0.0
 RecursiveArrayTools 0.1.2 0.4.0
```

KristofferC:
cc @carlobaldassi

carlobaldassi:
Yes, that assertion failure signals a bug. I'll try to reproduce and investigate this weekend.

nalimilan:
Should we merge this PR in the meantime anyway?

carlobaldassi:
Update: I have reproduced the error and I think I know what the problem is. I'll need more time to figure out what's the best way to fix it. In the meanwhile, I have verified that setting the `JULIA_PKGRESOLVE_ACCURACY` environment variable to `2` fixes the issue. So this can be merged, although it's likely that this error or some other solver failure will keep popping up in other PRs until `Base` is fixed.

nalimilan:
Thanks @carlobaldassi!

Though I've tried setting `JULIA_PKGRESOLVE_ACCURACY=2` on Travis, and it's so slow that it cancels the build after 10 minutes without any output. And I don't know how to bypass the check to merge the PR. I guess only an owner can do that.

carlobaldassi:
Ah, I see, that's annoying - it takes just 3 minutes on my laptop.
Another possible workaround which does not require a new version of Julia (but would require to rebase this PR) is to remove some problematic old versions of some packages, which introduce circular dependencies and create problems for the solver. That's probably fairly brittle, of course. Anyway, from what I've seen while testing, removing versions "0.4.7" and "0.5.0" from "Colors" (circular dependency with "Graphics"), and versions up to and including "0.1.0" of "URIParser" (circular dependency with "HttpCommon") should help. Here, "removing" may just mean "change their requires file to make them unavailable on julia 0.6 and later", like @TotalVerb did above.

nalimilan:
Thanks. I've checked and indeed the releases you mention do not even load on Julia 0.6, so it's not a problem to an an upper bound. Merging so that I can run Nanosoldier.

PR #12407, opened 2017-12-08
Put Julia version upper bound on a bunch of DataStructures versions

martinholters:
They don't even load on Julia versions>=the new upper bound.

I have no idea regarding the minimum Julia version for DataStructures v0.0.0 --- is `0.0.0` ok in that case?

And we should probably wait with merging this until a new DataStructures version is tagged, as the newest one installable on Julia master would otherwise be v0.2.6 which might result in unpleasant surprises, I could imagine.


aviks:
Ok, let us know when this is safe to merge.

martinholters:
CC @kmsquire Does this look ok to you? Would you mind tagging a new version?

martinholters:
Alright, new version 0.7.3 tagged, so this should be safe now.

kmsquire:
@martinholters, sorry, working through my Julia backlog and just saw this.  Thanks for working through it.  It generally seems fine, although I haven't checked all of the versions.

PR #7652, opened 2017-01-24
Tag FITSIO.jl v0.8.5

attobot:
Repository: [JuliaAstro/FITSIO.jl](https://github.com/JuliaAstro/FITSIO.jl)
Release: [v0.8.5](https://github.com/JuliaAstro/FITSIO.jl/releases/tag/v0.8.5)
Travis: [Travis Build Status](https://travis-ci.org/JuliaAstro/FITSIO.jl/builds/194697491)
Diff: [vs v0.8.4](https://github.com/JuliaAstro/FITSIO.jl/compare/990aacb09c87c78159aeee12faf47934189a7d4b...5079c542af0a91daa8da314808fd01286e316cdc)
`requires` vs v0.8.4: 
```diff
--- 0.8.4/requires
+++ 0.8.5/requires
@@ -1,3 +1,3 @@
-julia 0.3
+julia 0.4
 BinDeps 0.3.12
 Compat 0.8.6
```
cc: @kbarbary

PR #12431, opened 2017-12-10
Register new package Boot.jl v0.0.1

djsegal:
Repository: [djsegal/Boot.jl](https://github.com/djsegal/Boot.jl)
Release: [v0.0.1](https://github.com/djsegal/Boot.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/djsegal/Boot.jl)
cc: @djsegal

-----

notes:

+ tool for loading subdirectories within packages
+ related to: [Discourse #7652 (Packages, modules, and files)](https://discourse.julialang.org/t/packages-modules-and-files/7652/3)


KristofferC:
If you copy code written by other authors you need to comply with the license that code is under. 

It is also not advisable to include files programmatically because it prevents any sort of static analysis. Deleting methods are also not supported by Julia so this can corrupt the method cache. 

We already have a standard method of loading packages. Changing this break assumptions tools have so I don't think it is in the community's interest to have this registered.

djsegal:
> If you copy code written by other authors you need to comply with the license that code is under.

What does this mean? What license did I break? (The delete_method is the only copied thing)

-----

> Deleting methods are also not supported by Julia so this can corrupt the method cache.

Then why does Revise.jl do it?

-----

> I don't think it is in the community's interest to have this registered.

Why does your opinion on sharp knives seem to always dictate everything in the community?

-----

Excerpt from the [Rails Doctrine](http://rubyonrails.org/doctrine#provide-sharp-knives) about Sharp Knives:

+ >Because it’s always about other programmers when the value of sharp knives is contested. I’ve yet to hear a single programmer put up their hand and say “I can’t trust myself with this power, please take it away from me!”. It’s always “I think other programmers would abuse this”. That line of paternalism has never appealed to me.

KristofferC:
> What does this mean? What license did I break? (The delete_method is the only copied thing)

What license is the code you copied under? 

> Then why does Revise.jl do it? 

Revise.jl does not support deleting methods.

> Why does your opinion on sharp knives seem to always dictate everything in the community?

I am sorry you feel that way. If you feel you are being treated unjustly, I would recommend getting in touch with the community stewards (https://julialang.org/community/stewards/).

timholy:
> Then why does Revise.jl do it?

Revise doesn't do it. The code you copied is on an unmerged branch. The branch wasn't merged because it doesn't work properly.

djsegal:
Ok. Closing this sounds like the appropriate move for now then.

Bootloading is a feature that people will continue to ask for though. And it should be seen as more of a necessity than a nicety

PR #12444, opened 2017-12-11
Register BlackBoxOptimizationBenchmarking 

jonathanBieler:
https://github.com/jonathanBieler/BlackBoxOptimizationBenchmarking.jl.git

### DIFF
--- a/BlackBoxOptimizationBenchmarking/url
+++ b/BlackBoxOptimizationBenchmarking/url
@@ -1,0 +1,1 @@
+https://github.com/jonathanBieler/BlackBoxOptimizationBenchmarking.jl.git


aviks:
Tests fail with `LoadError: Package BlackBoxOptimizationBenchmarking has no tagged versions`

Please tag a version prior to registering.

jonathanBieler:
I made a new pull request, sorry about that.

https://github.com/JuliaLang/METADATA.jl/pull/12502

If I'm reading the doc correctly it says you should tag your package only after it's been registered, is that a mistake (I'll open an issue if so) ?

>Once the package URL for FooBar is registered in the official METADATA repo, people know where to clone the package from, but there still aren't any registered versions available. You can tag and register it with the PkgDev.tag() command:

https://docs.julialang.org/en/latest/manual/packages/

aviks:
Yeah, that advise sounds outdated.

PR #12489, opened 2017-12-14
Tag AutoGrad.jl v0.0.10

attobot:
Repository: [denizyuret/AutoGrad.jl](https://github.com/denizyuret/AutoGrad.jl)
Release: [v0.0.10](https://github.com/denizyuret/AutoGrad.jl/releases/tag/v0.0.10)
Travis: [Travis Build Status](https://travis-ci.org/denizyuret/AutoGrad.jl/builds/316491980)
Diff: [vs v0.0.9](https://github.com/denizyuret/AutoGrad.jl/compare/0c27dd85875a8c0cdbc9a4dd11ffbf72493ecbc2...823ea162c829402b0aaf7a7d9e4145f170fdd79b)
`requires` vs v0.0.9: 
```diff
--- 0.0.9/requires
+++ 0.0.10/requires
@@ -1,2 +1 @@
 julia 0.4
-Compat 0.8.1 # allunique
```
cc: @denizyuret

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #12505, opened 2017-12-15
Add AutoGrad v0.0.10 to the grandfather list

omus:
Addresses issue introduced in https://github.com/JuliaLang/METADATA.jl/pull/12489


vchuravy:
thanks! I missed that adding it to this list was necessary.

tkelman:
Travis was set to required to pass for this reason, so it's not possible to screw up other people's pending PR's here. whoever disabled that should put it back.

PR #9919, opened 2017-06-21
Tag Convex.jl v0.6.0

attobot:
Repository: [JuliaOpt/Convex.jl](https://github.com/JuliaOpt/Convex.jl)
Release: [v0.6.0](https://github.com/JuliaOpt/Convex.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaOpt/Convex.jl/builds/245410695)
Diff: [vs v0.5.0](https://github.com/JuliaOpt/Convex.jl/compare/a98ec6f7b1ba108e7920ea045ccc9cda3a747207...2b123e95641e699f364c18969bec51d0c633a3fd)
`requires` vs v0.5.0: 
```diff
--- 0.5.0/requires
+++ 0.6.0/requires
@@ -1,3 +1,4 @@
 julia 0.5
 MathProgBase 0.5 0.7
 DataStructures
+Compat 0.24
```
cc: @madeleineudell

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds are up to date.

PR #12506, opened 2017-12-15
Tag Convex v0.6.0 [https://github.com/JuliaOpt/Convex.jl]

omus:
Diff vs v0.5.0: https://github.com/JuliaOpt/Convex.jl/compare/a98ec6f7b1ba108e7920ea045ccc9cda3a747207...2b123e95641e699f364c18969bec51d0c633a3fd

Registering a tagged version of Convex.jl which has been tagged since June 21st but wasn't published. See https://github.com/JuliaOpt/Convex.jl/issues/211


mlubin:
The previous tag PR was closed: https://github.com/JuliaLang/METADATA.jl/pull/9919. I think there was a question about some breaking changes in the tag that never got addressed.

omus:
Thanks for pointing that out. I'll close this PR and keep the conversation in the original PR.

PR #12588, opened 2017-12-21
Register new package Adapt.jl v0.1.0

attobot:
Repository: [MikeInnes/Adapt.jl](https://github.com/MikeInnes/Adapt.jl)
Release: [v0.1.0](https://github.com/MikeInnes/Adapt.jl/releases/tag/v0.1.0)
cc: @MikeInnes

Please make sure that:

@MikeInnes This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


ChrisRackauckas:
Are we sure about the name? If you can't agree "quickly" then we might as well go with it, but I know that I and @SimonDanisch discussed other namings (but didn't necessarily agree on what it should be). Since Pkg2 is leaving soon anyways I don't think we're in any danger of "taking a good name", but we should try to get it right.

MikeInnes:
Going with it for now. If there are other/better suggestions it's pretty easy to re-register.

In general, I don't think that the name has to serve as a docstring that explains the full idea, it just needs to be memorable once you do understand. I think `adapt` serves that function quite well but would of course be open to other suggestions.

PR #12621, opened 2017-12-26
Register fstformat.jl

xiaodaigh:


aviks:
The general convention is to capitalise package and module names. Is that not a possibility here? Also, the tests seem to require `RCall`, but it is not declared as a dependency in `REQUIRE`. If it is required only for tests, and not for normal operations, then you could add it to `tests/REQUIRE`.

diegozea:
Also, RCall and DataFrames are used in this module and should be in the REQUIRE file of the package. It would be great if the name of the package indicates that this is a tiny wrapper of the R "fst" package rather than a Julia implementation. However,  I don't know what would be the best name for it.

xiaodaigh:
@aviks The reason why this package is uncapitalized is because the format `fst` is also not capitalized. I actually thought about it and I thought that "Fstformat" is not as good as "fstformat"

@aviks @diegozea I updated the REQUIRE

aviks:
Fair enough, I can appreciate that argument, but you should be cognizant of the fact that you will confuse many users who are used to typing upper case names for packages. Overall, I'll tend go with what you feel, but I'll wait to see if anyone else has stronger opinions. 

However, do you want to re-tag the package with the new commit, since you have updated the REQUIRE file? All of which might be easier if you use Attobot.

PR #12653, opened 2017-12-29
Blacklist the original Logging module for julia-0.7

c42f:
Unfortunately this module can't really be installed with julia-0.7, as there's a clash with the new stdlib module of the same name.  See https://github.com/JuliaLang/julia/issues/25182

Alternative and better solutions to this conundrum welcome.

To make `Logging` installable again, I imagine we would need to rename it.

CC @kmsquire, apologies for this mess.


aviks:
@kmsquire  hope you don't mind, I've merged this quickly without much discussions, since otherwise `0.7` is broken for many people. I'll revert/re-fix it if we can come up with a solution.

kmsquire:
No worries!  I’ve been following Chris’s logging development in Base, and
thought something like this would be needed. I haven’t really had time to
work on the Logging package much, and I’m glad Base now has better logging
functionality. Cheers!

PR #12642, opened 2017-12-28
Register new package fstformat.jl v0.0.1

attobot:
Repository: [xiaodaigh/fstformat.jl](https://github.com/xiaodaigh/fstformat.jl)
Release: [v0.0.1](https://github.com/xiaodaigh/fstformat.jl/releases/tag/v0.0.1)
cc: @xiaodaigh

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

@xiaodaigh This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.

PR #12689, opened 2018-01-02
Revert "Register new package fstformat.jl v0.0.2"

aviks:
Reverts JuliaLang/METADATA.jl#12644


aviks:
This should not have been merged since the @xiaodaigh agreed to change the name in  https://github.com/JuliaLang/METADATA.jl/pull/12642

There were two PR's open for the same package, and hence I missed that conversation.

kmsquire:
@tkelman (or anyone else) suggestion on how to handle this?  Will this cause problems?  

The new name only changes the case of two characters, which might cause problems on case-insensitive file systems if someone has installed the package. It’s unlikely many people have done that at this point, so my vote is to go ahead and merge, but wondering if I’m missing something...

tkelman:
Renaming anything that's been checked-in (including to metadata for anyone who has run `Pkg.update` today) in case only doesn't work well on case-insensitive filesystems, so it's entirely possible people will end up with this under a stale name in their local metadata and it can cause problems, yeah.

xiaodaigh:
sorry guys. can we revert that merge? I messed up as i didnt how to delete
a pull request.

kmsquire:
@xiaodaigh, unfortunately, we have to be careful about how we do this, be changes like this have caused problems in the past. 

The main thing now is that the new name of the package has to change, and the new name can’t just change the case of the letters of fstformat (which is unfortunate, because FstFormat was a solid choice).

Some possibilities. 

1. Fst (discouraged as too short/uninformative)
2. FstFileFormat
3. FstDataFrame (it is a DataFrame format, right?)
4. DataFrameFstFormat
5. FstPackage (not very Julian, but mimics upstream somewhat). 

Given that it can’t be FstFormat, I like 2 above.

andreasnoack:
So the package has a new name now https://github.com/xiaodaigh/FstFileFormat.jl. What should be done here? Merge or close?

xiaodaigh:
close it and use thr new name

PR #12709, opened 2018-01-03
Tag Optim.jl v0.12.0

attobot:
Release: [v0.12.0](https://github.com/JuliaNLSolvers/Optim.jl/releases/tag/v0.12.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/Optim.jl/builds/324544172)
Diff: [vs v0.11.0](https://github.com/JuliaNLSolvers/Optim.jl/compare/77858908144b3b852f2ad8e84d52921eea5e69a5...9b75e9058173edbf0f625600afc1c97f05374d73)
`requires` vs v0.11.0: 
```diff
--- 0.11.0/requires
+++ 0.12.0/requires
@@ -3,5 +3,5 @@
 Compat 0.18.0
 LineSearches 3.2.0
 Calculus
-NLSolversBase 3.0.0 4.0
+NLSolversBase 4.0.0
 ForwardDiff 0.5.0
```
cc: @pkofod

Please make sure that:


pkofod:
@jonathanBieler I think the problem with JuliaCIBot can be solved by an upper bound on your benchmarking package. The `Optimizer` should be `AbstractOptimizer` from v0.12.0

pkofod:
I see another pkg depending on Optimizer (unexported) in DSGE and a pkg using a type signature that unfortunately got removed by accident (so that has to be fixed) in Microeconometrics, but the problem in IncrementalInference is only a problem in JuliaCIBot, that wouldn't happen on master (this is because it's trying to checkout Optim 0.12.0 (from this PR) and NLsolve at the same time and NLsolve is not yet ready for NLSolversBase 4.0.0). So I will place an upper bound on DSGE, fix the signature Microeconometrics, and then we can refire JuliaCIBot.

pkofod:
I still need the Microeconometrics thing. I'm PRing bounds, and then I hope we're done!

pkofod:
Hmm, why is DSGE 4.0 using Optim 0.12 on JuliaCIBot, shouldn't this have excluded Optim v0.12 https://github.com/JuliaLang/METADATA.jl/commit/aa3a2c75abfef218f9d313eee4590969867ebc79 ?

pkofod:
anybody wants to close/open this? thanks

pkofod:
Alright, I don't understand why DSGE is failing here. Can anyone weigh in here?

pkofod:
Who knows how JuliaCIBot works? @ararslan ? @aviks ?

aviks:
So right now there is no released version of DSGE that will work with Optim 0.12.0? Is that correct? In which case, this might be a artifact of how we run the tests. 

(Sorry for the delay in looking into this... been swamped)

pkofod:
> (Sorry for the delay in looking into this... been swamped)

No problem, I just don't know how JuliaCIBot actually works. Yes, it is true that DSGE cannot currently be installed with Optim v0.12.0 unless you check it out yourself

aviks:
OK, so in that case I think we should merge this.

pkofod:
I've already notified DSGE.jl, but I'm not sure how active they are publicly.

PR #12729, opened 2018-01-05
Freeze METADATA for 0.5

ararslan:
After this, new tags and packages cannot be registered if they support Julia 0.5, as 0.5 is being declared as officially unmaintained.


aviks:
I'll close this for now (afraid of slipped fingers). Please reopen as necessary, once we have consensus.

andreasnoack:
There is a test error. Would the list need to be updated?

ararslan:
Looks like things are all good now.

PR #12741, opened 2018-01-05
Register hoccfs (was: Pull request/ce4c03d5)

arbenson:


aviks:
Hi @arbenson, we generally prefer packages to start with a capital letter. That is standard accross the ecosystem, and most user expect it to be so.

arbenson:
OK, so is the solution to this problem for me to just re-register a new package following the naming convention (copying and pasting my code).

Also, on a related note, is there a standard guide on registering packages (that would include this information)? I was following a couple of third-party blog posts.

PR #12808, opened 2018-01-09
Update to latest changes in DataFrames

andreasnoack:
With version 0.11.4 of `DataFrames` this version of `RDatasets` works again

### DIFF
--- a/RDatasets/versions/0.2.0/requires
+++ b/RDatasets/versions/0.2.0/requires
@@ -1,7 +1,6 @@
 julia 0.4
-DataFrames 0.0.0 0.11.0
+DataFrames 0.0.0 0.11.0 0.11.4
 Reexport
 Compat
 RData
 FileIO
-DataArrays 0.0.0 0.7.0


nalimilan:
Funny, I've just pushed a branch to do the same thing. I think you can drop the DataArrays dependency altogether since it's not used at all in RDatasets. Also, I wouldn't add an upper bound on DataFrames 1.0: if we do that, we should apply that change to all packages. Doing this only for one package can only create confusion.

nalimilan:
Cc: @randyzwitch

andreasnoack:
I think we should start to add upper bounds on major versions generally such that we don't need to do it retrospectively in METADATA.

nalimilan:
> I think we should start to add upper bounds on major versions generally such that we don't need to do it retrospectively in METADATA.

I think Pkg3 has a mechanism to handle this (e.g. https://github.com/JuliaLang/Juleps/issues/3). In the meantime, better only apply minimal changes.

randyzwitch:
Thanks for pinging me on this. This change makes the tests for the current RDatasets work, without needing the CSV fixes? Or is this PR https://github.com/johnmyleswhite/RDatasets.jl/pull/48 separately needed?

nalimilan:
The tests already work with DataFrames 0.11.4, this PR just allows the package manager to upgrade DataFrames to 0.11.4 when RDatasets is installed. johnmyleswhite/RDatasets.jl#48 will still be needed to silence the deprecation warnings (and to ensure the package will work once deprecations are removed from DataFrames).

randyzwitch:
So do I need to tag a new version of RDatasets with REQUIRES DataFrames 0.11, to override the current METADATA cap of 0.10?

nalimilan:
It would make sense to update REQUIRE in the RDatasets repo to match the one in this PR, but no need to tag a new release.

randyzwitch:
Sorry, I didn't look at what this PR was actually doing. Now I see that it's fixing RDatasets versioning on METADATA, not changing DataFrames.jl. Got it now.

andreasnoack:
> So do I need to tag a new version of RDatasets with REQUIRES DataFrames 0.11, to override the current METADATA cap of 0.10?

@randyzwitch No, it shouldn't be necessary to tag `RDatasets` now. You can just wait until https://github.com/johnmyleswhite/RDatasets.jl/pull/48 has been merged. However, Before tagging a new version you should add `DataFrames 0.11.4 1.0.0` to the `REQUIRE` file.

@nalimilan The current approach where the person who happens to tag a package which other depends on has to add (possibly unnecessary) upper bounds on all the dependent packages or check all the packages one-by-one is so broken that anything else is better than that. There is almost zero chance that this `1.0.0` upper bound will have an effect though.

nalimilan:
> @nalimilan The current approach where the person who happens to tag a package which other depends on has to add (possibly unnecessary) upper bounds on all the dependent packages or check all the packages one-by-one is so broken that anything else is better than that. There is almost zero chance that this 1.0.0 upper bound will have an effect though.

I agree on the goal, but please apply this to all DataFrames dependency, or to none. Having spent a full afternoon on adding these bounds, I'm aware the current system is far from ideal, but let's not make it even less consistent.

nalimilan:
@andreasnoack I've pushed changes, is that OK with you?

randyzwitch:
Thanks everyone

PR #12893, opened 2018-01-16
Tag Yeppp v0.1.2 [https://github.com/JuliaMath/Yeppp.jl]

panlanfeng:
Fix deprecation warnings in julia v0.6.


martinholters:
```
ERROR: LoadError: New tag 0.1.2 of package Yeppp requires julia 0.5.0, but version 0.1.0 of Yeppp requires julia 0.3.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.2.0 using `PkgDev.tag("Yeppp", :minor)`.
```

PR #12942, opened 2018-01-18
publish ZOOjl

AlexLiuyuren:
ZOOjl is the distributed version of ZOOpt. In order to improve the efficiency of handling distributed computing, we use Julia language to code the client end for its high efficiency and Python-like features. Meanwhile, the servers are still coded in Python. Therefore, users programs their objective function in Python as usual, and only need to change a few lines of the client Julia codes (just as easy to understand as Python).


ararslan:
The repository will need to have a .jl suffix. That is, it should be ZOO.jl rather than ZOOjl.

andreasnoack:

PR #12945, opened 2018-01-18
Only run tests on the current release version

aviks:
cherry pick one commit from #12891

### DIFF
--- a/.test/ci.sh
+++ b/.test/ci.sh
@@ -12,21 +12,21 @@ git checkout -b localbranch
 
 mkdir -p $CI_TMP_DIR
 cd $CI_TMP_DIR
-for ver in 0.4 0.5 0.6 nightly; do
-  if [ $ver = "nightly" ]; then
+
+# Only run on the current release version of Julia
+ver=0.6
+
+if [ $ver = "nightly" ]; then
     url="https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz"
     ver=0.7
-  else
+else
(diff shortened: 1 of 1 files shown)


.test/ci.sh
```diff
@@ -12,21 +12,21 @@ git checkout -b localbranch
 
 mkdir -p $CI_TMP_DIR
 cd $CI_TMP_DIR
-for ver in 0.4 0.5 0.6 nightly; do
-  if [ $ver = "nightly" ]; then
+
+# Only run on the current release version of Julia
+ver=0.6
```
omus: Why not just change the loop to: `for ver in 0.6; do`? Maybe in the future we'll want to test on `1.0 1.1`.

.test/ci.sh
```diff
@@ -12,21 +12,21 @@ git checkout -b localbranch
...
-if ! [ -e success-0.4 -a -e success-0.5 -a -e success-0.6 ]; then # nightly allowed to fail
-  exit 1
 fi
+mkdir -p $JULIA_PKGDIR/v$ver/.cache julia-$ver
+ln -s $BUILD_DIR $JULIA_PKGDIR/v$ver/METADATA
+curl -A "$CI_NAME for METADATA tests $(curl --version | head -n 1)" -L --retry 5 $url | \
+    tar -C julia-$ver --strip-components=1 -xzf -
+julia-$ver/bin/julia -e 'versioninfo(); include("$(ENV["BUILD_DIR"])/.test/METADATA.jl")'
+touch success-$ver &
+wait
```
omus: Not much of a point in backgrounding the `touch` command then waiting for it.

PR #13011, opened 2018-01-22
Support testing on private METADATA again

omus:
The changes introduced in https://github.com/JuliaLang/METADATA.jl/pull/12891 break support for private METADATA which run tests on other CI systems (GitLab CI for example).

Key take aways for maintaining support for alternative CI testing:
- Avoid requiring Travis CI specific environment variables "TRAVIS_*"
- Avoid interacting with "https://github.com/JuliaLang/METADATA.jl.git" and instead use `origin`
- Use `origin/HEAD` instead of `origin/metadata-v2` as private METADATAs may use a different remote head name

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -74,7 +74,7 @@ changed, added = cd(BUILD_DIR) do
     (filter_diff("M"), filter_diff("A"))
 end
 
-if isempty(changed) && isempty(added) && ENV["TRAVIS_EVENT_TYPE"] == "pull_request"
+if isempty(changed) && isempty(added) && get(ENV, "TRAVIS_EVENT_TYPE", "unknown") == "pull_request"
     warn("No changes between the PR and the base branch have been detected, which is " *
          "probably wrong.")
 else
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -1,6 +1,10 @@
 # The full path where the repository is cloned and where the job is run
 const BUILD_DIR = ENV["BUILD_DIR"]
 
+# Avoid using the merge commit when checking for changes as sometimes this can result
+# in extra changes being found during the diff.
+const HEAD = get(ENV, "TRAVIS_PULL_REQUEST_SHA", "HEAD")
+
 function get_remote_tags(url)
     ls = try
         readchomp(`git ls-remote --tags -q $url`)
@@ -64,7 +68,7 @@ function get_local_tags(dir)
     return localtags
 end
 
-function filter_diff(filt, commit1="origin/HEAD", commit2="HEAD")
+function filter_diff(filt, commit1="origin/HEAD", commit2=HEAD)
     split(readchomp(`git diff --name-only --diff-filter=$filt $commit1 $commit2`), '\n')
 end
 
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -1,8 +1,6 @@
 # The full path where the repository is cloned and where the job is run
 const BUILD_DIR = ENV["BUILD_DIR"]
 
-const PR_COMMIT_SHA = ENV["TRAVIS_PULL_REQUEST_SHA"]
-
 function get_remote_tags(url)
     ls = try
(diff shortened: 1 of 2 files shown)


omus:
cc: @ararslan 

> This adjusts the logic for checking whether policy 8 (don't change the SHA of an existing tag) is violated. In particular, it explicitly adds JuliaLang/METADATA.jl as a remote, fetches from it, and compares its head commit against the current commit SHA in the PR, as given by Travis. Hopefully this will improve the reliability of this sporadically incorrect check.

If you can provide me some information with how you were testing this check I can verify this PR doesn't re-introduce the issue.

.test/check-pr.jl
```diff
@@ -64,32 +64,22 @@ function get_local_tags(dir)
...
-    run(`git remote add _upstream https://github.com/JuliaLang/METADATA.jl.git`)
-    run(`git fetch _upstream`)
-    upstream_commit = readchomp(`git rev-parse _upstream/metadata-v2`)
-
-    # Compare the current commit with the default branch upstream, returning
-    # a string containing a newline-delimited list of files changed. We only
-    # care about additions (A) and modifications (M).
-    _changed = filter_diff(upstream_commit, PR_COMMIT_SHA, "M")
-    _added = filter_diff(upstream_commit, PR_COMMIT_SHA, "A")
+function filter_diff(filt, commit1="origin/HEAD", commit2="HEAD")
```
ararslan: Can we make the `commit2` default here `get(ENV, "TRAVIS_PULL_REQUEST_SHA", "HEAD")`?

ararslan:
> If you can provide me some information with how you were testing this check I can verify this PR doesn't re-introduce the issue.

As long as the Travis log here correctly reflects which files were added and/or modified, I think we're good. It was picking up random garbage before.

.test/check-pr.jl
ararslan: Or maybe this is fine as-is.

omus: `"HEAD"` should be set to be the same as `"TRAVIS_PULL_REQUEST_SHA"` for the Travis CI PR build. I would be concerned with switching the default as I _think_ this may also make the Travis CI push build also use the PR SHA.

ararslan: `TRAVIS_PULL_REQUEST_SHA` should only be defined for PR builds; it's empty for push builds. If `HEAD` is definitely the same then I'm fine with using it.

omus: Since it's not hard to validate this I figured I would just run a quick CI test: https://github.com/JuliaLang/METADATA.jl/pull/13016

ararslan: Ah, they are in fact different, since the commit Travis uses is the merge commit with the base branch.

omus: Results are different than I expected. I pushed the commit `eae304c081a2400c891b0152fc0d06c50ed0ccb0` and:

```
# Travis CI PR
TRAVIS_PULL_REQUEST_SHA: eae304c081a2400c891b0152fc0d06c50ed0ccb0
HEAD: 767aa62302fe17449e490b52da9260b97cb34fa8
```

```
# Travis CI Push
TRAVIS_PULL_REQUEST_SHA: 
HEAD: eae304c081a2400c891b0152fc0d06c50ed0ccb0
```
It looks we definitely do not want to use "TRAVIS_PULL_REQUEST_SHA".

ararslan: No, we do--that's how we make sure we don't pick up changes accidentally. The current branch's HEAD can be different from the merge commit with the base branch. Perhaps what we should do is determine whether the current CI provider is Travis and if not manually check out the merge commit like we do on Circle for the Julia repo.

omus: Ok, I think we're finally on the same page. The one part I'm not clear on is why we the PR SHA is picking up extra changes.

I'll add in:
```julia
# Avoid using the merge commit when checking for changes
const HEAD = get(ENV, "TRAVIS_PULL_REQUEST_SHA", "HEAD")
```
This should cover your concerns

omus:
@ararslan we good to merge this?

PR #13104, opened 2018-01-26
Disable policy 8 check

ararslan:
cc @aviks and @omus

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -68,31 +68,36 @@ function get_local_tags(dir)
     return localtags
 end
 
-function filter_diff(filt, commit1="origin/HEAD", commit2=HEAD)
-    split(readchomp(`git diff --name-only --diff-filter=$filt $commit1 $commit2`), '\n')
-end
-
-# Compare the current commit with the default branch upstream, returning a list of files
-# changed. We only care about additions (A) and modifications (M).
-changed, added = cd(BUILD_DIR) do
-    (filter_diff("M"), filter_diff("A"))
-end
-
-if isempty(changed) && isempty(added) && get(ENV, "TRAVIS_EVENT_TYPE", "unknown") == "pull_request"
-    warn("No changes between the PR and the base branch have been detected, which is " *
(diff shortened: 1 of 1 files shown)


tkelman:
Your changes may have made it worse, but the failures on CI were a race condition. Not a good idea.

aviks:
cc: @KristofferC I'm on board with this. Not ideal, but at this point it will make things better.

aviks:
Things were pretty bad before Alex's changes. We'd discussed turning it off then as well.

ararslan:
It seemed to be working correctly after my changes, but @omus noted that they were incorrect for private instances of METADATA, and I think https://github.com/JuliaLang/METADATA.jl/pull/13011 reintroduced the failures.

tkelman:
Don't disable it entirely, make it retry on failure to verify. Outright deletion of a tag also needs to be flagged. And people merging to a repo that's pulled from by hundreds or thousands of people a day need to pay closer attention, avoid merging multiple pr's related to the same tag where CI will be inaccurate. You're going to need to make the pkg3 registries more automated and prevent user/maintainer mistakes, not require more things to be verified manually.

tkelman:
You also *really* need to be running the consistency checks on all versions of julia that tags are accepted for, which you shut off for no reason and have completely ignored.

aviks:
At this point we have a bit of crisis at the moment, we need to get things fixed in the short term, so that metadata can continue to function. In this case, no tests are better than bad tests, as the current situation demonstrates, since bad tests directly lead to human mistakes.

omus:
I have a simple fix I think will address the problem and not disable the tests. I wasn't done validating it but I'll push it right away.

omus:
Potential fix I mentioned: https://github.com/JuliaLang/METADATA.jl/pull/13107

omus:
> At this point we have a bit of crisis at the moment

My preference is we try my potential fix first. If that doesn't resolve the issues then we should consider rolling back some changes then.

aviks:
Closed in favour of #13107

PR #13076, opened 2018-01-25
Tag Diversity.jl v0.4.2

attobot:
Repository: [richardreeve/Diversity.jl](https://github.com/richardreeve/Diversity.jl)
Release: [v0.4.2](https://github.com/richardreeve/Diversity.jl/releases/tag/v0.4.2)
Diff: [vs v0.4.1](https://github.com/richardreeve/Diversity.jl/compare/79cecfc9992c2ef65e4f40bd230fa7fa99b464e0...e008d04e282999113d2484c380ce8c38c2598233)
`requires` vs v0.4.1: 
```diff
--- 0.4.1/requires
+++ 0.4.2/requires
@@ -2,3 +2,4 @@
 DataFrames 0.9.0
 Phylo 0.2.0
 Missings
+Compat 0.47.0
```
cc: @richardreeve

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #13107, opened 2018-01-26
Use common ancestor for git diff

omus:
> "git diff A...B" is equivalent to "git diff $(git-merge-base A B) B"

When running the "check-pr.jl" tests we use `git diff` to verify that no METADATA files were modified which should remain constant. When we use `git diff A B` we can potentially see changes which occurred outside of the scope of the PR.

I believe (I haven't fully validated this) that it's caused from the `origin/HEAD` being updated and being newer than the HEAD.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -3,7 +3,7 @@ const BUILD_DIR = ENV["BUILD_DIR"]
 
 # Avoid using the merge commit when checking for changes as sometimes this can result
 # in extra changes being found during the diff.
-const HEAD = get(ENV, "TRAVIS_PULL_REQUEST_SHA", "HEAD")
+const DIFF_HEAD = get(ENV, "TRAVIS_PULL_REQUEST_SHA", "HEAD")
 
 function get_remote_tags(url)
     ls = try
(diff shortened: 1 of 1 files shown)


omus:
[This would be the line](https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/ci.sh#L10) which is causing `origin/HEAD` to be updated. I added this in #13011 which was equivalent to [this line](https://github.com/JuliaLang/METADATA.jl/pull/12891/files#diff-50c5778d18a48506ec4655c379777ef0R73) added in https://github.com/JuliaLang/METADATA.jl/pull/12891.

Really pulling from the origin shouldn't be necessary but there are some cases where `origin/HEAD` isn't available in the local repo.

aviks:
Can this be rebased on current master? That might help with the push test failure

ararslan:
I've rebased this branch against metadata-v2.

omus:
Thanks Alex. Was in too much of a hurry.

omus:
Squash when ready

aviks:
Looks good to me. lets go for it.

omus:
Restarting Travis CI on https://github.com/JuliaLang/METADATA.jl/pull/13076 with the new changes.

omus:
Restarted PR build is looking good already. I think we did it!

omus:
Changes passed. I'm restarting the Travis CI PR jobs which will be fixed by this change.

omus:
Ok, I think I restarted all the PRs that were effected.

martinholters:
Great!

Can we now please also re-enable the checks on older Julia versions? Why did we have to disable anyway?

omus:
> Can we now please also re-enable the checks on older Julia versions? Why did we have to disable them anyway?

I made a PR to re-enable checks on older version of Julia: https://github.com/JuliaLang/METADATA.jl/pull/13149

PR #13146, opened 2018-01-29
ZOOclient

AlexLiuyuren:
ZOOclient is the client part of Distributed ZOOpt. In order to improve the efficiency of handling distributed computing, we use Julia language to code the client end for its high efficiency and Python-like features. Meanwhile, the servers are still coded in Python. Therefore, users programs their objective function in Python as usual, and only need to change a few lines of the client Julia codes (just as easy to understand as Python).


aviks:
Closed in favour of #13159

PR #13149, opened 2018-01-29
Perform tests on multiple versions of Julia again

omus:
Reverts https://github.com/JuliaLang/METADATA.jl/pull/12945

### DIFF
--- a/.test/ci.sh
+++ b/.test/ci.sh
@@ -15,21 +15,21 @@ git checkout -b localbranch
 
 mkdir -p $CI_TMP_DIR
 cd $CI_TMP_DIR
-
-# Only run on the current release version of Julia
-ver=0.6
-
-if [ $ver = "nightly" ]; then
+for ver in 0.4 0.5 0.6 nightly; do
+  if [ $ver = "nightly" ]; then
     url="https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz"
     ver=0.7
-else
+  else
     url="https://julialang-s3.julialang.org/bin/linux/x64/$ver/julia-$ver-latest-linux-x86_64.tar.gz"
-fi
-mkdir -p $JULIA_PKGDIR/v$ver/.cache julia-$ver
-ln -s $BUILD_DIR $JULIA_PKGDIR/v$ver/METADATA
-curl -A "$CI_NAME for METADATA tests $(curl --version | head -n 1)" -L --retry 5 $url | \
-    tar -C julia-$ver --strip-components=1 -xzf -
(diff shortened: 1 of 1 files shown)


omus:
cc: @aviks and @ararslan

martinholters:
Why not check on 0.4, BTW?

martinholters:
So the CI failure seems to be because ODE 1.0.0 claims Julia 0.5 as lower bound but requires DiffEqBase 3.0.0 which has Julia 0.6 as lower bound. That looks like a legitimate concern. Too bad testing on Julia 0.5 was disabled when ODE 1.0.0 was tagged (#13059). Now what? Change the minimum Julia version of ODE 1.0.0 to 0.6? CC @ChrisRackauckas

ChrisRackauckas:
Yes, that was a mistake.

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/13151

omus:
> Why not check on 0.4, BTW?

I opted to not run the Julia 0.4 tests as Julia 0.5 is now required for new registered packages (and I believe tags). At some point we should drop support for testing on older Julia versions and I thought this was a reasonable cutoff.

martinholters:
So here's why I'd even like to test on Julia 0.3: Reproducibility. Someone---possibly me---may have old Julia code lying around somewhere, the results of which got published in some paper. Years later, I need to recreate the results for some reason. I install some old Julia version, do `Pkg.init()`, add the required packages, maybe pin them to old versions, and off I go. Except if at any point `Pkg` barks at me because it uses the newest METADATA commit that is indigestible to the old Pkg..

ararslan:
I don't think that scenario can happen. When METADATA is frozen for a particular version of Julia, you won't be able to install package versions that are newer than the cutoff, so it will effectively be in the last known-good state for that version of Julia forever. (That's how I understand it, anyway.)

omus:
Rebased to verify changes work with merged #13151

tkelman:
Technically any time you merge an oldest-allowed-julia exception tag you should verify it doesn't break these tests.

martinholters:
Theoretically, it might be possible that some subtlety in a `requires` file trips off older Julia versions, but goes unnoticed on newer ones. (I'm thinking of unusual whitespaces, a stray NUL byte at the end, stuff like that.) Improbable? Yes. Impossible? I don't know. Do you?

Also, why should we _not_ run the tests on older Julia versions? Is that just because of CI time?

omus:
Mostly it is CI time. I think ideally the way to address this is to test on all versions of Julia on and after the specified Julia version in the REQUIRE file (which was modified/added in the branch).

martinholters:
Closing/reopening to trigger CI against current ~master~ metadata-v2; will merge if that comes out green. We can think about refinements/better coverage for older Julia later.

PR #13159, opened 2018-01-30
Pull request/32623c6d

AlexLiuyuren:
ZOOclient is the client part of Distributed ZOOpt. In order to improve the efficiency of handling distributed computing, we use Julia language to code the client end for its high efficiency and Python-like features. Meanwhile, the servers are still coded in Python. Therefore, users programs their objective function in Python as usual, and only need to change a few lines of the client Julia codes (just as easy to understand as Python).


aviks:
I see that you have fixed the test to not fail after this. Do you want to update this PR with the new tag?

AlexLiuyuren:
But i failed JuliaCIBot test. It says:
ERROR: LoadError: ArgumentError: Module ZOOjl not found in current path.
My runtest.jl only has one valid line: using ZOOclient
ZOOjl is the previous git repository name, I have change it to ZOOclient now. I have checked all of my code in https://github.com/eyounx/ZOOclient.jl  and find that the word `ZOOjl` does not exist. This Error confuses me and i'm struggling to eliminate it.

andreasnoack:
I assume this is superseded by https://github.com/JuliaLang/METADATA.jl/pull/13180

PR #13164, opened 2018-01-30
Tag FixedEffectModels 0.5.0

matthieugomez:


andreasnoack:
Looks like this breaks https://github.com/jmboehm/RegressionTables.jl

matthieugomez:
What am I supposed to do? I am not sure where the failure comes from.
The JuliaCIBot reports a log with 
```
ERROR: unknown package RegressionTables
```

jmboehm:
Strange, it seems that the install script is not cloning `RegressionTables.jl.git`. I've re-set travis.yml to use the default script... let's see whether this helps.

PR #13185, opened 2018-01-31
Register new package TriangleMesh.jl - Pull request/d93b7ebf

konsim83:
Julia package for generation and refinement of triangular meshes. Interfaces C-library "Triangle". Name "TriangleMesh" chosen because "Triangle" already exists (unregistered, with different functionality though). Package is documented and tested (win, linux, osx).


fredrikekre:
What is the difference between this and https://github.com/cvdlab/TRIANGLE.jl? You say that the functionality is different, but the same C-library is wrapped?

konsim83:
True, the same C-library is wrapped but TriangleMesh has a much more flexible interface that allows users to mesh polygons or point clouds in various ways by passing parameter keywords to the mesh to the 'create_mesh' function. The package TRIANGLE.jl does not allow this flexibility and also does not return a struct with all the information that is sometimes necessary for FEM/DG applications (markers, attributes, neighbors etc). Secondly, TriangleMesh, in contrast to TRIANGLE.jl supports local refinement. In principle this can be extended to handle also adaptive refinement. Point three: TriangleMesh allows even more flexibility for users familiar with passing command line switches the Triangle C-library. Point four: It has a thorough documentation that describes the workflow for both the convience routines and direct passing of command line switches.

PR #13241, opened 2018-02-05
Register Dojo [https://github.com/VIVelev/Dojo.jl.git]

VIVelev:


andreasnoack:
`ERROR: LoadError: Package Dojo has no tagged versions`

PR #13293, opened 2018-02-08
Tag ScikitLearn.jl v0.4.0

attobot:
Repository: [cstjean/ScikitLearn.jl](https://github.com/cstjean/ScikitLearn.jl)
Release: [v0.4.0](https://github.com/cstjean/ScikitLearn.jl/releases/tag/v0.4.0)
Travis: [Travis Build Status](https://travis-ci.org/cstjean/ScikitLearn.jl/builds/339041954)
Diff: [vs v0.3.0](https://github.com/cstjean/ScikitLearn.jl/compare/95ee7137d368e59de51be5009d5bb4e1f043b077...f82da50940cb4ef98f7f6277972f3565055a29cd)
`requires` vs v0.3.0: 
```diff
--- 0.3.0/requires
+++ 0.4.0/requires
@@ -7,4 +7,4 @@
 StatsBase 0.8.0
 Iterators 0.1.9
 Compat 0.18.0
-DataFrames 0.0.0 0.11.0+DataFrames 0.11
```
cc: @cstjean

Please make sure that:


andreasnoack:
There are quite a bit test failures here

cstjean:
I'm not sure how to fix this. I do need to drop 0.5 support, but this is moot, since DataFrames 0.11 doesn't support it anyway. The main issue is

```
ERROR: resolve is unable to satisfy package requirements.
  The problem was detected when trying to find a feasible version
  for package DataFrames.
  However, this only means that package DataFrames is involved in an
  unsatisfiable or difficult dependency relation, and the root of
  the problem may be elsewhere.
```

In order to upgrade to DataFrames 0.11, I had to checkout a bunch of packages (including DataFrames.jl), and I'm still not clear on which package was holding back DataFrames. Compounding the problem is that GaussianProcesses has an upper bound on Optim 0.8, which causes more package resolving weirdness.

All I can say is that the tests pass locally. If this is not sufficient for METADATA standards, I may have to wait until the ecosystem progresses on DataFrames and Optim.

```
 - Atom                          0.6.9
 - BenchmarkTools                0.2.5              master
 - ClobberingReload              0.2.1+             master
 - Compose                       0.5.4
 - Conda                         0.7.1
 - DataFrames                    0.11.5+            master
 - DecisionTree                  0.6.5
 - GaussianMixtures              0.1.0
 - GaussianProcesses             0.4.0
 - IJulia                        1.7.0
 - Images                        0.12.0
 - JLD2                          0.0.5
 - JuMP                          0.18.0
 - LightGraphs                   0.11.1
 - LowRankModels                 0.2.0+             master
 - Memoize                       0.2.0
 - NBInclude                     1.2.0
 - NBTesting                     0.0.2+             master
 - PkgDev                        0.1.6
 - PlotlyJS                      0.10.0
 - Plots                         0.15.0
 - ProfileView                   0.3.0
 - ProgressMeter                 0.5.3
 - PyCallJLD                     0.0.2
 - PyPlot                        2.3.2+             master
 - QuartzImageIO                 0.3.1
 - QuickTypes                    0.2.0+             master
 - RDatasets                     0.3.0
 - Requires                      0.4.3
 - Revise                        0.1.1
 - ScikitLearn                   0.3.0+             master
 - TypedTables                   0.2.0
 - WAV                           0.9.0
```

andreasnoack:
Try to free all the checked out packages and then do `Pkg.update("DataFrames")` then `Pkg` often tells you exactly which package is holding back the update. (My guess is Gadfly)

cstjean:
```
> Pkg.free("LowRankModels")
LowRankModels's requirements can't be satisfied because of the following fixed packages: DataFrames
```

That's probably it. The current LowRankModels master was updated for missing values, and it's only a TEST/REQUIRE requirement, so 99% of users won't even notice any issue. However, even once it's tagged, Travis will still fail because of the PyPlot ZLib issue https://github.com/JuliaPy/PyPlot.jl/issues/151#issuecomment-361058512 (it used to work fine)

cstjean:
So what's the solution here? I'm waiting for LowRankModels to be tagged (again - only a test/REQUIRE dependencies). Most of my tests come from running the example notebooks, which is IMO a nice combination, but the PyPlot ZLib issue is unlikely to be solved anytime soon. I don't have time or inclination to copy-paste all the notebook code into tests/runtests.jl. If passing tests is a prerequisite for tagging, then I'll disable the notebook tests, though that hardly seems like a win to me.

I could look into ways to run the notebooks without plotting, but in the meantime, it sucks to keep DataFrames down for ScikitLearn users.

andreasnoack:
It looks like `LowRankModels` master passes on 0.6 so why not get a new release registered in METADATA? We have been ignoring test failures related to the zlib issue so that shouldn't be a problem for the release.

andreasnoack:
Notice also that since this PR was opened we have changed the minimum requirement to Julia 0.6 so the METADATA tests fail with
```julia
ERROR: LoadError: METADATA/ScikitLearn/versions/0.4.0/requires: oldest allowed julia version 0.5.0 too old (>= 0.6.0 needed)
Stacktrace:
```

PR #13297, opened 2018-02-08
Tag SubsetSelection v0.0.4 [https://github.com/jeanpauphilet/SubsetSelection.jl]

jeanpauphilet:
New release provides: 
- better documentation
- tests/exceptions when data is not formatted appropriately
- warnings if the algorithm fails to converge efficiently and suggestions to fix it
- accurate computation of the resulting regressor (recover_primal function)

Diff vs v0.0.3: https://github.com/jeanpauphilet/SubsetSelection.jl/compare/091f03fa529ebf91a758abb964a4fdce337d3667...556afe8887260ceb035697fb8631ac96cd2e937b


andreasnoack:
There seems to be an issue with the dependency versions here
```
    SubsetSelection v0.0.4 – no valid versions exist for package 
```

jeanpauphilet:
Indeed. I require julia 0.5 0.7- since I have not tested it on julia 0.7 yet. Is there a way not to run the test on julia 0.7?

martinholters:
The problem occurs on Julia 0.5 because you have `LIBLINEAR 0.3.0` in your REQUIRE, but that in turn requires `julia 0.6-`, making your package uninstallable.

andreasnoack:
It looks like the tag here and the one in your repo are now out of sync. The best solution would probably be to enable https://github.com/attobot/attobot for your repo. Then remove the release and the 0.0.4 tag and the make a new 0.0.4 release which will then be handled by Attobot.

andreasnoack:
No activity for three months so closing.

PR #13298, opened 2018-02-09
Pull request/cdefc996

currymj:
This repository provides a sampler using the Distributions.jl interface for the Pólya-Gamma distribution, which is used in Bayesian statistics, especially when you have a Bernoulli or binomial distribution parameterized by a real value squashed by the logistic function.


andreasnoack:
Please consider using https://github.com/attobot/attobot when tagging new releases of the package.

PR #13310, opened 2018-02-10
Register Commutator [https://github.com/goerz/Commutator.jl.git]: v0.1.0

goerz:


goerz:
Is it typical that pull requests aren't merged for over a week?

andreasnoack:
When registering new packages, we typically leave some time for comments about the name and possible overlaps with other packages. However, typically such PRs don't stay open for as long as this one.

PR #13340, opened 2018-02-13
Tag Gadfly.jl v0.6.5, Compose.jl v0.5.5

bjarthur:


bjarthur:
CI reports as failing, yet clicking through reveals the "current" is the same as the "new", so i think this is okay to merge.

andreasnoack:
Have you ever seen the specific test error locally or on Travis?

bjarthur:
no, but i see [the problem](https://github.com/GiovineItalia/Gadfly.jl/blob/master/test/runtests.jl#L115) now.  `compare_examples.jl` should only be run in an interactive session.  moreover, the checks on {genned,cached}output/ should ignore the .mkdir file.  i'll submit a PR.

PR #13345, opened 2018-02-13
Register BayesNN and update GradDescent

jacobcvt12:
Created a package for fitting bayesian neural networks with variational inference; updated `GradDescent` to include some recent gradient optimizers.


andreasnoack:
Would be great if you could split registration and tagging into separate PRs as well as using https://github.com/attobot/attobot. It makes the process easier.

martinholters:
In case you missed it in the log: `Error: Repository name BayesNN does not end in .jl`

andreasnoack:
No activity for three months so closing

PR #13367, opened 2018-02-15
Clean up the log

andreasnoack:
Merging ~~https://github.com/JuliaLang/julia/pull/25662~~ https://github.com/JuliaLang/julia/pull/25472 ~~broke the version testing here~~ printed an error in the log which is confusing when the tests fail for some other reason or because of timeout.

The `Pkg` deprecation warning also added a lot of noise. Finally, I've disabled running the tests in parallel. It completely messes up the log and was the original reason why I missed an error. It will make each build slower but I guess it shouldn't have an effect on the overall throughput since we can just run more builds concurrently. Is that a correct interpretation of Travis?

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -153,7 +153,9 @@ for (pkg, versions) in Pkg.Read.available()
                     same_minor(x::VersionNumber) = (majmin(x) == majmin(ver) &&
                         juliaver_in_require(pkg, x; check=false) < juliaver)
                     ind_same_minor = findfirst(same_minor, sortedversions)
-                    ind_same_minor == 0 && continue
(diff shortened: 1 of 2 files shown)


nalimilan:
The PR you mention is about `indexin`. The `findfirst` changes are much older, and we should be able to use `Compat.findfirst` for that.

andreasnoack:
Thanks. I've updated the link. I'm not sure if using packages in the package testing script is feasible/practical.

tkelman:
> It will make each build slower but I guess it shouldn't have an effect on the overall throughput since we can just run more builds concurrently. Is that a correct interpretation of Travis?

No it is not. Each Travis VM has multiple virtual cores, which is what this had been making use of. Now it's only using a single core and letting the others that are allocated sit idle. Travis counts concurrency based on separate jobs, not how many cores are used within a single job.

PR #13381, opened 2018-02-16
Revert "Tag DiffEqDiffTools.jl v0.4.0 [https://github.com/JuliaDiffEq…

andreasnoack:
…/DiffEqDiffTools.jl/releases/tag/v0.4.0] (#13372)"

This reverts commit ffee6a377a62d3cf94634b27d77ec34e08ac4aaa.


ChrisRackauckas:
That's odd. All of the tests passed. These were the three from yesterday:

https://github.com/JuliaLang/METADATA.jl/pull/13371
https://github.com/JuliaLang/METADATA.jl/pull/13372
https://github.com/JuliaLang/METADATA.jl/pull/13373

The resolver probably broke so we can probably just remove the upper bounds...

andreasnoack:
There is a difference between each PR passing pairwise against `metadata-v2` and them passing jointly.

KristofferC:
Yes, when you tag a bunch of related packages in different PRs you would need to run the tests incrementally. Ideally, this would be done automatically by some bot but we don't have that.
I think we should do this if it brings back stability to METADATA and then get this tag in again when we figure out what the problem is.

ChrisRackauckas:
But we already know what the problem is. The upper bound makes the resolver have to choose between bounding DiffEqDiffTools, getting the latest version of the packages, or going to pre DiffEq v4.0. The resolver is broken around DiffEq v4.0, so it just stops doing anything. If we just get rid of that upper bound we should be fine. That's a quick fix for now which isn't too breaking to the packages anyways, though technically incorrect.

andreasnoack:
Superseded by https://github.com/JuliaLang/METADATA.jl/pull/13384

tkelman:
> Ideally, this would be done automatically by some bot but we don't have that

See https://github.com/barosl/homu

KristofferC:
When we looked at this before, the lack of AppVeyor support was blocking. Do you know if anything has changed in that area?

tkelman:
Appveyor matters for base julia, it doesn't matter for metadata

PR #13383, opened 2018-02-16
Revert "bound OrdinaryDiffEq for DiffEqDiffTools"

ChrisRackauckas:
Reverts JuliaLang/METADATA.jl#13371 . While technically this upper bound is needed to stop breakage in the Rosenbrock algorithms, it seems that the resolver has problems with more upper bounds so we should just allow it to happen.


ChrisRackauckas:
That had extra commits for some reason. I'll just do it manually.

PR #13384, opened 2018-02-16
remove upper bounds on ordinarydiffeq to fix resolver

ChrisRackauckas:


ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/13381

PR #13393, opened 2018-02-17
Release xtensor 0.0.11

SylvainCorlay:


SylvainCorlay:
The same build error occured for earlier versions and our cxxwrap dependency.

PR #13493, opened 2018-02-23
A few compat issues and drop 0.5 since it doesn't work anymore

quinnj:
cc: @aviks 

As far as I can tell, no METADATA travis checks are passing while we continue to try and check w/ 0.5, so this PR drops it.

### DIFF
--- a/.test/METADATA.jl
+++ b/.test/METADATA.jl
@@ -69,7 +69,7 @@ function juliaver_in_require(pkg, version; check=true)
                         error("$requires_file: oldest allowed julia version not specified (>= $minjuliaver needed)")
                     end
                 else
-                    juliaver = max(juliaver, convert(VersionNumber, tokens[2]))
+                    juliaver = max(juliaver, VersionNumber(tokens[2]))
                     hasjuliaver = true
                 end
                 if check && juliaver < minjuliaver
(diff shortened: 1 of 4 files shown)


omus:
Would be nice to enforce that new tags require Julia 0.6 or above as we aren't running tests on 0.5 anymore.

quinnj:
Yeah, we could do that. I wonder if attobot could help in enforcing this at all (i.e. fail to even PR if 0.5 is still the minimum), @simonbyrne, does attobot do checks like that?

andreasnoack:
@ararslan Isn't there some more steps required here if we are dropping 0.5 support?

quinnj:
Just pushed a commit to to require 0.6 as the minimum julia version.

simonbyrne:
It's not completely trivial, but we could add it to AttoBot: essentially just need a way to parse the REQUIRE file.

quinnj:
I think we should merge this as-is and figure out proper dropping of 0.5 later. Holding up the entire package ecosystem is not ok, IMO.

ararslan:
See https://github.com/JuliaLang/METADATA.jl/pull/12729. @aviks did not want to require 0.6 at a minimum in METADATA yet.

aviks:
Yes, I had disagreed with  that then, but our hands are forced now, I
suppose.

ararslan:
Okay. I've reopened https://github.com/JuliaLang/METADATA.jl/pull/12729, which will need some updating.

PR #13544, opened 2018-02-27
Pull request/daf300ac

sswatson:


andreasnoack:
Is it on purpose that you are releasing two different packages and two different tags for one of them with this PR? Please consider using https://github.com/attobot/attobot when making releases. Also, it appears that `ConformalMaps` doesn't pass tests. See http://juliarun-ci.s3.amazonaws.com/daf300ac98703f22144a4a7b6f88fb10ae83e5d5/ConformalMaps_reverse_dependency_tests.html

sswatson:
It was not particularly intentional that the two updates (one tag and one release) appeared on the same PR...I think this is what happens automatically if you do `Pkg.register("AsyPlots")`, `Pkg.tag("ConformalMaps")` and then `Pkg.publish()` or even `Pkg.publish("AsyPlots")`. (I'm at the instruction-following level with this process). I guess the problem with my workflow was not using Attobot to handle tagging. I went ahead and installed it, and I also updated the ConformalMaps tests. 

I guess I should close this PR and figure out how to open one with just AsyPlots? Thanks.

andreasnoack:
> I guess I should close this PR and figure out how to open one with just AsyPlots?

Yes. I think that would be the easiest thing to do.

PR #13545, opened 2018-02-27
Tag OptiMimi v0.0.4

jrising:
This update (amongst other changes) support Julia 0.6.  Also in the push is a tag for the last commit that supports Julia 0.4.


andreasnoack:
From the log:
```
ERROR: LoadError: New tag 0.0.4 of package OptiMimi requires julia 0.6.0, but version 0.0.1 of OptiMimi requires julia 0.4.0. Use a new minor package version when support for an old version of Julia is dropped.
```
Also, please consider using https://github.com/attobot/attobot when making releases.

jrising:
I'm withdrawing this and pushing two new PRs for OptiMimi v0.0.3 and v1.0.0.

PR #13551, opened 2018-02-27
Tag Pyramids v0.0.2

loganwilliams:


andreasnoack:
From the log:
```
ERROR: LoadError: New tag 0.0.2 of package Pyramids requires julia 0.6.0, but version 0.0.1 of Pyramids requires julia 0.4.0. Use a new minor package version when support for an old version of Julia is dropped.
```
Please consider using https://github.com/attobot/attobot for releases

PR #13595, opened 2018-03-01
Tag GR v0.27.0 [https://github.com/jheinen/GR.jl]

jheinen:
Diff vs v0.26.0: https://github.com/jheinen/GR.jl/compare/f8cf54180ce3eddf8d9d95c2d46498feb492e17c...aae8cd89307395927a11ce2abbdf584072e36f89


andreasnoack:
We have now bumped the minimum requirement to Julia 0.6
```
ERROR: LoadError: METADATA/GR/versions/0.27.0/requires: oldest allowed Julia version 0.5.0 too old (>= 0.6.0 needed)
```
Please also consider using https://github.com/attobot/attobot when releasing new versions

PR #13658, opened 2018-03-05
Tag Gaston v0.7.4 [https://github.com/mbaz/Gaston.jl]

mbaz:
Diff vs v0.7.3: https://github.com/mbaz/Gaston.jl/compare/48fc465620f8862d3d4c5cda1e393374388ebf04...e3174d7998f2dd1217885cd5af730485f0f00755


andreasnoack:
Please use https://github.com/attobot/attobot when making releases

mbaz:
@andreasnoack I thought attobot use was optional. Does it cause more work for you if I don't use it?

andreasnoack:
It easier to avoid mistakes at our end if people use Attobot and it is easier for package maintainers to use Attobot.

mbaz:
I don't think it buys me a lot personally, but if it makes things easier at your end, then I'll use it. I would also recommend changing the documentation; right now, attobot is just suggested.

andreasnoack:
> I don't think it buys me a lot personally

Have you tried to use it yet? It really is convenient.

mbaz:
I just took a quick look at the docs. I'm sure I'll end it loving it :)

PR #13659, opened 2018-03-05
Tag DriftDiffusion v0.1.0 [https://github.com/agbondy/DriftDiffusion.jl]

agbondy:
Diff vs v0.0.1: https://github.com/agbondy/DriftDiffusion.jl/compare/e8425940807d4f842edf0fc74b7d1a9d8c3cdce1...1604172e2d910e91c033b6bbb4b36d509d3d539a


andreasnoack:
Tests fail. From the log:
```
ERROR: LoadError: LoadError: ArgumentError: Module MAT not found in current path.
Run `Pkg.add("MAT")` to install the MAT package.
```
so you probably need to add MAT to the REQUIRE file.

andreasnoack:
Also, please use https://github.com/attobot/attobot for releases

agbondy:
OK, thanks for the feedback. Attobot is not mentioned in the Julia documentation about package development.

andreasnoack:
It is mentioned here https://docs.julialang.org/en/stable/manual/packages/#Tagging-and-Publishing-Your-Package-1 but I agree that the recommendation could be clearer. The package documentation will soon be updated as we move to Pkg3. In any case, attobot is really useful. In particular in a case like this, where you need to make changes to a release PR.

andreasnoack:
Superseded #13686

PR #13768, opened 2018-03-11
Update package URLs after move to JuliaRobotics organization.

tkoolen:
@Affie (and @dehann), I took the liberty of changing the AprilTags.jl URL as well.


dehann:
Thanks @tkoolen !

PR #13947, opened 2018-03-22
Fixedeffectmodels

matthieugomez:


andreasnoack:
Is it on purpose that you tag two different versions? Please consider using https://github.com/attobot/attobot when making releases.

matthieugomez:
The v0.5.0 commit had already been merged with metada. I think I wrote this new commit for v0.5.1 in the same branch as the one with a commit to tag v0.5.0. Not sure why the past commit shows up since it is already merged (https://github.com/JuliaLang/METADATA.jl/commit/c369d5011f4dac34e87e8fd7957031395e29612e)

PR #13970, opened 2018-03-25
Register & Tag mmtf.jl v0.0.1

joels94:


kmsquire:
Hi, @joels94, thanks for submitting this package. Generally, registered Julia packages follow certain naming conventions.  See this link:

https://docs.julialang.org/en/stable/manual/packages/#Guidelines-for-naming-a-package-1

joels94:
Hi @kmsquire, thanks for the suggestion. How about MMTF.jl? MMTF stands for MacroMolecular Transmission Format. Its a file format used to store data in binary format. I have written a package to parse and write MMTF file format. So, like other file format packages (BSON.jl, JSON.jl, CSV.jl, etc), can we have it as MMTF.jl?

bicycle1885:
I think MMTF.jl is much better.

bicycle1885:
But, why are you going to register it as a separated package? I think it can be a submodule of BioStructures.jl.

joels94:
Thanks, @bicycle1885 for the suggestion! In that case, let me close this pull request and create a new PR once I have renamed the package.

joels94:
@bicycle1885 I was having a discussion with Joe Greener about having it as a separate package similar to other MMTF libraries such as [mmtf-python](https://github.com/rcsb/mmtf-python), [mmtf-javascript](https://github.com/rcsb/mmtf-javascript), etc. Maybe we can further discuss it in an issue in our repo https://github.com/BioJulia/mmtf.jl.

jgreener64:
I feel it makes more sense as its own package in line with similar packages from other languages. BioStructures.jl can then import and extend the decoding/encoding functionality.

PR #14002, opened 2018-03-27
Tag OnlineStatsBase.jl v0.7.0

attobot:
Repository: [joshday/OnlineStatsBase.jl](https://github.com/joshday/OnlineStatsBase.jl)
Release: [v0.7.0](https://github.com/joshday/OnlineStatsBase.jl/releases/tag/v0.7.0)
Diff: [vs v0.6.1](https://github.com/joshday/OnlineStatsBase.jl/compare/1fcbc8ede10b41311ccab0bb5cb672d0717ee693...75bf29f67cc780609ddc210ce0edcbfb42d98ddd)
`requires` vs v0.6.1: 
```diff
--- 0.6.1/requires
+++ 0.7.0/requires
@@ -1,3 +1,4 @@
 julia 0.6
 LearnBase
-NamedTuples+NamedTuples
+Compat
```
cc: @joshday

Please make sure that:


joshday:
This breaks OnlineStats, IndexedTables, and JuliaDB.  PRs for fixes/releases are incoming in the next few days.

andreasnoack:
Should probably add some upper bounds on those.

joshday:
I'm hesitant to add upper bounds to IndexedTables and JuliaDB.  After this round of releases, both packages will be future proof against changes to OnlineStats[Base].  The integration is much simpler now and is based around three methods (`fit!`, `merge`, and `copy`) that every OnlineStat has. I think upper bounds will make releases harder and add version conflicts that aren't needed.

OnlineStats should probably have them, but it was discussed at one point that upper bounds would need to be added to all previous OnlineStats versions in order for the resolver to work correctly.

shashi:
> I'm hesitant to add upper bounds to IndexedTables and JuliaDB. After this round of releases, both packages will be future proof against changes to OnlineStats[Base]. ... I think upper bounds will make releases harder and add version conflicts that aren't needed.

Version conflict is probably more informative than having packages throw mysterious `UndefVarError`s. Also version conflicts are unlikely once the dependent packages are also released. We should add upper bounds for existing versions. Otherwise rolling back to previous versions of JuliaDB / IndexedTables will break them.

joshday:
This can be closed.

joshday:
Sorry, ignore that.  This is good to go.

PR #14009, opened 2018-03-27
Tag Reduce v0.3.1

chakravala:
Critical parser bug fix + additional `sum`, `prod`, `horner` methods  https://github.com/chakravala/Reduce.jl

Diff vs v0.3.0: https://github.com/chakravala/Reduce.jl/compare/8313a3a5f236d8541ae41876092cf64174ef371e...eaa70267ef8a67cbc2cf69b1971704f834e95c09


andreasnoack:
It looks like tests are failing. Please also consider using https://github.com/attobot/attobot for package releases.

chakravala:
Tests are all working fine on my end, I have restarted the failed tests, they should all be working.

THe problem is, I have decided to switch some `String` with `IOBuffer` and this makes the `Reduce` package unstable for some reason. Most times the test will pass, but there is a statistical chance that a test will fail after I switched to `IOBuffer` apparently. I was trying to improve the speed and performance by minimizing the number of `String` used.

So the tests typically pass, but sometimes, there seems to be a statistical chance of the `IOBuffer` making a mistake. I'm not sure why `IOBuffer` is so much less reliable than `String`.

I had this issue at the beginning too, although I completely eliminated it after much testing. Now, the issue of a statistical chance of failure seems back, since I tried switching to `IOBuffer`.

It might take some time to figure out how to completely eliminate it from this new version, but 99% of the time everything should work fine, really wish I  knew what makes `IOBuffer` unstable.

I don't want to use attobot, I'd like to do things manually myself.

chakravala:
Okay, the instability seems to go away when you set `Reduce.SubCall(false)`, so the source of the statistical problem is the `Reduce.SubReplace` function.

chakravala:
Tests are passing now...

chakravala:
It might be that the statistical likelihood of failure is due to making too many calls to reduce in a very short period of time. In that case, the problem would be fixed by puting a wait timer in for each call, but that would make things a bit slower on each call...

chakravala:
UPDATE:

the `master` branch now has a fix that solves the statistical issue with the clogged pipe by detecting and circumventing the issue, so the issue of having any chance of failed tests should go away now.

PR #14003, opened 2018-03-27
Tag OnlineStats.jl v0.17.0

attobot:
Repository: [joshday/OnlineStats.jl](https://github.com/joshday/OnlineStats.jl)
Release: [v0.17.0](https://github.com/joshday/OnlineStats.jl/releases/tag/v0.17.0)
Travis: [Travis Build Status](https://travis-ci.org/joshday/OnlineStats.jl/builds/358979112)
Diff: [vs v0.16.0](https://github.com/joshday/OnlineStats.jl/compare/0b9a268cecfef098ba4822cca0d5f3132991b8e5...c91622399c9a2b893928bdd9797e17c0894b6a6a)
`requires` vs v0.16.0: no changes
cc: @joshday

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #14024, opened 2018-03-29
OnlineStatsBase 0.7.0 and OnlineStats 0.17.0 with upper-bounds in dependents

shashi:
subsumes #14002 and #14003


andreasnoack:
@joshday Let me know if you are okay with this and I'll merge.

joshday:
I'm going to fight a little bit more against upper bounds for OnlineStats[Base] in IndexedTables and JuliaDB.  After this round of changes, it will be nearly impossible for OnlineStats[Base] to break either package. 

If we add upper bounds, any new release of OnlineStats can't be used (even though the code works) until new releases of IndexedTables/JuliaDB are out.  Committing myself to an extra two package releases for every OnlineStats release is unnecessary added work.

I propose I cherry pick @shashi's `upper bound OnlineStatsBase in OnlineStats` commit but leave off the bounds in IndexedTables/JuliaDB.  Thoughts?

joshday:
This can be closed.

PR #14038, opened 2018-03-30
Register new Package QuConst.jl

Roger-luo:
A Package for physical constants


garrison:
Why is it called QuConst?  The README does not specify.

Roger-luo:
@garrison I was a sub module for something I'm developing: QMTK.jl, but I cannot rename it to Const.jl
If there is better.

Roger-luo:
I'll rename it and re-submit a PR later.

PR #14076, opened 2018-04-02
Register Constants [https://github.com/Roger-luo/Constants.jl.git]

Roger-luo:
v0.0.1


Roger-luo:
@kmsquire Although currently, it only includes physical constants but I have included a set of types that could handle arbitrary constants, I just don't know what else people would like to use at the moment. But I'm happy to include all possible constants in other situation in future versions.

I guess it would be quite inconvenient to separate such simple package (for constants) into physical and other type of constants. 

What do you think?

kmsquire:
I honestly don't have a good feel, since I don't really have need for most of these (or most other) constants myself in every day work.  

The only other constants I'm aware of in Julia are mathematical constants like π, e, γ, etc.  These few are irrational and are built in, but do you plan to handle these types of constants in your package as well?  You might know this, but they're somewhat special, since they're (almost?) [arbitrary precision](https://julialang.org/blog/2017/03/piday) (also [here](https://github.com/JuliaLang/julia/issues/26550)).  

My inclination would still be to focus on physical constants (and name the package PhysicalConstants.jl), since I think that still covers a large set of possible constants of interest, and makes it easy understand what the package contains.

Roger-luo:
I see. How about rename it to **PhyConsts.jl**?

kmsquire:
Looking at https://pkg.julialang.org, there are two physics-related packages which start or contain `Phys`, but none (that I found) that just contain `Phy`. 

How about `PhysConsts.jl` or `PhysConstants.jl`, for consistency with the other packages?  

(To be clear, these are just suggestions.  If you strongly prefer `PhyConsts.jl`, that would probably be fine.)

When you resubmit, you should submit the version with passing tests.  Cheers!

Roger-luo:
Thanks, I'll rename it to PhysConsts.jl. I'll re-submit a PR later.

garrison:
It might make sense to leave this PR open for a bit while the package name bikeshedding is performed, rather than open a new PR each time a new name is suggested (see also previous discussion in #14038).

PR #14054, opened 2018-03-31
Tag LineSearches.jl v4.0.0

attobot:
Repository: [JuliaNLSolvers/LineSearches.jl](https://github.com/JuliaNLSolvers/LineSearches.jl)
Release: [v4.0.0](https://github.com/JuliaNLSolvers/LineSearches.jl/releases/tag/v4.0.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/LineSearches.jl/builds/360562668)
Diff: [vs v3.2.5](https://github.com/JuliaNLSolvers/LineSearches.jl/compare/6873fa5a2264a40f6e5b9fa2df2681a2847eec72...b58f286a484a83f7963d7e017cecbd2bf6951519)
`requires` vs v3.2.5: no changes
cc: @pkofod

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #14123, opened 2018-04-03
Upper bound Optim

pkofod:
ref #14112 #14054

this time actually commit the Optim changes...

The commits look weird, but as long as this is squashed, it should be good.


pkofod:
.. yeah

Issue #10873, opened 2017-08-23
Pkg.update() gives errors for older versions of Julia


TheOriginalSoni:
I am currently running Julia Version 0.4.5 x86_64-linux-gnu.

When I run Pkg.update() on it, I get the following error - 
```

INFO: Initializing package repository /home/username/.julia/v0.4
INFO: Cloning METADATA from git://github.com/JuliaLang/METADATA.jl
fatal: unable to connect to github.com:
github.com[0: 192.30.253.112]: errno=Connection refused
github.com[1: 192.30.253.113]: errno=Connection refused

ERROR: failed process: Process(`git clone -q -b metadata-v2 git://github.com/JuliaLang/METADATA.jl METADATA`, ProcessExited(128)) [128]
 in anonymous at ./pkg/dir.jl:52
```

What is the fix to this error?

P.S. I got this version of Julia after running sudo apt-get install julia on Ubuntu (16.04)

fredrikekre:
Not a direct fix to the problem, but you really shouldn't use such an old (and now unmaintained) release. You can download the latest 0.6.0 release here: https://julialang.org/downloads/

TheOriginalSoni:
That I realised after a couple hours. But in that case, how does one go about changing what version of Julia one gets when they do `apt-get install julia`? That is how I installed Julia 0.4 on my system, and I believe a good number of people might use this way to test the language

fredrikekre:
The recommended way to get julia is to download [generic binaries](https://julialang.org/downloads/) (or compile yourself). Using `apt-get` is not recommended, see https://github.com/JuliaLang/julialang.github.com/pull/475 for some more discussion.

bjarthur:
perhaps not the same problem, but related possibly, is that package CI tests are failing on julia 0.5: 

```
INFO: Cloning METADATA from https://github.com/JuliaLang/METADATA.jl
ERROR: GitError(Code:ERROR, Class:Net, SSL error: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version)
 in macro expansion at ./libgit2/error.jl:99 [inlined]
 in clone(::String, ::String, ::Base.LibGit2.CloneOptions) at ./libgit2/repository.jl:191
 in #clone#112(::String, ::Bool, ::Ptr{Void}, ::Nullable{Base.LibGit2.AbstractCredentials}, ::Function, ::String, ::String) at ./libgit2/libgit2.jl:327
[...]

PR #14175, opened 2018-04-07
Tag EEG v0.2.1 [https://github.com/codles/EEG.jl]

rob-luke:
Diff vs v0.2.0: https://github.com/codles/EEG.jl/compare/49aeb6759e3c0ea822a2515279f5d7395e1d79ee...49aeb6759e3c0ea822a2515279f5d7395e1d79ee


andreasnoack:
From the log
```julia
ERROR: LoadError: METADATA/EEG/versions/0.2.1/requires: oldest allowed julia version 0.5.0- too old (>= 0.6.0 needed)
```
Remember that you'll need to bump the minor version when changing the Julia requirement.

It looks like tests are failing and there are many deprecation warnings in the test output.

Also, please consider using https://github.com/attobot/attobot for releases.

codles:
Hi @andreasnoack ,

Thanks for the advice. I will use atoobot from now on.

I wanted to release a final version that I know works with 0.5 and then commit the next version which is for >= 0.6 (in a day or two). I would like to have a final version for 0.5 as I have released papers using this version and would like to be able to go back and reproduce the results easily.

codles:
Also the tests fail due to METADATA git error. Which I believe is reported here https://github.com/JuliaLang/METADATA.jl/issues/10873

andreasnoack:
It is no longer possible to tag releases for 0.5. There is more information in https://discourse.julialang.org/t/errors-for-git-pkg/9351. In short, with Julia 0.5 on Linux, people won't be able to install any packages from GitHub anyway so it wouldn't help much to make a final release for 0.5.

PR #14222, opened 2018-04-11
Register new package Systems.jl v0.1.0

attobot:
Repository: [JuliaReach/Systems.jl](https://github.com/JuliaReach/Systems.jl)
Release: [v0.1.0](https://github.com/JuliaReach/Systems.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaReach/Systems.jl/builds/365107439)
cc: @mforets

Please make sure that:

@mforets This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


mforets:
The package itself doesn't require `LazySets`, but we have some unit tests which use that package (see #14191). In Travis we workaround this with

```julia
...
script:
  - julia -e 'Pkg.clone(pwd()); Pkg.build("Systems"); Pkg.clone("https://github.com/JuliaReach/LazySets.jl"); Pkg.test("Systems"; coverage=true)'
...
```

What should i do here (if i don't want to add the `LazySets` dependency)? Remove those tests? Use a `@requires` block?

dpsanders:
Use the `test/REQUIRE` file (i.e. in the `test` directory).

Is the package name rather too general? (Although supposedly this won't be an issue with Pkg3?)

mforets:
> Use the test/REQUIRE file (i.e. in the test directory).

Great, I'll try that! thanks.

> Is the package name rather too general? (Although supposedly this won't be an issue with Pkg3?)

Yep, we chose it general because we just wanted to give the bare bones to work with some mathematical systems. But there exist related packages some of them more specialized or that offer more functionality, eg. [see this list](https://juliareach.github.io/Systems.jl/latest/about.html#Related-projects-1). If `Systems.jl` is a too general name for a community package we could change it, or maybe submitting to here was not a good idea, i don't know :)

andreasnoack:
A specific name would be better.

blegat:
Would a name like SystemTypes be specific enough ? The generality in the name is kind of on purpose, it is meant to be the definition of a common interface for packages working with systems and a few types for common systems.

dpsanders:
The word "Systems" means many different things in different domains.
Without looking at the package, I have no idea what the context is from the name.

blegat:
What about "MathSystems" or "MathematicalSystems" ?

andreasnoack:
`MathematicalSystems` sounds reasonable to me. @dpsanders thoughts?

mforets:
> Use the test/REQUIRE file (i.e. in the test directory).

Done, see  #14598.

> MathematicalSystems sounds reasonable to me. @dpsanders thoughts?

Ok. If you don't disagree, i will rename the package to `MathematicalSystems.jl`.

mforets:
This PR should be closed, see now PR #14608.

PR #14234, opened 2018-04-12
Register new package Morton.jl

bjarthur:

PR #14235, opened 2018-04-12
Tag Morton.jl v0.0.1

bjarthur:
i tagged b/c travis [gave an error](https://github.com/JuliaLang/METADATA.jl/pull/14234) because i hadn't.  i don't understand why one must tag before registering because [the docs say to register, publish, tag, and publish again](https://github.com/JuliaLang/julia/blob/master/doc/src/manual/packages.md#tagging-and-publishing-your-package).


aviks:
Our tests have required tagging before publishing for a while now. maybe the docs need updating (I'd assumed that was done). In general, attobot does all of this correctly now.

PR #14238, opened 2018-04-12
Bound BinaryProvider versions on Ipopt

mlubin:
Recent changes appear to have broken the package (https://github.com/JuliaOpt/Ipopt.jl/issues/110). This should prevent further user-visible breakage until we can address the underlying issue. I picked 0.2.5 as the version of BinaryProvided that was released when we last tagged Ipopt.
@staticfloat


mlubin:
Doesn't seem to be as widespread as I feared, so I'll close this and let the discussion continue at https://github.com/JuliaOpt/Ipopt.jl/issues/110.

PR #14404, opened 2018-04-25
Tag StatsModels.jl v0.2.4

attobot:
Repository: [JuliaStats/StatsModels.jl](https://github.com/JuliaStats/StatsModels.jl)
Release: [v0.2.4](https://github.com/JuliaStats/StatsModels.jl/releases/tag/v0.2.4)
Travis: [Travis Build Status](https://travis-ci.org/JuliaStats/StatsModels.jl/builds/371205664)
Diff: [vs v0.2.3](https://github.com/JuliaStats/StatsModels.jl/compare/9bbc2a7404e75250f37f854d16233f0b0ce9bf15...dc070dc73f2de92e6c9816c3b696388307295031)
`requires` vs v0.2.3: 
```diff
--- 0.2.3/requires
+++ 0.2.4/requires
@@ -2,3 +2,4 @@
 DataFrames 0.11.5
 StatsBase 0.20.1
 Compat 0.63
+ArgCheck
```
cc: @kleinschmidt

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #14429, opened 2018-04-27
StatsModels/v0.2.4 (with upper bounds)

kleinschmidt:
add upper bounds to StatsModels dependents and tag v0.2.4 (supersedes #14404)


nalimilan:
Thanks! Just to be sure, have you added a dummy dependency on StatsModels to older versions of packages whose recent versions require StatsModels? That's not ideal, but if we don't do that the package manager will downgrade packages to very old versions to satisfy requirements.

PR #14522, opened 2018-05-04
Revert "Tag RecipesBase.jl v0.3.0 

KristofferC:
This reverts commit 6e4c4411dba098643787c330a999032f8127be53.

This commit doesn't exist anymore so the release is useless and causing problems on julia 0.7 where this release is "active".


KristofferC:
cc @mkborregaard

mkborregaard:
I deleted that tag in error, sorry for my mess. Could I just republish a v0.3.0 release?

KristofferC:
The tag doesn't matter, it is what is registered in METADATA that does. We either delete this one or you just tag a new v0.3.1 release and that would probably work just as well (assuming no one upper bounds to 0.3.0 but that won't happen).

PR #14523, opened 2018-05-04
Register MathOptInterfaceMosek

ulfworsoe:
[https://github.com/JuliaOpt/MathOptInterfaceMosek.jl.git]

### DIFF
--- a/MathOptInterfaceMosek/url
+++ b/MathOptInterfaceMosek/url
@@ -1,0 +1,1 @@
+https://github.com/JuliaOpt/MathOptInterfaceMosek.jl.git


ulfworsoe:
Tests failing: This is a bit tricky. Tests fails because no MOSEK license is available.

andreasnoack:
...and please add a comment here when opening [another PR](https://github.com/JuliaLang/METADATA.jl/pull/14525/files) for the same package.

ulfworsoe:
Thank you and sorry about the mess! I intent to do both, but registering the package caused me some problems.

andreasnoack:
So do you still want version 0.0.1 from https://github.com/JuliaLang/METADATA.jl/pull/14525/files to be available in METADATA?

ulfworsoe:
Yes. But the tests are not currently succeeding. I have a solution, but I need to test it, then I will make a new release through attobot.

PR #14549, opened 2018-05-06
Tag CxxWrap.jl v0.6.0

attobot:
Repository: [JuliaInterop/CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl)
Release: [v0.6.0](https://github.com/JuliaInterop/CxxWrap.jl/releases/tag/v0.6.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaInterop/CxxWrap.jl/builds/375569089)
Diff: [vs v0.5.0](https://github.com/JuliaInterop/CxxWrap.jl/compare/9b5be121ce09047add2309bdcfe1797e582b6322...64735e756b9ccfd38717a009e846816af0d077fc)
`requires` vs v0.5.0: no changes
cc: @barche

Please make sure that:


SylvainCorlay:
@barche thanks for the heads up!

SylvainCorlay:
We are about to release a backward-incompatible version of xtensor-core, so the next Xtensor.jl will get both the new cxxwrap and xtensor 0.16.

andreasnoack:
Meanwhile, Xtensor would need upper bounds here in METADATA

SylvainCorlay:
I guess that we should add "0.6-" and tag another version of Xtensor.jl


On May 7, 2018 11:49, "Andreas Noack" <notifications@github.com> wrote:

Meanwhile, Xtensor would need upper bounds here in METADATA

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaLang/METADATA.jl/pull/14549#issuecomment-387000976>,
or mute the thread

andreasnoack:
The more urgent part is to add the bounds here in METADATA to all the existing versions. You can then just fix you package for the new version of CxxWrap and bump the lower bound in the package instead.

andreasnoack:
@barche Something weird is happening when I try to rerun the tests. Would you mind deleting and reapplying the tag? It might help?

barche:
@andreasnoack OK, done. Should the upper bound of Xtensor also be updated in this PR?

andreasnoack:
> Should the upper bound of Xtensor also be updated in this PR?

No, it shouldn't be needed.

Do you know what could cause this CMake error http://juliarun-ci.s3.amazonaws.com/1d1ff377cd4375f76fc3e3dc860f6daa68cfcc7d/Xtensor_with_pull_request_of_CxxWrap_on_julia_0_6.log? I'm a bit confused why this would happen for the "new" but not the "current" build since they are supposed to be using the same version of CxxWrap.

SylvainCorlay:
@wolfv I think that there is a "-" missing in the xtensor.jl metadata (in my understanding of the syntax).

wolfv:
@andreasnoack specifically said that the `-` is not needed and reviewed the other PR. 
Does it need to be changed?

andreasnoack:
Yes. The bounds should be fine. Notice in the top of the logs that the installed versions of the packages are the same.

SylvainCorlay:
Well, the new version of cxxwrap that is backward incompatible has not been
released yet right?

andreasnoack:
> Well, the new version of cxxwrap that is backward incompatible has not been released yet right?

Correct. That is what this PR is about.

SylvainCorlay:
I meant it is expected that xtensor does not pick it up...

andreasnoack:
> I meant it is expected that xtensor does not pick it up...

I don't follow. There are currently upper bounds on xtensor in METADATA which prevent that this new version of CxxWrap is getting installed as part of our CIBot testing. If the bounds weren't there, the test of xtensor in CIBot would use CxxWrap v0.6.0 which this release is about. The thing I don't get is why the CIBot tests are currently failing since version v0.5.0 of CxxWrap is being used. In particular, because the "Current" test is using passing. It might be somethign spurious with CIBot so if you could confirm that xtensor works with CxxWrap v0.5.0, we can go ahead and release the new version of CxxWrap since it should affect xtensor (because of the upper bounds).

wolfv:
Ok, I just had a look at the CI output (i didn't realize that we get to see xtensor CI output here, very cool!). Problem seems to be that cmake doesn't find `FindJlCxx.cmake` or similar for that package, but the version number is picked correctly. I'll investigate.

barche:
It's actually taking the new CxxWrap for Xtensor, there's a line referring to JlCxx version v0.2.1, which is the new one.

jstrube:
For what it's worth, I'm not sure why the tests for LCIO are failing, but I've just updated master to work with CxxWrap 0.6, and it looks fine from my perspective. I'm going to push a release as soon as this one is out.

andreasnoack:
@nkottary Are you able to tell what is going on with CIBot? Initially, it looks like "new" and "current" are using the same package versions but at some point, they diverge.

nkottary:
LCIO.jl has upper bound for CxxWrap see here: [requires](https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/LCIO/versions/0.8.0/requires)

jstrube:
That's because I haven't made a release,  yet. I can do that later today. I
wanted to wait until this one is out,  because the new one will need a new
lower bound. Will do when I get back to my computer.

andreasnoack:
@nkottary But `XTensor` has an upper bound at 0.6 for CxxWrap, see https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/Xtensor/versions/0.1.0/requires, so why is it able to install it anyway?

barche:
I don't understand where this line in the [log](http://juliarun-ci.s3.amazonaws.com/1d1ff377cd4375f76fc3e3dc860f6daa68cfcc7d/Xtensor_with_pull_request_of_CxxWrap_on_julia_0_6.log) comes from:

```
Archive:  /home/jrun/Xtensor/.julia/v0.6/CxxWrap/deps/downloads/v0.2.1.zip
```

That is part of CxxWrap 0.6, while at the top it says it's getting CxxWrap 0.5.

SylvainCorlay:
> > Should the upper bound of Xtensor also be updated in this PR?
>
> No, it shouldn't be needed.
> 
> Do you know what could cause this CMake error http://juliarun-ci.s3.amazonaws.com/1d1ff377cd4375f76fc3e3dc860f6daa68cfcc7d/Xtensor_with_pull_request_of_CxxWrap_on_julia_0_6.log? I'm a bit confused why this would happen for the "new" but not the "current" build since they are supposed to be using the same version of CxxWrap.****

@andreasnoack It appears that CxxWrap 0.6 is being picked up instead of 0.5.

So the upper bound added for CxxWrap in xtensor's metadata does not appear to have an effect. Are you sure that the upper bound should not have a `-` sign?

andreasnoack:
Yes. See https://docs.julialang.org/en/stable/manual/packages/#Requirements-Specification-1. Adding the `-` shouldn't make a difference. We add these upper bounds (without `-`) all the time and it usually works. I don't know what is going on here but I suspect there is some kind of issue with CIBot. Hopefully, @nkottary will be able to figure out what is going on here.

andreasnoack:
I've just tried to merge this locally and the upper bounds works as expected so this is an issue with CIBot. Hene, it shouldn't delay this PR any longer and I'll go ahead an merge.

PR #14571, opened 2018-05-08
Add upper bound to xtensor CXXWrap version

wolfv:
Is it ok to add this upper bound now, or do we have to release a new version of xtensor-julia?

### DIFF
--- a/Xtensor/versions/0.0.1/requires
+++ b/Xtensor/versions/0.0.1/requires
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.4.2
 Compat 0.21
-CxxWrap 0.4
+CxxWrap 0.4 0.5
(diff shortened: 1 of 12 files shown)


andreasnoack:
For now, you don't have to tag a new release of XTensor. Adding upper bounds here in METADATA is just fine. However, you'll have to do it for all versions registered in METADATA. Otherwise, the package manager will just install a lower version of XTensor instead of holding back `CxxWrap`. Please also remove the `-` from `0.6` since the right endpoints of the interval are already excluded.

wolfv:
Hi andreas, thanks for your help on this. 

I just added endpoints to all versions. Hope that's sufficient

SylvainCorlay:
THanks @wolfv & @andreasnoack for fixing this!

PR #14580, opened 2018-05-09
Pull request/5a31489f

reesepathak:
Introducing PSPlot.jl, a package will allows you to access psfrag plotting functionality in Julia, with the convenience of plotting in PyPlot. 

Currently only supports OS X due to a binary dependency (Inkscape). Hope to add additional unix support and then windows support in the future.


reesepathak:
Importantly, note that JuliaCIBot fails because of non-osx systems. Currently this package only supports OS X, as mentioned in the original pull request.

andreasnoack:
In the future, please consider using https://github.com/attobot/attobot when releasing new versions of this package or registering another one.

reesepathak:
Yes, I will have to look into that. Thanks!

reesepathak:
@andreasnoack What else do I need to do for this package to be accepted?

andreasnoack:
From my side, it is good to go but we usually let package registration PRs stay open a couple of days to allow people to comment.

PR #14584, opened 2018-05-09
Register Blobs

jamii:


jamii:
> Looks like tests are failing.

The tests can't be run with `--check-bounds=yes` because I test that `@inbounds` works. Is there a way to disable that option? It doesn't seem to appear explicitly in the 

> It also looks like you are registering several versions of the package.

This is a rename of ManualMemory (as discussed [here](https://discourse.julialang.org/t/renaming-packages/10433)) plus a new minor version v0.2.0.

> Please consider using https://github.com/attobot/attobot for future releases and package registrations.

Ok.

jamii:
I didn't find a way to change to CI options so I've just added a runtime check and disabled those tests unless `--check-bounds=default`

PR #14613, opened 2018-05-11
Add upper bounds on Compat in StatsBase to avoid stderr error

andreasnoack:


andreasnoack:
It is necessary to remove version 0.21.0 since it had a lower bound on Compat at 0.61.0 but an upper bound at 0.55 is required to avoid the issue with ambuity between `stderr` in `StatsBase` and in `Compat`

nalimilan:
Do we really need these upper bounds, or is it just to avoid the annoyance of having to fully qualify functions? I'm asking because @ChrisRackauckas noted the dependency resolver can completely break because of these upper bounds.

ChrisRackauckas:
I did find out about this because the dependency resolver broke in my case, with the curious fact that Compat downgraded which is very rare and a quick search brought me to here.  The resolver couldn't find out that it could just upgrade Compat and all stats packages together to jump over it, and instead got stuck with the old versions of some things and failed to do any Pkg operation (erroring due to the resolver).

nalimilan:
OK, looks like there's no good solution to this. It seems we should recommend explicitly importing functions you depend on, as it's too easy for any package to break its dependencies just by adding a function which creates a name conflict.

PR #14622, opened 2018-05-12
Move InternedStrings.jl to JuliaString Org

oxinabox:
I transferred the Repo to the Org,
so now I need to update this URL before I can publish anymore releases


oxinabox:
https://github.com/JuliaString/InternedStrings.jl/issues/17

PR #14633, opened 2018-05-13
Register ImageHistogram [https://github.com/t37w/ImageHistogram.jl.git]

t37w:
Hello,
following the process description within the julia documentation I want to register this package.
The goal is to create 2D and 3D gray and colored histograms and plot them.
Cheers,
Thomas

### DIFF
--- a/ImageHistogram/url
+++ b/ImageHistogram/url
@@ -1,0 +1,1 @@
+https://github.com/t37w/ImageHistogram.jl.git


t37w:
Anybody an idea how to get out of the deadlock of "travis CI build failed with empty details".
I have no idea how to step forward.

andreasnoack:
The easy solution is to use https://github.com/attobot/attobot for the releases

t37w:
Thank you @andreasnoack for the hint.
Is this possible at this state in the process of initial registration ?
By the way, I followed the process description available for julia stable and latest.  It would be good if the preferred use of `attobot`  would be mentioned there.

andreasnoack:
> Is this possible at this state in the process of initial registration ?

Yes. Just enable Attobot for your repository and make a release on GitHub. If the package isn't registered then Attobot will handle that.

>  It would be good if the preferred use of attobot would be mentioned there.

It is mentioned as a "tip" but we should probably make it clear that Attobot is the preferred way of making releases.

t37w:
Using Attobot, package registration succeeded now.
Updating the related documentation in Julia would avoid such a discussion.  Especially as there is a detailed description about using PgkDev.
Nevetheless, @andreasnoack thank you.

I submit this comment using the "close and comment" button, assuming this is the correct step to get this pull request out of the list of open pull.

Cheers,
Thomas

PR #14638, opened 2018-05-14
Register SMSSVD [https://github.com/rasmushenningsson/SMSSVD.jl.git]:… 

rasmushenningsson:
… v0.0.1


andreasnoack:
We usually recommend that acronyms are avoided, see https://docs.julialang.org/en/stable/manual/packages/#Guidelines-for-naming-a-package-1.

Please also consider using https://github.com/attobot/attobot

rasmushenningsson:
The name refers to a method presented in the paper "SMSSVD - SubMatrix Selection Singular Value Decomposition" (https://arxiv.org/abs/1710.08144).
Everyone I have discussed with refers to it as SMSSVD, and that's why I choose that name.
I could change it to SubMatrixSelectionSVD - do you think that is better?

andreasnoack:
> Everyone I have discussed with refers to it as SMSSVD

The reason we have the guideline is to help people who are *not* familiar with the acronym. Hence, I think that `SubMatrixSelectionSVD` is a better name.

rasmushenningsson:
That makes sense. I'll remove this one and create a new pull request.

PR #14694, opened 2018-05-17
Add upper bound to NLsolve

pkofod:
This was a mistake. The upper bounds was only present in METADATA and not the repo as such. It should have been there all along.


pkofod:
Is this the way to do this, or should I tag a new version of NLsolve that reintroduces the bound?

andreasnoack:
This is the way to do it but as long as master of `NLsolve` needs the upper bound, it should also be added to the REQUIRE file.

PR #14777, opened 2018-05-22
Register new package TracyWidom.jl v0.0.2

AustenLamacraft:
Package providing the Tracy-Widom distribution from random matrix theory.


nkottary:
The url is incorrect and so tests are failing. It should be `https://github.com/AustenLamacraft/TracyWidom.jl`

andreasnoack:
Have you considered contributing this to the https://github.com/JuliaMath/RandomMatrices.jl/blob/master/src/densities/TracyWidom.jl package instead?

AustenLamacraft:
Will ask

andreasnoack:
Superseded by https://github.com/JuliaMath/RandomMatrices.jl/pull/44

PR #14831, opened 2018-05-24
update URL for new organization

Datseris:
Otherwise attobot cannot tag a release in metadata...


Datseris:
See https://github.com/JuliaMusic/MIDI.jl/issues/97

PR #14835, opened 2018-05-24
Register new package NURVG.jl v0.0.1

attobot:
Repository: [awllee/NURVG.jl](https://github.com/awllee/NURVG.jl)
Release: [v0.0.1](https://github.com/awllee/NURVG.jl/releases/tag/v0.0.1)
cc: @awllee

Please make sure that:

@awllee This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


andreasnoack:
Acronyms are generally discouraged. See https://docs.julialang.org/en/stable/manual/packages/#Guidelines-for-naming-a-package-1. Also, have you considered contributing this to `Distributions` instead of creating a separate package. I believe your methods for Gamma and Beta are already the default in `Distributions`.

awllee:
I'm happy to rename the package, if necessary. The resulting non-acronym version would be quite long, which is I suppose not a big deal. I did see that there are many existing packages that are acronyms.

The motivation for this package differs from what is provided by Distributions, as far as I can tell, in a few ways that are important for the code I'm writing that uses these routines:

1.  As far as I can see Distributions favours creating a distribution object dist and then calling rand(dist) which adds unnecessary overhead when sampling many times from distributions with different parameters.

2. Distributions uses calls to RMath to generate samples for these routines, which precludes the provision of user-specified AbstractRNG inputs. This is a major issue for writing multi-threaded code in which random number generation takes a significant amount of time.

The algorithms for gamma and beta are different to the R routines, as is the binomial algorithm, which are called by Distributions.

I would also be happy for some of the code to be incorporated in Distributions eventually, but I'm not clear on where the developers of that package are w.r.t. native Julia implementations for samplers and whether interests (esp. relating to performance) are compatible.

awllee:
I deleted this release and re-released using the name NonUniformRandomVariateGeneration.jl

I will definitely try to contribute to Distributions.jl, time permitting. I think largely this would be a question of providing an additional sampling interface for high performance situations rather than making any changes to their very nice approach for general use.

andreasnoack:
Superseded by #14884

PR #15061, opened 2018-06-07
Remove stdlib packages from PtFEM require file

andreasnoack:
Fixes https://github.com/JuliaLang/julia/issues/27469


fredrikekre:
Why are Travis so happy when we get
```
fatal: error thrown and no exception handler available.
InitError(mod=:LibGit2, error=ErrorException("could not load library "libgit2
```
?

andreasnoack:
I think the issue is only on master even though it is hard to see from the log.

PR #15087, opened 2018-06-08
Register MAVs.jl

huckl3b3rry87:
….2, v0.0.3


andreasnoack:
Please consider renaming this to `MichiganAutonomousVehicles` to make it easier for users unfamiliar with the name MAV to reason about the content of the package. See https://docs.julialang.org/en/v0.6.2/manual/packages/#Guidelines-for-naming-a-package-1.

Also from the log
```julia
ERROR: LoadError: New tag 0.0.3 of package MAVs requires julia 0.6.0, but version 0.0.1 of MAVs requires julia 0.5.0. Use a new minor package version when support for an old version of Julia is dropped. Re-tag the package as 0.1.0 using `PkgDev.tag("MAVs", :minor)`.
```

Finally, please consider using Attotbot for managing package registrations and releases.

huckl3b3rry87:
Please delete this, I renamed it as you suggested and will make another pull request when I am done

PR #15157, opened 2018-06-13
Set Julia upper bound for Dates package

omus:
In Pkg3 the UUID for the Dates stdlib and the Uncurated Dates.jl package (this package) are identical.

```bash
$ head -n2 stdlib/Dates/Project.toml 
name = "Dates"
uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
$ head -n2 ~/.julia/registries/Uncurated/D/Dates/Package.toml 
name = "Dates"
uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
```

Having the same UUIDs could be problematic here as the two versions of Dates are now significantly different. In order to address this issue I've set an upper bound on Julia such that this package is not compatible with Julia 0.7 and thereby should not be included in a Pkg3 style registry making the issue go away (as suggested by @StefanKarpinski).

Note: for Dates.jl versions ≤ v"0.3.0" I've set the upperbound to Julia 0.4- as this upper bound was already used for the 0.3.1 and 0.3.2 versions. For the other versions I set the upper bound to Julia 0.4 as this version of Julia already included Dates in base.


omus:
@quinnj could you verify that the upper bounds are set appropriately?

quinnj:
Looks fine by me; I don't understand the travis failures; something about an `Immerse` package not having satisfiable versions or something. @aviks, any idea?

martinholters:
I don't know why the problem shows up for `Immerse`, but e.g. `ExcelReaders` 0.8.0 requires `Dates` and `julia 0.6.0-rc1` and there are many more packages requiring `Dates` and `julia 0.4` or `julia 0.5`. All of these would become uninstallable. Presumable, `Immerse` (maybe indirectly) depends on one of them?

omus:
I'll review the packages referencing Dates in the REQUIRE and make some adjustments

omus:
Interesting development. The [Dates.jl](https://github.com/quinnj/Dates.jl) package only works on version of Julia ≤ 0.4 since it uses things like [`Union(args...)`](https://github.com/quinnj/Dates.jl/blob/1764fc048a2186438be22dc7a3a34e87b333f296/src/dates/types.jl#L151) which was was deprecated in Julia 0.4 in favor of `Union{args...}`.

Also, packages that list "Dates" in the REQUIRE file will always install and build the non-base version. However, even if the Dates.jl package is installed on Julia versions ≥ 0.4 the base version of Dates is the only version that can be loaded.

What this means is that any package version that requires a minimum version of Julia ≥ 0.4 and Dates should _not_ include Dates in its REQUIRE file.

omus:
Package versions which require a minimum version of Julia < 0.4 and Dates are trickier and I'm trying to find the best way forward.

omus:
My plan is to set a higher Julia upper bound on Dates 0.4.4 which should allow older versions of packages to continue to work as they did. However if by some magic a package managed to support Julia from version 0.3 to 0.7 it would now break under Julia 0.7

For most packages that still reference Dates in their REQUIRE file this should not be an issue as these references are from older versions of the package:

- Escher
- ExcelReaders
- Gadfly
- GitHub
- HttpCommon
- Humanize
- JavaCall
- Lumberjack
- MarketData
- MySQL
- ODBC
- OnlineStats
- PyCall
- Quandl
- Redis
- SGP4
- TermWin
- TimeData
- TimeModels
- TimeSeries
- Timestamps
- Winston
- YAML

There are a few however where the latest version of the package still references Dates. It appears that these packages are no longer maintained:

- Augur
- Biryani
- Bokeh
- EconDatasets
- Faker
- GoogleCharts
- FinancialMarkets

andreasnoack:
Bump

omus:
Sorry about that. This fell off my radar. I'll upper bound the specified packages and I think we'll be good to go.

omus:
Set the Julia upper bound to `0.7-` as this should include all Julia 0.6 versions. Some packages didn't have a Julia lower bound so I set those to `0.0-`. And finally, FinancialMarkets already had a Julia upper bound of `0.6` so I didn't need to make any modifications there.

omus:
Looking over the packages changed here I found three which need PRs to remove Dates from their REQUIRE files:
- [ ] [Escher.jl](https://github.com/JuliaGizmos/Escher.jl/blob/a461bf87a7bf6904ad45c98e13aa699f2dea3b41/REQUIRE#L5)
- [ ] [GitLab.jl](https://github.com/JuliaComputing/GitLab.jl/blob/f0b3f056d66e132f5dcd91efa220eb932782494c/REQUIRE#L7)
- [ ] [SGP4.jl](https://github.com/crbinz/SGP4.jl/blob/70a7be3515cb3254bc81c5244106d5c3ecc09329/REQUIRE#L4)

PR #15162, opened 2018-06-13
Tag Optim.jl v0.15.0

attobot:
Release: [v0.15.0](https://github.com/JuliaNLSolvers/Optim.jl/releases/tag/v0.15.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaNLSolvers/Optim.jl/builds/391934787)
Diff: [vs v0.14.1](https://github.com/JuliaNLSolvers/Optim.jl/compare/6b412863e4119965375d31043b72bc40b9cc51db...7efe46184252625cb1272470d8f3e564699f323a)
`requires` vs v0.14.1: 
```diff
--- 0.14.1/requires
+++ 0.15.0/requires
@@ -1,9 +1,9 @@
 julia 0.6.0
 PositiveFactorizations
 Compat 0.18.0
-LineSearches 3.2.4 4.0.0
-Calculus
-NLSolversBase 4.0.0 5.0.0
+LineSearches 6.0.1
+NLSolversBase 6.1.1
 ForwardDiff 0.5.0
 DiffEqDiffTools 0.4.0
 Calculus # TODO: Remove when DiffEqDiffTools release Hessian PR
+Parameters
```
cc: @pkofod

Please make sure that:


andreasnoack:
Looks like this is breaking a couple of packages

pkofod:
Always interesting to see packages break, that I've never heard of. I'll PR upper bounds here, and update their REQUIREs.

pkofod:
It's not clear to me what the error is in RoMe - it also looks to me as Optim is downgraded to 0.14.1, so is it really this tag that breaks RoMe? Variography seem to pass an inconsistent combination of initial values and bounds to Fminbox. Should I upper bound both, or what's the solution ?

andreasnoack:
Let's ask @juliohm and @dehann

pkofod:
Ah, I misread Variography's error, that should definitely be upper bounded, as the order of arguments have changed. That's why the input is inconsistent (it was `x0, lower, upper` but is now `lower, upper, x0`)

pkofod:
https://github.com/juliohm/Variography.jl/pull/1 for REQUIRE update. I think we might still need @dehann s input on RoMe

pkofod:
@andreasnoack  if you look at the RoMe log, doesn't it first install this tag and then it downgrades it during setup of the REQUIRE packages?

andreasnoack:
I think the failure might be spurious. I cannot reproduce it. I've restarted the tests. Let's see what happens.

dehann:
RoME.jl tests still seems to pass on the rerun -- I will keep my eye open and test thoroughly in the next couple of days.  Thanks for the heads up!

pkofod:
Thanks for verifying that we didnt break RoMe.

PR #15188, opened 2018-06-15
Pull request/34a91bd2

brilhana:


andreasnoack:
It looks like the package mainly contains general-purpose optimization algorithms. Why not use Optim.jl?

brilhana:
For two reasons: the most efficient algorithm is not implemented by `Optim` and the package is going to include discrete choice models, as well as the solvers, in the foreseeable future.

andreasnoack:
But why not contribute the missing optimization algorithms to Optim.jl (which do you have in mind?) and just focus on the discrete choice models?

brilhana:
Trust region method using the truncated conjugate gradient but you have a valid point. I'll see if they're open to the idea.

PR #15198, opened 2018-06-16
Update SuiteSparse path

andreasnoack:
cc: @KristofferC


tkelman:
This is an entirely different package and shares none of the commit history, this breaks anything that was using the previous package.

PR #15260, opened 2018-06-21
Pull request/6bbe1a42

damiendr:


andreasnoack:
Please consider avoiding the acronym to help users not familiar with ROS understand what the package is about. See https://docs.julialang.org/en/v0.6.2/manual/packages/#Guidelines-for-naming-a-package-1.

Please also consider using Attobot when registering and releasing packages.

damiendr:
ROS is universally referred to as ROS in the field.

I tried using Attobot but it gave me an error with the hooks. As for the two pull requests they're for a 0.0.1 and 0.0.2. I'm not familiar with the package registration process, what should I do instead?

andreasnoack:
> ROS is universally referred to as ROS in the field.

The point is to make this meaningful for people **outside** the field, cf the naming guidelines in the link.

If you get an error from Attobot then you can ask Slack or discourse first and if it is a genuine issue then file it in the repo for Attobot. Using Attobot is not required, though, it is just much simpler for you as well as us, so I'd strongly recomment that you use it. If you choose to open a PR like then please leave a comment. In particular if you open a second. Otherwise, it is hard for us to know if this was on purpose or not.

damiendr:
> The point is to make this meaningful for people outside the field, cf the naming guidelines in the link.

I appreciate your point about acronyms. But this package is a parser for data formats that are known as ROS bags and ROS messages. People who don't know what ROS is have no reason to be interested. People who might want to use this package, on the other hand, will google something like "read ros bag julia". Likewise we have `HDF5.jl` and not `HierarchicalDataFormat5.jl`.

> If you get an error from Attobot then you can ask Slack or discourse first and if it is a genuine issue then file it in the repo for Attobot. Using Attobot is not required, though, it is just much simpler for you as well as us, so I'd strongly recomment that you use it. If you choose to open a PR like then please leave a comment. In particular if you open a second. Otherwise, it is hard for us to know if this was on purpose or not.

All right, I'll give the bot another go. I've just been following the instructions [here](https://docs.julialang.org/en/stable/manual/packages/#Creating-a-new-Package-1), which don't say much about the purpose of the pull request or what happens behind the scene, or whether you should make one for every tag. Sorry for the confusion!

damiendr:
Attobot worked this time. Shall I close the two PR I opened manually?

andreasnoack:
> Attobot worked this time.

Great.

> Likewise we have `HDF5.jl` and not `HierarchicalDataFormat5.jl`.

Indeed we have developed an exception for specific file formats over time. It wasn't clear to me, though, that `ROSData` was for file formats.

PR #15285, opened 2018-06-23
RecurrenceAnalysis 0.2.0

heliosdrm:
New package version supporting Julia 0.7.0-alpha


andreasnoack:
There is an issue with the tag. Please consider using Attobot when releasing packages. That would have avoided the issue.

heliosdrm:
I found the issue, sorry. I'll retry with attobot installed.

heliosdrm:
It seems that the issue is fixed now. It should be ready to merge.

PR #15373, opened 2018-06-28
Register new package Optionals

simonfxr:


andreasnoack:
From the log
```julia
ERROR: LoadError: METADATA/Optionals/versions/0.1.1/requires: no julia entry (>= 0.6.0 needed)
```

PR #15389, opened 2018-06-28
Tag Memento.jl v0.8.0

attobot:
Repository: [invenia/Memento.jl](https://github.com/invenia/Memento.jl)
Release: [v0.8.0](https://github.com/invenia/Memento.jl/releases/tag/v0.8.0)
Travis: [Travis Build Status](https://travis-ci.org/invenia/Memento.jl/builds/397897712)
Diff: [vs v0.7.0](https://github.com/invenia/Memento.jl/compare/c9305dec6b3ba8b6bcc694144bb3622b4f4867ba...2bf2f59c994ce64af697adef1018964ad233e651)
`requires` vs v0.7.0: no changes
cc: @iamed2

Please make sure that:


ararslan:
Looks like this removes `getparent` without deprecation, which is breaking MIPVerify.

iamed2:
`getparent` was not part of the public API

andreasnoack:
cc @vtjeng

vtjeng:
@andreasnoack I should be able to fix it - what should I change the call to? Sorry if it's obvious --- responding to this on my phone.

iamed2:
@vtjeng No worries, it's not obvious. The way to avoid breakage in the future would be to keep track of the loggers you use and `setlevel!` on those, or to have a known parent logger that you call `setlevel!(...; recursive=true)` on (it recurses down to children). But if you want to keep the current logic and don't mind the unstable private API I'd do `setlevel!.(Memento.getpath(MIPVerify.LOGGER)[2:end], level)`.

rofinn:
@vtjeng Using only the public Memento API your function is equivalent to:

```
setlevel!(getlogger(MIPVerify), level)
setlevel!(getlogger(), level)
```

However, I believe running `Memento.config!("notice"; recursive=true)` at runtime would more appropriately solve your problem by recursively setting all child loggers.

omus:
I've filed an [issue](https://github.com/vtjeng/MIPVerify.jl/issues/4) with MIPVerify.jl. Since no registered packages depend on MIPVerify and that he shouldn't have been using `getparent` to begin with I'll merge this PR.

vtjeng:
@omus and @rofinn, just wanted to make sure that my REQUIRE is appropriately updated - I need to be using `Memento>=0.8.0` for `Memento.config!` to be available, right?

rofinn:
@vtjeng Yeah, that would work, but isn't strictly necessary. `Memento.config!` was introduced in 0.7.0 with a deprecation for `Memento.config`. Also, `Memento.config!` and `setlevel!` started accepting the `recursive` keyword. In 0.8.0, the default logging level changed from `:warn` to `:info`.

PR #15469, opened 2018-07-02
Add Humanize v0.5.0 for Julia 0.7, upperbound previous versions.

IainNZ:


Humanize/versions/0.5.0/requires
```diff
@@ -0,0 +1,2 @@
+julia 0.7-beta
+Dates
```
Keno: stdlibs should not go into REQUIRE. This breaks the old pkg manager.

IainNZ: Oh interesting. I got complaints I thought from 0.7-beta if I didn't have that in my package's REQUIRE. Should they go in there, but not here?

Keno: Stdlibs are automatically picked up by the pkg3 system by parsing the source code. You likely needed a `Pkg.resolve` after editing your source code. In any case, putting it here is bad.

IainNZ: Got it. Will delete this PR and attempt the attobot way for good measure.

PR #15484, opened 2018-07-03
Register RequiredKeywordArguments 

timkittel:
Hi,
I would like to register this package https://github.com/timkittel/RequiredKeywordArguments.jl.git
Cheers,
Tim

### DIFF
--- a/RequiredKeywordArguments/url
+++ b/RequiredKeywordArguments/url
@@ -1,0 +1,1 @@
+https://github.com/timkittel/RequiredKeywordArguments.jl.git


timkittel:
It seems the error might not be part of my package. Or how could I fix that?

The [ci error](https://travis-ci.org/JuliaLang/METADATA.jl/builds/399462881#L502) seems to be due to the `Pkg.dir()` call in [`METADATA.jl`](https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/.test/METADATA.jl#L7). The package builds and passes all tests on my [travis setup](https://travis-ci.org/timkittel/RequiredKeywordArguments.jl/branches). Did I do something wrong?

timkittel:
okai, it seems the tag was necessary, so I submitted a PR with a tag #15489

PR #15489, opened 2018-07-03
Tag RequiredKeywordArguments v0.1.1

timkittel:


Keno:
Since this feature is available on 0.7, perhaps the best thing to do would be to integrate the functionality of this macro into Compat.jl's `@compat`?

timkittel:
Thanks for the hint. I opened a PR JuliaLang/Compat.jl#586 and hope everything works that way. Could you check it?

Also, do you know the time frame until this is merged into a version tagged in METADATA? I would like to use this functionality in a few projects that I am working on at the moment. If it takes a while, could one accept this package temporarily?

KristofferC:
If you just temporarily need it, you can just check out Compat master. A new version should of Compat shouldn't be far off.

timkittel:
True, I will go for that, thanks!

So I close this PR.

PR #15555, opened 2018-07-05
Add Package Optionals

simonfxr:


andreasnoack:
Are you sure that you want to register a new package just for 0.6 support? Soon nobody will care about 0.6.

simonfxr:
@andreasnoack My intension was to support 0.6 and up. I thought the line `julia 0.6` in `REQUIRE` means that at least 0.6 is needed, but does not exclude 0.7 and above. `Pkg3` is also supported via `Project.toml`. Do you have any advice for me to mark my package for 0.7?

andreasnoack:
You write that

> After dropping support for v0.6 and older, `Optional{T}` can simply be replaced with `Union{T, Missing}` (hopefully) without further changes.

so my impression was that this package was mainly useful if you want to support 0.6. My comment was that soon nobody will care much about 0.6 so it might not be worth the trouble to create a package just to support functionality across 0.6 and newer.

simonfxr:
I see. I mainly wrote that package for making upgrading some of my own code easier, since it will probably take some time before I can go 0.7/1.0 only. I thought there might be other people who might care about simultaneously supporting 0.6 and up.

andreasnoack:
No problem. I just wanted to ask such that you were aware but if you think there is a use case then it's fine to register this. Currently, there are a lot of package maintainers that drop 0.6 support to avoid spending time on supporting a version of Julia that will soon be obsolete. But if you still want this just let me know and I'll merge.

simonfxr:
Yes, please merge this PR. Its good that you make it clear to people that they should upgrade as soon as possible. I'm exited about going 1.0 only, until then I have to support a physics simulation running 0.6...

PR #15558, opened 2018-07-05
update Clipper url to use the JuliaGeometry fork

SimonDanisch:
as agreed with @maxvoxel8 in https://github.com/Voxel8/Clipper.jl/pull/29, we want to make JuliaGeometry/Clipper.jl the official repository.
Let me know if there is more to do than just changing the URL!


andreasnoack:
I think it would be a good idea to move the main repo. It will be confusing that official repo is a fork.

PR #15564, opened 2018-07-05
Register ForneyLab [https://github.com/biaslab/ForneyLab.jl.git]: v0.8.0

marcocox:


andreasnoack:
Please consider using Attobot when registering and releasing packages.

PR #15594, opened 2018-07-07
Tag Dendriform v0.1.1

chakravala:
https://github.com/chakravala/Dendriform.jl

updated for v0.7


andreasnoack:
Looks like something went wrong here. Please consider using Attobot for this. It will make it much easier to make releases.

PR #15645, opened 2018-07-10
update 0.7 tests to use OldPkg

KristofferC:


KristofferC:
Now gets to the point where PkgDev is broken on 0.7.

PR #15698, opened 2018-07-12
Register Preconditioners [https://github.com/mohamed82008/Preconditio…

mohdibntarek:
…ners.jl.git]: v0.1.0


andreasnoack:
From the log
```julia
ERROR: LoadError: LoadError: The following tags for Preconditioners have not been pushed to upstream repository: 0.1.0.

andreasnoack:
Please consider using Attobot instead. That makes it much easier to avoid such issues when registering and releasing packages.

mohdibntarek:
I will check out Attobot thanks. `PkgDev` wouldn't accept my Github credentials so I used the manual method.

PR #15870, opened 2018-07-18
Register new package GSReg.jl v0.1.0

vmari:
GSReg.jl -Global Search Regression- (https://github.com/ParallelGSReg/GSReg.jl), is a new package to perform the all-subset-regression approach exploiting Julia’s parallel capabilities and allowing users to choose between a simple GUI (https://github.com/ParallelGSReg/GSRegGUI) or an R/Stata-friendly command line interface.
--
Econometrics allows researchers to deal with competing theories about complex processes which could lead to very different diagnosis and even opposite policy advises. Advances in model selection techniques, driven by the increasing number of available HPC algorithms, constitutes one of its major contributions to Economic Theory. In-sample model selection has benefited from using mathematical developments to reduce the search space (e.g. Branch and Bound theorems) and efficiently find the best subset regression (in terms of defined information criteria).
It is also possible to use heuristic approaches like Genetic algorithms, or different dimension reduction methods like Stepwise, Lasso or Ridge estimators. While failing to guarantee global optimality, they are fast and well-suited for Big and/or Sparse Data. On the contrary, out-of-sample model selection techniques require more computing resources. In a Machine Learning environment (e.g. problems focusing on predictive analysis) there is an increasing universe of “training/test” algorithms (many of them showing very interesting performance in Julia) to compare alternative results and find-out a suitable model.
In Econometrics (e.g. problems focusing on causal inference) we require five important features which narrow the set of available algorithms: 1) Parsimony (to avoid very large atheoretical models); 2) Interpretability (for causal inference, rejecting “intuition-loss” transformation and/or complex combinations); 3) Across-models sensitivity analysis (economic theory is preferred against “best-model” unique results);
4) Robustness to time series and panel data information (preventing the use of raw bootstrapping or random subsample selection for training and test sets); and 5) advanced residual properties (e.g. going beyond the i.i.d assumption and looking for additional panel structure properties -for each model being evaluated-, which force a departure from many algorithms).
For these reasons, most economists prefer flexible all-subset-regression approaches, choosing among alternative models by means of some out-of-sample criteria, model averaging results, theoretical limits on covariates coefficients and residual constraints.


andreasnoack:
Thanks for contributing this package. Please consider renaming the package to e.g. `GlobalSearchRegression` to make it easier for people to reason about the content of this package, cf https://docs.julialang.org/en/stable/manual/packages/#Guidelines-for-naming-a-package-1.

vmari:
Thanks. We will work on your suggestion and resend it later.

PR #16240, opened 2018-08-02
Registering BoltzmannMachines.jl and BoltzmannMachinesPlots.jl

stefan-m-lenz:
Hello!
I would like to register the BoltzmannMachines-package in the official repo.
Version 0.1.0 is compatible with 0.6, but hopefully we will be able to upgrade it to work with Julia 0.7 soon after registering.

The package will also be presented at the JuliaCon 2018:
https://juliacon2018.sched.com/event/FQ4e/boltzmannmachinesjl-a-julia-native-package-for-training-and-evaluating-multimodal-deep-boltzmann-machines

Thank you very much!


stefan-m-lenz:
I think I messed something up, I will try again with attobot.

PR #16298, opened 2018-08-05
Register new package TRIANGLE.jl v0.1.1

attobot:
Repository: [cvdlab/TRIANGLE.jl](https://github.com/cvdlab/TRIANGLE.jl)
Release: [v0.1.1](https://github.com/cvdlab/TRIANGLE.jl/releases/tag/v0.1.1)
cc: @furio

Please make sure that:

@furio This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


andreasnoack:
Are you sure that you'd like the name to be all caps? Usually, Julia packages are only all caps if they wrap a library that is written in all caps and even then, people often avoid all caps. In this case, it even looks like the wrapped library doesn't use all caps.

furio:
Ok seems fair! I'll change where needed and resubmit. Thx for the feedback.

PR #16429, opened 2018-08-08
Register new package MPIFiles.jl v0.1.1

attobot:
Repository: [MagneticParticleImaging/MPIFiles.jl](https://github.com/MagneticParticleImaging/MPIFiles.jl)
Release: [v0.1.1](https://github.com/MagneticParticleImaging/MPIFiles.jl/releases/tag/v0.1.1)
Travis: [Travis Build Status](https://travis-ci.org/MagneticParticleImaging/MPIFiles.jl/builds/413665175)
cc: @tknopp

Please make sure that:

@tknopp This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


tknopp:
0.7 is still not fully working but I want to register this now to support both my 0.6 and 0.7 users.

andreasnoack:
Same as #16431

tknopp:
I see your point but at this point I cannot rename this package anymore since it is referenced in quite some publications (e.g. https://github.com/MagneticParticleImaging/MDF).

andreasnoack:
> I see your point but at this point I cannot rename this package anymore since it is referenced in quite some publications

I don't think that is a problem since GitHub will handle the redirect.

tknopp:
The name `MPIFiles` is well known in our research community. We cannot change this without confusing people.

andreasnoack:
The point of the guidelines is not to help people within the community of some subfield but to make it possible for most people to reason about the content of a package. You can choose to ignore the community guidelines but please be aware that our community wouldn't work if everybody did so. If you think the guidelines are wrong, I'd suggest that you propose an alternative that will work in general.

tknopp:
I think its not easy to 100% confirm to the guidelines. Packages like `FFTW.jl` or `NFFT.jl` exist and do not fulfill the requirements

andreasnoack:
`FFTW.jl` **does** follow the guidelines. Specifically, it matches 6. in https://docs.julialang.org/en/v0.6.2/manual/packages/#Guidelines-for-naming-a-package-1. NFFT is more questionable. I think FFT could be considered covered by 1. so `NonEquidistantFFT` had been a better name in my opinion. It would be much more informative. You could probably point to other examples since we have occasionally merged PRs that maybe shouldn't have been merged but that is not a good argument for not following the guidelines in general.

In this specific example, the abbreviation is extra problematic because MPI means something completely different to many Julia users.

tknopp:
I understand the general namespace issue but my packages MPIFiles and MPIReco are not ambiguous. The suffices "Files" and "Reco" make them unique. You can also google for "MPIFiles" and "MPIReco" you will find that its the first hit.

It a little frustrating for me that I don't get my packages in. As I have said, renaming is not an option since the packages are already in use and recognized by my userbase with the name.

andreasnoack:
The problem isn't that they aren't unique. The problem is that they are confusing and uninformative about what the packages do. The consensus is that acronyms aren't good package names which is why I suggested that you changed them. And again, github redirects so it's not a problem that you've already used the current names. If you insist on ignoring the guidelines and use these names anyway then we won't stop you. I just don't understand why you'd choose to do that. (Please put correct bounds on the package in any case such that tests pass).

tknopp:
I don't see how a github redirect fixes the issue. This would imply that all user code needs to be changed since the `MPIFiles` prefix is not valid anymore. And I certainly don't want to write `MagneticParticleImagingFiles.myfunction` all the time
```
If you insist on ignoring the guidelines and use these names anyway then we won't stop you. I just don't understand why you'd choose to do that. 
```
This sounds like I am doing something very negative. If I would be to create a new package then I would welcome name suggestions but maybe you can understand that it is a lot less convenient if you have an established package which is known to a larger userbase. I have talked about the package in conferences and yes the name `MPIFiles.jl` is a brand that I use to promote my open source package.

So its not that I am ignoring the guidelines. I am just not agreeing with them. In my impression it is the authors right to chose a name that he/she thinks is best. The registry should not judge over the author's choice unless there is a name conflict.

andreasnoack:
> And I certainly don't want to write MagneticParticleImagingFiles.myfunction all the time

You can always define

```julia
const MPIFiles = MagneticParticleImagingFiles
```
locally.

Anyway, just to avoid any misunderstandings here: even though I think the name should be changed, this PR isn't blocked on the name issue but because tests don't pass.

tknopp:
I fixed the tests for 0.7 and 1.0 and https://github.com/JuliaLang/METADATA.jl/pull/19040
is now the release that should be considered.

andreasnoack:
Next time, please just update the existing PR. See https://github.com/attobot/attobot.

PR #16517, opened 2018-08-09
Pull request/7aa58472

nbavafa:
Adding SMD, NextGenSeqUtils, RAD, and DPMeansClustering packages.


andreasnoack:
Please add packages once at a time and consider using Attobot for registering them. Notice that the guidelines for packages, https://docs.julialang.org/en/v0.6.2/manual/packages/#Guidelines-for-naming-a-package-1, prescribes avoiding abbreviations so you might want to consider renaming some of the packages.

nbavafa:
While we understand abbreviations are not preferred, these are the names referenced in [the following manuscript](https://www.biorxiv.org/content/early/2018/08/03/383794) and we would strongly prefer to keep everything consistent.

andreasnoack:
GitHub will handle redirects after renaming a package so having used the name in a publication shouldn't be an issue. I'd strongly recommend that you follow the guidelines. Names such as RAD and SMD could mean anything and probably match names in many subfields.

PR #16520, opened 2018-08-10
Tag Reduce v1.0.0

chakravala:
[Reduce.jl](https://github.com/chakravala/Reduce.jl)

Compatible with julia 0.7 and 1.0, also some minor enhancements were added too.

Diff vs v0.4.1: https://github.com/chakravala/Reduce.jl/compare/fcc9459a099b9bab2ef2e958438b92903061d92d...382b5ff84bb66067f2f087156c6bd930c297318a


andreasnoack:
Please consider using Attobot. It makes it much easier for us to check that things are right (and it is also easier for you).

chakravala:
Yes, I understand, I will be switching to it, this is probably the last time I am doing a manual release, I wanted to make sure that the upper bounds for previous versions are labeled correctly, I dont know if attobot would have done it correctly or not for this package.

fredrikekre:
You generally don't need to add the upper bounds.

PR #16640, opened 2018-08-11
Tag BoltzmannMachines.jl v1.0.0

attobot:
Repository: [stefan-m-lenz/BoltzmannMachines.jl](https://github.com/stefan-m-lenz/BoltzmannMachines.jl)
Release: [v1.0.0](https://github.com/stefan-m-lenz/BoltzmannMachines.jl/releases/tag/v1.0.0)
Diff: [vs v0.1.0](https://github.com/stefan-m-lenz/BoltzmannMachines.jl/compare/cc778a5a73699b668568793320ffdded2f92610d...e69e9100d91e1969bd1ce06db805d9694fd49503)
`requires` vs v0.1.0: 
```diff
--- 0.1.0/requires
+++ 1.0.0/requires
@@ -1,2 +1,2 @@
-julia 0.6
+julia 0.7
 Distributions
```
cc: @stefan-m-lenz

Please make sure that:


andreasnoack:
Please enable Travis testing and add test badges to your README

stefan-m-lenz:
Thank you for your efforts!
It took me some time to get the CI running. It was especially difficult because the tests run a parallelized stochastic algorithm and test its accuracy. Due to the stochastic nature of the algorithm, tests may fail sometimes and so I had to think of a way to handle these tests more relaxed.

andreasnoack:
Not sure exactly why the tests are failing when run on CIBot but I suspect it might be related to setting to `LOAD_PATH` in https://github.com/stefan-m-lenz/BoltzmannMachines.jl/blob/master/test/runtests.jl#L9. Is that really necessary?

stefan-m-lenz:
I have all my test methods that are called in runtests.jl in a module BMTest which is located in /test/BMTest.jl.

Also, I wanted to remove the plotting package from the BoltzmannMachines test dependencies to avoid circular dependencies. To be able to execute the examples, which use some plotting, without using the actual plotting package, I prepared a mock plotting package located in /test/mock/BoltzmannMachinesPlots.jl.

By modifying the LOAD_PATH, I am able to load these modules with `import`/`using`.
How would it be possible to achieve this in another, compatible way? Thank you very much!
I also looked at the capabilities of the new package manager, but I wasn't able to figure out how to do this via toml specification or something like that.

andreasnoack:
I don't think you need https://github.com/stefan-m-lenz/BoltzmannMachines.jl/blob/master/test/runtests.jl#L9 at all and instead of https://github.com/stefan-m-lenz/BoltzmannMachines.jl/blob/master/test/runtests.jl#L13 I think you can just do `include(joinpath(@__DIR__(), "test", "BMTest.jl"))`.

stefan-m-lenz:
This solution is not meeting my requirements because I can't mock the plots package with it.


    module Bla
        function f()
            println("Bla")
        end
    end

    using Bla

Will result in an error because `using` looks up a package in the path.
So I would have to get rid of the `using BoltzmannMachinesPlots` in my `examples.jl`.
This is complicated...

But since I tested with the default CI script and used behaviour specified in the docs: Is this really my fault or isn't this an error of the JuliaCIBot then?

I see in https://juliarun-ci.s3.amazonaws.com/bb9f2442c537a13594a849b18fb55ce3b95fc4d1/pull_request_of_BoltzmannMachines_on_julia_0_7.log:

`ERROR: LoadError: ArgumentError: Package Distributions not found in current path`.

This should not be the case and it should not be affected by appending something to the `LOAD_PATH`. It must be something else that goes wrong.

stefan-m-lenz:
What also is weird: Why is 

    Environment:
        JULIA = /opt/julia-0.6/bin/julia

when using Julia 1.0 and 0.7?
(https://juliarun-ci.s3.amazonaws.com/bb9f2442c537a13594a849b18fb55ce3b95fc4d1/pull_request_of_BoltzmannMachines_on_julia_0_7.log)

andreasnoack:
> This solution is not meeting my requirements because I can't mock the plots package with it.

You can do
```julia
module Bla
    function f()
        println("Bla")
    end
end

using .Bla
```
Would that be okay?

> and used behaviour specified in the docs

Could you direct me to where in the docs this is described? I'd like to know how this is described.

> Is this really my fault or isn't this an error of the JuliaCIBot then?

It is fair to blame `CIBot` here. It is still using the old package manager which is causing some problems. However, we need some time to upgrade CIBot and reverse dependency is really valuable so if it only requires minor adjustments to your package, I'd say that it would be worth it. However, if you don't like the changes, we can skip the reverse dependency testing for your package. However, typically, users shouldn't mess around with `LOAD_PATH` so it might be worth avoiding it anyway.

nkottary:
I have fixed the `JULIA` environment variable.

andreasnoack:
> I have fixed the `JULIA` environment variable.

Could you elaborate? I'm not sure what this refers to in this issue.

nkottary:
I was replying to @stefan-m-lenz comment about the `JULIA` environment variable always pointing to 0.6. This is fixed now and should reflect on future tests.

stefan-m-lenz:
@nkottary Thank you for that!
Would it be possible to re-run the test? Maybe that fixes the issue.

stefan-m-lenz:
I run in all kinds of issues when trying without temporarily modifying the LOAD_PATH variable. It really is very ugly and I haven't got it to work.

[Here](https://docs.julialang.org/en/v1/manual/modules/#Module-file-paths-1)  it is said that adding something to the LOAD_PATH variable is a valid way to load modules.

I think the issue is more with the `@everywhere` and the usage of the Distributed package than with modifying the LOAD_PATH.

andreasnoack:
I'll restard the tests.

Did you try the `include` + "prefix with `.`" method mentioned in https://github.com/JuliaLang/METADATA.jl/pull/16640#issuecomment-416499887?

andreasnoack:
@stefan-m-lenz Have you been changing the tags? Please see the log. Might be best to remove and reapply the tag through GitHub. That should update this PR and trigger the tests again.

stefan-m-lenz:
> Did you try the include + "prefix with ." method mentioned in #16640 (comment)?

This doesn't really solve my problem as I would like to avoid modifying the examples. And replacing `import BoltzmannMachinesPlots` by `import .BoltzmannMachinesPlots` is still a modification. 
An ugly workaround would be to load the file as string, throw this line out while using `include` before and including the file with `include_string`. But I like to avoid such ugliness if possible.

But also the BMTest module depends on BoltzmannMachines and needed to import it. I tried the dot thing there with `..` but then I ran into `MethodError: no method matching dirname(::Nothing)`.
This is all very confusing and I really do not want to waste too much of our time on this. I will reapply the tag and hope that it works. Otherwise it's probably best to wait until the CIBot has matured and is expected to behave in the same way as the standard Travis CI script.

andreasnoack:
> Otherwise it's probably best to wait until the CIBot has matured and is expected to behave in the same way as the standard Travis CI script.

I agree. While reverse dependency testing is nice it is mainly important for packages with many dependent packages and hopefully, it won't take too long to upgrade CIBot.

stefan-m-lenz:
@nkottary Oh it works now. Great! But it still says 

    Environment:
        JULIA = /opt/julia-0.6/bin/julia

in https://juliarun-ci.s3.amazonaws.com/60a5c93256e125305cd45c988f7306ee175b38ba/pull_request_of_BoltzmannMachines_on_julia_0_7.log

PR #16714, opened 2018-08-12
Register new package LARLIB.jl v0.1.0

attobot:
Repository: [cvdlab/LARLIB.jl](https://github.com/cvdlab/LARLIB.jl)
Release: [v0.1.0](https://github.com/cvdlab/LARLIB.jl/releases/tag/v0.1.0)
cc: @furio

Please make sure that:

@furio This PR will remain open for 24 hours for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


andreasnoack:
Are you sure that you'd like the name to be all caps? It quite non-standard for Julia packages.

furio:
Hi @andreasnoack 

LAR stands for Linear Algebraic Representation it would be quite long to put in the name (LinearAlgebraicRepresentationLibrary) hence LARLIB... do you think we should change to something else? LARlib ? LARLib? only LAR?

andreasnoack:
Ideally, the name would be something like `LinearAlgebraicRepresentations` since it would be easy to figure out what it actually did. Using just `LAR` would violate out naming guidelines and `LAR(LIB/Lib/lib)` would be close to. Remember that very few people would have a clue what `LAR` refers to. If you insist on `LAR`-something then I think that `LARLib` or `LARlib` are better than `LARLIB`.

furio:
Ok I'm linking this PR to the repository so we can decide :)

Thx for the feedback! We will do another PR.

PR #16813, opened 2018-08-14
Release Xtensor 0.4.0

SylvainCorlay:


andreasnoack:
Why is it that you don't want to use Attobot? It makes our work quite a bit simpler. E.g. I can see the tests status in the PR here.

You should consider testing on the latest bugfix release instead of 0.6.0 and 0.6.1.

SylvainCorlay:
I tend to reduce the number of tools I use for work :) and I can make releases with git without learning another tool.

PR #16836, opened 2018-08-14
Register DPMeansClustering [https://github.com/MurrellGroup/DPMeansCl…

nbavafa:
…ustering.jl.git]: v0.0.1, v0.0.2


andreasnoack:
Would be great if you could use Attobot for this one as well.

PR #17060, opened 2018-08-19
Tag ImageTransformations.jl v0.5.0

attobot:
Repository: [JuliaImages/ImageTransformations.jl](https://github.com/JuliaImages/ImageTransformations.jl)
Release: [v0.5.0](https://github.com/JuliaImages/ImageTransformations.jl/releases/tag/v0.5.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaImages/ImageTransformations.jl/builds/417833716)
Diff: [vs v0.4.2](https://github.com/JuliaImages/ImageTransformations.jl/compare/93efe78157c47861ea3f5f3a233839b9bae0f116...23fbd5f8c3536fa055abad23806eb47da8f8121a)
`requires` vs v0.4.2: 
```diff
--- 0.4.2/requires
+++ 0.5.0/requires
@@ -1,5 +1,5 @@
-julia 0.6
-ImageCore 0.1.2
+julia 0.7
+ImageCore 0.7
 CoordinateTransformations 0.4.0
 Interpolations 0.4
 AxisAlgorithms
@@ -10,5 +10,3 @@
 Colors 0.7.0
 ColorVectorSpace 0.2
 FixedPointNumbers 0.3
-ShowItLikeYouBuildIt
-Compat 0.27
```
cc: @Evizero

Please make sure that:


Evizero:
Ok so this is a bit of an unfortunate situation (which is mostly my fault)

The package itself works. i.e. all the requirements it needs are tagged an available.

The tests however won't and can't pass here, because it has `ReferenceTests.jl` as a test dependency, which itself has the full `Images.jl` as a normal depdency, which in turn depends on `ImageTransformations.jl`. Obviously that is not a great situation (especially when there is a breaking julia version change where every involved package breaks), and its something we will have to improve in the future.

That said, given that the current tagged version of ImageTransformations doesn't work on 0.7 anyway, merging this can't do any more harm than good.

Evizero:
cc: @timholy

quinnj:
Thanks for the explanation @Evizero; hopefully you folks can sort out the circular dependency in testing.

PR #17104, opened 2018-08-20
Set upper bound for NamedTuples

omus:
Package does not work correctly on Julia 0.7 and not at all with 1.0. Currently there aren't plans to support these Julia versions as NamedTuple support has been integrated into Base.


omus:
cc: @davidanthoff

quinnj (approved):

quinnj:
Looks like a bunch of queryverse packages will need to be updated as well; I think they already have 0.7 tags but still depend on NamedTuples?

omus:
I probably need to set the upper bound to 1.0 and make sure there is a version of NamedTuples.jl which work correctly with 0.7.

davidanthoff:
I'm all in favor of merging this PR.

I'm just not done with porting Queryverse, but it won't depend on NamedTuples.jl anymore for the 0.7 versions.

andreasnoack:
I think the solution is to add upper bounds at Julia 0.7 or 1.0 for the affected packages in this PR. It should be fine, since the packages don't work on 0.7/1.0.

omus:
I think we're ready to go

iamed2:
@omus This broke any package which uses NamedTuples and supports 0.6, 0.7, and 1.0. LibPQ as an example.

omus:
That is unfortunate. Since LibPQ is using the Base NamedTuple implementation on Julia 0.7 and above you only need to be able to install NamedTuples.jl on Julia 0.7 and above in order to have a single version that support Julia 0.6 onwards.

I would prefer not to revert this change as the versions of NamedTuples.jl so far do not work with Julia 1.0 there have been quite a few issues reported to NamedTuples.jl because of this. For new users it is definitely better if we don't allow them to install a package on Julia 1.0 which doesn't work.

Another option would be to make a new minor release of LibPQ which is only for 0.7 and up. That allows you to drop the NamedTuples.jl requirement and everything would work again. I suspect like me you want to avoid having to support multiple versions of a code base unless it is absolutely necessary.

The last option which I think will work best is to make a new release of NamedTuples.jl which is for Julia 0.7 onwards only. What I'll do is just have the package re-export the Base NamedTuple implementation which will encourage users to use the Base implementation but still allows packages to work.

omus:
I'm not sure if people were more upset that NamedTuples.jl wasn't working on Julia 1.0 or that I did this upper bound. In either case I've made a Julia 0.7+ only release of NamedTuples that works and reduces pain points for users using packages that haven't fully switched to 1.0. See https://github.com/JuliaLang/METADATA.jl/pull/17804

PR #17216, opened 2018-08-22
Upper bound Julia version in IntervalAritmetic.jl

dpsanders:


lbenet:
There seems to be aproblem with travis. The important lines are:
```
INFO: Verifying METADATA...
    PolynomialZeros v0.0.1 – no valid versions exist for package 
    PolynomialZeros v0.1.0 – no valid versions exist for package 
```
Any idea @dpsanders?

lbenet:
I think the problem that travis detected in the dependencies, in particular in `PolynomialZeros.jl`, is due to the fact that travis is checking that everything works *also* Julia 0.7 and Julia 1.0; see for example [this line](https://travis-ci.org/JuliaLang/METADATA.jl/builds/419330605?utm_source=github_status&utm_medium=notification) of travis report, which is running in Julia 0.7; the next error occurs when checking Julia 1.0. My guess is that this happens because  `PolynomialZeros.jl` does not include an upper bound in REQUIRE for the applicable version of Julia. 

@andreasnoack Is this interpretation correct? If so, what can we do about this?

Cc @dpsanders , @jverzani

jverzani:
The inclusion of ValidatedNumerics in PolynomialZeros is no longer necessary. I will remove it with the next release. Would this help?

andreasnoack:
Just bound `PolynomialZeros` at 0.7 as well in this PR since the package doesn't work on 0.7. I believe that should fix the issue.

lbenet:
@jverzani Thanks a lot for your help! Would it be possible to release a new (patch?) version of `PolynomialZeros` with the upper bound for the Julia version as suggested by @andreasnoack? This would allow us to go on in the process of releasing versions that work with Julia 0.7 and 1.0.

jverzani:
Sure, I have it ready to go. Will get to it today.

andreasnoack:
My suggestion is actually just, in this PR, to add upper bounds to the existing released versions in METADATA since they don't work on 0.7 anyway.

jverzani:
Okay, then I won't do anything with PolynomialZeros for v0.7, unless I'm told I'm reading this incorrectly.  Thanks for your attention.

lbenet:
@andreasnoack But, isn't precisely that what @dpsanders did?

andreasnoack:
> But, isn't precisely that what @dpsanders did?

I don't think so. I only see upper bounds made in `IntervalAritmetic` in this PR.

lbenet:
Sorry, my mistake, you are right! I'll try to remedy this.

PR #17228, opened 2018-08-22
Adding ABC package

eford:


eford:
WHy is JuliaCIBot taking so long (6 hours and counting)?  What is it doing?

dpsanders:
JuliaCIbot is problematic at the moment. 

Please read the package naming guidelines - this package name would usually be far too generic / not comprehensible to the standard Julia user.

eford:
>
> JuliaCIbot is problematic at the moment.
>
Do you know if moment is likely to be closer to a day or a week?
I've been happily using unregistered packages since v0.2.  However, the new
package manager makes it very painful to use unregistered packages.
Realistically, I won't be able to update these packages to work with 1.0
until these get registered (or someone provides much better documentation
for how to use unregistered packages simply).

> Please read the package naming guidelines - this package name would
> usually be far too generic / not comprehensible to the standard Julia user.
>
I have.   ABC is a standard abbreviation for Approximate Bayesian
Computation (https://en.wikipedia.org/wiki/Approximate_Bayesian_computation
).
Writing it out is unreasonably long, I don't see that as a viable
alternative.  Any half-abbreviation would be non-standard and worse than
just using ABC.

Thanks,
Eric

dpsanders:
ApproximateBayesian.jl seems like a good name to me.

It is easy to use unregistered packages with the new package manager - you just do 

add URL#master 

where URL is the correct URL, or dev with a local directory.

eford:
>
> ApproximateBayesian.jl seems like a good name to me.
>
Approximate Bayesian Computing and ABC have a technical meaning.  It's not
just two adjectives in front of computing.
"Approximate Bayesian" does not have a technical meaning, so that would
risks creating confusion.

> It is easy to use unregistered packages with the new package manager -
>
Oh, if only that were true!  I've watched the video from JuliaCon and still
spent hours just trying to make sense of the bizzare error messages that it
gives.

> you just do
>
> add URL#master
>
> where URL is the correct URL, or dev with a local directory.
>
I specified a new JULIA_DEPOT_PATH and started julia 0.7
(I wish someone would have warned me about that in advance.  Now my .julia
directory is a confused mess with both v0.6 and v0.7 stuff, and my goofing
around trying to get things to work with 0.7.)

Then
(v0.7) pkg> add https://github.com/eford/ABC.jl#master
julia> using ABC
That gives a bunch of depicated warnings.  Now I want to try to fix them.
I've started a v1 branch, but I know there's more to do.  So I try
(v0.7) pkg> dev ABC
ERROR: The following package names could not be resolved:
 * ABC (985726d4-534b-5288-8e35-7537d599df55 in manifest but not in project)
Please specify by known `name=uuid`.

Now what?  I'm hoping that if I can get the package registered then the
"package names could be resolved".
From the limited documentation that's avaliable at this time, it seems that
manifests and uuids are complicated barely-human-readable things, but Pkg3
is supposed to create them automatically.  But it's not working for any of
my packages.
Any ideas?

Thanks,
Eric

dpsanders:
I don't see anything wrong with long package names - indeed I have a published one that's about as long as approximate bayesian computing. You can use tab completion and only need the name once a session anyway.

For Pkg issues you should post in discourse or on the pkg-usage channel in slack.

eford:
I use the module name explicitly in several places, so it shows up repeatedly in codes that use the pacakage.  I'd prefer the shorter name for readability, compatability, etc. 

I already appealed for help with the new pacakage manager via Discourse, but didn't get any help.  Googling turns up lots of other confused people asking for help figuring out the new package manager.

dpsanders:
Just put

const ABC = ApproximateBayesianComputation

Try the slack channel. I'm afraid I can't help with those package problems.

eford:
Thanks, that will be useful!  
I do see the benefits of using 
const ABC = ApproximateBayesianComputing
to have both a shorthand for coding and a longer package name.
I've created a new repo named ApproximateBayesianComputing and an associated pull request to METADATA.  
My research group has been using ABC.jl for 5 years, so I'm a little worried that this will break collaborators repos/workflows.  
I think my plan is to just leave ABC.jl as is (an unregistered package) meant for existing users using Julia <v0.6.  I can tell people that when moving to v0.7, they should switch to ApproximateBayesianComputing.jl.  
Any suggestions for minimizing the pain of the name change?

WRT Slack:  Slack is not a good way to provide documentation on how people should use the new package manager.  There's not a good way for others to find the tips.  Much better would be a blog post, discourse thread or FAQ page, so that the few people who have already figured out how to use it (and hopefully have developed a practical workflow) don't need to answer the same questions over and over again.  

Thanks,
Eric

andreasnoack:
Superseded by #17270

PR #17337, opened 2018-08-25
Request to add Combo.jl to Julia's repo [Pull request/84e6b33c]

uripatish:
Following my talk at the last JuliaCon on Combo.jl, the package is now available on GitHub, and I was hoping that it could be added to Julia's official repo.


quinnj:
Please add tests [here](https://github.com/uripatish/Combo.jl/blob/master/test/runtests.jl) and enable CI testing via [travis](https://travis-ci.org/). It would also be ideal to support julia 0.7, as that is what many users are transitioning too; if that's not realistic, it would be better to have your REQUIRE file say `julia 0.6 - 0.7` to ensure users can't install it on 0.7/1.0.

Let me know if any of that doesn't make sense; happy to clarify or help explain anything in particular. Also, with new packages, there's a standard 24-hour wait period to allow general comments on package name, contents, and suitability for officially registering. It can also help the release process to enable https://github.com/attobot/attobot for your repo, so that creating a new release automatically takes care of the METADATA work for you.

uripatish:
I've updated the tests, but adding `julia 0.6 - 0.7` instead of `julia 0.6` to REQUIRE makes Pkg.build("Combo") fail. I've also enabled Travis, and the build is passing (I've added the Travis badge to README.md). I've also enabled attobot, but not sure if anything happend, I have no experience with it. As for adapting the package to 0.7.0/1.0.0, that is going to have to wait for a while as a have few deadlines that are coming up really soon, and I have to attend to these first.

andreasnoack:
To use attobot, please add a tag from GitHub as explained in the documentation. You might want to remove some of the existing commits although it isn't required.

PR #17366, opened 2018-08-27
Release RDatasets version 0.5.0

andreasnoack:
Apparently, Attobot isn't enabled for the repo and the owner isn't active so I'm releasing this manually while we are waiting for a proper solution.


andreasnoack:
CIbot fails because of a project file in Nullables. See https://github.com/JuliaAttic/Nullables.jl/pull/18

andreasnoack:
CIBot now fails because of the project file in Mocking so let's ask @omus if he'd be okay with deleting the file for the time being until CIBot has been updated to the new package manager.

omus:
It should be fine to delete (although it seems a bit silly). Can I just delete it on `master` or do I need a new tag?

andreasnoack:
Thanks. It would have to be a new release. I agree that this isn't optimal but I think reverse dependency testing is important and we don't know yet when CIBot will work with the new package manager.

omus:
Should be addressed by: https://github.com/JuliaLang/METADATA.jl/pull/17490

PR #17394, opened 2018-08-27
Tag Miniball.jl v0.2.0

attobot:
Repository: [JuliaFEM/Miniball.jl](https://github.com/JuliaFEM/Miniball.jl)
Release: [v0.2.0](https://github.com/JuliaFEM/Miniball.jl/releases/tag/v0.2.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaFEM/Miniball.jl/builds/421247351)
Diff: [vs v0.1.2](https://github.com/JuliaFEM/Miniball.jl/compare/5b994939e1bde4a8fa77b3bcd4d54dc55d5c8495...689a86ff3a2ce61ec70d15a24a47690ce3a8c4e2)
`requires` vs v0.1.2: 
```diff
--- 0.1.2/requires
+++ 0.2.0/requires
@@ -1 +1 @@
-julia 0.6
+julia 0.6 0.6
```
cc: @TeroFrondelius

Please make sure that:


andreasnoack:
The entry `julia 0.6 0.6` looks wrong since the interval is closed-open, i.e. `[0.6,0.6)` which is empty. Did you mean to make it `0.6` only? That would be `julia 0.6 0.7-`

TeroFrondelius:
@jw3126 can you reply? Did we make a mistake or do we want that nobody can install this? 

@andreasnoack FYI we are trying to depreciate this package.

jw3126:
@andreasnoack is the effect of this, that nobody can `Pkg.add("Miniball")` anymore, then yes this is what we want. Or is the effect, that everybody who adds Miniball, just gets an older version. Then this is a mistake indeed.

andreasnoack:
It's the latter. People will just get the older version. I theory, it might be possible to make the old version uninstallable but I wouldn't recommend doing so. Instead, I'd just add upper bounds on the existing releases in METADATA at `0.7-`. That would make it impossible to install the package on versions of Julia starting with 0.7 but without affecting existing installations of the package on Julia 0.6.

jw3126:
Ok thanks for the help, I opened a PR to set the upper bounds.

andreasnoack:
Sounds good. I'll close this one.

PR #17459, opened 2018-08-28
Try splitting the travis tests into Julia versions

andreasnoack:
This should speed up the tests and also make it easier to see which versions of Julia that are failing.

### DIFF
--- a/.test/ci.sh
+++ b/.test/ci.sh
@@ -16,30 +16,21 @@ git checkout -b localbranch
 mkdir -p $CI_TMP_DIR
 cd $CI_TMP_DIR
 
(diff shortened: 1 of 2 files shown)


andreasnoack:
cc @omus since you made the changes to make the tests run outside Travis. Would you prefer that we set a `CI_JULIA_VERSION` instead of relying on `TRAVIS_JULIA_VERSION` here?

omus:
I think it would be preferable to use a `CI_JULIA_VERSION`. Thanks for considering this :)

omus:
I've updated out GitLab CI script and this work great 👍

PR #17487, opened 2018-08-29
Tag ImageView.jl v0.7.0

attobot:
Repository: [JuliaImages/ImageView.jl](https://github.com/JuliaImages/ImageView.jl)
Release: [v0.7.0](https://github.com/JuliaImages/ImageView.jl/releases/tag/v0.7.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaImages/ImageView.jl/builds/422109665)
Diff: [vs v0.6.0](https://github.com/JuliaImages/ImageView.jl/compare/c02143410325f0879f24b13f01f5252d7992cca3...c18c64ce1d9d3178f1a14b85f6f2cebc880bb2fd)
`requires` vs v0.6.0: 
```diff
--- 0.6.0/requires
+++ 0.7.0/requires
@@ -1,6 +1,6 @@
-julia 0.6
+julia 0.6 0.7
 Images 0.6.0
-GtkReactive 0.2
+GtkReactive 0.4.2
 FixedPointNumbers 0.3
 Colors
 ColorVectorSpace
```
cc: @timholy

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #17488, opened 2018-08-29
Place julia version upperbound on ImageView

timholy:
Companion to #17487

### DIFF
--- a/ImageView/versions/0.0.0/requires
+++ b/ImageView/versions/0.0.0/requires
@@ -1,4 +1,4 @@
-julia 0.2-
+julia 0.2.0- 0.3.0
 Tk
 Images 0.0.0- 0.6.0
 Winston
--- a/ImageView/versions/0.0.1/requires
+++ b/ImageView/versions/0.0.1/requires
@@ -1,4 +1,4 @@
-julia 0.2-
+julia 0.2.0- 0.3.0
 Tk
 Images 0.0.0- 0.6.0
 Winston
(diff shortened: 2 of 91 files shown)


andreasnoack:
The log says
```
    PGFPlots v1.4.3 – no valid versions exist for package 
    PGFPlots v1.5.0 – no valid versions exist for package 
    PGFPlots v1.5.1 – no valid versions exist for package 
    PGFPlots v1.6.0 – no valid versions exist for package 
    PGFPlots v1.7.0 – no valid versions exist for package 
    PGFPlots v2.0.0 – no valid versions exist for package 
    Reel v0.2.0 – no valid versions exist for package 
    Reel v0.2.1 – no valid versions exist for package 
    Reel v0.2.2 – no valid versions exist for package 
    Reel v1.0.0 – no valid versions exist for package 
    Reel v1.0.1 – no valid versions exist for package
```
so could you add bounds on them as well?

timholy:
Yes, but CC @shashi and @mykelk. (I think we should give them a chance to review this before merging.) It's not entirely obvious to me how the PGFPlots dependency is in there, it could be a resolver issue?

FYI the way I am doing this is adding the bound on the most recent release, and then using `PkgDev.propagate_upperbound("Reel", "julia")`. It's a very useful utility that was [never merged](https://github.com/JuliaLang/PkgDev.jl/pull/94). (The perfect became the enemy of the good and I ran out of interest.)

PR #17510, opened 2018-08-29
Add upper bounds on deleted packages

andreasnoack:
cc: @bramtayl, @miguelraz


Chunks/versions/0.0.2/requires
```diff
@@ -1,2 +1,2 @@
-julia 0.6
+julia 0.6 0.6
```
ararslan, line 1: Shouldn't this be `0.6 0.7-`, since the versions form the interval `[lower, upper)`? If the interval is closed at 0.6 and open at 0.6, it's empty, so the package won't be installable.

andreasnoack: The repo has been deleted so it's on purpose.

ararslan: Oh, I see.

ararslan (approved):

Chunks/versions/0.0.2/requires
andreasnoack: I'm actually in doubt if completely deleting the entry would be better in this case.

ararslan: You mean removing the `julia` line altogether?

andreasnoack: No. I was actually wondering if it would be better to completely delete all the files in METADATA related to the repo.

ararslan: This PR seems like a better solution to me since it's more traceable for any existing code that might be referencing these.

PR #17516, opened 2018-08-30
add Julia v1.0 upperbound to old ForwardDiff releases

jrevels:
ref #17515 and https://github.com/JuliaDiff/ForwardDiff.jl/issues/352

I have no idea whether this is the right solution, especially w.r.t. Pkg3 @KristofferC @StefanKarpinski

### DIFF
--- a/DelayDiffEq/versions/0.10.0/requires
+++ b/DelayDiffEq/versions/0.10.0/requires
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.6 1.0.0
 DiffEqBase 1.21.0 2.0.0
 OrdinaryDiffEq 2.17.0 2.19.0
 DataStructures 0.4.6
--- a/DelayDiffEq/versions/0.5.0/requires
+++ b/DelayDiffEq/versions/0.5.0/requires
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.6 1.0.0
 DiffEqBase 0.15.0 2.0.0
 OrdinaryDiffEq 2.9.0 2.15.0
 DataStructures 0.4.6
--- a/DelayDiffEq/versions/0.5.1/requires
+++ b/DelayDiffEq/versions/0.5.1/requires
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.6 1.0.0
 DiffEqBase 1.15.0 2.0.0
 OrdinaryDiffEq 2.10.0 2.15.0
 DataStructures 0.4.6
--- a/DelayDiffEq/versions/0.5.2/requires
+++ b/DelayDiffEq/versions/0.5.2/requires
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.6 1.0.0
 DiffEqBase 1.15.0 2.0.0
 OrdinaryDiffEq 2.10.0 2.15.0
 DataStructures 0.4.6
--- a/DelayDiffEq/versions/0.6.0/requires
+++ b/DelayDiffEq/versions/0.6.0/requires
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.6 1.0.0
 DiffEqBase 1.16.0 2.0.0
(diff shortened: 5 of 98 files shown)


andreasnoack:
Looks like this is causing a chain reaction of extra upper bounds. It seems that the process can be automated a bit. See https://github.com/JuliaLang/METADATA.jl/pull/17488#issuecomment-416996296.

jrevels:
Well, that's presumably because this PR makes it so that there is no longer any version of ForwardDiff in METADATA that supports Julia 1.0.

IIUC, we just need to run CI again after #17515 is merged and things should turn up green.

jrevels:
Restarted failing builds here now that #17515 has been merged, let's see how it goes.

jrevels:
Okay, looks like some packages are upper-bounded to only support ForwardDiff versions that now have a 1.0.0 upper-bound.

Thus, these packages cannot technically support Julia 1.0 while depending on ForwardDiff unless they update to use ForwardDiff 0.9.0.

For example, here's the REQUIRE file for ReverseDiffSparse v0.8.2:

```
julia 0.6.0
Calculus
DataStructures
MathProgBase
NaNMath 0.2.1
ForwardDiff 0.5 0.9
Compat 0.59
```

For version ranges like these, I feel pretty comfortable bumping bounds for the package author (e.g. to `ForwardDiff 0.5 0.10`) - if you support ForwardDiff 0.8.0+, then you should automatically support 0.9.0 with no problems AFAICT. 

However, I'm not sure whether it's considered poor form for me to go in and alter their VERSION requirements. It would probably be more consistent for package authors to tag new versions themselves, right?

andreasnoack:
I think it's fine to alter these packages in their released versions since it's not possible to install them anyway with the bounds that you add on your package.

jrevels:
Okay, updated this PR to bump ReverseDiffSparse's ForwardDiff upper bound as previously mentioned.

I can't do a simple bump for the relevant DiffEqCallbacks versions, or for other packages in a similar situation - their ForwardDiff requirements are upper-bounded at an earlier version which is not compatible with Julia 1.0/ForwardDiff 0.9.0. I think it's correct for those versions to no longer be installable on Julia 1.0; they wouldn't work anyway.

There still some others I have to look at, I'll probably get to that tomorrow.

jrevels:
cc @mlubin since this is resulting in bumps to ReverseDiffSparse/JuMP's ForwardDiff upperbound requirement

mlubin:
JuMP shouldn't be installable on 1.0 but that's a pre-existing issue that doesn't need to be fixed here.

jrevels:
Okay, AFAIK this should be merged. I don't know how to do this, though, since CI is "failing" and a pass is required to merge.

Here are a list of package versions that are causing CI to fail because they are no longer installable on Julia 1.0 due to this change (only listing the highest uninstallable version here):

OrdinaryDiffEq v2.26.0
DelayDiffEq v1.3.0
DiffEqCallbacks v0.4.0
Klara v0.9.2
ReverseDiffSparse v0.8.0
StochasticDiffEq v2.11.1

None of these package versions should've been installable on Julia 1.0 in the first place given their ForwardDiff requirement.

In other words, this PR fixes existing problems by correctly rendering these old package versions uninstallable on Julia 1.0. AFAICT, all of these packages have releases that depend on the correct version of ForwardDiff to support Julia 1.0, so users should just update to those releases if they're on Julia 1.0. Most (but not all) of these are fairly old versions of these packages anyway.

andreasnoack:
You'd have to either add upper bounds on Julia for the packages and versions listed in the log. The only alternative is to modify the resolver. Otherwise, we'd break METADATA. Between adding uncontroversial upper bounds on other people's packages and breaking METADATA I prefer the former.

jrevels:
> either add upper bounds on Julia for the packages

Ah, good idea, didn't think of this

jrevels:
Whew, finally. Thanks for your help @andreasnoack

PR #17636, opened 2018-09-03
Register new package Tags.jl v0.0.1

attobot:
Repository: [rafaqz/Tags.jl](https://github.com/rafaqz/Tags.jl)
Release: [v0.0.1](https://github.com/rafaqz/Tags.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/rafaqz/Tags.jl/builds/423923769)
cc: @rafaqz

Please make sure that:

@rafaqz This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


nalimilan:
Wouldn't it make more sense to call the package FieldTags or something more explicit like that?

rafaqz:
Hah I've been struggling for a good name for whatever these things are for months.

It was called MetaFields.jl but that was overloading the "meta" word too much, and was really not very clear what they did either.

They are more similar to _tags_ in go than anything else I've found, and in go they are just called Tags, so I was going with that.
Although these actually do a lot more than struct tags in go, which are just a single string for each field.
https://medium.com/golangspec/tags-in-golang-3e5db0b8ef3e

They also refer to them as "struct tags" which would also work in julia.

So the options are:
Tags.jl
StructTags.jl
FieldTags.jl

nalimilan:
I don't have a clear opinion about the best name, but "field" or "struct" sound like key words here. I really hadn't guessed what the package was about after seeing Tags.jl.

rafaqz:
I'm not sure anyone will guess what it does no matter what I call it. 

I was trying for a simple/catchy name because its a simple but powerful bit of tooling one you get the concept, and people should just be able to say tags like they do in go.

But I think you are right, FieldTags.jl is more descriptive. Don't merge this I'll think about it. This would be name number 4 so it has to be the right one this time.

andreasnoack:
Sorry. Didn't notice that you'd asked that this wasn't merged. It will be reverted in https://github.com/JuliaLang/METADATA.jl/pull/17816 as soon as tests pass.

rafaqz:
Maybe its just the decisive move required. 

Any opinon on Tags.jl vs FieldTags.jl?

andreasnoack:
Haven't thought much about this but Julia's types are often called tags so using tags here might not be the best choice. You write

> This package lets you define metadata about fields in a struct

so maybe `FieldMetadata` or `FieldDescriptions` are considering. If you insist on using tags then I think `FieldTags` is the better choice.

rafaqz:
Oh man now I have even more choices to make! but it does seem like Tags.jl isn't specific enough. 

Close this and I'll resuibmit as something else

PR #17676, opened 2018-09-04
Register new package LocalCoverage.jl v0.1.0

attobot:
Repository: [tpapp/LocalCoverage.jl](https://github.com/tpapp/LocalCoverage.jl)
Release: [v0.1.0](https://github.com/tpapp/LocalCoverage.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/tpapp/LocalCoverage.jl/builds/424249161)
cc: @tpapp

Please make sure that:

@tpapp This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


andreasnoack:
CIBot fails because of the project file. Since the project file isn't needed, I'd recommend that you delete it because it will make it much easier to verify that the package works.

tpapp:
@andreasnoack: Thanks for the info. I would rather fix the project file, but I can't figure out what's wrong with it (I would appreciate help).

In fact, the JuliaCI build logs appear to install Coverage just fine, so I am not sure what is going on.

andreasnoack:
It's because CIBot still uses the old package manager which mostly works except when packages have a project file. Hence it is not possible to fix the project file. The only option currently compatible with CIBot is to delete the file. We don't require that it is deleted but I'd recommend it since it makes it easier for us to check that the package works.

tpapp:
Thanks. I am considering just waiting until the new infrastructure that uses the project files is in place.

Is there an open issue about it I can subscribe to?

andreasnoack:
> Is there an open issue about it I can subscribe to?

No, but we'll announce once project files are supported by CIBot.

andreasnoack:
I might have been able to implement a workaround for the project file issue. I've restarted the tests to see if it works.

tpapp:
It seems it cannot find itself. Perhaps a path issue?

andreasnoack:
You package explicitly uses the new package manager. CIBot isn't able to handle that yet but we are working on it.

PR #17690, opened 2018-09-04
Add upper bounds at Julia 0.7 in GLVisualize

andreasnoack:
We are currently wasting resources on reverse dependency testing of GLVisualize so this PR adds upper bounds on all released versions of the package to avoid it being tested on 0.7 and 1.0.

cc: @SimonDanisch

### DIFF
--- a/CDDLib/versions/0.0.1/requires
+++ b/CDDLib/versions/0.0.1/requires
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.7
 BinDeps
 MathProgBase
 Polyhedra 0.0.1 0.1-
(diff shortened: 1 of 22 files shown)


SimonDanisch:
fair enough, thank you!

CDDLib/versions/0.0.1/requires
```diff
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.4 0.7
```
SimonDanisch, line 1: is this an oversight?

andreasnoack: No. It's on purpose. The resolver complains if this bound isn't added. However, it shouldn't have any practical effect since this version is very old.

PR #17707, opened 2018-09-04
Add upper bounds in JuliaFEM at Julia 0.7

andreasnoack:
This will avoid that the package is tested as a reverse dependency of the JuliaFEM family of packages and it will also help users to avoid installing the package on Julia 1.0 before it is ready.

cc: @ahojukka5


ahojukka5:
Looks good to me.

PR #17724, opened 2018-09-05
Register new package Mirrors.jl v1.0.0

attobot:
Repository: [sunoru/Mirrors.jl](https://github.com/sunoru/Mirrors.jl)
Release: [v1.0.0](https://github.com/sunoru/Mirrors.jl/releases/tag/v1.0.0)
cc: @sunoru

Please make sure that:

@sunoru This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


nalimilan:
How about calling the package PkgMirrors to be more specific?

sunoru:
Yes it makes sense... However I think `Mirrors` should be obvious enough?

andyferris:
FWIW it wasn't obvious to me. Even if you discount *physical* mirrors (and mirror transformations, etc) this is only a subset of possible network mirrors, supporting only `Pkg` and that's about it.

sunoru:
OK. I renamed the package: https://github.com/JuliaLang/METADATA.jl/pull/17893

andyferris:
Thanks. Let's close this one.

PR #17732, opened 2018-09-05
Move README.md

andreasnoack:
Because GitHub limits the number of files processed from the root directory, it's been a while since the README has been visible. This PR works around the issue by moving README.md to the `.github` folder. Notice that list of files is only contracted if you don't have write access to the repo so you might want to try this out in incognito mode.

I've also added a header.


aviks (approved):

andreasnoack:
I reported the issue that caused this move to GitHub support and unfortunately they have now decided to disable the contraction behavior for the time being while looking into the issue. So even with this change users would have to scroll all the way down to the bottom. Hopefully, the contraction behavior will be back soon such that the README becomes visible again.

PR #17804, opened 2018-09-06
Tag NamedTuples.jl v5.0.0

attobot:
Repository: [JuliaData/NamedTuples.jl](https://github.com/JuliaData/NamedTuples.jl)
Release: [v5.0.0](https://github.com/JuliaData/NamedTuples.jl/releases/tag/v5.0.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaData/NamedTuples.jl/builds/425425335)
Diff: [vs v4.0.2](https://github.com/JuliaData/NamedTuples.jl/compare/4d03148530531cc1f455179db2ff1a401e283f7c...e688c56b2e74bd521dee4271c2c28bb5020d3711)
`requires` vs v4.0.2: 
```diff
--- 4.0.2/requires
+++ 5.0.0/requires
@@ -1 +1 @@
-julia 0.6 0.7
+julia 0.7
```
cc: @omus

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #17970, opened 2018-09-12
Upper bound IterableTables and Query at NamedTuples 5.0.0

ararslan:
The versions of IterableTables and Query which depend on NamedTuples but do not already impose an upper bound on it now have an explicit upper bound at NamedTuples 5.0.0, which will be introduced in #17804.


GunnarFarneback:
Travis has it wrong. This PR introduces the additional breakage seen in the latest run on #17804.

PR #17981, opened 2018-09-13
Make LazyQuery uninstallable because the repo has been deleted

andreasnoack:
@bramtayl Please stop deleting repos you've registered in METADATA without first making them uninstallable. Just deleting them causes problems for our infrastructure.


bramtayl:
Ok will do

PR #18026, opened 2018-09-14
Tag AmplNLWriter.jl v0.4.1

attobot:
Repository: [JuliaOpt/AmplNLWriter.jl](https://github.com/JuliaOpt/AmplNLWriter.jl)
Release: [v0.4.1](https://github.com/JuliaOpt/AmplNLWriter.jl/releases/tag/v0.4.1)
Travis: [Travis Build Status](https://travis-ci.org/JuliaOpt/AmplNLWriter.jl/builds/428765732)
Diff: [vs v0.4.0](https://github.com/JuliaOpt/AmplNLWriter.jl/compare/6a81ccb2434eee5aca69feaedd77bf7b38f4502f...c3ab02096a96122bcad0b04067990611d0492c41)
cc: @odow

Please make sure that:


andreasnoack:
Tests are failing on 0.7 and 1.0.

odow:
This patch is a bug fix for existing users on 0.6. 

Updating to 0.7 an 1.0 would be nice, but I don't have the time to do so right now, and I don't see why this should be a blocker for an existing bug affecting existing users.

andreasnoack:
That is fine but then I'd recommend adding an upper bound on Julia. It is much nicer to get an error when trying to install the package. (It would also make it possible to merge this PR without human interaction since tests would pass.)

odow:
If I add an upper bound to this tag, won't it just install an earlier version without the upper bound? In which case, won't I need to modify all of the existing tags in METADATA to add an appropriate upper bound?

andreasnoack:
Yes. Adding bounds on all previously released versions is the proper way of doing this.

andreasnoack:
No activity for a month so closing

PR #18067, opened 2018-09-17
Upper bound for Hecke

thofma:
Preparation for new releases.


andreasnoack:
If you plan to make a chance to Nemo that will break Hecke then it is probably better style to make it `0.10.0` instead of `0.9.2`.

thofma:
OK, will do.

PR #17800, opened 2018-09-06
Tag NLPModels.jl v0.5.0

attobot:
Repository: [JuliaSmoothOptimizers/NLPModels.jl](https://github.com/JuliaSmoothOptimizers/NLPModels.jl)
Release: [v0.5.0](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/releases/tag/v0.5.0)
Diff: [vs v0.4.3](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/compare/b0e2a2975577e1ca0b70e80bddbc04a1dd554a32...27d948f714b9443d161e7afb21fb61e9c9d1cf19)
`requires` vs v0.4.3: 
```diff
--- 0.4.3/requires
+++ 0.5.0/requires
@@ -1,2 +1,3 @@
 julia 0.6
 LinearOperators 0.4.0
+ForwardDiff 0.6.0
```
cc: @abelsiqueira

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #16242, opened 2018-08-03
Tag AmplNLReader.jl v0.2.0

attobot:
Repository: [JuliaSmoothOptimizers/AmplNLReader.jl](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl)
Release: [v0.2.0](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/releases/tag/v0.2.0)
Diff: [vs v0.1.0](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/compare/ec4c1f77562193e5b917218d45ac8e1f097bbe2a...90f498fc38bbf90dadd98d7aa9ee5327bbd5adc5)
`requires` vs v0.1.0: no changes
cc: @dpo

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #18099, opened 2018-09-18
Tag AmplNLReader.jl v0.2.1

attobot:
Repository: [JuliaSmoothOptimizers/AmplNLReader.jl](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl)
Release: [v0.2.1](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/releases/tag/v0.2.1)
Diff: [vs v0.1.0](https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/compare/ec4c1f77562193e5b917218d45ac8e1f097bbe2a...5af4c4d1550ec496f02bb09e25b3ee74468537a3)
`requires` vs v0.1.0: 
```diff
--- 0.1.0/requires
+++ 0.2.1/requires
@@ -1,5 +1,4 @@
-julia 0.4
+julia 0.6 0.7
 BinDeps
-Compat 0.7.15
 @osx Homebrew 0.4.0
-NLPModels 0.0.1
+NLPModels 0.0.1 0.5.0
```
cc: @abelsiqueira

Please make sure that:


dpo:
Not sure why the JuliaCIBot is testing Julia ≥ 0.6 if we have set an upper bound?!

abelsiqueira:
True, though I think these are allowed fails.
However, it is running 0.6 on Linux, and thus is failing for 'new'. But we don't intend to support Linux 0.6 anymore. Any way to mark that?

andreasnoack:
This was not what I meant in https://github.com/JuliaLang/METADATA.jl/pull/17800#issuecomment-422081556. Please check the PR that I linked to. You should just add an upper bound on NLPModels in https://github.com/JuliaLang/METADATA.jl/blob/metadata-v2/AmplNLReader/versions/0.1.0/requires. I'll close this PR. Please delete the tag 0.2.1 tag in the ApmlNLReader repo.

dpo:
AmplNLReader is only failing because of NLPModels for anecdotal reasons. The C interface breaks on certain Linux machines with Julia 0.6 (and not 0.7) but we haven't been able to determine the cause. Tag 0.2.1 removes support for the Linux / Julia 0.6 combination. Linux support will return together with Julia 0.7 support.

andreasnoack:
Okay. Please remove and reapply the tag if you still want to release this version. (Btw why 0.2.1 when the previous version was 0.1.0?)

abelsiqueira:
v0.2.0: https://github.com/JuliaLang/METADATA.jl/pull/16242

We could delete both and re-release 0.2.0. @dpo is that ok?

andreasnoack:
Yes. It usually preferred to update tags that haven't been released to METADATA.

dpo:
Sure, that sounds good. Thanks both.

PR #18131, opened 2018-09-19
Register new package JuMPNLPModels.jl v0.1.0

attobot:
Repository: [JuliaSmoothOptimizers/JuMPNLPModels.jl](https://github.com/JuliaSmoothOptimizers/JuMPNLPModels.jl)
Release: [v0.1.0](https://github.com/JuliaSmoothOptimizers/JuMPNLPModels.jl/releases/tag/v0.1.0)
cc: @abelsiqueira

Please make sure that:

@abelsiqueira This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


mlubin:
@abelsiqueira, would you consider tweaking the name of this package? The current name could easily be read as meaning that the repository is part of the JuMP project, which it is not. Since it's the interface between NLPModel and JuMP, maybe something like `NLPModelsJuMPInterface`?

abelsiqueira:
cc @dpo

dpo:
@mlubin This package used to be called NLPJuMPModels. Would that have been better to avoid the potential confusion?

mlubin:
@dpo I don't want to make a rash judgement on this, so let me check with the JuMP developers and we'll get back to you quickly. For reference, the only other registered packages with JuMP in the name that are unaffiliated with the JuMP developers are [MultiJuMP](https://github.com/anriseth/MultiJuMP.jl) and [ConditionalJuMP](https://github.com/rdeits/ConditionalJuMP.jl). In both cases we asked the authors to put a statement in the readme clarifying that they're not developed or maintained by the JuMP developers.

dpo:
@mlubin Ok, thanks. I personally like the current name but we're of course happy to add a note that the repo isn't affiliated with / maintained by the JuMP developers.

abelsiqueira:
README is updated with the disclaimer. I'll wait for confirmation before updating the tag.

mlubin:
Given the precedent, we are ok if

1. the README contains a disclaimer similar to what you already added, and
2. the package name does not begin with JuMP

So `NLPJuMPModels` is fine, but we'd object to `JuMPNLPModels`.

dpo:
I'd go with NLPModels_JuMP or something like that then, because NLPModels is the name of another package and shouldn't be split. @abelsiqueira Does that sound ok?

abelsiqueira:
I think `NLPModelsJuMP` follows some unspoken guideline of avoiding `_`. But I'm happy either way.

dpo:
Ok for me.

abelsiqueira:
New PR: https://github.com/JuliaLang/METADATA.jl/pull/18291

This can be closed.

dpo:
@abelsiqueira You can probably close this to avoid any confusion.

abelsiqueira:
I can't because @attobot is the PR owner. Maybe @andreasnoack can close for us, please.

PR #18161, opened 2018-09-20
Tag AWSS3.jl v0.5.0

attobot:
Repository: [samoconnor/AWSS3.jl](https://github.com/samoconnor/AWSS3.jl)
Release: [v0.5.0](https://github.com/samoconnor/AWSS3.jl/releases/tag/v0.5.0)
Diff: [vs v0.4.0](https://github.com/samoconnor/AWSS3.jl/compare/7d1c9a84180b91638b6f72bf30213ded1ef49eef...8bfbfb59cc28965170d004ef86cdf621823fe1c9)
`requires` vs v0.4.0: 
```diff
--- 0.4.0/requires
+++ 0.5.0/requires
@@ -1,5 +1,5 @@
-julia 0.7.0-beta
-AWSCore 0.3.4
+julia 1.0
+AWSCore 0.5
 HTTP
 SymDict
 Retry
```
cc: @samoconnor

Please make sure that:


samoconnor:
CI passing now

andreasnoack:
Please don't comment on passing PRs unless you want to avoid them being merged. It breaks the automatic merging functionality.

samoconnor:
ok, i only commented because everything was green but it had not automatically merged. Maybe it would help if attobot posted some verbose comments to make it clear if it is waiting on me, or if it is just waiting for some background processing of its own...

andreasnoack:
Currently, we merge at regular intervals but once I have some cycles it should instead be triggered by the tests passing so at that point the time between green and merging should be really short.

samoconnor:
ok, sounds good.
it sure would be nice if the travis scripts were smart enough to:
 - notice that AWSS3 is not building because the required version of AWSCore does not exist yet
 - cancel the build instead of failing it
 - restart the build later on when AWSCore has beed CI-ed and merged into METADATA

I feel like every time I do some updates to half a dozen packages I have to spend a few hours babysitting them through the process of getting all the CI passing and all the METADATA PRs happy. That is all after I have all the test running and passing locally. It makes travis feel like a burdensome hurdle rather than a useful tool.

All that being said. I do appreciate all the time that has been invested in tooling recently. Keep it up!

andreasnoack:
Being smart about restarting PRs could be really useful for dependent PRs but I also suspect it would require some work to get right. Hence, the short-term goal is to make it simpler for package authors to get PRs through without interacting with me or another METADATA maintainer. I think we are almost there. Once the merging is automatically triggered, we just need to document it a bit. It is also worth mentioning that, eventually, we'll migrate to a release process with PRs directly against a registry instead of METADATA so while we hope that the tooling carries over to the new system, we'll probably not implement new features (beyond what I just mentioned) before we have transitioned.

PR #18180, opened 2018-09-20
Register new package List.jl v0.1.0

attobot:
Repository: [c-p-murphy/List.jl](https://github.com/c-p-murphy/List.jl)
Release: [v0.1.0](https://github.com/c-p-murphy/List.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/c-p-murphy/List.jl/builds/431155804)
cc: @c-p-murphy

Please make sure that:

@c-p-murphy This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


dpsanders:
This package name is too generic. How about MutableLinkedList.jl? Or maybe this functionality could go into the DataStructures.jl package?

c-p-murphy:
Sure, I am happy to do either of those. I first put this together in response to an [issue](https://github.com/JuliaCollections/DataStructures.jl/issues/443) over in DataStructures.jl, but it didn't seem like it was what they were looking for so I figured I would just publish it as a stand-alone package. I can submit it as a PR over in DataStructures.jl if that makes more sense though.

dpsanders:
There's also 
https://github.com/ChrisRackauckas/LinkedLists.jl
which may be similar?

c-p-murphy:
Yeah, I saw that but it seemed like it doesn't work with the update to v0.7/v1.0, so I wasn't sure if it was being actively maintained. I guess another option would be to open a PR over there.

c-p-murphy:
I updated the name and it opened another PR here instead of updating this, but I am still open to doing either of the other options if one of those seems like it makes more sense. Just let me know.

ararslan:
> but it didn't seem like it was what they were looking for

Seems fine to include in DataStructures with a different type name. AFAICT the only objection was to changing the existing linked list type to be mutable. Adding a `MutableLinkedList` type to that package rather than splitting it into its own seems preferable.

c-p-murphy:
Ok, that makes sense. I'll do that and submit a PR over there.

PR #18376, opened 2018-09-25
Register HurdleDMR [https://github.com/AsafManela/HurdleDMR.jl.git]: …

AsafManela:
…v0.0.1-alpha.1, v0.0.2, v0.1.0

Registering now so that v0.1.0 will be available to julia 0.6 users while subsequent versions will support julia 0.7 and up.


andreasnoack:
Please use https://github.com/attobot/attobot for registering your package.

PR #18440, opened 2018-09-27
Tag StaticPolynomials.jl v1.0.0

attobot:
Repository: [JuliaAlgebra/StaticPolynomials.jl](https://github.com/JuliaAlgebra/StaticPolynomials.jl)
Release: [v1.0.0](https://github.com/JuliaAlgebra/StaticPolynomials.jl/releases/tag/v1.0.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaAlgebra/StaticPolynomials.jl/builds/434049895)
Diff: [vs v0.3.0](https://github.com/JuliaAlgebra/StaticPolynomials.jl/compare/e9c99143c23e151ade76a85fdae36e879662c1c4...b7be41e801d0040d9cce7df92b1a5efe40e68b24)
`requires` vs v0.3.0: no changes
cc: @saschatimme

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #18523, opened 2018-10-01
Cap dependencies of HomotopyContinuation

saschatimme:
This caps the dependencies of HomotopyContinuation for all versions which support 0.7 to address the CI failure in #18440.

### DIFF
--- a/HomotopyContinuation/versions/0.3.0/requires
+++ b/HomotopyContinuation/versions/0.3.0/requires
@@ -1,7 +1,7 @@
 julia 0.7
 FixedPolynomials 0.4.0
 ProgressMeter 0.6.0
-StaticPolynomials 0.3.0
+StaticPolynomials 0.3.0 0.4.0
 DynamicPolynomials 0.1.0
 MultivariatePolynomials 0.2.0
 TreeViews 0.3.0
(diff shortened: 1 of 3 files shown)


andreasnoack:
Please just cap `StaticPolynomials` at version 1.0. It should be sufficient.

saschatimme:
But I don't want to run into the same problems again. What is the proper way to deal with this?

andreasnoack:
Having too many upper bounds might cause problems for users of your package since they might be incompatible with the requirements of other packages so I'd try to keep the number of upper bounds low. In particular, adding upper bounds after a release should only be done once breakage has been confirmed. You can add upper bounds to master of your package but please try not to make them too narrow.

saschatimme:
> Having too many upper bounds might cause problems for users of your package since they might be incompatible with the requirements of other packages so I'd try to keep the number of upper bounds low.

Isn't this what the new environments in Pkg are supposed to solve?

> In particular, adding upper bounds after a release should only be done once breakage has been confirmed.

I agree that this is not great (and I changed it for now), but even if I add upper bounds to master (which I already did), won't the old releases still hunt me for every breaking change?

andreasnoack:
> Isn't this what the new environments in Pkg are supposed to solve?

Yes. But it only works if there aren't conflicting requirements within an environment. This might not always be the case for individual projects and much REPL work will continue to happen in the default (global) environment.

saschatimme:
I disagree with this, so I made a post at [discourse](https://discourse.julialang.org/t/dependency-constraints-in-packages/15739) to have a more proper discussion place.
For now I can live with just capping StaticPolynomials for the moment (the PR already reflects this).

PR #18748, opened 2018-10-10
Register new package StructuralInheritance.jl v0.1.0 [https://github.…

WschW:
…com/WschW/StructuralInheritance.jl]


andreasnoack:
Please consider using Attobot for registering your package. It is the recommended way of doing it since it is less error prone and your PR will be merged automatically.

WschW:
Thanks, I will look into that

PR #18778, opened 2018-10-11
Tag MatrixMarket.jl v0.3.0

tkonolige:


andreasnoack:
Please use Attobot for releases. Then the PR will get merged automatically and it will be much easier for all of us.

tkonolige:
I would but I don't have permission to add it to the repository. I'll ping someone who does.

PR #18955, opened 2018-10-20
Register new package NumericalAnalysis.jl v0.0.1

attobot:
Repository: [chakravala/NumericalAnalysis.jl](https://github.com/chakravala/NumericalAnalysis.jl)
Release: [v0.0.1](https://github.com/chakravala/NumericalAnalysis.jl/releases/tag/v0.0.1)
Travis: [Travis Build Status](https://travis-ci.org/chakravala/NumericalAnalysis.jl/builds/443948383)
cc: @chakravala

Please make sure that:

@chakravala This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


dpsanders:
This seems like a package name that is too generic. Is there something more specific that you could name it?

chakravala:
The ongoing intentions of this package are encapsulated by the name `NumericalAnalysis`, I'm not sure it makes sense to give it any other name, since numerics and analysis are the main usage, (not in the applied sense but in the theoretical sense), it is used for analysis of error bounds and for observing mathematical properties, quite literally, Numerical Analysis in the sense of the abstract topic that pure mathematicians study, to determine theoretical properties of error bounds (not applied math, but theoretical investigation of numerical error bounds and their analysis).

This first version of 0.0.1 only handles one type of analysis yet, relating to the rounding error propagation of floating point calculations for polynomials. However, other types of analysis will be added to it. It is now time to start releasing the package, because I am publishing a paper which uses this analysis.

I have already submitted my paper to the Advances in Engineering Mathematics journal and used the name NumericalAnalysis.jl there already, linking to it on github.

So I would prefer to keep the name NumericalAnalysis, since it is specifically intended to expand on the Numerical Analysis topic and the name is already submitted in my paper.

This package will also host some teaching and learning material for NumericalAnalysis students, so it is also intended to be a complementary classroom tool for Numerical Analysis students to help visualize the concepts that are learned in that type of course-work.

Hope that clarifies why the name is chosen.

Is your only concern that the name is generic? If that's the only concern about it, then I would prefer to keep this name, since it was an intentional choice. If there is another aspect behind your concerns with the naming convention, please elaborate.

PS: is there an issue with depending on `Printf` from `stdlib`?

simonbyrne:
The problem is that your package only targets a single problem (rounding error propagation of floating point calculations for polynomials), using a single method. While I don't doubt your intention to expand the functionality, numerical analysis is an incredibly broad topic, well beyond what could be captured by a single package.

Although we don't have formal policies, we do encourage packages to use names which describe the functionality rather than the field or topic. There are some exceptions, such as DifferentialEquations.jl, but these tend to be extensive compendiums of other packages.

> I have already submitted my paper to the Advances in Engineering Mathematics journal and used the name NumericalAnalysis.jl there already, linking to it on github.

I'm sure the editors would be happy to let you change the name, and GitHub automatically redirects renamed repositories.

ViralBShah:
This is quite a broad name. Almost every package in Julia does numerical analysis! 

I guess another option is to not register the package if it must have this name  and have users clone it directly.

chakravala:
> This is quite a broad name. Almost every package in Julia does numerical analysis!

There seems to be some confusion about the meaning of the word *analysis* here. While many packages of Julia rely on Numerical Analysis *concepts* and use Numerical Computations, all these packages are not concerned with the **Analysis** aspect, hence they are not `NumericalAnalysis` packages.

This package is named Numerical**Analysis** because it specifically is for the analysis of the mathematical theory of floating point arithmetic and approximation error itself. So this is an actual NumericalAnalysis package, while the other julia packages are typically *applications* of numerical analysis.

The other packages you refer to don't typically do Analysis, as this package does. They assume concepts obtained from Analysis, but they don't fundamentally explore the mathematical analysis and theory. That's why this package is appropriately named.

So, `NumericalAnalysis` here is used in the Mathematician sense of the term, not in the *applied* sense of the term. This is a package for mathematicians who wish to gain more insight about theory and foundations of `NumericalAnalysis` itself, as opposed to applying it specifically.

This is not a package that *does* numerical tasks, rather it is a package that explores underlying *theory*

timholy:
There are a ton of packages concerned with "the analysis of the mathematical theory of floating point arithmetic and approximation error": see IntervalArithmetic, DoubleDouble, ArbNumerics, etc. The name here is definitely too broad.

chakravala:
Would `ErrorBounds` be an acceptable name?

The package does a lot more than just compute error bounds, so it's not a name that reflects it.

This is gonna be a problem, because if I change the name, then the code from my paper no longer works, and the reviewers may not recognize that the name change has occured. The name change might need to wait until the review period is over then.

timholy:
ErrorBounds sounds much better, though I'd defer to people who've spent more time working in this area.

GitHub does do [redirects](https://help.github.com/articles/renaming-a-repository/), so hopefully you should be fine. Or wait to register until review is over, as you say.

ViralBShah:
ErrorBounds.jl is certainly much better!

chakravala:
Actually, I think the name will be changed to either `Wilkinson` or  `WilkinsonBounds`, in honor of the pioneer of the type of error analysis used, since `ErrorBounds` is kind of general also.

https://en.m.wikipedia.org/wiki/James_H._Wilkinson

ViralBShah:
I think `WilkinsonBounds` is better than `Wilkinson`. Are the techniques used generally referred to as Wilkinson Bounds in the literature, or something similar?

andreasnoack:
No activity for a month so closing

PR #19050, opened 2018-10-23
Add StaticArrays v0.9 upper cap for Plots

blegat:
See https://github.com/JuliaPlots/Plots.jl/issues/1802.
I don't understand why this wasn't catched by the CIBot in https://github.com/JuliaLang/METADATA.jl/pull/19043.
It [seems it didn't run the Plots tests](https://github.com/JuliaLang/METADATA.jl/pull/19043) while StaticArrays is in its dependencies.

### DIFF
--- a/Plots/versions/0.12.1/requires
+++ b/Plots/versions/0.12.1/requires
@@ -4,7 +4,7 @@ RecipesBase 0.2.0
 PlotUtils 0.4.1
 PlotThemes 0.1.3
 Reexport
-StaticArrays 0.5
+StaticArrays 0.5 0.9
 FixedPointNumbers 0.3
 Measures
 Showoff
--- a/Plots/versions/0.12.2/requires
+++ b/Plots/versions/0.12.2/requires
@@ -4,7 +4,7 @@ RecipesBase 0.2.0
 PlotUtils 0.4.1
 PlotThemes 0.1.3
 Reexport
-StaticArrays 0.5
+StaticArrays 0.5 0.9
 FixedPointNumbers 0.3
 Measures
 Showoff
--- a/Plots/versions/0.12.3/requires
+++ b/Plots/versions/0.12.3/requires
@@ -4,7 +4,7 @@ RecipesBase 0.2.0
 PlotUtils 0.4.1
 PlotThemes 0.1.3
 Reexport
-StaticArrays 0.5
+StaticArrays 0.5 0.9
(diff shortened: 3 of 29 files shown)


mkborregaard:
Great. It's probably due to https://github.com/JuliaCI/PackageEvaluator.jl/issues/164 that it didn't run.

fredrikekre:
Wouldn't it be better to release StaticArrays 0.9.1?

blegat:
StaticArrays has deprecated its FixedSizeArrays submodule which breaks Plots until it is updated.
Currently, all users doing `] up` will end up with Plots not working anymore so this PR at least prevent this until Plots is updated. Plots is a widely used package so this PR should prevent many breakages.

fredrikekre:
Right but it is only deprecated so it shouldnt break stuff. Should fix StaticArrays too at least.

blegat:
> Right but it is only deprecated so it shouldnt break stuff. Should fix StaticArrays too at least.

Yes but they renamed it `FixedSizeArraysWillBeRemoved`. Apparently they wanted to break code but leave the opportunity to still use the submodule. I agree that a deprecation warning might have made the transition smoother.

mkborregaard:
Daniel pushed a fix for this in Plots https://github.com/JuliaPlots/Plots.jl/pull/1803

c42f:
> Apparently they wanted to break code but leave the opportunity to still use the submodule

```julia
julia> using StaticArrays.FixedSizeArrays
WARNING: StaticArrays.FixedSizeArrays is deprecatedStaticArrays.FixedSizeArrays is deprecated. Use StaticArrays directly.
 in module StaticArrays

julia> Vec(1,2,3)
3-element Vec{3,Int64}:
 1
 2
 3
```

We *do* continue to provide a `StaticArrays.FixedSizeArrays` binding via `@deprecate_binding`.

So I don't understand what went wrong here, yet.

c42f:
Proposed fix: https://github.com/JuliaArrays/StaticArrays.jl/pull/527

c42f:
The reason for renaming these modules at all was simply so that we can get a proper deprecation warning via `@deprecate_binding` which is not exactly designed for deprecating bindings which are going to be removed entirely, and requires a "new" name. So I gave the old modules a new (nonsense) name which is what has backfired here, and pointed the old name at the new one using `@deprecate_binding`.

PR #19081, opened 2018-10-24
Register Unidecode

matthieugomez:
I wrote this package because I was looking for a way to revert unicode characters to UTF-8, but please tell me if there is a better way!
https://github.com/matthieugomez/Unidecode.jl/blob/master/src/Unidecode.jl


andreasnoack:
It would be great if you could use Attobot instead. See https://discourse.julialang.org/t/the-current-metadata-release-process/16672

PR #19114, opened 2018-10-25
Tag GroupedErrors.jl v0.2.0

attobot:
Repository: [piever/GroupedErrors.jl](https://github.com/piever/GroupedErrors.jl)
Release: [v0.2.0](https://github.com/piever/GroupedErrors.jl/releases/tag/v0.2.0)
Travis: [Travis Build Status](https://travis-ci.org/piever/GroupedErrors.jl/builds/446172575)
Diff: [vs v0.1.4](https://github.com/piever/GroupedErrors.jl/compare/46325e8053888f5946f01e1472f61b3962bb861f...2f8c13c3004a6402ad8650047ebebe7e8a26a668)
`requires` vs v0.1.4: 
```diff
--- 0.1.4/requires
+++ 0.2.0/requires
@@ -1,12 +1,13 @@
-julia 0.6
+julia 0.7
 
 StatsBase
-KernelDensity
+KernelDensity 0.5.1
 Loess
 IndexedTables 0.4.1
 IterableTables 0.5.1
 TableTraitsUtils 0.1.0
 TableTraits
 MacroTools
-Lazy
+Lazy 0.13.2
 ShiftedArrays 0.3.0
+DataValues
```
cc: @piever

Please make sure that:


piever:
Not sure what's happening with the CI. It seems it didn't pick the right version of GroupedErrors to run:  for instance it is complaining about using `NamedTuples` without an explicit dependency but I've removed `using NamedTuples` in this release.

andreasnoack:
For some reason it doesn't pick the latest version of `KernelDensity`. Maybe because something requires latest `Interpolations` which isn't supported by `KernelDensity` 0.5.1 https://github.com/JuliaLang/METADATA.jl/blob/aa47041b1065afc9fd26ec005a6b35c2a85162f9/KernelDensity/versions/0.5.1/requires#L5.

but I can really figure out which package that is requiring `Interpolations`. Hopefully, it isn't the resolver that gets it wrong.

piever:
That's strange: the configuration must be solvable, because I have it locally on my machine (with GroupedErrors master, KernelDensity 0.5.1 and Interpolations 0.9.2). If that's helpful, there's a valid configuration on Travis too (see [latest build](https://travis-ci.org/piever/GroupedErrors.jl/jobs/446172577#L544)).

andreasnoack:
I'll try to go through the same steps that CIBot does to generate the registry. Maybe that will show what is causing this.

andreasnoack:
Hm. This doesn't happen locally. It's really puzzling. I'll try to restart the tests. If it happens again then I'd have to try this locally in a clean environment.

andreasnoack:
@piever Could you try to remove and reapply the tag?

andreasnoack:
I've tried to update the registry generation to use the same version of `Pkg` that the actual registry uses. This has been tested in https://github.com/andreasnoack/METADATA.jl/pull/14 and for it seemed to fix the issue on 1.0 but the issue remains on 0.7. I don't know why it happens but suspect it is related to the upper bound on `Interpolations` in `KernelDensity`.

If this passes on 1.0 then I'll merge and then we can try it out on 0.7 once it is in the registry and see if there is a real conflict.

andreasnoack:
Let's see what happens.

andreasnoack:
It's not just a CIBot thing. It's real. Locally, I just got
```julia
(v0.7) pkg> test GroupedErrors
   Testing GroupedErrors
 Resolving package versions...
 Installed KernelDensity ── v0.5.0
 Installed GroupedErrors ── v0.1.4
 Installed Interpolations ─ v0.10.5
    Status `/var/folders/h6/f69m8fqj1md41j4dndvs8cnw0000gn/T/tmplCnRew/Manifest.toml`
```
i.e. the same downgrading as we saw on CIBot.

andreasnoack:
I think I can make some sense of this now. So the reason is that because of 
https://github.com/JuliaLang/METADATA.jl/blob/aa47041b1065afc9fd26ec005a6b35c2a85162f9/KernelDensity/versions/0.5.1/requires#L5 there are two solution which probably don't have a clear ordering between them. The one we expected is
```julia
KernelDensity  v0.5.1
Interpolations v0.9.2
GroupedErrors  v0.2.0
```
but there is also
```julia
KernelDensity  v0.5.0
Interpolations v0.10.5
GroupedErrors  v0.1.4
```
and actually the latter has the highest bugfix version sum (with the same minor version sum). What contributes to the confusion is that the resolver picks the first on when `Pkg.add` is run and the other when `Pkg.test` is run. Probably because of the completely unrelated test dependency.

piever:
Thanks for figuring this out! I guess the issue arises because I didn't specify in the REQUIRE of previous versions that they are incompatible with Julia 0.7 and above

PR #19071, opened 2018-10-24
Tag TimeSeries.jl v0.14.0

attobot:
Repository: [JuliaStats/TimeSeries.jl](https://github.com/JuliaStats/TimeSeries.jl)
Release: [v0.14.0](https://github.com/JuliaStats/TimeSeries.jl/releases/tag/v0.14.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/445450733)
Diff: [vs v0.13.0](https://github.com/JuliaStats/TimeSeries.jl/compare/93d1890820ddfe1abc30c997aa437707ffa15efc...b98c9681c6281770c86b58da14b2c1aa49798410)
`requires` vs v0.13.0: no changes
cc: @iblis17

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #19137, opened 2018-10-26
Add upper bound for MarketData and TimeSeriesResampler at TimeSeries

iblislin:
To fix JuliaCIBot failure at https://github.com/JuliaLang/METADATA.jl/pull/19071
cc @femtotrader

### DIFF
--- a/MarketData/versions/0.8.0/requires
+++ b/MarketData/versions/0.8.0/requires
@@ -1,4 +1,4 @@
 julia 0.6.3
-TimeSeries 0.9.0
+TimeSeries 0.9.0 0.14.0
 HTTP 0.6.12
 Reexport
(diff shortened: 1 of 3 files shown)


iblislin:
TimeSeries v0.14.0 contains breaking changes https://github.com/JuliaLang/METADATA.jl/pull/19071
So I set the upper bound to v0.14.0.

iblislin:
Thank @andreasnoack for patient guiding!

andreasnoack:
Shouldn't there have been a bound on version 0.10.0 as well?

iblislin:
@andreasnoack you are right.
(I'm not ware that there is 0.10.0 under `ls` :man_facepalming:

PR #19233, opened 2018-10-31
DataDeps requires HTTP 0.7

oxinabox:
I updated the Project.toml without updating the REQUIRE

this is reconning
https://github.com/oxinabox/DataDeps.jl/pull/77


oxinabox:
@fredrikekre  this is what I need to do, right?

fredrikekre (approved):

PR #19526, opened 2018-11-14
Tag StatPlots.jl v0.8.2

attobot:
Repository: [JuliaPlots/StatPlots.jl](https://github.com/JuliaPlots/StatPlots.jl)
Release: [v0.8.2](https://github.com/JuliaPlots/StatPlots.jl/releases/tag/v0.8.2)
Travis: [Travis Build Status](https://travis-ci.org/JuliaPlots/StatPlots.jl/builds/455208943)
Diff: [vs v0.8.1](https://github.com/JuliaPlots/StatPlots.jl/compare/05c7cc3c1f88de334c0eac0e7cf9f57ea082fa3f...51b23123fef0bd84a8d6a957bee5238d1830c72b)
`requires` vs v0.8.1: no changes
cc: @mkborregaard

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #19570, opened 2018-11-16
cap statplots on forestbiometrics

piever:
I've added a cap to StatPlots in the ForestBiometrics package, to prevent breakage from https://github.com/JuliaLang/METADATA.jl/pull/19526

### DIFF
--- a/ForestBiometrics/versions/0.1.1/requires
+++ b/ForestBiometrics/versions/0.1.1/requires
@@ -1,6 +1,6 @@
 julia 0.6.0
 DataFrames 0.10.1
 Plots 0.13.1
-StatPlots 0.5.0
+StatPlots 0.5.0 0.8.1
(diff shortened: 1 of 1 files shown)


ForestBiometrics/versions/0.1.1/requires
```diff
@@ -1,6 +1,6 @@
 julia 0.6.0
 DataFrames 0.10.1
 Plots 0.13.1
-StatPlots 0.5.0
+StatPlots 0.5.0 0.8.1
```
andreasnoack, line 4: Shouldn't this be `0.8.2`? The right end point is excluded.

piever: 0.8.1 also breaks ForestBiometricst (but the PR got stale in METADATA).

PR #19557, opened 2018-11-15
Tag Unitful.jl v0.13.0

attobot:
Repository: [ajkeller34/Unitful.jl](https://github.com/ajkeller34/Unitful.jl)
Release: [v0.13.0](https://github.com/ajkeller34/Unitful.jl/releases/tag/v0.13.0)
Diff: [vs v0.12.0](https://github.com/ajkeller34/Unitful.jl/compare/0526f0c40599b82407e68246512506d38c227632...f88830253ecbce93420d5543c48cc731cf91f3e8)
`requires` vs v0.12.0: no changes
cc: @ajkeller34

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #19575, opened 2018-11-16
Upper-bound Unitful version for ImagineFormat

ajkeller34:
This PR is prep work for a version of Unitful with breaking changes (https://github.com/JuliaLang/METADATA.jl/pull/19557). I will make individual PRs for the [very few] packages that the breaking changes affect.

### DIFF
--- a/ImagineFormat/versions/0.2.0/requires
+++ b/ImagineFormat/versions/0.2.0/requires
@@ -3,5 +3,5 @@ AxisArrays
 ImageMetadata 0.1.1
 FixedPointNumbers 0.3.0
 FileIO
-Unitful
+Unitful 0.0 0.13-
 Ranges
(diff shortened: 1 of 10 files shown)


andreasnoack:
What about the previous versions of ImagineFormat?

ajkeller34:
Since the previous versions had no bounds on Unitful, is it fair to give the first Unitful release as the lower bound? Or do I need to go back and see whether each version of ImagineFormat actually works with the oldest versions of Unitful?

andreasnoack:
You can just use `0.0` since that is what an empty entry means.

timholy:
Depending on the number of releases (I honestly can't remember), [`propagate_upperbounds`](https://github.com/JuliaLang/PkgDev.jl/pull/94) can sometimes be useful. Requires julia-0.6 and checking out that branch.

ajkeller34:
Okay, sorry for the noise and thanks for the help. I've used 0.0 as the lower bound and 0.13- as the upper bound for all versions using Unitful, since none of them had bounds previously.

PR #19576, opened 2018-11-16
Upper-bound Unitful version for UnitfulAngles

ajkeller34:
This PR is prep work for a version of Unitful with breaking changes (#19557). I will make individual PRs for the [very few] packages that the breaking changes affect.


andreasnoack:
Do all previous versions have upper bounds?

ajkeller34:
Now they do. Lower bound is 0.0 since previously there were no bounds on the Unitful versions.

PR #19641, opened 2018-11-19
Tag StateSpaceRoutines.jl v0.1.2

attobot:
Repository: [FRBNY-DSGE/StateSpaceRoutines.jl](https://github.com/FRBNY-DSGE/StateSpaceRoutines.jl)
Release: [v0.1.2](https://github.com/FRBNY-DSGE/StateSpaceRoutines.jl/releases/tag/v0.1.2)
Diff: [vs v0.1.1](https://github.com/FRBNY-DSGE/StateSpaceRoutines.jl/compare/1fd8e4e279b082413ef88947026f7d3dcbbbe926...e9d21bee06e6786de02b0dc704c22654334baf54)
`requires` vs v0.1.1: 
```diff
--- 0.1.1/requires
+++ 0.1.2/requires
@@ -3,3 +3,4 @@
 Roots
 QuantEcon
 HDF5
+JLD
```
cc: @caimichael

Please make sure that:


caimichael:
Hello,

I was hoping someone could help me clarify why the JuliaCIBot says that the tests fail for Julia 0.6 (previous and current), even if the output of the CI builds indicates that the tests passed on both builds. Also, this version is not meant to work on Julia 0.7 or Julia 1.0, so I don't think the CI should try to build on those versions.

Michael

andreasnoack:
Unfortunately, we have had quite a few of these cases over the last couple of days. We still don't know why.

Another thing, if the package doesn't support Julia 0.6 then please add an upper bound in the REQUIRE file.

caimichael:
Hi Andreas. I've added the upper bound to the REQUIRE file. Is the proper protocol for me to create a new "patch" tag, which includes this new commit so the CI can run again? Or is there a way to just amend the current tag with this new commit?

andreasnoack:
The latter. Since this PR hasn't been merged yet you can just delete the release and use the new commit for the same release.

PR #19433, opened 2018-11-10
Tag StaticArrays.jl v0.10.0

attobot:
Repository: [JuliaArrays/StaticArrays.jl](https://github.com/JuliaArrays/StaticArrays.jl)
Release: [v0.10.0](https://github.com/JuliaArrays/StaticArrays.jl/releases/tag/v0.10.0)
Travis: [Travis Build Status](https://travis-ci.org/JuliaArrays/StaticArrays.jl/builds/453305584)
Diff: [vs v0.9.2](https://github.com/JuliaArrays/StaticArrays.jl/compare/1683f798e0dcb8725768407053a45ae48434fed7...a46d711c568e4a42b755947267b6297c5ea8913c)
`requires` vs v0.9.2: no changes
cc: @ChrisRackauckas

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #19688, opened 2018-11-22
Upper-bound StaticArrays for BEAST

timholy:
Ref #19433. CC @krcools


krcools:
Thank you for fixing this!

PR #19689, opened 2018-11-22
Register new package QRUpdate.jl v0.1.0

attobot:
Repository: [haampie/QRUpdate.jl](https://github.com/haampie/QRUpdate.jl)
Release: [v0.1.0](https://github.com/haampie/QRUpdate.jl/releases/tag/v0.1.0)
Travis: [Travis Build Status](https://travis-ci.org/haampie/QRUpdate.jl/builds/458352828)
cc: @haampie

Please make sure that:

@haampie This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


KristofferC:
This broke the registry by registering a package with an identical name (modulo casing) as an existing package https://github.com/JuliaLang/METADATA.jl/pull/471)7. cc @andreasnoack @haampie

andreasnoack:
Is this an intended limitation in the registry? They wouldn't get the same UUID so shouldn't the registry be fine?

fredrikekre:
The problem is that some systems can't handle two folders named `QRupdate` and `QRUpdate`. It is not a limitation in the registry, but in the conversion script I would say. Instead of producing
```
$uuid1 = { name = "QRupdate", path = "Q/QRupdate" }
$uuid2 = { name = "QRUpdate", path = "Q/QRUpdate" }
```
it would have to produce
```
$uuid1 = { name = "QRupdate", path = "Q/QRupdate" }
$uuid2 = { name = "QRUpdate", path = "Q/QRUpdate2" }
```
or something, and everything should be fine.

andreasnoack:
I was actually not sure if we intended to allow multiple packages with the same name in the same registry. I just revisited the Pkg Julep and it seems that it isn't the intension so maybe we just want to disallow repeated names after normalizing the casing.

I would have to check but I would think that we check for exact match of names in the METADATA tests but that the casing here made them pass.

haampie:
Whoops 😅, I can rename the package, so removing it from metadata now is fine for me.

StefanKarpinski:
We can eventually handle this fine in registries, but as @fredrikekre said, we'd need to choose a different way of assigning a path to packages with clashing names on the registry side which the conversion script doesn't currently handle since package names in METADATA are supposed to be unique. Note that cloning METADATA with this situation on a case-insensitive system would have the same problem since `QRupdate` and `QRUpdate` would be the same directory.

KristofferC:
I think we should revert this quite quickly since everyone who uses the package manager now gets their registry poisoned AFAIU.

haampie:
@StefanKarpinski already did that: https://github.com/JuliaLang/METADATA.jl/commit/450af89455aee12cbb6710da5627bb87d5538b49

StefanKarpinski:
Yes, should be fixed now. Let me know if it's not.

KristofferC:
Oh alright. Good.

aviks:
Even if the system can technically handle packages with the same name, I think we should not allow this in the registry by default. It will the cause of too much confusion. While we are at it, we should also disallow new packages that are within an edit distance of 2 from an existing package. The NPM fiasco of today reminded me of the previous npm fiasco.

00vareladavid:
Nix does something like "$name-$uuid" to name their directories. I think that would work well to avoid this issue.

StefanKarpinski:
I think we can just use `$(name[1])/$name/$slug` in cases where there are collisions.

PR #19815, opened 2018-12-01
uncap ShowItLikeYouBuildIt

magerton:
Uncaps ShowItLikeYouBuildIt so that Interpolations@0.8.0 is compatible with julia 1.0

### DIFF
--- a/ShowItLikeYouBuildIt/versions/0.2.0/requires
+++ b/ShowItLikeYouBuildIt/versions/0.2.0/requires
@@ -1,1 +1,1 @@
-julia 0.6
+julia 0.6 2-


timholy:
Because ShowItLikeYouBuildIt effectively migrated to Base but with subtle changes, I think it's a little complicated to support on 1.0. (I could be wrong, but ImageCore tried this and it was kind of a mess.)

Why do you want Interpolations 0.8? Newer versions are significantly better---faster, more flexible, and easier to understand.

magerton:
Thanks -- yes, in a project of mine I wrote a few [small extensions](https://github.com/magerton/ShaleDrillingModel.jl/blob/upgrade-to-v0.7/src/BSplineExtensions.jl#L34-L92) for Interpolations@0.8.0 to do in-place interpolation in parallel and also compute the directional derivative in a single dimension. I was hoping to avoid digging into the internals of the new Interpolations.jl to revise my existing codebase. If it's an issue to support ShowItLikeYouBuildIt, no worries at all -- I've tweaked my local clone of the registry so that I can keep using Interpolations@0.8.0.

timholy:
I understand where you're coming from: I had several private packages that did something similar. Of course it was easier for me to port my own packages since I was also responsible for the redesign of Interpolations, but my experience may nevertheless be informative. Once I rewrote them, the Interpolations-related customizations were only 30% of the length of the original (it's a rare day you get to delete more than half a big chunk of code), easier to understand, and one was also considerably faster. The new design has essentially one [key concept](https://github.com/JuliaMath/Interpolations.jl/blob/35de5f9bb6ef2c7516ff385402d225918eabc72c/src/Interpolations.jl#L145-L187), though of course there are a lot of specific method calls to grok.

Sorry for any inconvenience, but I hope in the long run it will make your life better.

magerton:
Thanks for the response, @timholy! I'll have a look -- 70% shorter and easier to understand would be dreamy!

PR #19871, opened 2018-12-04
Tag Compose.jl v0.7.2

attobot:
Repository: [GiovineItalia/Compose.jl](https://github.com/GiovineItalia/Compose.jl)
Release: [v0.7.2](https://github.com/GiovineItalia/Compose.jl/releases/tag/v0.7.2)
Travis: [Travis Build Status](https://travis-ci.org/GiovineItalia/Compose.jl/builds/463310091)
Diff: [vs v0.7.1](https://github.com/GiovineItalia/Compose.jl/compare/f37d1fe1fcce1bf6fcdc4a76f69e71ab88a7b474...84afc6c16f100c94c7337eafc3664bb3b9c7413b)
`requires` vs v0.7.1: no changes
cc: @bjarthur

Please make sure that:


andreasnoack:
It looks like the regressions tests don't pass and that it's not just `id` differences.

bjarthur:
the differences are just a change in how opacity is specified.  the visual appearance of the regression tests remain the same when looked at manually.

bigger question though is why the regression tests are being performed by CI.  is TRAVIS no longer defined in the test environment?  see https://github.com/GiovineItalia/Compose.jl/blob/master/test/runtests.jl#L31

andreasnoack:
> bigger question though is why the regression tests are being performed by CI.

The tests are not run on Travis. You should probably use the variable `CI` variable instead. It's set on CIBot and it's also set on Travis, see https://docs.travis-ci.com/user/environment-variables/#default-environment-variables.

andreasnoack:
I'll merge for now but it would be great if you could make that change before next release.

PR #19894, opened 2018-12-05
Register new package GCSObjectStores.jl v0.2.0

attobot:
Repository: [JockLawrie/GCSObjectStores.jl](https://github.com/JockLawrie/GCSObjectStores.jl)
Release: [v0.2.0](https://github.com/JockLawrie/GCSObjectStores.jl/releases/tag/v0.2.0)
cc: @JockLawrie

Please make sure that:

@JockLawrie This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


dpsanders:
Please don't use abbreviations in package names.

JockLawrie:
Agreed it's not ideal. Happy to change. But to what?

The aim of the name is to convey that the package follows the API defined in [ObjectStores.jl](https://github.com/JockLawrie/ObjectStores.jl) and uses _Google Cloud Storage_ as the storage back-end.

For reference, [LocalDiskObjectStores.jl](https://github.com/JockLawrie/LocalDiskObjectStores.jl) implements the `ObjectStore` API and uses the local file system as the storage back-end. Following this naming structure we get `GoogleCloudStorageObjectStores.jl`, which I think is too long.

Perhaps `GoogleCloudObjectStores.jl`? This name seems to imply that there's only 1 GCP product that can handle object storage, which is not strictly the case - you could write an `ObjectStore` that uses _Google DataStore_ or _Google Big Table_ for example. But this probably doesn't matter.

I also don't want to give the impression that this package wraps the entire _Cloud Storage_ API. Users who require that can use `GoogleCloud.jl`. Instead this package provides a limited subset of the _Cloud Storage_ functionality (the most commonly used parts) in an API that is compatible with other storage back-ends, which allows back-ends to be swapped out without changing code. With suitable messaging in the README I think this risk can be mitigated.

So the current candidates are `GoogleCloudObjectStores.jl`, or more precisely, `GoogleCloudStorageObjectStores.jl`.

Thoughts?
Alternatives?

andreasnoack:
I think either of `GoogleCloudObjectStores.jl` or `GoogleCloudStorageObjectStores.jl` would be big improvement.

JockLawrie:
OK renaming to `GoogleCloudObjectStores.jl`. Thanks

JockLawrie:
`GoogleCloudObjectStores.jl` is PR #19966. This PR (#19894) can be discarded.
Thanks again,
Jock

PR #19913, opened 2018-12-05
Register new package DFOLS.jl v0.1.0

attobot:
Repository: [QuantEcon/DFOLS.jl](https://github.com/QuantEcon/DFOLS.jl)
Release: [v0.1.0](https://github.com/QuantEcon/DFOLS.jl/releases/tag/v0.1.0)
cc: @arnavs

Please make sure that:

@arnavs This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


dpsanders:
Naming the package without abbreviations would be preferable.

arnavs:
So `DerivativeFreeOptimizationforLeastSquares.jl`? `:-)`

The algorithm in this case is known by the abbreviation, as is the underlying Python package, so I think this should be OK. Open to suggestions though.

arnavs:
(Just to elaborate, I think what you suggest would be preferable if this was a general `DerivativeFree.jl` or even `DerivativeFreeLeastSquares.jl` or something, but if it's a wrapper around a specific Python package then it should match the name of the package is my view).

fredrikekre:
> but if it's a wrapper around a specific Python package then it should match the name of the package is my view

Yes, see the guidelines, point 6: https://julialang.github.io/Pkg.jl/dev/creating-packages/#Package-naming-guidelines-1

arnavs:
Perfect @fredrikekre. Is there anything else we need to discuss before this PR is actionable?

fredrikekre:
No, but new packages registrations are open 3 days for comments etc.

arnavs:
For sure, not rushing you, just wanted to check if I've left anything undone.

arnavs:
@fredrikekre Is this good to go now? Would help to tag since I could use it in deps for other projects

PR #20011, opened 2018-12-11
Tag release of AuditoryFilters.jl

simonbyrne:
This was a very old package which never had a release, and thus broke attobot.

cc: @jfsantos, @aviks


jfsantos:
Even though the tests for this version failed, can we add it as is? I just committed fixes for Julia 0.6 so that all my tests are passing on 0.6, 0.7, and 1.0, but I thought it might be better to wait until this version is there and retag using attobot once it is happy with my package.

simonbyrne:
Probably the best option is to delete the old tag and create a new one, then I can update this PR.

jfsantos:
Thanks, I just deleted and created a new tag for the updated code.

andreasnoack:
Is this ready?

jfsantos:
@andreasnoack Yes. The package has most of the functionality it will ever have and it matches equivalent packages in MATLAB/Python. I'll add more detailed documentation eventually but so far users of the equivalent packages in MATLAB/Python will know how to use it as I've kept the API similar.

simonbyrne:
Cool. Attobot should work in future.

PR #20151, opened 2018-12-19
update NNlib

MikeInnes:


MikeInnes:
Turns out upper bounds are exclusive, so that wasn't quite what I meant.

PR #20218, opened 2018-12-23
Tag InstantiateFromURL.jl v0.2.1

attobot:
Repository: [QuantEcon/InstantiateFromURL.jl](https://github.com/QuantEcon/InstantiateFromURL.jl)
Release: [v0.2.1](https://github.com/QuantEcon/InstantiateFromURL.jl/releases/tag/v0.2.1)
Diff: [vs v0.2.0](https://github.com/QuantEcon/InstantiateFromURL.jl/compare/5fb5dc917481bbb7ba1ddd02f59274e09eaff333...f8e1c41101e6fca8e80e9ce08afc4958085f90f4)
`requires` vs v0.2.0: no changes
cc: @arnavs

Please make sure that:


arnavs:
@andreasnoack Actually, could this be reverted? I forgot that I had this PR open- would like to test this a bit more before deploying to METADTA.

arnavs:
cc @jlperla. I think we should still tag this commit at some point (removes a `@suppress`, for ex.), but no need to do so now for the reasons discussed.

jlperla:
Sure. After we test on every operating system and make sure we like the lack of suppression on the output. We can revisit after you are back and we have more testing resources

andreasnoack:
Ideally, we don't revert releases unless they unintentionally break other packages or out infrastructure. However, if this is still a big issue for you then you can open a PR to METADATA which reverts this.

PR #20240, opened 2018-12-27
Register new package Jfire.jl v0.0.4

attobot:
Repository: [orangeSi/Jfire.jl](https://github.com/orangeSi/Jfire.jl)
Release: [v0.0.4](https://github.com/orangeSi/Jfire.jl/releases/tag/v0.0.4)
cc: @orangeSi

Please make sure that:

@orangeSi This PR will remain open for three days for feedback (which is optional). If you get feedback, please let us know if you are making changes, and we'll merge once you're done.


dpsanders:
It is strongly preferred not to include J in the name, and to use a name that is informative about the content of the package.

orangeSi:
> It is strongly preferred not to include J in the name, and to use a name that is informative about the content of the package.

thanks for point out that~ I was a little hesitating when I gived the name Jfire, but I don't want to just use the same name fire with python package fire which inspired my package development. I want to left fire name to package of other people.

dpsanders:
What I mean is that using the name "fire" doesn't tell us what the package actually does. It could be something like CommandLineInterface.jl

orangeSi:
> What I mean is that using the name "fire" doesn't tell us what the package actually does. It could be something like CommandLineInterface.jl

my package is inspired by https://github.com/google/python-fire, so I want to keep the fire name while with a little difference. I will consider to chanage the name in later release. If you have name suggestion, welcome to tell me, thanks~

ronisbr:
I thought we already have something like Python Fire in this package: https://github.com/ylxdzsw/Fire.jl

orangeSi:
> I thought we already have something like Python Fire in this package: https://github.com/ylxdzsw/Fire.jl

I knew that before I start to write this

dpsanders:
What is the difference between the packages? Could they be integrated into a single one?

orangeSi:
> What is the difference between the packages? Could they be integrated into a single one?

I woluld hope to integrate into a single one. But the main purpose  of writing this package is to practice my julia skill, because  I  started to learned julia a few weeks ago and still be a fresh bird to julia.

The difference of  both is the implement way. The function of both packages have not too much difference. But the code of Fire.jl is hard to understand for me now, So I tried  another implement way.

orangeSi:
@andreasnoack what do I need to do now? It is my first time to do this, thanks for help~

andreasnoack:
If you'd still like to have this package registered then let us know and we can merge this. However, it is not clear to me why you'd like to register the package if the purpose of this is just to practice your Julia skills.

orangeSi:
> If you'd still like to have this package registered then let us know and we can merge this. However, it is not clear to me why you'd like to register the package if the purpose of this is just to practice your Julia skills.

yes, I still want to register this, thanks~
Rome wasn't build in one day. I think if I pulish my package, I have more motivation to practive my Julia skill . The function of my package is still developing, but it has worked in some ways which I often use with. Maybe my words mislead you, if codes cann't help  some people to do somethings easily, I will not want to publish it.

PR #20284, opened 2019-01-01
Pin FourierFlows

navidcy:


GeophysicalFlows/versions/0.1.0/requires
```diff
@@ -1,7 +1,7 @@
 julia 0.7
 
 FFTW
-FourierFlows
+FourierFlows 0.2.2
```
andreasnoack:
```suggestion
 FourierFlows 0.0 0.3
```
Your version would add 0.2.2 as the lower bound where the suggestion here adds 0.3 as the upper but excluded bound.

navidcy: Thanks for correcting that.

PR #20307, opened 2019-01-02
Adds minimum requirements for FourierFlows

navidcy:


navidcy:
julia 0.7 failed but it seems some sort of time-out error while downloading something
can we trigger travis-ci to run again?

PR #20441, opened 2019-01-08
Check that Project.toml version matches tag

ararslan:
This adds a check for PRs that the tagged version matches the version specified in the Project.toml at that revision, if applicable. This will catch issues such as the Nabla 0.3.0 tag, which had 0.2.0 as its version in its Project.toml and necessitated a 0.3.1 to get the versions matching.

Stefan, I was hoping you could speak to the validity of my implementation of the check since it uses Pkg internals, and Andreas, I was hoping you could opine on whether this is a worthwhile check to have in METADATA.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -146,3 +146,28 @@ for pkg in keys(modified)
         error(msg)
     end
 end
+
+if VERSION >= v"0.7.0"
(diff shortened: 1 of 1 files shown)


andreasnoack:
Thanks for adding this check

PR #20490, opened 2019-01-10
Tag MiniLogging v0.0.3 [https://github.com/colinfang/MiniLogging.jl]

colinfang:
Diff vs v0.0.2: https://github.com/colinfang/MiniLogging.jl/compare/8185dac196aa94c76fefdd0760097c6460802e89...7496182ff496d0b0f51beb2676027d528d50da8a


colinfang:
Can I update the version of my package, while still only targeting julia v0.5 & v0.6 for the moment?

colinfang:
@andreasnoack Can I update the version of my package, while still only targeting julia v0.5 & v0.6 for the moment? Or what is the best way to fix the error here?

andreasnoack:
Julia 0.5 is no longer supported so I'd recommend that you increase the minimum version required. Would that be possible? If you *need* to continue support for Julia 0.5 then there is a process to opt out of testing but I'd have to check how it works.

PR #20764, opened 2019-01-22
Require Julia >=0.7 for new registrations

ararslan:
And clean out the vestiges of the old, now unmaintained 0.6.


ararslan:
Oh, the grandfathered list needs to be updated. :expressionless:

StefanKarpinski:
I object to this change. I don't see why we would not allow anyone to register versions for older Julia versions. It requires us to do nothing at all. What is the harm?

ararslan:
After discussion we've decided to leave it alone

StefanKarpinski:
After some discussion on Slack, we've come to the conclusion that there's no reason to do this now. In the future, we can run each new version with every Julia version that we still have around that it claims to be compatible with. We may want manual intervention in cases where it claims compatibility with Julia versions that we are no longer able to run tests against. For reverse dependency testing, we should only test against actively maintained versions of Julia, which should keep CI demand within reason. I believe that the [original reasoning](https://github.com/JuliaLang/METADATA.jl/pull/6146) for disallowing registering package versions claiming compatibility for old versions of Julia was because we believed that people would incorrectly claim compatibility. That is less of an issue now since we do considerably more automatic testing of new registered versions.

andreasnoack:
I don't have strong opinions here. It might save a little labor not to bother with old versions but I'm not sure it would be that much. However

> was because we believed that people would incorrectly claim compatibility.

It's pretty safe to assume that people don't claim correct compatibility. That happens all the time. 

> That is less of an issue now since we do considerably more automatic testing of new registered versions.

We currently don't do much testing since CIBot is no longer running. Hence we wouldn't find out if a new version of Compat broke all packages on 0.6.

PR #20773, opened 2019-01-23
Delete sha1

kaarthiksundar:
Remove all files relating v2.0.0 - the latest available version is v1.0.0


andreasnoack:
It's a little cleaner to do as a revert. See https://github.com/JuliaLang/METADATA.jl/pull/20784

PR #20774, opened 2019-01-23
Delete requires

kaarthiksundar:
Remove all files relating to v2.0.0. Latest version is v1.0.0


andreasnoack:
It's a little cleaner to do as a revert. See #20784

PR #20740, opened 2019-01-22
Tag Dubins.jl v2.0.0

attobot:
Repository: [kaarthiksundar/Dubins.jl](https://github.com/kaarthiksundar/Dubins.jl)
Release: [v2.0.0](https://github.com/kaarthiksundar/Dubins.jl/releases/tag/v2.0.0)
Travis: [Travis Build Status](https://travis-ci.org/kaarthiksundar/Dubins.jl/builds/482687787)
Diff: [vs v0.1.1](https://github.com/kaarthiksundar/Dubins.jl/compare/ed3a7d781735fc0f157ece4c1373bef80dd2c209...a7e317a91989ac30d58e8a421ed3a3a678ce8319)
`requires` vs v0.1.1: 
```diff
--- 0.1.1/requires
+++ 2.0.0/requires
@@ -1,3 +1,3 @@
-julia 0.6
-Memento 0.5.0
-Compat 0.26.0
+julia 0.7
+Memento 0.8
+Compat 1.0
```
cc: @kaarthiksundar

Please make sure that:
- CI passes for supported Julia versions (if applicable).
- Version bounds reflect minimum requirements.

PR #20784, opened 2019-01-23
Revert "Tag Dubins.jl v2.0.0 [https://github.com/kaarthiksundar/Dubin…

andreasnoack:
…s.jl/releases/tag/v2.0.0] (#20740)"

This reverts commit aa440164a09c183d8ea3368fd71fc4a33a24e2a7.


andreasnoack:
Notice that if somebody happens to be on the v2.0.0 release then their project will break and they'd have to fix it. Something like `free Dubins` might be sufficient in that case.

PR #20885, opened 2019-01-25
Register Unicode2LaTeX v1.0.0

ghost:
Hand-crafted since I am hosting all new code over at GitLab since the autumn of 2017. Thus, please be extra careful when reviewing the syntax.


ghost:
Okay, that is better. All tests passed after I mucked up the UUID twice.

PR #21242, opened 2019-02-07
Update Resource Addresses

### LINKED ISSUE #15092: Updating repository URL
Would it be a welcomed PR to update the repository URL?
For example,
For `Nulls.jl` the repository on file is

`https://github.com/JuliaData/Nulls.jl.git`, but actually redirects to 
`https://github.com/JuliaData/Missings.jl`.

Nosferican:
Fix #15092


Nosferican:
What is the recommended approach here? https://github.com/MikeInnes/Tokamak does not end in `.jl` should the package be removed?

andreasnoack:
Let's just leave it now that it's there.

andreasnoack:
Could you leave Tokamak as is? Then it should be good to go.

PR #21342, opened 2019-02-11
Revert "Tag BedgraphFiles.jl v2.0.1"

StefanKarpinski:
Reverts JuliaLang/METADATA.jl#21009

This version refers to a commit that cannot be found when checking out the BedgraphFiles.jl repo, which breaks a lot of stuff. Usually when a commit is tagged it is visible in git. @CiaranOMara did you change the commit that the v2.0.1 tag referred to after registering it?

### DIFF
--- a/BedgraphFiles/versions/2.0.1/requires
+++ b/BedgraphFiles/versions/2.0.1/requires
@@ -1,9 +1,0 @@
-julia 0.7
-Bedgraph 1.0.0
-DataFrames 0.9.0
-FileIO 1.0.1
-IterableTables 0.9.0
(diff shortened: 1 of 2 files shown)


CiaranOMara:
@StefanKarpinski, this looks serious. Yes, I retagged BedgraphFiles.jl v2.0.1 and had not got around to creating a pull request to METADATA.jl to update the SHA. Sorry, I was not aware that this would break a lot of stuff.

StefanKarpinski:
No problem. Ping me on the PR you make to put this version back since I'll need to make sure it doesn't break things.

PR #21479, opened 2019-02-15
Check for stdlibs in REQUIRE

ararslan:
This adds a check that will error if a package lists a stdlib module in its REQUIRE file.

### DIFF
--- a/.test/check-pr.jl
+++ b/.test/check-pr.jl
@@ -189,3 +189,33 @@ if VERSION >= v"0.7.0"
         end
     end
 end
+
+# Ensure no stdlibs are listed in REQUIRE
+if VERSION >= v"0.7.0"
+    const STDLIBS = readdir(Pkg.Types.stdlib_dir())
+    for (pkg, versions) in modified, ver in versions
+        reqs = joinpath(BUILD_DIR, pkg, "versions", string(ver), "requires")
+        # Do a first pass over the file to get the Julia version. We'll do this as two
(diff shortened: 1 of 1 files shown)


.test/check-pr.jl
```diff
@@ -189,3 +189,46 @@ if VERSION >= v"0.7.0"
...
+    "Logging",
+    "Markdown",
+    "Mmap",
+    "Pkg",
+    "Printf",
+    "Profile",
+    "Random",
+    "REPL",
+    "Serialization",
+    "SHA",
```
oxinabox: Are we still allowing 0.6 packages to update their Metadata? If so banning them from depending on SHA.jl is not going to be ok.

.test/check-pr.jl
```diff
@@ -189,3 +189,46 @@ if VERSION >= v"0.7.0"
...
+    "SharedArrays",
+    "Sockets",
+    "SparseArrays",
+    "Statistics",
+    "SuiteSparse",
+    "Test",
+    "Unicode",
+    "UUIDs",
+]
```
omus: Probably better to use Pkg3 to get this list? That would mean this check couldn't occur on Julia 0.6 checks

ararslan:
Concerns addressed. The list of stdlib modules is now obtained directly from Pkg, and this allows SHA in REQUIRE files so long as the minimum required Julia version is before it was added as a stdlib.

andreasnoack:
Great. Thanks for adding this check.

PR #21525, opened 2019-02-18
Revert "Tag JuliennedArrays.jl v0.1.0"

KristofferC:
Reverts JuliaLang/METADATA.jl#17012

This release has been deleted from the repo and is causing problems https://discourse.julialang.org/t/install-package-dependence-from-specific-repository/20930

### DIFF
--- a/JuliennedArrays/versions/0.1.0/requires
+++ b/JuliennedArrays/versions/0.1.0/requires
@@ -1,2 +1,0 @@
-julia 0.6
-Keys
--- a/JuliennedArrays/versions/0.1.0/sha1
+++ b/JuliennedArrays/versions/0.1.0/sha1
@@ -1,1 +1,0 @@
-75dc7dec61714feed2d28f438dad5dd44c8591a3


KristofferC:
And 0.0.3 apparently bumped julia version but this was hidden due to the existence of this 0.1.0. So now this fails. Great.

ferrolho:
I think merging this may have caused following PR and merges to fail, according to Travis

KristofferC:
Yeah, that is possible. We should probably revert this and instead make the 0.1.0 release the same commit as the 0.3.0 release.

ferrolho:
Yup, this seems to have fixed it: [JuliaLang/METADATA.jl/builds/500404291](https://travis-ci.org/JuliaLang/METADATA.jl/builds/500404291)

I tried re-submitting my PR but I still get an error related to this, I suppose:
```
ERROR: LoadError: New tag 0.0.3 of package JuliennedArrays requires julia 1.0.0, but version 0.0.1 of JuliennedArrays requires julia 0.6.0. Use a new minor package version when support for an old version of Julia is dropped.
```

PR #21573, opened 2019-02-19
Remove Currencies and CurrenciesBase

EricForgy:
Not sure what is the procedure for removing packages. Hope this is alright.

I would like these to live in JuliaFinance registry.


dpsanders:
You just need to transfer ownership to the JuliaFinance organization and then make a PR to METADATA to update the urls.

EricForgy:
Hi @dpsanders 

The ownership is already with JuliaFinance so no need to make a PR to update the urls. I was thinking to "de list" these packages from `General`. 

I plan to maintain my own registry and want users to 

```julia
pkg> registry add https://github.com/JuliaFinance/JuliaFinance.git
```

before they can

```julia
pkg> add Currencies
```

I probably will not submit further PRs to `General`, so the risk is someone will be getting stale versions if these packages remain here.

Does that make sense?

EricForgy:
Bump 👋

EricForgy:
Bump 👋

EricForgy:
This is causing me some problems. Could this please be merged or can someone please explain to me why it can't be merged?

EricForgy:
Bump 👋

dpsanders:
I don't have access to do this. I'm guessing everybody with access is unavailable.

EricForgy:
Thanks for pinging on Slack. It is still the weekend, so hopefully someone can look soon 🙏 

I imagine this might create some issue when translating to `General`, but hope not 🙏

aviks:
Yeah, I don't think this will get done on the weekend. At the very least, as a non-standard operation, it needs to be done manually, and as something might break downstream actions, and so needs to be considered and acted upon at a more active timeperiod.  (Most metadata operations are merged automatically these days)

As a separate issue, I would invite you to consider the social effects of this decision. You are asserting that all dependencies of these packages will need to live in your own registry. That is a pretty strong statement, and so while there is no technical limitation any more to doing that, I hope you have thought through the consequences.

EricForgy:
Thanks @aviks 

No worries about getting this done over the weekend. I understand. 

I am also aware of the ramifications, but am hopeful there will be some policy reconsiderations with `General` that would someday allow packages to be registered even if they have dependencies on external registries.

I think you are aware of this discussion: [Decentralized Package Manager](https://discourse.julialang.org/t/decentralized-package-manager/20931).

If this requires manual effort, I'm happy to do that myself if I can. For example, if it helps, I could submit a corresponding PR to `General`.

I definitely don't want to create a lot of work for someone.

EricForgy:
PS: The urgency on this is gone now that I found a workaround. I (and anyone who wants to use this) basically need to modify the local version of `General`. That is not ideal, but if it helps, we can hold off on this until METADATA is retired as long as that happens fairly soon.

KristofferC (changes requested):
This will break peoples code and shouldn't be done until there is a workaround.

If the current maintainer no longer wants to host the package we should mirror it and repoint the url.

KristofferC:
> That is not ideal, but if it helps, we can hold off on this until METADATA is retired

👍

PR #21665, opened 2019-02-23
Rename MathOptInterfaceMosek to MosekTools

blegat:
See discussion here: https://github.com/JuliaOpt/Mosek.jl/issues/168
cc @ulfworsoe


blegat:
We should keep the existing tags to avoid breaking code and just create a new package with the new name.

PR #21692, opened 2019-02-24
Revert "Tag Grassmann.jl v0.0.4"

chakravala:
Reverts JuliaLang/METADATA.jl#21690

Force pushed the commit, so you're going to have to revert. sorry. won't happen again.

it was due to an issuw with backwards compatibility for 0.7


chakravala:
@andreasnoack could you please merge this? otherwise the METADATA is going to be messed up

PR #21711, opened 2019-02-25
Put an upper bound on HTTP version in Coverage <= v0.8.0

martinholters:
A breaking API change in HTTP 0.8.0 was only respected in Coverage 0.8.1 (with https://github.com/JuliaCI/Coverage.jl/pull/197), so earlier Coverage versions have to be used together with earlier HTTP versions. As Coverage 0.8.1 requires Julia 1.0, but HTTP 0.8.0 is compatible with Julia 0.7, without this PR, coverage submission from CI is broken unless one takes extra measures to downgrade HTTP.

Ref https://github.com/JuliaCI/Coverage.jl/issues/202, cc @ararslan 

Unfortunately, I have no idea what reasonable lower bounds are. I choose the lowest ones claimed to be compatible with Julia 0.6 or 1.0, respectively. Should they rather be 0.0.0 to make it clear that they are not real, intentional lower bounds?

### DIFF
--- a/Coverage/versions/0.5.0/requires
+++ b/Coverage/versions/0.5.0/requires
@@ -1,5 +1,5 @@
 julia 0.6
 JSON
-HTTP
+HTTP 0.4 0.8.0
 Git
 Compat 0.44.0
--- a/Coverage/versions/0.5.1/requires
+++ b/Coverage/versions/0.5.1/requires
@@ -1,5 +1,5 @@
 julia 0.6
 JSON
-HTTP
+HTTP 0.4 0.8.0
 Git
 Compat 0.44.0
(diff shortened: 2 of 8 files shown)


ararslan:
Seems fine to me. I assume it isn't easier to fix the problem in Coverage?

martinholters:
>  I assume it isn't easier to fix the problem in Coverage?

We'd have to lower Coverage's minimum required Julia version to 0.7 and then tag 0.8.2. I'd expect that to be relatively easy, considering the few changes between Julia 0.7 and 1.0. I don't know whether that's desired, however. There's probably a reason the minimum version was bumped straight to 1.0 and not just 0.7. Also, older Coverage versions would obviously sill not work, they just wouldn't get installed by default, which would suffice to make standard CI setups work.

No hurry here as coverage submission is _also_ broken by https://github.com/JuliaWeb/MbedTLS.jl/issues/193 which is unrelated to the HTTP version bounds here, but unless objections are voiced, I'll merge in a few days.

PR #21715, opened 2019-02-25
restore Currencies.jl

KristofferC:
The Currencies.jl package has been made nonfunctional (git history destroyed). This PR repoints the URL to a mirror with the last functioning copy (currently on my own account). We should find a better place to host it but for now, it will do.

### DIFF
--- a/Currencies/url
+++ b/Currencies/url
@@ -1,1 +1,1 @@
-git://github.com/JuliaFinance/Currencies.jl.git
+https://github.com/KristofferC/Currencies.jl
--- a/Currencies/url
+++ b/Currencies/url
@@ -1,1 +1,1 @@
-https://github.com/KristofferC/Currencies.jl.git
+https://github.com/JuliaPackageMirrors/Currencies.jl.git
--- a/Currencies/url
+++ b/Currencies/url
@@ -1,1 +1,1 @@
-https://github.com/KristofferC/Currencies.jl
+https://github.com/KristofferC/Currencies.jl.git


EricForgy:
Hi @KristofferC. The git history wasn't deleted. I hope nothing is implied by that statement. The old git history can be found at [CurrencyBaskets.jl](https://github.com/JuliaFinance/CurrencyBaskets.jl).

The original `Currencies.jl` was renamed `CurrencyBaskets.jl` to preserve it and I created a new repo with the new design for going forward. I don't see why this is an issue. It is basically a breaking redesign / change.

I am open to suggestions how to resolve things. One thing I can do is delete `JuliaFinance/Currencies.jl` and rename `JuliaFinance/CurrenciesBaskets.jl` back to `JuliaFinance/Currencies.jl`. This will preserve the git history if that is important and then implement the changes again. 

This is a breaking change and it would not be great if breaking changes are not allowed for packages. There is absolutely nothing sinister going on here.

KristofferC:
The git history of https://github.com/JuliaFinance/Currencies.jl has indeed been completely clobbered.

> One thing I can do is delete `JuliaFinance/Currencies.jl` and rename `JuliaFinance/CurrenciesBaskets.jl` back to `JuliaFinance/Currencies.jl`. This will preserve the git history if that is important and then implement the changes again.

The way you rename a package is to fork it and re-register it (or since you don't want to register it anymore, you can just push whatever you want to your own registry). If you don't want to develop the package under Currencies.jl anymore it can be moved to https://github.com/JuliaAttic/, which is equivalent to this PR with the exception of the organization name. 

Since no more development will be done on Currencies.jl, I suggest merging this PR and reevaluate the situation in a few weeks to see if the original package is functional again.

EricForgy:
> The git history of https://github.com/JuliaFinance/Currencies.jl has indeed been completely clobbered.

No it hasn't. It is all there in CurrencyBaskets.jl. Like I said, I can rename it back if the history is so important. 

This is a redesign / breaking change. I don't see the problem. I renamed the previous design as a courtesy in case someone wanted that old functionality. I regret that now. I should have just overwritten it and we wouldn't be having these issues.

There should just be one `Currencies` in JuliaFinance and it would kind of suck if users had to manually pick which one in order to use any other JuliaFinance packages because they all pretty much depend on Currencies.jl.

KristofferC:
> I renamed the previous design as a courtesy in case someone wanted that old functionality.

That's not how you do it. You fork the package and keep developing it. If dependencies want to change to the fork they do this by changing what they depend on.

EricForgy:
> That's not how you do it. You fork the package and keep developing it. If dependencies want to change to the fork they do this by changing what they depend on.

That makes sense. I didn't know that. I could try.

What if I do this?

1. Delete the newly created `Currencies.jl` since it doesn't have the git history.
1. Rename `CurrencyBaskets.jl` back to `Currencies.jl` to retain the history. So now we have the correct repo with the correct history.
1. Fork the old `Currencies.jl` somewhere else.
1. Implement the new design in `Currencies.jl`

Does that work for you?

KristofferC:
An alternative is to force push the commit of the latest commit of where Currencies.jl was called Currencies and then leave it alone. Dependencies can then start move over to CurrencyBaskets.jl if they wish to.

This is equivalent to this PR with the exception of the url.

KristofferC:
> Implement the new design in Currencies.jl

As long as there is a Currencies.jl repo, with the correct git history, that loads a package called Currencies and the url in METADATA.jl points to that repo, it is ok.

EricForgy:
> An alternative is to force push the commit of the latest commit of where Currencies.jl was called Currencies and then leave it alone. Dependencies can then start move over to CurrencyBaskets.jl if they wish to.
> 
> This is equivalent to this PR with the exception of the url.

Hi @KristofferC ,

I just force pushed the last tagged version CurrencyBaskets.jl back to Currencies.jl. This restores the git history, but only for the master branch 🤔 

Is there an easy way to force push everything?

Would it be easier to just delete it and rename CurrencyBaskets -> Currencies? This would restore everything as if I never renamed it in the first place.

KristofferC:
Master branch (and possibly tags) is all that really matters. Renaming CurrencyBaskets.jl to Currencies.jl and reverting the commit that renamed it is another way, sure.

EricForgy:
Ok. I will try the rename (🙏), but it means Currencies.jl will not exist for a few seconds / minutes, but I think that is ok because you already merged the mirror, right?

EricForgy:
Currencies.jl is fully restored 🙏

KristofferC:
> reverting the commit that renamed it is another way, sure.

FYI, this has not been done yet.

EricForgy:
Right. Sorry. Trying now, but I got some error:

```
POST git-receive-pack (201 bytes)
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: Cannot force-push to a protected branch        
Pushing to https://github.com/JuliaFinance/Currencies.jl.git
To https://github.com/JuliaFinance/Currencies.jl.git
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/JuliaFinance/Currencies.jl.git'
```

KristofferC:
You don't have to force push it. The revert commit can just be on top of master. If you look at the fork url in this PR you can see how it looks.

EricForgy:
Got it. Thank you 🙏

KristofferC:
:+1:

PR #21737, opened 2019-02-26
FactCheck.jl: Update URL

Divided-Pi:
Hi, I've been working on bringing FactCheck.jl up-to-date to be compatible with Julia v1.0 and Julia v1.1. I've edited the source to use the updated syntax and have done my best to get runtests.jl to run and succeed in the travis.cl config file. As far as I can tell, my package is using the new Pkg format correctly in terms of the Project.toml and Manifest files. 

If I'm missing anything or if the updated package is not up to snuff please let me know and I'll try to correct any issues. 

Thanks you

### DIFF
--- a/FactCheck/url
+++ b/FactCheck/url
@@ -1,1 +1,1 @@
-git://github.com/JuliaAttic/FactCheck.jl.git
+https://github.com/Divided-Pi/FactCheck.jl.git


Divided-Pi:
This is my first attempt at contributing to opensource, so apologies in advance

andreasnoack:
What is your motivation behind reviving the package instead of using the `Test` standard library?

Divided-Pi:
> What is your motivation behind reviving the package instead of using the `Test` standard library?

The original motivation was to get caught up with julia v1.0+ with a hands on project, and originally I was going to attempt to update a second package which used FactCheck in it's testing. That second deprecated project does have an active successor which doesn't rely on FactCheck, so I'm going to try and help there anyway. 
I continued with it thinking if there were other old packages which also used FactCheck it might be nice to have an updated version to make the updating process a bit easier for others. But to be honest I don't have an idea how often FactCheck was used while it was active. 

If it's seen as redundant and confusing to make it an active project again that's understandable. My only question would be if the package is currently set up properly to be an active project. This was my first time using pkg's functionality such as Project/Manifest.toml files so it would be nice to know if I handled that correctly

Thanks for any feedback

andreasnoack:
>  This was my first time using pkg's functionality such as Project/Manifest.toml files so it would be nice to know if I handled that correctly

It looks right to me.

Unless you really want to, I'd recommend that we don't revive `FactCheck`. I think it would be better if any packages that still use `FactCheck` were transitioned to use `Test`.

Divided-Pi:
Thanks Andreas, I agree it makes sense to move people towards Test in general. And if anyone needs FactCheck they're welcome clone my repo and use it as a local/develop package as needed. Nice to know I had everything set up correctly though. It was a fun learning experience

PR #21776, opened 2019-02-28
Fix v0.11.0 sha1 for Currencies

EricForgy:
Oh man 😔 

I turned off Attobot for this package so I thought I was safe to do some experiments with tagging, but it looks like Attobot did not deactivate and submitted a PR anyway 😔 

[Edit: More likely, it is entirely possible that I didn't "Save" since I was in a rush.]

Really sorry about that 🙏 

This package currently has no v0.11.0 and that git-tree-sha1 is not valid. I thought when I remove this package from the Attobot integration, it wouldn't create a PR and didn't notice until now 😔

Edit: What is the best solution? Instead of deleting the version, I created an actual v0.11.0 and updated the sha1. Open to suggestions.


EricForgy:
```
ERROR: LoadError: LoadError: Do not modify existing published tag Currencies/versions/0.11.0/sha1 (policy 8).
```

I'll try something else. Please let me know the proper way to deal with this :pray: I am not a Git expert (obviously) and shooting in the dark.

Issue #22416, opened 2019-03-25
Should catch disagreement between Project.toml:[compat] and REQUIRE


timholy:
#22412 allowed a mistake to slip through, see https://github.com/timholy/Revise.jl/issues/256. Ideally we should catch such problems automatically.

tpapp:
Or, ideally, even ignore `REQUIRE` if `Project.toml` is available.

DilumAluthge:
METADATA.jl is no longer maintained, so I am closing this issue.

If this issue is still relevant, please re-open a new issue on the Julia General registry: https://github.com/JuliaRegistries/General

PR #22417, opened 2019-03-25
Fix bounds on newest Revise

timholy:
Will this automatically propagate to JuliaRegistries? I infer from https://github.com/timholy/Revise.jl/issues/256 that the info in REQUIRES took precedence over what was in the Project.toml file. See #22416.


timholy:
In the interests of speed (bug reports are starting to pile up) I'm merging this now. We can look into other fixes later.

timholy:
The answer is, this does propagate to JuliaRegistries/General, so I think this should be fixed.

PR #21942, opened 2019-03-06
Tag Unicode2LaTeX.jl v1.1.0

ghost:
Good old manual commit from a GitLab user.

PR #22464, opened 2019-03-27
Tag Unicode2LaTeX.jl v1.2.0

ghost:
I submitted a [PR to tag v1.1.0](https://github.com/JuliaLang/METADATA.jl/pull/21942) roughly three weeks ago and it is still open. Now it is time for v1.2.0 given some upstream changes. What can I do to make PR merging run more smoothly so that we avoid this kind of situation in the future?


ghost:
I had forgot to push the right tag to the GitLab repo, could someone restart the CI? I can only do so using a rebase at this point since the change was external to the METADATA commit.

